Issue #2217, opened 2016-08-07
'alpha' parameter is unused in `draw_networkx_labels`


sanghack81:
Each of `draw_networkx_labels` and `draw_networkx_edge_labels` in `networkx.drawing.nx_pylab` defines `alpha` as a parameter, but the passed argument is not used in the code failing to generate transparent text.

aryamccarthy:
Seems like @Zsailer fixed this; can we close it?

sanghack81:
Closed #2269

PR #2269, opened 2016-09-30
alpha kwarg not used in pylab label drawing, added it here.  

Zsailer:
Fixed issue #2217. 

Both `draw_networkx_labels` and `draw_networkx_edge_labels` functions in the `nx_pylab` module take an `alpha` keyword argument, but don't use it when generating the text objects. I just added the bridge between the two here.


dschult:
Thanks!

Issue #2235, opened 2016-08-29
Abandon All pydotplus, Ye Who Graph Here


leycec:
NetworkX 1.11 unilaterally [dropped support](https://github.com/networkx/networkx/pull/1924) for**[`pydot`](https://github.com/erocarrera/pydot)**in favour of**[`pydotplus`](https://github.com/carlos-jenkins/pydotplus)**– breaking backward compatibility with upstream projects requiring the former.

Breaking backward compatibility isn't _necessarily_ a problem per se. Sometimes you gotta [crack the world's shell](https://www.youtube.com/watch?v=McKR0IgCM34) in order to make a tender omelette worthy of Iron Chef. :egg: 

Dropping support for the _only_ well-maintained pure-Python GraphViz library (`pydot`) in favour of its worst-maintained alternative (`pydotplus`) smells like a rotting wound waiting to fester, however.
## Then There Were Three

There are three `pydot`-like projects. In order of increasing commit and release activity, these are:
1. **[`pydotplus`](https://github.com/carlos-jenkins/pydotplus),** a `pydot` fork circa 2007. `pydotplus` is moribund, on life support, and effectively dead – and has been for nearly two years. It lasted seven glorious years before unceremoniously kicking the bit bucket. Only the good forks die young. This fork's:
   - [Last commit](https://github.com/carlos-jenkins/pydotplus/commits/master?page=1) was at **December 8th, 2014.** 
   - [Last stable release](https://pypi.python.org/pypi/pydotplus) was at **December 9th, 2014.**
2. **[`pydot-ng`](https://github.com/pydot/pydot-ng),** yet _another_ `pydot` fork also circa 2007. `pydot-ng` is _also_ moribund, on life support, and effectively dead – and has been for nearly a year. It lasted eight glorious years and then silently fell into a [sarlacc pit](https://www.youtube.com/watch?v=oezZTXzXs90), one-upping `pydotplus` by several months. All forks succumb to GitHub winter, in the end. This fork's:
   - [Last commit](https://github.com/pydot/pydot-ng/commits/master) was at **March 15th, 2015.** 
   - [Last stable release](https://pypi.python.org/pypi/pydot-ng) was at **February 26th, 2015.**
[...]

PR #2272, opened 2016-10-05
Make PyDot Support Great Again

leycec:
This pull request refactors the `networkx.drawing.nx_pydot` submodule to support the well-maintained [`pydot`](https://github.com/erocarrera/pydot) project in lieu of the effectively dead [`pydotplus`](https://github.com/carlos-jenkins/pydotplus) fork, resolving [issue #2235](https://github.com/networkx/networkx/issues/2235) (_Abandon All pydotplus, Ye Who Graph Here_) and restoring basic sanity.

See #2235 for full-frontal details. A brute-force **tl;dr** follows:

* The last commit for [`pydotplus`](https://github.com/carlos-jenkins/pydotplus) was at **[December 8th, 2014](https://github.com/carlos-jenkins/pydotplus/commits/master?page=1)** – nearly two years ago. Long-standing open issues and pull requests remain unresolved. `pydotplus` is a face-rotting zombie.
* The last commit for [`pydot`](https://github.com/erocarrera/pydot) was at **[July 2nd, 2016](https://github.com/erocarrera/pydot/commits/master)** – only a little over two months ago. Issues and pull requests are quickly dispatched with, as evidenced by a [closed issue](https://github.com/erocarrera/pydot/issues/133) I foolishly opened two days ago.

**`pydot` lives.**

## Not all that Glitters is PyDot Gold

`pydot` 1.2.0 recently broke [backward API compatibility](https://github.com/erocarrera/pydot/blob/master/ChangeLog) with pre-1.0.29 forks (like `pydotplus`) in mostly beneficial ways. Let's talk a bit about that.

### Dot File Deserialization Gone Wild

In `pydot` 1.2.0, the `pydot.graph_from_dot_data()` function now _unconditionally_ returns a list of all deserialized `pydot.Dot` objects rather than _conditionally_ returning a list if the passed data contains two or more graphs or a single `pydot.Dot` object otherwise.

This is a positive change. The prior approach complicated caller handling and encouraged caller errors. It didn't help that the prior approach was poorly documented and that everyone downstream made simplistic assumptions about that approach. How do we know this?

Because NetworkX currently assumes the `pydot.graph_from_dot_data()` function to _always_ return a single `pydot.Dot` object. Which was never the case. Maybe nobody cares about `pydot` integration with dot files containing multiple graphs? It is pretty edge-case. <sup>_Or maybe my tenuous deathgrip on sanity has slipped another notch lower._</sup>

NetworkX functionality calling the `pydot.graph_from_dot_data()` function includes:

* `networkx.drawing.nx_pydot.read_dot()`.
* `networkx.drawing.nx_pydot.pydot_layout()`.
* `networkx.drawing.tests.test_pydot.TestPydot.pydot_checks()`.

This functionality has all been refactored to accept lists of `pydot.Dot` objects. In the case of both `pydot_layout()` and `pydot_checks()`, a list of exactly one object is _always_ returned. Life is simple here.

In the case of `read_dot()`, however, a list of two or more objects are returned if the passed dot file contains two or more graphs. This was _always_ the case, of course. NetworkX currently ignores this discrepancy by returning "sight-unseen" whatever `pydot` returns _without_ documenting that a list of graphs rather than a single graph might be returned.

Two options exist here:

1. **Break NetworkX backwards compatibility.** Refactor `read_dot()` to unconditionally return a list of `pydot.Dot` objects – just like the underlying `pydot.graph_from_dot_data()`. Orthogonality is preserved, which is mildly good. All downstream logic is broken, which is strongly bad.
1. **Mostly preserve NetworkX backwards compatibility.** Refactor `read_dot()` to unconditionally return only a single `pydot.Dot` object if the passed dot file contains only one graph _or_ raise an exception otherwise.

Under the safe assumption that 99.9999% of all calls to `read_dot()` pass dot files containing only one graph, the latter _"Don't break everything!"_ approach is arguably the sane approach. I like sanity, so I went with that. 

If anyone ever requires deserialization of multiple `pydot.Dot` objects from a single dot file, subsequently adding a new NetworkX function (e.g., `read_dots()`) _or_ a new optional keyword argument to the existing `read_dot()` function (e.g., `return_list=False`) would be trivial. This is left as an exercise to the reader.

### String Decoding and What It Can Do for You

In `pydot` 1.2.0,<sup>..._maybe earlier?_</sup> the `pydot.Dot.create_dot()` method now returns an encoded byte array (`bytes` in Python 3) rather than a probably-incorrectly-decoded Unicode string (`str` in Python 3).

This is a positive change, too. The prior approach masked inevitable encoding errors with a Jedi handwave: _"These are the character-like bytes you might be looking for. No further questions!"_

This does require, however, that NetworkX functionality calling `create_dot()` explicitly decode its return value into a proper Unicode string. My simplistic assumption is that GraphViz encodes this return value via the current user's preferred locale (e.g., `LC_*` environment variables on POSIX-compatible platforms), which is safely decodable by NetworkX via the same encoding. This assumption appears to hold. May the [Spaghetti Monster](https://s-media-cache-ak0.pinimg.com/564x/66/68/18/6668181a59aaf6846cfcb90f2a780e18.jpg) preserve us all.

The `pydot.Dot.create_dot()` method is only called by the `networkx.drawing.nx_pydot.pydot_layout()`,  which has been refactored to decode the returned byte array via the current `locale.getpreferredencoding()`. <sup>..._which is probably just `utf-8`, in most cases._</sup>

## We Can Build a Better PyDot Integration

The newer 1.2.0 `pydot` API is a non-trivial departure from the older pre-1.0.29 `pydot` API. Since both cannot be reasonably supported at the same time, this pull request revises all `networkx.drawing.nx_pydot` functions to explicitly require `pydot >= 1.2.0`. To enforce this constraint in a human-readable manner:

* A new private `networkx.drawing.nx_pydot._import_pydot()` function has been defined, importing and returning the `pydot` module if the currently installed version of that module is `>= 1.2.0` _or_ raising an `ImportError` exception. For [robustness](https://stackoverflow.com/a/21065570/2809027), the ubiquitous `pkg_resources.parse_version()` function  is called to implement this version comparison. Since NetworkX requires setuptools at installation time, the setuptools-bundled `pkg_resources` package is assumed to be available at runtime. This assumption is almost certainly safe for most definitions of "almost," "certainly," and "safe."
* Most `import pydotplus` statements in the codebase now read `pydot = _import_pydot()`.

## Are You Quite Done Yet?

**Nope.** Unit tests have also been improved to explicitly test the `networkx.drawing.nx_pydot.pydot_layout()` function. Because this is a safety-first pull request. For maintainability, I also obsessively documented everything in these tests. Because obsessiveness.

Thanks for the hard volunteerism, fellow NetworkX-ers! This graph paper is for you. :chart_with_upwards_trend:

### DIFF
--- a/networkx/drawing/tests/test_pydot.py
+++ b/networkx/drawing/tests/test_pydot.py
@@ -1,8 +1,7 @@
 """Unit tests for pydot drawing functions."""
-import os
+import sys
 import tempfile
-from nose import SkipTest
-from nose.tools import assert_true
+from nose.tools import assert_equal, assert_is_instance, assert_true
 import networkx as nx
 from networkx.testing import assert_graphs_equal
 
@@ -10,37 +9,83 @@ from networkx.testing import assert_graphs_equal
 class TestPydot(object):
     @classmethod
     def setupClass(cls):
-        global pydotplus
-        try:
-            import pydotplus
-        except ImportError:
-            raise SkipTest('pydotplus not available.')
+        '''
+        Fixture defining the `pydot` global to be the `pydot` module if both
+        importable and of sufficient version _or_ skipping this test.
+        '''
+        global pydot
+        pydot = nx.nx_pydot.setup_module(sys.modules[__name__])
+        assert pydot is not None
 
-    def pydot_checks(self, G):
+    def pydot_checks(self, G, prog):
+        '''
+        Validate :mod:`pydot`-based usage of the passed NetworkX graph with the
+        passed basename of an external GraphViz command (e.g., `dot`, `neato`).
+        '''
+
+        # Set the name of this graph to... "G". Failing to do so will
+        # subsequently trip an assertion expecting this name.
+        G.graph['name'] = 'G'
+
+        # Add arbitrary nodes and edges to the passed empty graph.
         G.add_edges_from([('A','B'),('A','C'),('B','C'),('A','D')])
         G.add_node('E')
+
+        # Validate layout of this graph with the passed GraphViz command.
+        graph_layout = nx.nx_pydot.pydot_layout(G, prog=prog)
+        assert_is_instance(graph_layout, dict)
+
+        # Convert this graph into a "pydot.Dot" instance.
         P = nx.nx_pydot.to_pydot(G)
+
+        # Convert this "pydot.Dot" instance back into a graph of the same type.
         G2 = G.__class__(nx.nx_pydot.from_pydot(P))
+
+        # Validate the original and resulting graphs to be the same.
         assert_graphs_equal(G, G2)
 
+        # Serialize this "pydot.Dot" instance to a temporary file in dot format.
         fname = tempfile.mktemp()
         assert_true(P.write_raw(fname))
-        Pin = pydotplus.graph_from_dot_file(fname)
 
+        # Deserialize a list of new "pydot.Dot" instances back from this file.
+        Pin_list = pydot.graph_from_dot_file(path=fname, encoding='utf-8')
+
+        # Validate this file to contain only one graph.
+        assert_equal(len(Pin_list), 1)
+
+        # The single "pydot.Dot" instance deserialized from this file.
+        Pin = Pin_list[0]
+
+        # Sorted list of all nodes in the original "pydot.Dot" instance.
         n1 = sorted([p.get_name() for p in P.get_node_list()])
+
+        # Sorted list of all nodes in the deserialized "pydot.Dot" instance.
         n2 = sorted([p.get_name() for p in Pin.get_node_list()])
-        assert_true(n1 == n2)
 
-        e1 = [(e.get_source(),e.get_destination()) for e in P.get_edge_list()]
-        e2 = [(e.get_source(),e.get_destination()) for e in Pin.get_edge_list()]
-        assert_true(sorted(e1) == sorted(e2))
+        # Validate these instances to contain the same nodes.
+        assert_equal(n1, n2)
 
+        # Sorted list of all edges in the original "pydot.Dot" instance.
+        e1 = sorted([
+            (e.get_source(), e.get_destination()) for e in P.get_edge_list()])
+
+        # Sorted list of all edges in the original "pydot.Dot" instance.
+        e2 = sorted([
+            (e.get_source(), e.get_destination()) for e in Pin.get_edge_list()])
+
+        # Validate these instances to contain the same edges.
+        assert_equal(e1, e2)
+
+        # Deserialize a new graph of the same type back from this file.
         Hin = nx.nx_pydot.read_dot(fname)
         Hin = G.__class__(Hin)
+
+        # Validate the original and resulting graphs to be the same.
         assert_graphs_equal(G, Hin)
 
     def testUndirected(self):
-        self.pydot_checks(nx.Graph())
+        self.pydot_checks(nx.Graph(), prog='neato')
 
     def testDirected(self):
-        self.pydot_checks(nx.DiGraph())
+        self.pydot_checks(nx.DiGraph(), prog='dot')
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,15 +102,10 @@ install:
       source activate test-environment;
       echo "Testing that gdal installed ok by if osgeo imports";
       python -c "from osgeo import ogr;print('importing ogr worked')";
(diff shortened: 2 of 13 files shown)


networkx/drawing/tests/test_pydot.py
```diff
@@ -1,46 +1,91 @@
...
+        # Sorted list of all edges in the original "pydot.Dot" instance.
+        e1 = sorted([
+            (e.get_source(),e.get_destination()) for e in P.get_edge_list()])
 
+        # Sorted list of all edges in the original "pydot.Dot" instance.
+        e2 = sorted([
+            (e.get_source(),e.get_destination()) for e in Pin.get_edge_list()])
+
+        # Validate these instances to contain the same edges.
+        assert_equal(e1, e2)
```
MridulS: Looks like this is not working for python2 (e2 is empty)

leycec:
**Bizarre,** but sadly typical of my life. Appveyor shows [Python 2.7 passing on Windows](https://ci.appveyor.com/project/dschult/networkx-pqott/build/1.0.633/job/1wjxj2a5u3nff0dt), suggesting a platform-specific quandry. <sup>..._encoding, maybe?_</sup>

@MridulS Thanks for the quick catch there. I'll get this patched up, rebased, and repushed in a jiffy. In the meanwhile, enjoy this festive pumpkin. :jack_o_lantern:

leycec:
Oh, I see. Appveyor _of course_ skips all `pydot` tests entirely:

    SKIP: pydot not available

It shouldn't be, however, because both `GraphViz` and `pydot` are trivially installable with [Miniconda under Windows](https://anaconda.org/anaconda/graphviz). Tragically, NetworkX's [Appveyor configuration](https://github.com/networkx/networkx/blob/master/appveyor.yml) fails to leverage Miniconda.

It probably should, arguably. I can see that I have a long night of Red Bull-fuelled productivity ahead of me.

leycec:
**Fixed it.** Upstream `pydot` contains a trivial Python 2.7-specific show-stopper. Unsurprisingly, string types were detected incorrectly. Lines 357 through 363 of the upstream `dot_parser.py` module should be refactored from:

```
    elif isinstance(toks[2][0], type('')):

        for n_next in [n for n in tuple(toks)[2::2]]:

            if (isinstance(n_next, P_AttrList) or
                    not isinstance(n_next[0], type(''))):
                continue
```

...to:

```
    elif isinstance(toks[2][0], str_type):

        for n_next in [n for n in tuple(toks)[2::2]]:

            if (isinstance(n_next, P_AttrList) or
                    not isinstance(n_next[0], str_type)):
                continue
```

I'll post an upstream pull request. With any luck, they'll hastily push out a new stable release. When they do, we'll revisit this downstream pull request.

Thanks again, all!

MridulS:
@leycec That was fast! 🍻  Thanks for debugging pydot, was scratching my head for a long time to find it!

leycec:
> :beers: Thanks for debugging pydot

Thank _you_ for the tasty clinking beers.

I caught [johnyf](https://github.com/johnyf) at a propitious time: he was mere hours away from releasing [pydot 1.2.3](https://github.com/erocarrera/pydot/commit/294b7315153fafe5076a74ab9aa164e80bb57f89). With uncanny speed reminiscent of [ keyboard-bound [Daredevil](https://images-na.ssl-images-amazon.com/images/M/MV5BNzUwMDEyMTIxM15BMl5BanBnXkFtZTgwNDU3OTYyODE@._V1_UY1200_CR89,0,630,1200_AL_.jpg), he graciously [accepted](https://github.com/erocarrera/pydot/commit/97061a8bf17d0ed205ff2c7fe92d389603d04558) my changes _and_ pushed out [1.2.3](https://pypi.python.org/pypi/pydot).

We should be good to go here, so I'll trigger a retest. Let's cross fingers on worky.

leycec:
Since pydot >= 1.2.3 resolves a long-standing Python 2.x issue required for sane NetworkX behaviour, all `nx_pydot` functionality now mandates at least pydot >= 1.2.3 rather than merely pydot >= 1.2.0. Because no one wants additional bug reports.

**All tests locally pass both Python 2.x and 3.x.** Travis, [what say you?](https://travis-ci.org/networkx/networkx/builds/165712043)

leycec:
We have triumphant lift-off. :helicopter:

dschult:
Thanks a lot!!  hopefully this will settle the pydot import trouble for a while. :)

PR #2273, opened 2016-10-06
Create bridges.py

rmsyed:
Contains three simple functions for detection and retrieval of bridges or local bridges in a undirected network.

### DIFF
--- a/networkx/algorithms/bridges.py
+++ b/networkx/algorithms/bridges.py
@@ -1,164 +1,238 @@
 import networkx as nx
 
-
 __author__ = "Rohail Syed <rohailsyed@gmail.com>"
 
 __all__ = [
-	'bridges_exist',
-	'all_bridges',
-	'all_local_bridges'
+    'bridges_exist',
+    'all_bridges',
+    'all_local_bridges'
 ]
 
 
-
 def bridges_exist(G):
-	""" Checks if any bridges exist in this network. We will simply call the all_bridges()
-	function with the first-match stop parameter. Since the search for bridges will terminate
-	after finding one instance, this function is faster if we just want to know if at least one
-	bridge exists in the network and what it is.
-	
-	Parameters
-	----------
-	G : Undirected Graph object
-	
-	Returns
-	----------
-	boolean
-		True if we found at least one bridge.
-		False otherwise.
-	
-	Notes
-	----------
-	This function can be useful to quickly determine whether or not bridges exist in a given network.
-	"""
-	
-	results = all_bridges(G, first_match = True)
-	if len(results)>0:
-		return True
-	else:
-		return False
+    """ Checks if any bridges exist in this network. We will simply call the
+    all_bridges() function with the first-match stop parameter. Since the
+    search for bridges will terminate after finding one instance, this function
+    is faster if we just want to know if at least one bridge exists in the
+    network and what it is.
 
-		
-		
+    Parameters
+    ----------
+    G : Undirected Graph object
 
+    Returns
+    ----------
+    boolean
+        True if we found at least one bridge.
+        False otherwise.
 
+    Examples
+    --------
+    >>> G = nx.cycle_graph(5)
+    >>> print (bridges_exist(G))
+    False
+    >>> G.remove_edge(1,2)
+    >>> print (bridges_exist(G))
+    True
+
+    Notes
+    ----------
+    This function can be useful to quickly determine whether or not bridges
+    exist in a given network.
+    """
+
+    results = all_bridges(G)
+    if len(results) > 0:
+        return True
+    else:
+        return False
 
 
 def all_local_bridges(G, first_match=False):
-	""" Looks through the graph object G for all local bridges.
-	We formally define a local bridge to be any edge AB such that the removal
-	of the edge results in a distance strictly greater than 2 between nodes A and B.
-	Note that all bridges are local bridges.
-	
-	Parameters
-	----------
-	G : Undirected Graph object
-	first_match : boolean
-		Tells us if we should only look for at least one instance of a local bridge (True) or look for
-		all (False).
-	
-	Returns
-	----------
-	dict
-		edges : List of edges that are local bridges
-		spans : List of corresponding spans of the ith local bridge
-	
-	Notes
-	----------
-	This function can be useful to quickly determine what local bridges exist in a given network
-	and what their spans are.
-	"""
-	
-	# For each node A, remove the edge it has with each of its neighbors and determine
-	# A's new shortest path to that node. If none exists, we have a bridge (span=-1 represents infinite)
-	# Otherwise record the finite and strictly greater than 2 span of the local bridge.
-	nodeset = G.nodes()
-	allBridges = []
-	allSpans = []
-	usedNodes = []
-	for startNode in nodeset:
-		neighbors = G.neighbors(startNode)
-		for endNode in neighbors:
-			if endNode in usedNodes:
-				# this edge has already been checked. move on
-				continue
-			if [startNode,endNode] in allBridges:
-				# this edge is already known to be a local bridge
-				continue
-			G.remove_edge(startNode, endNode)
-			try:
-				pathLength = nx.shortest_path_length(G, startNode, endNode)
-				G.add_edge(startNode, endNode)
-				if pathLength>2:
-					# found a local bridge
-					allBridges.append([startNode, endNode])
-					allSpans.append(pathLength)
-					if first_match:
-						return {"edges":allBridges, "spans":allSpans}
-						
-			except nx.NetworkXNoPath:
-				# found a bridge
-				allBridges.append([startNode, endNode])
-				allSpans.append(-1)
-				G.add_edge(startNode, endNode)
-				if first_match:
-					return {"edges":allBridges, "spans":allSpans}
-		usedNodes.append(startNode)
-		
-	return {"edges":allBridges, "spans":allSpans}
+    """ Looks through the graph object G for all local bridges.
+    We formally define a local bridge to be any edge AB such that the removal
+    of the edge results in a distance strictly greater than 2 between
+    nodes A and B. Note that all bridges are local bridges.
+
+    Parameters
+    ----------
+    G : Undirected Graph object
+    first_match : boolean
+        Tells us if we should only look for at least one instance of a local
+        bridge (True) or look for all (False).
+
+    Returns
+    ----------
+    dict
+        edges : List of edges that are local bridges
+        spans : List of corresponding spans of the ith local bridge
+
+    Examples
+    --------
+    >>> G=nx.cycle_graph(5)
+    >>> bridges_dict = all_local_bridges(G)
+    >>> print ("Edges: ",bridges_dict["edges"])
+    >>> print ("Spans: ",bridges_dict["spans"])
+    ('Edges: ', [[0, 1], [0, 4], [1, 2], [2, 3], [3, 4]])
+    ('Spans: ', [4, 4, 4, 4, 4])
+
+    Notes
+    ----------
+    This function can be useful to quickly determine what local bridges exist
+    in a given network and what their spans are. The function finds the local
+    bridges as follows. For each node A in G, capture the list of all
+    neighbors it has. For each of these neighbors, remove its edge with A
+    and determine A's new shortest path to that node. If none exists, we have
+    a bridge(which we represent as having span=-1). Otherwise, if the span
+    is strictly greater than 2, that edge is a local bridge. Repeat this
+    process for all edges.
+    """
+
+    nodeset = G.nodes()
+    allBridges = []
+    allSpans = []
+    usedNodes = []
+    for startNode in nodeset:
+        neighbors = G.neighbors(startNode)
+        for endNode in neighbors:
+            if endNode in usedNodes:
+                # this edge has already been checked. move on
+                continue
+            if [startNode, endNode] in allBridges:
+                # this edge is already known to be a local bridge
+                continue
+            G.remove_edge(startNode, endNode)
+            try:
+                pathLength = nx.shortest_path_length(G, startNode, endNode)
+                G.add_edge(startNode, endNode)
+                if pathLength > 2:
+                    # found a local bridge
+                    allBridges.append([startNode, endNode])
+                    allSpans.append(pathLength)
+                    if first_match:
+                        return {"edges": allBridges, "spans": allSpans}
+
+            except nx.NetworkXNoPath:
+                # found a bridge
+                allBridges.append([startNode, endNode])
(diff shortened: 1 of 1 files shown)


MridulS:
Thanks for this!
Could you please add tests and reformat your code to PEP8 specs.

jfinkels:
@rmsyed Thanks for your contribution! Before you spend the time to reformat your code, I'd like to propose implementing the bridge-finding algorithm given in the Wikipedia article, https://en.wikipedia.org/wiki/Bridge_(graph_theory), which is linear time as opposed to the cubic time of the proposed implementation here. Are interested or able to do that?

rmsyed:
Hi guys,

Okay, so I've updated the bridges.py file to implement Tarjan's algorithm and I've also updated it to include examples for each function. I ran the pep8 check on the code and it fully validated.

The problem is that when I commit it, the Travis CI build gives an error for the bridges_exist() function's examples. I'm not sure why considering that it works perfectly fine when I run it locally and considering that all the other examples in the code seem to work. Does anyone have an idea why this is happening?

Thanks.

aryamccarthy:
When you run your tests, make sure you're doing `nosetests --with-doctest`. Also, it looks like your output is different between Python 2 and Python 3.

aryamccarthy:
This is a really useful feature that you've implemented. I notice that `all_local_bridges()` is iterating over the edges, so it's cleaner to do that like this:

```python
    bridges = {}
    for e in G.edges():
        G.remove_edge(*e)
        try:
            (u, v) = e           
            path_length = nx.shortest_path_length(G, u, v)
        except nx.NetworkXNoPath:
            bridges[e] = -1  # found a bridge
            if first_match:
                return bridges
        else:
            if path_length > 2:  # found a local bridge
                bridges[e] = path_length
                if first_match:
                    return bridges
        finally:
            G.add_edge(*e)
    return bridges
```

aryamccarthy:
You can also avoid global variables if you alter `all_bridges()` to use an inner function:

```python
def bridges(G):
    """ Looks through the graph object `G` for all bridges.

    We formally define a bridge to be any edge `(u, v)` such that the removal
    of the edge increases the total number of connected components.

    Parameters
    ----------
    G : Undirected Graph object

    Returns
    ----------
    iterable
        Edges that are bridges

    Examples
    --------
    >>> G = nx.cycle_graph(5)
    >>> bridges = bridges(G)
    >>> list(bridges)
    []
    >>> G.remove_edge(0,1)
    >>> list(bridges(G))
    [(2, 1), (3, 2), (4, 3), (0, 4)]

    Notes
    ----------
    This function can be useful to quickly determine what bridges exist
    in a given network. We use an implemenation of Tarjan's Bridge-finding
    algorithm to do this.

    The algorithm is described in [1]_.

    References 
    ----------
    .. [1] R. Endre Tarjan, 
        "A note on finding the bridges of a graph", *Information Processing 
        Letters*, **2** (6): 160--161

    """

    visited = set()
    depths = {}
    low = {}
    parent = {node: None for node in G}

    def bridge_util(u, depth):
        visited.add(u)
        depth += 1
        depths[u] = low[u] = depth
        for v in G[u]:
            if v not in visited:
                parent[v] = u
                for e in bridge_util(v, depth):
                    yield e
                # Check if subtree rooted at v has connection
                # to an ancestor of u.
                low[u] = min(low[u], low[v])
                # If the lowest vertex reachable from the subtree under v
                # is below u in the DFS tree, then `u-v` is a bridge.
                if low[v] > depths[u]:
                    yield (u, v)
            elif v != parent[u]:
                low[u] = min(low[u], depths[v])

    for u in G:
        if u not in visited:
            for e in bridge_util(u, depth=0):
                yield e
```

This version is a generator, so it would make your `bridges_exist()` function even faster.

aryamccarthy:
Hi @rmsyed—looking through the Travis logs, your problem is here:
```
      File "/home/travis/virtualenv/pypy3-2.4.0/site-packages/networkx/algorithms/bridges.py", line 217, in all_bridges
              startnode = allnodes[0]
    TypeError: 'generator' object is not subscriptable
```
The comment I posted above circumvents that, after which your build should pass.

rmsyed:
Thanks Arya! Those look like much cleaner implementations and likely more efficient. I made some small tweaks to the code that would ensure that the bridge-finding algorithm ran on all connected components in the network (before it was only running on one). However, it still seems that every time I update and commit the code, the Travis build fails on the same issue.

I removed the print statements as suggested and ran the doctest locally and, at least for Python 2.7, it didn't give any issues (I've attached the output here). So I'm still not sure why its giving errors on the github build.
[doctest_trace.txt](https://github.com/networkx/networkx/files/517339/doctest_trace.txt)

aryamccarthy:
Hey @rmsyed—Travis is comparing against the dev version of networkx, but I bet that when you run `import networkx as nx`, it gets your local version of networkx, which is (probably) the older 1.11.

A lot of the main library functions are now generators. One example is `G.nodes()`. Now that it's a generator, you can duplicate the functionality you want with `next(allnodes)`, which gets the next (in this case, first) item from the generator. There's actually a function in `nx.utils` called `arbitrary_element`, which uses that pattern while making your intent a bit clearer.

Of course, you can sidestep that whole process with the algorithm posted above.

aryamccarthy:
One more thought: since `all_local_bridges()` may return just the first match, I might rename it `local_bridges(). For symmetry, I'd rename `all_bridges()` to `bridges()`. Again, I'm really impressed that you tackled something missing from the codebase. It's always scary to venture forth in new directions.

networkx/algorithms/bridges.py
```diff
@@ -68,20 +70,33 @@ def all_local_bridges(G, first_match=False):
                 G.add_edge(startNode, endNode)
                 if pathLength > 2:
                     # found a local bridge
-                    allBridges.append([startNode, endNode])
-                    allSpans.append(pathLength)
+                    bridgeObjs.append(
+                        {"edge": [startNode, endNode], "span": pathLength})
```
aryamccarthy: Instead of making a list of dictionaries like this, why not have a dictionary of edge -> span? 
`bridges[(start_node, end_node)] = path_length`

The spans aren't meaningful without their corresponding edges, and this lets people who just need the edges to say `for edge in all_local_edges()`

networkx/algorithms/bridges.py
```diff
@@ -68,20 +70,33 @@ def all_local_bridges(G, first_match=False):
...
+                bridgeObjs.append({"edge": [startNode, endNode], "span": -1})
                 G.add_edge(startNode, endNode)
                 if first_match:
+                    allBridges.append(sorted([startNode, endNode]))
+                    allSpans.append(-1)
                     return {"edges": allBridges, "spans": allSpans}
         usedNodes.append(startNode)
 
+    # sort the list in ascending order (for consistency in output)
+    for bridge in bridgeObjs:
```
aryamccarthy: The output will be consistent from run to run either way, so let the client sort this if they want.

networkx/algorithms/bridges.py
```diff
@@ -202,15 +217,58 @@ def all_bridges(G):
...
+    results = all_local_bridges(G, first_match=True)
+    if len(results["spans"]) > 0:
+        return True
+    else:
+        return False
+
+
 def bridges_exist(G):
     """ Checks if any bridges exist in this network. We will simply call the
-    all_bridges() function with the first-match stop parameter. Since the
```
aryamccarthy: Don't forget that you can use Markdown syntax in your docstrings, which makes the generated docs look really nice. As an example: \`all_bridges()\` (with back-ticks) becomes `all_bridges()`.

networkx/algorithms/bridges.py
```diff
@@ -68,20 +70,33 @@ def all_local_bridges(G, first_match=False):
                 G.add_edge(startNode, endNode)
                 if pathLength > 2:
                     # found a local bridge
-                    allBridges.append([startNode, endNode])
-                    allSpans.append(pathLength)
```
aryamccarthy: Could you formally define spans in your docstring?

networkx/algorithms/bridges.py
rmsyed: I did the sort because my original code sometimes gave different orderings of edges and for consistency with the doctest in Travis CI, I thought it would be necessary to make the order of output edges exactly match. Running your code suggestion for at least the local bridges function seems to have solved that issue as well. So yes, the ordering isn't needed anymore.

networkx/algorithms/bridges.py
```diff
@@ -295,8 +213,8 @@ def bridges_exist(G):
     exist in a given network.
     """
 
-    results = all_bridges(G)
```
aryamccarthy: Hey, your latest build looks like it's going to pass!

One last thing from me: you can leverage the fact that `bridges()` is a generator to make this more efficient. Right now, your code constructs the entire list. You can actually dodge that—a lot like how you short-circuit `local_bridges()` using `first_match`. Here's how I'd do it:

```python
    try:
        next(bridges(G))
    except StopIteration:
        return False
    return True
```
If you do something like this, be sure to update the docstring accordingly. Does this approach make sense?

aryamccarthy:
Looks like this isn't the first PR of Tarjan's bridge-finding algorithm. See the discussion at #903.

rmsyed:
Oh, good catch! Was that code ever implemented in a public release of networkx? I couldn't find a bridge function on the networkx public documentation for a `bridges()` function. If it has been implemented, perhaps it would be good to close this pull request or maybe consider narrowing the scope to only the local bridges function.

jfinkels:
Can you adjust your pull request to create a `bridges()` function that is a wrapper around the existing code that's already in NetworkX? As discussed on the linked pull request above, see the `networkx/algorithms/components/biconnected.py` module. Note that module is for determining 2-node-connectivity, not 2-edge-connectivity.

rmsyed:
Ok, I've modified the code for the `bridges()` function to use the `biconnected_components()` function to find all bridges.

Edit: I just realized the notes in the latest commit were the notes from an earlier commit. I'll update those comments accordingly, but first I want to make sure the code is suitable and appropriate.

jfinkels:
The `local_bridges` function is needlessly complicated. From my understanding based on the definition you've provided for "local bridge", an edge is a "local bridge" if its two incident nodes do not share a common neighbor. So I think you could do this (assuming a graph with no self-loops):

```python
def local_bridges(G):
    for u, v in G.edges():
        if not (set(G[u]) & set(G[v])):
            yield u, v
```
Is my understanding correct?

rmsyed:
Yes, the definition is right and your code does detect local bridges but it doesn't return the span of those bridges which is often important to know. The original function was doing both but I suppose we could use your code, modified slightly to also compute the span. I do agree that your version is simpler although I think the worst case time complexity is the same once we include the code for computing the spans. I'll modify the code accordingly

jfinkels:
Oh, I see, you are also returning the distance between the two endpoints when the edge is removed. Well how about something like
```python
def local_bridges(G):
    for u, v in G.edges():
        if not (set(G[u]) & set(G[v])):
            H = G.copy()
            H.remove_edge(u, v)
            try:
                d = nx.shortest_path_length(H, u, v)
            except NetworkXNoPath:
                d = float('inf')
            yield u, v, d
```
If you need some kind of dictionary, you can do
```python
distances = defaultdict(dict)
for u, v, d in local_bridges(G):
    distances[u][v] = distances[v][u] = d
```

It would be nice if there were a faster algorithm for this. Do you have any references for these algorithms, or a reference for local bridges, with which I am not familiar?

rmsyed:
Your update looks good although I think it might be more efficient to simply remove the edge, check the distance, and then add the edge back rather than creating an entire copy each time. Also I think simply returning -1 for infinite distance might be suitable since span is always an integer. So this might look something like:

```
def local_bridges(G):
    for u, v in G.edges():
        if not (set(G[u]) & set(G[v])):
            G.remove_edge(u, v)
            try:
                d = nx.shortest_path_length(G, u, v)
                G.add_edge(u,v)
            except NetworkXNoPath:
                d = -1
                G.add_edge(u,v)
            yield u, v, d
```

As far as I'm aware, the algorithm for finding all local bridges is linear in O(V+E) and the algorithm for finding distance from some node `u` to `v` is linear in O(V+E) so the algorithm will be quadratic time. It would be interesting if there was a faster solution.

The concept of local bridges was first (I think) introduced by Mark Granovetter in his paper "The Strength of Weak Ties" (1973) on pages 1364-1365. He uses the term "degree" instead of span but Easley and Kleinberg in their book "Networks, Crowds and Markets" (2010) use the term "span" (page 52). I prefer to call it span because degree is an ambiguous term which could also refer to simply the number of edges on a node.

You can also find a quick reference to local bridges from Wikipedia: https://en.wikipedia.org/wiki/Bridge_(interpersonal)#Local_bridge

jfinkels:
The reason I suggested using a copy is so that we do not have to modify the
user's graph directly. This is something of a convention in NetworkX
functions (at least, that's what it seems to me). This is both slow and
memory-intensive, yes, but it doesn't perform any unexpected modification
of the user's data, and a less memory-hungry version can be implemented
directly by the user if needed.

As for the running time, it is quadratic if the graph is sparse, but if the
graph is dense, the running time is quartic (!), which is particularly
unpleasant for large graphs.

aryamccarthy:
You can increase the efficiency by only copying the user's graph once, outside of the loop over edges:

```python
def local_bridges(G):
    H = G.copy(with_data=False)
    for u, v in G.edges():
        ...
```

Then make sure to re-add the edge after removing it.

dschult:
You can avoid copying the graph and also avoid removing the edge by "hiding" the edge using a weight function in the shortest path routine.  The weight function would be something like:

    enodes=set((u,v))
    weight = lambda n, nbr, d: 1 if (n not in enodes or nbr not in enodes) else None

Then replace ``shortest_path_length(G,u,v)`` with ``shortest_path_length(G,u,v,weight)``

networkx/algorithms/bridges.py
```diff
@@ -0,0 +1,179 @@
...
+    Notes
+    ----------
+    This function can be useful to quickly determine what bridges exist
+    in a given network. We first construct the biconnected components
+    of the graph. Then, since any bridge must exist only in a two-node
+    biconnected component, we simply iterate and output the two-node
+    sets. This iteration is a linear time operation.
+    """
+
+    biconnects = nx.biconnected_components(G)
```
jfinkels: This should be
```python
for component in nx.biconnected_components(G):
    if len(component) == 2:
        yield tuple(component)
```
without the initial blank line in the body of the function.

Why? I use `component` as the variable of iteration over each biconnected component instead of `bridge_pair`, since the variable does not necessarily hold set of cardinality two on each iteration. I use `yield tuple(component)` because we already know the component is a set of cardinality two, so expressed as a tuple, the component is itself an edge.

Issue #2270, opened 2016-09-30
Add approximation algorithm for feedback vertex set


jfinkels:
The _minimum feedback vertex set problem_ is the problem of finding the smallest subset of vertices that, when removed, cause a graph to become acyclic. The problem is NP-complete in general, but there is [a 2-approximation algorithm](http://dx.doi.org/10.1137/S0895480196305124).

JamesClough:
I've had an attempt at this in #2279

babara2:
I was wondering if any progress has been made on this project. If not, I would like to give this a try and see what I can come up with.

dschult:
Look at #2316 for the current status. It's been almost a year since the last activity so you should give it a try.

PR #2279, opened 2016-10-11
Add approximation for minimal feedback vertex set

JamesClough:
First attempt at adding an approximation for finding a minimal feedback vertex set. Solves issue #2270 
Review and feedback very welcome.

### DIFF
--- a/doc/source/reference/algorithms.approximation.rst
+++ b/doc/source/reference/algorithms.approximation.rst
@@ -52,6 +52,14 @@ Dominating Set
    min_weighted_dominating_set
    min_edge_dominating_set
 
+Feedback Vertex Set
+-------------------
+.. automodule:: networkx.algorithms.approximation.feedback_vertex_set
+.. autosummary::
+   :toctree: generated/
+
+   feedback_vertex_set
+
 
 Independent Set
 ---------------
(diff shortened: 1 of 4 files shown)


jfinkels:
I think you are using the `cycle_basis` function as if it returned a list of all cycles in the graph, but that is not the correct usage of the function (it returns the "cycle basis" of the graph, a minimum set of cycles from which all other cycles can be constructed via "exclusive or").

JamesClough:
@jfinkels Yes you are correct. Do you have any ideas on the best approach to check for semidisjoint cycles? The ```simple_cycles``` method is only for directed graphs.
It does not matter what order we find them in so perhaps it is best to start with the cycles in the cycle basis, and once they are all checked then go through their exclusive-or combinations? That should hopefully minimises the number of times we have to go through that slow process.

jfinkels:
I'm about to create a pull request that has my attempt at an implementation of this. It is not quite working yet, but the code matches the paper's pseudocode fairly closely, so a fresh set of eyes should be able to determine where my code is incorrect. I'll cc you @JamesClough so you can take a look at it.

hagberg:
See #2316

PR #2280, opened 2016-10-11
MAYBE INCORRECT: fix links

thecapacity:
Sorry I'm not familiar with the docutils syntax but the links to the PDF's and downloads are broken / inactive when I view them from the main site: https://networkx.readthedocs.io/en/stable/

They do work from this page; https://github.com/networkx/networkx/blob/v1.11/doc/source/download.rst

However, that syntax is different / maybe-auto-generated ...

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -8,11 +8,11 @@ NetworkX documentation
     :Release: |version|
     :Date: |today|
 
-    Tutorial :download:`[PDF]  <networkx_tutorial.pdf>`
+    Tutorial :download:`[PDF]  _<networkx_tutorial.pdf>`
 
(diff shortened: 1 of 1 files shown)


jfinkels:
Hi @thecapacity, thanks for the report. The reStructured Text format cheat sheet is here: http://docutils.sourceforge.net/docs/user/rst/quickref.html, specifically, you want the section on [hyperlink targets][1].

However, it looks like those download links have been removed in the latest version of the documentation, https://networkx.readthedocs.io/en/latest/, so this is no longer an issue.

[1]: http://docutils.sourceforge.net/docs/user/rst/quickref.html#hyperlink-targets

PR #2284, opened 2016-10-13
Adds chain decomposition algorithm.

jfinkels:
The chain decomposition is essentially the non-overlapping initial segments of the fundamental cycles of a DFS search tree, when cycles are read as a list of edges starting from the nontree edge. This function induces a very simple bridge-finding algorithm (every edge not in a chain is a bridge), which I will propose later if this function is satisfactory.

### DIFF
--- a/networkx/algorithms/tests/test_chains.py
+++ b/networkx/algorithms/tests/test_chains.py
@@ -1,0 +1,130 @@
+# test_chains.py - unit tests for the chains module
+#
+# Copyright 2004-2016 NetworkX developers.
+#
+# This file is part of NetworkX.
+#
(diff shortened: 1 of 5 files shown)


networkx/algorithms/tests/test_chains.py
```diff
@@ -0,0 +1,127 @@
...
+    """Yields cyclic permutations of the given sequence.
+
+    For example::
+
+        >>> list(cycles('abc'))
+        [('a', 'b', 'c'), ('b', 'c', 'a'), ('c', 'a', 'b')]
+
+    """
+    n = len(seq)
+    yield from (tuple(islice(cycle(seq), i, i + n)) for i in range(n))
```
jfinkels: Good catch, thanks.

PR #1700, opened 2015-08-03
use dict comprehensions instead of dict()

SanketDG:
#1640

PR #1435, opened 2015-03-26
Pep8

gcetusic:
For programmers that use linter tools in their editors, code that  doesn't conform to PEP8 (https://www.python.org/dev/peps/pep-0008/) tends to become somewhat obtrusive. A perfect example is SublimeLinter. In any case, Python code should conform to PEP8 and NetworkX lacks this in some parts of its code. This PR more or less fixes this in base classes (classes folder). Any changes to other parts of the code should also incorporate pep8 style fixes around the area that's being edited.

PR #1355, opened 2015-02-16
Forces PEP8 style recommendations on Python files

jfinkels:
This branch was created using the [autopep8][1] tool. [PEP8][2] compliance is important because PEP8 is an almost-universal style convention, and universality implies predictability. Predictability is as important in coding style as it is in rules of the road: it prevents confusion and therefore prevents "dangerous" behavior. In this environment, that mostly just means improved readability.

[1]: https://pypi.python.org/pypi/autopep8/
[2]: https://www.python.org/dev/peps/pep-0008/

PR #2293, opened 2016-10-21
Unused variables in list comprehensions replaced by a wildcard

atombrella:
Variables initialized and immediately reassigned

Unused imports

PyCharm generates quite a lot of warnings with the introspection. This pull requests addresses a tiny fraction of them to determine. Would it be feasible to include something like `flake8` in the CI test suites?

### DIFF
--- a/doc/gh-pages.py
+++ b/doc/gh-pages.py
@@ -89,7 +89,7 @@ if __name__ == '__main__':
         except CalledProcessError:
             print("using development as label")
             tag = "development"   # Fallback
-    
+
     startdir = os.getcwdu()
     if not os.path.exists(pages_dir):
         # init the repo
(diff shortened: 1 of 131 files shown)


aryamccarthy:
Raymond Hettinger (the `itertools` guy) has a [talk](https://www.youtube.com/watch?v=wf-BqAjZb8M) about PEP8 for big projects, and his golden rule of PEP8 is "PEP8 unto thyself, not unto others."

By removing whitespace, you introduced a failed test. I'd be interested to see whether you've uncovered a bug in the original implementation, if you're willing to do that digging.

atombrella:
Yes. However, my point was getting rid of unused imports and unused variables, which is what ``flake8`` is good at.

I'm digging a bit, removing trailing whitespace shouldn't break tests ...

jfinkels:
Although we appreciate the effort, please see my comment about a similar change https://github.com/networkx/networkx/pull/1700#issuecomment-255016723 and the comments on pull request #1355 and #1435. The general rule being followed currently in NetworkX is to make these changes only when nearby code is being changed.

PR #2297, opened 2016-11-01
ENH: Added feature delete_edge_attributes and delete_node_attributes

Erotemic:
In one of my projects I have to maintain a graph with lots of dynamic state. Its often convenient to be able to remove attributes entirely from a graph instead of having set the attribute to None. It makes debugging much easier as extra keys don't get in the way if they've been deleted.

I initially made an internal helper function to get this functionality following the discussion in https://groups.google.com/forum/#!topic/networkx-discuss/S-rIdrNNrUY 

It seems like this may be something that the official library would want to support, hence the pull request. 

The added functions work similarly to get_*_attributes and set_*_attributes. Except they are not delete_*_attributes. But perhaps they may be better named as del_*_attributes?

Also, there is a question of implementation in delete_edge_attributes. The implementation in my commit differentiates between the multigraph and non-multigraph inputs. However, the following code should be universal to both:

```python
    for edge in edges:
        d = G
        for p in edge[:-1]:
            d = d[p]
        del d[edge[-1]]
```

However, it seems a bit more obfuscated, which is why I went with the more explicit implementation. 

Any comments? If this feature is wanted I can make any requested changes and add the appropriate tests. If not, then feel free to close the PR.

### DIFF
--- a/networkx/classes/function.py
+++ b/networkx/classes/function.py
@@ -28,9 +28,9 @@ __all__ = ['nodes', 'edges', 'degree', 'degree_histogram', 'neighbors',
            'is_directed', 'info', 'freeze', 'is_frozen', 'subgraph',
            'add_star', 'add_path', 'add_cycle',
            'create_empty_copy', 'set_node_attributes',
-           'get_node_attributes', 'set_edge_attributes',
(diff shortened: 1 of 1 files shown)


Erotemic:
Closing due to lack of attention. Might look into again if I need this functionality in the future.

Issue #2295, opened 2016-11-01
write_graph6 uses excessive amounts of memory


basimr:
I am working with a graph of roughly 650,000 edges, with each vertex being a short string of <100 characters. I found that calling the write_graph6() function to write my graph to disk uses too much memory and takes significantly longer than the most other formats (I've tried adjlist, gexf, gml, pickle, YAML, and more).

In my tests, write_graph6() uses gigabytes of memory and over two minutes to write a 20,000 edge graph to disk, other methods take significantly less memory and mere seconds for my original 650,000 edge graph.

More specifically, for 10,000 edges the function uses ~5GB and with 20,000 edges it uses ~20GB of memory. When I try to use my 650,000 edge graph, my system (with 32GB of memory) runs out of memory and starts thrashing.

I've attached the 20,000 edge graph as an adjlist (delimited by two pipes: ||) and YAML.
[graph.adjlist.zip](https://github.com/networkx/networkx/files/563135/graph.adjlist.zip)
[graph.yaml.zip](https://github.com/networkx/networkx/files/563136/graph.yaml.zip)

jfinkels:
Hi @basimr, thanks for the report. It looks like with a bit of review, the function can be rewritten to reduce its memory usage. Which version of Python are you using?

basimr:
I'm using Python 3.5.2 and NetworkX 1.11-1.

jfinkels:
I am working on a fix for this. The function seems to create the contents of the graph6 file as a string in memory before writing it to a file. Rewriting it so it makes better use of iterators and writes each chunk of data directly to the file should help.

hagberg:
It looks like your graph.yaml.zip example has 20,000 edges and only 19967 nodes.  That is very sparse. 

The fact that your nodes are strings is not relevant because a label information is discarded with these formats. 

If you want to use a format like graph6 I suggest using sparse6 which will easily be able to handle large sparse graphs quickly and without so much memory.

```python
In [21]: G = nx.gnm_random_graph(100000,200000)

In [22]: %timeit len(nx.generate_sparse6(G))
1 loops, best of 3: 1.96 s per loop
```

Issue #643, opened 2012-07-20
Restructure names of readwrite modules (migrated from Trac #655)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/655
Reported 2011-11-28 by @hagberg, assigned to @hagberg.

Consider changing names of graph readers and writers to standardize in modules.
Change names from, e.g. 

``` python
read_graphml()
write_graphml()
```

to 

``` python
graphml.read()
graphml.write()
```

Another benefit might be that all of the modules for reading and writing various formats wouldn't need to be automatically imported.  That is, we could import modules when needed and access them like

``` python
>>> from networkx import graphml
>>> G = graphml.read(path)
```

hagberg:
This is really just syntax and can wait for future release.

hagberg:
Closed due to inactivity.

PR #2299, opened 2016-11-02
Makes write_graph6 less memory-intensive.

jfinkels:
This commit changes the implementation of the `write_graph6` function so
that it makes better use of iterators and writes directly to file
instead of buffering the entire encoding of the graph as a string in
memory before writing. This commit also removes the `generate_graph6`
function, since it's functionality can be simulated by using the
`write_graph6` function to write to a file.

Fixes issue #2295.

### DIFF
--- a/doc/source/reference/readwrite.sparsegraph6.rst
+++ b/doc/source/reference/readwrite.sparsegraph6.rst
@@ -23,9 +23,9 @@ Graph6
 .. autosummary::
    :toctree: generated/
 
-   parse_graph6
+   from_graph6_bytes
    read_graph6
-   generate_graph6
+   to_graph6_bytes
    write_graph6
 
 Sparse6
@@ -34,8 +34,7 @@ Sparse6
 .. autosummary::
    :toctree: generated/
 
-   parse_sparse6
+   from_sparse6_bytes
    read_sparse6
-   generate_sparse6
+   to_sparse6_bytes
    write_sparse6
-
--- a/networkx/readwrite/graph6.py
+++ b/networkx/readwrite/graph6.py
@@ -20,20 +20,64 @@ For more information, see the `graph6`_ homepage.
 .. _graph6: http://users.cecs.anu.edu.au/~bdm/data/formats.html
 
 """
+from itertools import islice
+import sys
+
 import networkx as nx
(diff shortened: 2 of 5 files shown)


jfinkels:
Okay in order to do things the right way, these changes are now more substantial than I originally intended. This makes read/write graph6/sparse6 work only with `bytes` objects (which is `str` on Python 2). This is the correct way to do this because the graph6 encoding is a sequence of raw bytes (no Unicode necessary).

This also removes the `generate_graph6` function and some helper functions because they were unnecessary abstractions.

As for the Appveyor "Permission Denied" error when trying to write to a temporary file, I have no idea how to fix that, as I am unfamiliar with Microsoft WIndows.

jfinkels:
Okay, I can see wanting a `to_string` function. How about having a standard IO interface for writing to a string versus a file, something like `(to/from)_*_string` and `(to/from)_*_file` , where * is the format name (in this case `graph6`).

jfinkels:
As for the `bytes` issue, should I leave the file writing in "text" mode in this pull request? I do think this should be in bytes, but I can open a different issue for that.

hagberg:
I agree it should be bytes.  Doesn't matter to me if it is the same or different PR.

A long time ago (#643) I was proposing restructuring the read/write system.  I don't remember exactly why I was suggesting to do it that way - is that a good idea?  Your naming idea make me think of that old issue.

jfinkels:
I don't have very strong feelings about naming things like `graphml.read` versus `read_graphml`. I do think that if you are using `read` and `write`, then these should corresponding to reading from and writing to a file-like object (to match Python's `io` module), not from/to a string. That's why I propose `to_string`/`from_string`. How about in this pull request, I rework things so that there are functions `to_graph6_string`, `from_graph6_string`, `read_graph6`, `write_graph6`?

hagberg:
That sounds like a good plan.  It would be easy to implement a spelling change of e.g. to_graph6_string as graph6.to_string if we later decided that was better.

jfinkels:
I have found the source of the "Permission denied" issue for my test: https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file

I'll rework this so that I can have the unit test without this issue.

PR #2306, opened 2016-11-08
Adds linear-time bridge-finding algorithm.

jfinkels:
This bridge-finding algorithm is implemented using the
`chain_decomposition` function.

This is an alternate implementation to the approach given in pull request #2273. The implementation there is basically
```python
for component in nx.biconnected_components(G):
    if len(component) == 2:
        yield tuple(component)
```
I find this implementation based on chain decomposition a bit less mysterious.

### DIFF
--- a/doc/source/reference/algorithms.bridges.rst
+++ b/doc/source/reference/algorithms.bridges.rst
@@ -1,0 +1,9 @@
+Bridges
+=======
+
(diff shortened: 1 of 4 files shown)


jfinkels:
The test failure is unrelated and addressed by pull request #2329.

PR #2309, opened 2016-11-11
flip source target

scott-vsi:
Are the source and target flipped in this error message?


scott-vsi:
Oh yeah. good point. I'm working with a directed graph. The same change should be made to astar.py:astar_path().

PR #2316, opened 2016-11-22
Approximate minimum feedback vertex set

jfinkels:
Improved implementation of pull request #2279; fixes issue #2270.

Currently does not work but I'm not sure why. It's kind of a pain to debug. Hopefully someone can figure it out.


jfinkels:
cc @JamesClough

networkx/algorithms/approximation/fvs.py
```diff
@@ -0,0 +1,222 @@
...
+    move_and_cleanup(G, F, weights)
+    while G:
+        # Try to find a semidisjoint cycle in the graph.
+        try:
+            C = next(semidisjoint_cycles(G))
+        # If there are no semidisjoint cycles, update the degree-scaled
+        # weights of all nodes.
+        except StopIteration:
+            gamma = min(weights[v] / (d - 1) for v, d in G.degree())
+            weights -= {v: gamma * (d - 1) for v, d in G.degree()}
```
JamesClough: Tried running this to see what the problem was - my problem (running Python 2.7 - don't know if this is different in other versions) was this line where we're trying to subtract a dictionary from a Counter. Was this dictionary meant to be a Counter as well?

networkx/algorithms/approximation/fvs.py
```diff
@@ -0,0 +1,222 @@
...
+        # F. Clean up the graph by removing any dangling tails.
+        move_and_cleanup(G, F, weights, stack=stack)
+    # Remove extraneous nodes from the feedback vertex set. A node v is
+    # extraneous to the feedback vertex set F if F - {v} remains a
+    # feedback vertex set in the original graph. This not only ensures a
+    # minimal feedback vertex set but is also required to achieve the
+    # approximation guarantee.
+    while stack:
+        v = stack.pop()
+        if nx.is_forest(G_original.subgraph(F - {v})):
```
JamesClough: Here we're checking whether the FVS minus vertex v is a forest - but shouldn't we actually be checking whether the original graph minus (FVS - v) is a forest?

jfinkels:
@JamesClough You are correct in both of your comments, I'll make those changes and try again.

networkx/algorithms/approximation/fvs.py
jfinkels: This actually works on Python 3, but not on Python 2 it seems.

PR #2319, opened 2016-11-27
Fix test ordering issues for python3.6

hagberg:
The new dictionary implementation in Python3.6 exposes some test issues.  In test where ordering of the dictionaries is assumed the tests may/will fail.  The types of issues most common are

- Nonunique answer (e.g. shortest path on a grid)
- Need to use assert_edges_equal or assert_nodes_equal to handle ordering differences

- Test isomorphic istead of equal, e.g. we can't guarantee roundrip of networkx->scipy sparse -> networkx with the same labels with our current setup.   In order to do that we'd need to store networkx labels with the scipy sparse data.

Some test items were commented out and need inspection

 - non-unique chain test
 - kcomponents test depends on order

### DIFF
--- a/networkx/algorithms/approximation/tests/test_kcomponents.py
+++ b/networkx/algorithms/approximation/tests/test_kcomponents.py
@@ -147,36 +147,23 @@ def test_karate_1():
     assert_equal(karate_k_num, k_num)
 
 def test_example_1_detail_3_and_4():
-    solution = {
-        3: [set([40, 41, 42, 43, 39]),
-            set([32, 33, 34, 35, 36, 37, 38, 42, 25, 26, 27, 28, 29, 30, 31]),
-            set([58, 59, 60, 61, 62]),
-            set([44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 61]),
-            set([80, 81, 77, 78, 79]),
-            set([64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 80, 63]),
-            set([97, 98, 99, 100, 101]),
-            set([96, 100, 82, 83, 84, 85, 86, 87, 88, 89, 90, 91, 92, 94, 95])
-        ],
-        4: [set([40, 41, 42, 43, 39]),
-            set([42, 35, 36, 37, 38]),
-            set([58, 59, 60, 61, 62]),
-            set([56, 57, 61, 54, 55]),
-            set([80, 81, 77, 78, 79]),
-            set([80, 73, 74, 75, 76]),
-            set([97, 98, 99, 100, 101]),
-            set([96, 100, 92, 94, 95])
-        ],
-    }
     G = graph_example_1()
     result = k_components(G)
+    # In this example graph there are 8 3-components, 4 with 15 nodes
+    # and 4 with 5 nodes.
+    assert_true(len(result[3]) == 8)
+    assert_true(len([c for c in result[3] if len(c) == 15]) == 4)
+    assert_true(len([c for c in result[3] if len(c) == 5]) == 4)
+    # There are also 8 4-components all with 5 nodes.
+    assert_true(len(result[4]) == 8)
+    assert_true(all(len(c) == 5 for c in result[4]))
+    # Finally check that the k-components detected have actually node
+    # connectivity >= k.
     for k, components in result.items():
-        print(k)
-        print(components)
         if k < 3:
             continue
-# Skip tests for nonunique results
-#        for component in components:
-#            assert_true(component in solution[k])
+        for component in components:
+            assert_true(nx.node_connectivity(G.subgraph(component)) >= k)
 
 @raises(nx.NetworkXNotImplemented)
 def test_directed():
--- a/networkx/algorithms/approximation/tests/test_kcomponents.py
+++ b/networkx/algorithms/approximation/tests/test_kcomponents.py
@@ -170,10 +170,13 @@ def test_example_1_detail_3_and_4():
     G = graph_example_1()
     result = k_components(G)
     for k, components in result.items():
+        print(k)
+        print(components)
         if k < 3:
             continue
-        for component in components:
-            assert_true(component in solution[k])
+# Skip tests for nonunique results
+#        for component in components:
+#            assert_true(component in solution[k])
 
 @raises(nx.NetworkXNotImplemented)
 def test_directed():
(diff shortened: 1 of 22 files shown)


jtorrents:
Regarding the k-components tests, the problem is with the labels of the nodes of test graphs, which are quite complex and involve several uses of `nx.convert_node_labels_to_integers` and `nx.disjoint_union` in the process of building them. The tests that fail hardcode the sets of nodes that form 3-components and 4-components. With the new dictionary implementation these nodes have a different labels but the components found by the algorithms have the correct node connectivity value.  I'll prepare a PR against your branch shortly.

jtorrents:
Aric, I made a pull request against your branch with test fixes for k-components tests and node connectivity approximation doctests. See:

https://github.com/hagberg/networkx/pull/5

hagberg:
The travis-ci test are running the doctests and thus failing.  Aside from the larger question about what to do with the doctests I can't seem to turn off those tests.  I thought turning off the doctests in setup.cfg would do it https://github.com/networkx/networkx/pull/2319/commits/9e00af69b6f7ad3e37d6d051ed7e4f5d762d6541 
But apparently that isn't true when you run the tests with nose.main()?

networkx/algorithms/approximation/tests/test_kcomponents.py
```diff
@@ -147,33 +147,23 @@ def test_karate_1():
...
-            set([96, 100, 92, 94, 95])
-        ],
-    }
     G = graph_example_1()
     result = k_components(G)
+    # In this example graph there are 8 3-components, 4 with 15 nodes
+    # and 4 with 5 nodes.
+    assert_true(len(result[3]) == 8)
+    assert_true(len([c for c in result[3] if len(c) == 15]) == 4)
+    assert_true(len([c for c in result[3] if len(c) == 5]) == 4)
```
jfinkels: It would be nicer to use `assert_equals` in these lines I think.

networkx/algorithms/approximation/tests/test_kcomponents.py
```diff
@@ -147,33 +147,23 @@ def test_karate_1():
...
+    assert_true(len(result[4]) == 8)
+    assert_true(all(len(c) == 5 for c in result[4]))
+    # Finally check that the k-components detected have actually node
+    # connectivity >= k.
     for k, components in result.items():
         if k < 3:
             continue
         for component in components:
-            assert_true(component in solution[k])
+            assert_true(nx.node_connectivity(G.subgraph(component)) >= k)
```
jfinkels: `assert_greater_equals` works here.

networkx/algorithms/centrality/flow_matrix.py
```diff
@@ -16,9 +16,9 @@ def flow_matrix_row(G, weight='weight', dtype=float, solver='lu'):
     C = solvername[solver](L, dtype=dtype) # initialize solver
     w = C.w # w is the Laplacian matrix width
     # row-by-row flow matrix
-    for u,v,d in G.edges(data=True):
+    for u,v in sorted(sorted((u,v)) for u,v in G.edges()):
```
jfinkels: Just checking: are nodes guaranteed to be orderable here?

networkx/algorithms/connectivity/tests/test_kcomponents.py
```diff
@@ -126,34 +126,23 @@ def test_karate_component_number():
...
-            {76, 77, 78, 79, 80},
-            {93, 94, 95, 99, 100},
-            {96, 97, 98, 99, 100},
-        ],
-    }
     G = torrents_and_ferraro_graph()
     result = nx.k_components(G)
+    # In this example graph there are 8 3-components, 4 with 15 nodes
+    # and 4 with 5 nodes.
+    assert_true(len(result[3]) == 8)
```
jfinkels: Same comment as above, these can all be `assert_equals`, and `assert_greater_equals` below.

networkx/algorithms/shortest_paths/tests/test_weighted.py
```diff
@@ -151,18 +152,21 @@ def test_bidirectional_dijkstra_no_path(self):
...
-        assert_equal(sorted(dist.items()),
-                     [((0, 0), 0), ((0, 1), 1), ((1, 0), 1), ((1, 1), 2)])
-
+    def test_dijkstra_predecessor2(self):
+        # 4-cycle
+        G = nx.Graph([(0,1),(1,2),(2,3),(3,0)])
+        pred, dist = nx.dijkstra_predecessor_and_distance(G, (0))
+        assert_equal(pred[0],[])
+        assert_equal(pred[1],[0])
+        assert_true(pred[2] in [[1,3],[3,1]])
```
jfinkels: These is an `assert_in` I believe.

jfinkels:
I'll take a look at the chain decomposition test since I contributed that one. It is true that the decomposition is not unique. It may end up being too much of a pain to be in the test suite.

jtorrents:
Good points @jfinkels , I follow your suggestions for k-components tests in a new pull request against Aric's branch: https://github.com/hagberg/networkx/pull/6

Regarding turning out doctests by default, I think one of the nice features of NetworkX is it's up to date documentation. We risk missing outdated doctests if we do not test them by default. Given that there are only 12 failing doctests (and there are hundreds in total) would using `# doctest: +SKIP` be an option?

networkx/algorithms/centrality/flow_matrix.py
jtorrents: Not in general, but in this case there is a relabeling of the nodes of the input graph to integer labels according to rcm ordering (see lines 230 and 344 from `current_flow_betweenness.py
`). So I think it's safe to assume the nodes can be sorted safely.

networkx/algorithms/shortest_paths/tests/test_weighted.py
hagberg: Yes, that would be better.

hagberg:
I'd like to test the doctests.  Adding +SKIP isn't really optimal.  
I'd just like to be able to turn the doctests off for some of the travis-ci testing configurations.   

Testing the doctests will be tricky and maybe only possible for a fixed version of Python.  The doctests are meant as examples and shouldn't have a lot of (unnecessary) sorting or other output processing that confuses the simple use case.

jtorrents:
I agree, adding `+SKIP` to some doctests is not optimal, testing doctests for only some python versions is not much better IMO but I guess is the price to pay to support a wide variety of Python versions. I'm not sure how to handle that in Travis-ci but I guess it should be possible.

jtorrents:
Another downside of only testing docstrings for some python versions in our Travis-ci instance is that if some distributor (e.g debian) ships NetworkX for a python version for which we do not test docstrings, the tests will fail for them.

hagberg:
Continuied in #2413

PR #2320, opened 2016-11-27
Allow grid_graph generator to accept tuple dim argument

smopucilowski:
This refactors the grid_graph() generator to accept both a tuple as well as a list argument for its dimension argument.


smopucilowski:
Additionally, this commit prevents clobbering of the dim argument when passed as a list.

PR #2326, opened 2016-12-02
Use weight=None in Centrality Algorithms rather than weight="weight"

Michael-E-Rose:
Some functions use `None` as standard for the `weight` parameter, but most used `'weight'`.  I harmonized this such that all functions use `weight='weight'` if their computation differs for weighted graphs.   On the go I applied PEP8 and Networkx' Top of the Module standards.

There was also one misspecified test for the `global_reaching_centrality`, namely the one on undirected and weighted networks.


Michael-E-Rose:
The AppVeyor test for `PYTHON_VERSION=3.3.5, PYTHON_ARCH=32` fails on `test_asyn_lpa.test_two_communities`.  Could someone explain why?  I did not even touch the file.

Michael-E-Rose:
The AppVeyor test for `PYTHON_VERSION=3.3.5, PYTHON_ARCH=32` fails on `test_asyn_lpa.test_two_communities`.  Could someone explain why?  I did not even touch the file.

dschult:
Appveyor has been failing occasionally at that step for a little while now.
It is definitely not something you did.

Michael-E-Rose:
I updated the release note. Two centralities, that I have forgotten in the initial commit, were updated as well.

doc/source/reference/release_2.0.rst
```diff
@@ -106,3 +106,10 @@ API changes
    Most of the shortest_path algorithms now raise a NodeNotFound exception
    when a source or a target are not present in the graph.
 
+* [`#2326 <https://github.com/networkx/networkx/pull/2326>`_]
+   Centrality algorithms were harmonized with respect to the defaul behavior of
```
jfinkels: There is a typo here: defaul -> default.

doc/source/reference/release_2.0.rst
```diff
@@ -106,3 +106,10 @@ API changes
...
+   the weight parameter. The default behavior is now to use `'weight'` as weight
```
jfinkels: This second sentence is a little clunky. How about

```rest
The default value of the ``weight`` keyword argument has been changed from ``None`` to the string ``'weight'``.
```

doc/source/reference/release_2.0.rst
```diff
@@ -106,3 +106,10 @@ API changes
...
+   keyword. This affects `betweenness_centrality()`, `edge_betweenness_centrality()`,
```
jfinkels: Perhaps this would be easier to read as an alphabetized list? Also, these should be references to the actual functions:
```rest
This affects the following centrality functions.

- :func:`betweenness_centrality`
- :func:`edge_betweenness_centrality`
- ...
```

Michael-E-Rose:
Thanks for the review, @jfinkels! These are all valid points. I updated the release notes accordingly.

dschult:
I haven't looked at this carefully, but some shortest_path algorithms choose whether to use Dijkstra or not based on the weight parameter. Does this make the default using Dijkstra instead of using an unweighted algorithm?  If so, how does the user specify using the unweighted algorithm?

Michael-E-Rose:
The keyword for no weight is `None`, as it was already the case for some of the algorithms.

Whether to use Dijkstra or shortest path algorithm is determined by blocks like this:
```
if weight is None:  # use BFS
        S, P, sigma = _single_source_shortest_path_basic(G, s)
    else:  # use Dijkstra's algorithm
        S, P, sigma = _single_source_dijkstra_path_basic(G, s, weight)
```
So yes, it would make Dijkstra the default algorithm. Would this cause problems if the graph is unweighted and the default keyword (i.e. `"weight"`) is used?

dschult:
If there is no edge attribute ```"weight"``` for any edge then Dijkstra should return the same results as the unweighted methods, but would take longer.  

If the ```"weight"``` attribute exists and differs from ```1``` for any edge then certainly the results could differ from the unweighted methods.  I suppose so long as it is clear which is the default and how to get both BFS and Dijkstra, people will find how to call it to get what they want. 

Currently if we don't specify which algorithm and don't specify which attribute we get unweighted methods. If we use a method that implies a weight then the default weight attribute is ```"weight"``` and Dijkstra is used. So if no weight is needed, the whole idea of weight is not needed. I like the idea that users who don't care about weights don't have to think about weights when they call the functions.

What is the advantage of making Dijkstra the default algorithm?

Michael-E-Rose:
> What is the advantage of making Dijkstra the default algorithm?

I think there is none.  To be honest, I have not foreseen that this will be the case, and it's probably not that smart to make Dijkstra the default.

There are probably two ways: First, I implement a check in the betweenness centrality algorithm whether weights differ. Second, we do not make `'weight'` the default keyword for the weight parameter, but `None`.

jtorrents:
I think that the best approach is to use `None` as the default value for the `weight` parameter everywhere. Thus if the user wants to compute an unweighted version of the algorithm, she doesn't have to use the optional `weight` parameter even if the input graph has an edge attribute named weight. I think it would be confusing that the same call would yield different results depending on having or not an edge attribute named weight. I think that this is an instance of 'Explicit is better than implicit'. On the other hand, for the shortest paths functions, I think that defaulting to Dijkstra is not a good idea, as it's slower than BFS.

Michael-E-Rose:
Absolutely correct. I will work on this and change the existing commits.

dschult:
As I look at this in more detail it seems that having ```weight=None``` for the dijkstra methods doesn't make sense.  Maybe those shouldn't even have a default.  We should at least make the private functions have no default for the weight parameter.

Michael-E-Rose:
I removed the default for the private functions. What's with the changes that @jfinkels requested a while ago? I guess I worked them in as well.

dschult:
I think you already implemented @jfinkels suggestions.
This touched a lot of code.  Thanks for all your efforts.
Is this ready for merge or do you have more you are looking to do?

Michael-E-Rose:
No, from my side it's all fine.

PR #2329, opened 2016-12-09
Simplifies two community test for asyn-lpa.

jfinkels:
I'm hoping this should help with the occasional test failures while maintaining the spirit of the test.

### DIFF
--- a/networkx/algorithms/community/tests/test_asyn_lpa.py
+++ b/networkx/algorithms/community/tests/test_asyn_lpa.py
@@ -1,4 +1,10 @@
 from nose.tools import assert_equal
+try:
+    from nose.tools import assert_count_equal
+except ImportError:
+    from nose.tools import assert_items_equal as assert_count_equal
+
+import networkx as nx
 from networkx import asyn_lpa, Graph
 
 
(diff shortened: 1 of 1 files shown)


dschult:
The current failure occurs sometimes when a single community (all nodes) is found.
Is that still possible with this graph and this algorithm?  I'm not sure if this commit is making it less likely to occur or impossible to occur.

jfinkels:
I was assuming it would make it much less likely, but I'm not familiar enough with the algorithm to say.

dschult:
It can still fail the test -- the algorithm chooses nodes at random and they "vote" (change their label) to the label that is the majority of their neighbors (or arbitrarily for ties). The order in which nodes are chosen determines the outcome. So I think we should include both possible outcomes in our test.

I'll try to put together a PR leaving the old test, but including both possible answers.  What do you think?

    -    assert_equal(result, ground_truth)
    +    # probabilitistic result could be all nodes in one community. So test result is either.
    +    assert(result in [ground_truth, set(frozenset(range(16)))])

dschult:
I merged #2339 which is an alternate version of this.

Issue #2331, opened 2016-12-12
articulations_points returns repeated nodes


malbarbo:
```python
import networkx as nx
from networkx.algorithms.components.biconnected import articulation_points

g = nx.Graph()
g.add_edges_from([(0, 1), (1, 2), (1, 3)])
assert list(articulation_points(g)) == [1, 1]
```

leotrs:
After investigating the problem, node `1` is returned twice because the algorithm identifies two different biconnected components that split from the original graph after removing `1`. This is due to the fact that the algorithm proceeds by DFS traversal. Starting at `0`, node `1` will be identified as articulation point after traversing down its child `2` (and everything that follows from `2`), and once again after traversing down node `3`.

It seems to me that this behavior is already known, since [all tests](https://github.com/networkx/networkx/blob/6e20b952a957af820990f68d9237609198088816/networkx/algorithms/components/tests/test_biconnected.py) that involve `articulation_points` call `set` over the returned list (thus deleting the duplicate values).

@malbarbo I would advice just calling `set` over the return value, as it seems that there is some valuable information in the repeated values in the returned list of `articulation_points`.

@jfinkels On the other hand, if this is not expected behavior, it should be easy to add a quick memoification step inside `articulation_points` so that the same node is not yielded more than once. I could do this if there is interest.

jfinkels:
It seems the function currently yields the node once each time it is discovered to lie in the intersection of two maximal biconnected subgraphs. In this example, the node 1 is in the intersection of the two maximal biconnected components {0, 1} and {1, 2} and in the intersection of {0, 1} and {1, 3}. This is a little confusing if we just consider the name of the function, `articulation_points`, which seems to me to imply that it yields each articulation point once.

My feeling is that this "multiplicity" information about articulation points could be expressed more clearly in a different way (see the [Related structures][1] section of the Wikipedia article on biconnected components for some possibilities). Therefore, I think that this function should yield each articulation point only once:

```python
seen = set()
for v in _biconnected_dfs(G, components=False):
    if v not in seen:
        yield v
```

A test should also be added to check that node are yielded only once (the example code by @malbarbo is fine).

[1]: https://en.wikipedia.org/wiki/Biconnected_component#Related_structures

leotrs:
The proposed PR should take care of this.

[...]

PR #2333, opened 2016-12-14
change articulation_points so that it only returns every vertex once

leotrs:
Solves #2331.

### DIFF
--- a/networkx/algorithms/components/biconnected.py
+++ b/networkx/algorithms/components/biconnected.py
@@ -417,7 +417,11 @@ def articulation_points(G):
            Communications of the ACM 16: 372–378. doi:10.1145/362248.362272
 
     """
-    return _biconnected_dfs(G, components=False)
+    seen = set()
+    for articulation in _biconnected_dfs(G, components=False):
+        if articulation not in seen:
+            seen.add(articulation)
+            yield articulation
 
 
 @not_implemented_for('directed')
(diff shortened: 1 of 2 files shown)


leotrs:
The only failing test doesn't seem to depend on my changes. cc. @jfinkels

jfinkels:
That test failure is being addressed by pull request #2329.

networkx/algorithms/components/biconnected.py
```diff
@@ -417,7 +417,11 @@ def articulation_points(G):
            Communications of the ACM 16: 372–378. doi:10.1145/362248.362272
 
     """
-    return _biconnected_dfs(G, components=False)
+    seen = set()
+    for articulation in _biconnected_dfs(G, components=False):
+        if articulation not in seen:
+            seen.add(articulation)
+            yield articulation
```
jfinkels: Can you change the documentation from

```rest
     Returns
     -------
     articulation points : generator
         generator of nodes
```
to
```
     Yields
     ------
     node
         An articulation point in the graph.
```
since this function is now an iterator generator instead of a function that returns a generator.

leotrs: Done. Let me know if this is looking good.

leotrs:
So, are we waiting for something else before this gets merged?

leotrs:
@jfinkels ^

Issue #2291, opened 2016-10-20
PyPy on travis is very old


TiberiumPY:
https://travis-ci.org/networkx/networkx look here, PyPy is version 2.5, BUT there's already 5.4.1! Can you please update PyPy to the latest?
https://github.com/travis-ci/travis-ci/issues/5027 here this issue.
In this issue you can find a script to test against latest stable PyPy

PR #2334, opened 2016-12-15
Enables PyPy-5.3.1 in Travis builds

jfinkels:
Addresses issue #2291.


hagberg:
The build gets killed for some reason.  Any idea?

jfinkels:
You manually re-ran the Travis job right? I'm not sure what the problem is here. I don't know why travis would kill a job. In any case, there are test errors and failures, so if someone has access to this version of PyPy (or newer), he or she can at least start to diagnose some of the problems.

dschult:
see #2468

PR #2335, opened 2016-12-18
Allowing incomplete personalization

### LINKED ISSUE #2327: Personalization in PageRank
In https://github.com/networkx/networkx/blob/master/networkx/algorithms/link_analysis/pagerank_alg.py in all algorithms, the personalization dict accepted needs a value for every node. In my personal application, most values are 0. I suspect that in a lot of applications, including the default, most nodes will have the same probability.
Furthermore, for all but the PageRank algorithm, the docstring suggests that each of these values needs to be non-zero. As far as I know, there is no code-technical or mathematical basis for this and it should be "at least one non-zero value".

My proposal:

- Add a parameter specifying a default value for personalization.
- Do not force presence of all keys in personalization, instead use default value for non-existant values.
- Rewrite docstrings to specify "at least one non-zero value"

I'll be happy to implement this. Feedback appreciated. :)

pimotte:
Would close #2327 

Happy to process any feedback, if there is any.

### DIFF
--- a/networkx/algorithms/link_analysis/pagerank_alg.py
+++ b/networkx/algorithms/link_analysis/pagerank_alg.py
@@ -35,8 +35,9 @@ def pagerank(G, alpha=0.85, personalization=None,
 
     personalization: dict, optional
       The "personalization vector" consisting of a dictionary with a
-      key for every graph node and personalization value for each node.
+      key some subset of graph nodes and personalization value each of those.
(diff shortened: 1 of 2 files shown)


pimotte:
So it looks like the asyn_lpa test fails, but as far as I can see it is not related to my changes, and some work is being done on this. If it does have something to do with my changes, a pointer would be really helpful.

PR #1679, opened 2015-07-29
Allows arbitrary metric in geometric generators.

jfinkels:
This commit adds the `metric` keyword argument to three functions in
`networkx.generators.geometric`, namely

* `random_geometric_graph`,
* `waxman_graph`,
* `geographical_threshold_graph`.

This keyword argument allows the function to compare distances between
nodes according to arbitrary metric functions instead of hardcoded
Euclidean metrics. The default behavior if no `metric` argument is
specified remains the Euclidean metric.

This commit also cleans and simplifies the implementations of these
three functions, as well as adds more meaningful unit tests.

PR #2337, opened 2016-12-19
Use faster random geometric graph implementation.

aryamccarthy:
Replace the existing random geometric graph generator with one
that uses a k-d tree to efficiently determine which nodes are
adjacent, so the algorithm is now `O(n)` instead of `O(n^2)`.

Use p-norm number as parameter instead of function. The
scipy KDTree requires a number instead of distance function.

### DIFF
--- a/networkx/generators/geometric.py
+++ b/networkx/generators/geometric.py
@@ -9,6 +9,7 @@
 # Authors: Aric Hagberg (hagberg@lanl.gov)
 #          Dan Schult (dschult@colgate.edu)
 #          Ben Edwards (BJEdwards@gmail.com)
+#          Arya McCarthy (admccarthy@smu.edu)
 """Generators for geometric graphs.
 """
 from __future__ import division
@@ -20,6 +21,7 @@ from math import sqrt
 import math
 import random
 from random import uniform
+from scipy.spatial import KDTree
 
 import networkx as nx
 from networkx.utils import nodes_or_number
@@ -39,7 +41,7 @@ def euclidean(x, y):
 
 
 @nodes_or_number(0)
-def random_geometric_graph(n, radius, dim=2, pos=None, metric=None):
+def random_geometric_graph(n, radius, dim=2, pos=None, norm=2):
     """Returns a random geometric graph in the unit cube.
 
     The random geometric graph model places `n` nodes uniformly at
@@ -56,23 +58,12 @@ def random_geometric_graph(n, radius, dim=2, pos=None, metric=None):
         Dimension of graph
     pos : dict, optional
         A dictionary keyed by node with node positions as values.
-    metric : function
-        A metric on vectors of numbers (represented as lists or
-        tuples). This must be a function that accepts two lists (or
-        tuples) as input and yields a number as output. The function
-        must also satisfy the four requirements of a `metric`_.
-        Specifically, if *d* is the function and *x*, *y*,
-        and *z* are vectors in the graph, then *d* must satisfy
+    norm : float
+        Which Minkowski norm to use.  `p` has to meet the condition
+            ``1 <= p <= infinity``.
 
-        1. *d*(*x*, *y*) ≥ 0,
-        2. *d*(*x*, *y*) = 0 if and only if *x* = *y*,
-        3. *d*(*x*, *y*) = *d*(*y*, *x*),
-        4. *d*(*x*, *z*) ≤ *d*(*x*, *y*) + *d*(*y*, *z*).
-
-        If this argument is not specified, the Euclidean distance metric is
-        used.
-
-        .. _metric: https://en.wikipedia.org/wiki/Metric_%28mathematics%29
+        If this argument is not specified, the L2 norm (the Euclidean distance 
+        metric) is used.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


aryamccarthy:
This partially undoes an API change in #1679, which has never been in a release build—only the dev versions.

That pull request added the option of a custom distance metric. Scipy's KDTree implementation doesn't accept metrics as arguments; instead, it relies on the numeric norm, which we make into an argument to `random_geometric_graph()`.

networkx/generators/geometric.py
```diff
@@ -56,23 +58,12 @@ def random_geometric_graph(n, radius, dim=2, pos=None, metric=None):
...
     pos : dict, optional
         A dictionary keyed by node with node positions as values.
-    metric : function
-        A metric on vectors of numbers (represented as lists or
-        tuples). This must be a function that accepts two lists (or
-        tuples) as input and yields a number as output. The function
-        must also satisfy the four requirements of a `metric`_.
-        Specifically, if *d* is the function and *x*, *y*,
-        and *z* are vectors in the graph, then *d* must satisfy
+    norm : float
```
jfinkels: I think the name for this keyword argument should just be `p` for two reasons. First, it matches the use of the `p` keyword argument in the KDTree below, and second, the distance between two vectors is technically given by a metric, not a norm.

networkx/generators/geometric.py
```diff
@@ -56,23 +58,12 @@ def random_geometric_graph(n, radius, dim=2, pos=None, metric=None):
...
+        Which Minkowski norm to use.  `p` has to meet the condition
```
jfinkels: Since this must be a number greater than one, this should be the L_p metric, not the Minkowski metric. It would be nice to link to the definition of the L_p metric here.

networkx/generators/geometric.py
```diff
@@ -56,23 +58,12 @@ def random_geometric_graph(n, radius, dim=2, pos=None, metric=None):
...
+            ``1 <= p <= infinity``.
```
jfinkels: There is some unnecessary indentation here.

networkx/generators/geometric.py
```diff
@@ -56,23 +58,12 @@ def random_geometric_graph(n, radius, dim=2, pos=None, metric=None):
...
-        1. *d*(*x*, *y*) ≥ 0,
-        2. *d*(*x*, *y*) = 0 if and only if *x* = *y*,
-        3. *d*(*x*, *y*) = *d*(*y*, *x*),
-        4. *d*(*x*, *z*) ≤ *d*(*x*, *y*) + *d*(*y*, *z*).
-
-        If this argument is not specified, the Euclidean distance metric is
-        used.
-
-        .. _metric: https://en.wikipedia.org/wiki/Metric_%28mathematics%29
+        If this argument is not specified, the L2 norm (the Euclidean distance
```
jfinkels: The *p* should be rendered as a superscript. In math mode, this would be ```:math:`L^2` ```. Again, this is a metric, not a norm.

jfinkels:
Travis is telling us that the code fails when SciPy is not installed. Since NetworkX does not assume the user always has SciPy installed, perhaps we can try the SciPy version first, then fall back to the pure Python implementation if SciPy is not installed? Something like

```python
try:
    from scipy.spatial import KDTree
except ImportError:
    is_scipy_available = False
else:
    is_scipy_available = True

def _fast_geometric_graph(G):
    ...

def _slow_geometric_graph(G):
    ...

def geometric_graph(G):
   return _fast_geom_graph(G) if is_scipy_available else _slow_geom_graph(G)
```

Second, I believe there is a faster KDTree in SciPy called `cKDTree`. Can you test that one out instead? It should have the same interface.

networkx/generators/geometric.py
aryamccarthy: I avoided the use of `p` because I worry that someone may confuse this for the Erdos-Renyi graph; is that not a concern I should have?

jfinkels: You mean that someone would think `p` represents a probability? Well, I guess there's a possibility of that, but in this case clear documentation seems sufficient to me.

aryamccarthy: Yeah, that and `r` as a required argument should differentiate them.

networkx/generators/geometric.py
aryamccarthy: Fixed.

aryamccarthy:
@hagberg My only failed test was in algorithms\community\tests, which I didn't mess with. Could you restart?

dschult:
I restarted the tests, but you should not worry about it -- we are trying to fix the issue with a different PR.  We can merge your PR  even if it fails the Appveyor tests. :)

aryamccarthy:
Should be good to go, then, @dschult.

dschult:
Can you mention in the doc_string that this uses KDTree when Scipy is available.
Might also be good to mention that the algorithm changes from O(n^2) to O(n).

networkx/generators/geometric.py
```diff
@@ -38,8 +45,37 @@ def euclidean(x, y):
...
+    """
+    pos = nx.get_node_attributes(G, 'pos')
+    nodes, coords = list(zip(*pos.items()))
+    kdtree = KDTree(coords)  # Cannot provide generator.
+    edge_indexes = kdtree.query_pairs(radius, p)
+    edges = ((nodes[u], nodes[v]) for u, v in edge_indexes)
+    G.add_edges_from(edges)
+
+
+def _slow_construct_edges(G, radius, p):
```
aryamccarthy: Is there an efficient, networkx-friendly way to do this? I worry about the GIL, and also whether networkx assumes joblib is installed.

jfinkels: I'm not sure I understand. A way to do what?

aryamccarthy: Sorry—to parallelize.

jfinkels: Oh, I see. I have made a habit of marking when loops can be parallelized with a note like `# TODO This can be parallelized.` However, this is just a note for future developers working at a time when Python will have better parallel primitives. You can ignore it for now.

aryamccarthy:
Incorporated your changes, @dschult and @jfinkels. Should be ready now.

dschult:
Thanks very much!

PR #2341, opened 2016-12-24
*Really* allow community asyn_lpa test to have two answers

aryamccarthy:
This small fix improves on #2339 by incorporating the nosetests asserts, which give richer output upon failure.

EDIT: No, it actually fixes the fix to that testing bug we've had.


aryamccarthy:
Incidentally, while we're all here—there's a [new Github option](https://github.com/blog/2141-squash-your-commits#whats-changing) that lets you (the owners) squash commits when you merge.

aryamccarthy:
I intended this PR to just shift to more nosetests so failed tests would be better-explained, but I found a bug in the testing that this PR fixes.

Issue #2312, opened 2016-11-16
Networkx can not read the attributes from .gexf file


sdrgroup:
1) networkx only supports .gexf of version 1.1, but now .gexf has already been version 1.3
2) after I change the .gexf to 1.1(by changing the head of xml), networkx can load the file but the attributes of node are only color, size and position, the other attributes between <attributes></attributes> are missing.

I use networkx of version 1.11
The operation system is ubuntu 16.04

Michael-E-Rose:
Yes, networkx' standard version (xml namespace) is still on `'1.1draft'`, which is [deprecated](http://www.gexf.net/format/schema.html).

Would it be sufficient to simply change the default value of the version parameter?

hagberg:
The gexf reader needs updating to version 1.3.

Michael-E-Rose:
PR is live.

But why do you want version 1.3?  According to http://www.gexf.net/format/schema.html (January 2017) the recommended version is 1.2draft.

PR #2355, opened 2017-01-24
Use version 1.2draft by default and apply PEP8 to gexf.py

Michael-E-Rose:
Addresses #2312 (not sure if it fixes all of it).

* Use 1.2draft as default xml version
* Apply PEP8 were useful


Michael-E-Rose:
The first two commits are not mine - how do I get rid of them?

OrkoHunter:
@Michael-E-Rose You could try rebasing with the updated master. Probably that would help.

Michael-E-Rose:
No it doesn't look like this solved the issue...

OrkoHunter:
@Michael-E-Rose I only see one commit now :)

Michael-E-Rose:
Oh great! So this PR is ready to be merged :)

dschult:
I think this (#2355) is ready to go. 
There were other GML fixes recently that made two small conflicts.  
@Michael-E-Rose do you have time to fix them? If not I can do it.
Thanks.

Michael-E-Rose:
The checks fail.  Seems I did something wrong in the merge. Is there a way to undo the merge and try again?

dschult:
Based on #2436 I think you need to remove the ```viz``` part of line 293. 
You undid the change from #2436 when you merged. That change in a new commit might be easier than undoing the merge and redoing it.  But git allows you to revert and try again (google git revert).  You can also start a new branch from just before the merge with ```git checkout HEAD~1``` followed by ```git checkout -b new_name``` and try thr merge again...   still seems easier to change line 293 in ```test_gexf.py```

Michael-E-Rose:
I pushed again with a minor change and thus could merge again. Hope this time I made no errors.

Michael-E-Rose:
Thanks! I tested above merge solution on nosetests, but while they worked there, they did not work here. Maybe wrong branch...

Issue #2332, opened 2016-12-14
JSON exports no longer follow D3's preferred format as of D3 v4


bjlange:
The D3 [node/link JSON example](https://bl.ocks.org/mbostock/4062045), linked in the documentation, now expects source and target IDs in the link objects that are an attribute of the node objects which you specify when setting up the force-directed graph (e.g. `d.id`). Previously they expected indexes into the node list, you can see the change in the example gist [here](https://gist.github.com/mbostock/4062045/revisions?diff=split#diff-8143eda70191fa184211ddbe84974fbdL81). 

Should make things simpler for networkx users but assuming `node_link_data` is designed to target d3, the behavior of that should probably change (or at least, the link in the documentation should be taken out until it does).

Related to #2328 as well.

Michael-E-Rose:
It's easy to change this behaviour by removing the mapping (i.e. `mapping = dict(zip(G, count()))` and subsequent use of this).

The question is: Should it still be possible to use indexes corresponding to the node list?  I could implement this with an optional parameter.

hagberg:
I think we should just make it match the D3 v4 approach.  Integer indices were just there to make this work with D3.

rhilenova:
I'm attempting to create a PR for this, but I'm not sure how to handle tuple nodes. The current code creates a node with the name `[0, 1]`, and then can successfully create edges by index. However, a quick test with [the example](https://bl.ocks.org/mbostock/4062045#index.html) doesn't look like it handles that format well.

Michael-E-Rose:
Don't you just need to remove the mapping?  Because then the algorithm will just use the id's which are already there.

I had nearly finished a PR but wanted to wait for the others to be accepted.  It seem you're more advanced, @rhilenova, so go ahead with a PR and we have a look on the code!

rhilenova:
If I just remove the mapping, `TestNodeLink.test_graph_with_tuple_nodes` fails (unhashable type list). I haven't tried to fix that issue yet, because it made me wonder what D3 would expect and if it even made sense for tuple nodes to be output to json.

Michael-E-Rose:
Yes of course you need to adapt the test files. They expect integers but this must change.

Issue #2328, opened 2016-12-07
json_graph.node_link_graph only accepts integer indicies in links


rhilenova:
The d3.js example [here](http://bl.ocks.org/mbostock/4062045) has the node names for the sources and targets. However, this graph fails to load using json_graph.node_link_graph:
```
.../anaconda/lib/python2.7/site-packages/networkx/readwrite/json_graph/node_link.py in node_link_graph(data, directed, multigraph, attrs)
    168             edgedata = dict((make_str(k), v) for k, v in d.items()
    169                             if k != source and k != target and k != key)
--> 170             graph.add_edge(mapping[src], mapping[tgt], ky, **edgedata)
    171     return graph

TypeError: list indices must be integers, not unicode
```

Shouldn't node_link_graph only use `mapping` if it gets integer indices?

Michael-E-Rose:
This is the opposite of #2332, which is about producing a json graph, while this issue is about reading a json graph.

It's best to answer both issues together:  Should the mapping onto indexes still be possible (optional) or not?

hagberg:
See #2332.  No need to map on to indices.

PR #2366, opened 2017-02-13
Fix for issues #2328 and #2332

rhilenova:
Fix for issues #2328 and #2332.

Changes JSON node-link format to no longer export or import expecting links to be by node index. Instead, it expects node IDs like D3's current examples.

### DIFF
--- a/networkx/readwrite/json_graph/node_link.py
+++ b/networkx/readwrite/json_graph/node_link.py
@@ -1,10 +1,10 @@
 #    Copyright (C) 2011-2016 by
-# 
+#
 #    Aric Hagberg <hagberg@lanl.gov>
 #    Dan Schult <dschult@colgate.edu>
 #    Pieter Swart <swart@lanl.gov>
 #    Michael E. Rose <Michael.Ernst.Rose@gmail.com>
-#    
+#
 #    All rights reserved.
 #    BSD license.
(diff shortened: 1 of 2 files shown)


networkx/readwrite/json_graph/node_link.py
```diff
@@ -182,10 +179,18 @@ def node_link_graph(data, directed=False, multigraph=True, attrs=None):
...
+            graph.add_edge(src, tgt, **edgedata)
         else:
             ky = d.get(key, None)
             edgedata = dict((make_str(k), v) for k, v in d.items()
                             if k != source and k != target and k != key)
-            graph.add_edge(mapping[src], mapping[tgt], ky, **edgedata)
+            if (isinstance(src, list)):
+                src = tuple(src)
+            if (isinstance(tgt, list)):
+                tgt = tuple(tgt)
```
MridulS: Instead of doing this twice you can do
``` python
src = tuple(d[source]) if isinstance(d[source], list) else d[source]       
tgt = tuple(d[target]) if isinstance(d[target], list) else d[target]
```
while accessing the values in the starting of the for loop

hagberg:
Is this OK now @MridulS?

MridulS:
@hagberg :+1:

PR #2375, opened 2017-03-02
Raise an Exception for disconnected Graphs in bipartite.sets

### LINKED ISSUE #2127: Issues calculating max independent set
`Python 2.7.11`
`networkx 1.11`

I have the following directed acyclic graph:

DAG

I wanted to calculate the max independent set of the transitive closure of this DAG (the nodes colored in green).

``` python
from networkx import Graph, DiGraph, transitive_closure
from networkx.algorithms.bipartite import hopcroft_karp_matching, to_vertex_cover

# Build the DAG
G = DiGraph()
G.add_edge("A", "C")
G.add_edge("A", "B")
G.add_edge("C", "E")
G.add_edge("C", "D")
G.add_edge("E", "G")
G.add_edge("E", "F")
G.add_edge("G", "I")
G.add_edge("G", "H")

tc = transitive_closure(G)
btc = Graph()

# Create a bipartite graph based on the transitive closure of G
for v in tc.nodes_iter():
    btc.add_node((0, v))
    btc.add_node((1, v))

for u, v in tc.edges_iter():
    btc.add_edge((0, u), (1, v))

matching = hopcroft_karp_matching(btc)
vertex_cover = to_vertex_cover(btc, matching)
independent_set = set(G) - {v for _, v in vertex_cover}
```

I am expecting independent set to be `set(['B', 'D', 'F', 'I', 'H'])` but it is `set(['A', 'B', 'D', 'F', 'I', 'H'])` instead.  I am either doing something wrong, or there is a bug in the `to_vertex_cover` function.

References:
- http://cs.stackexchange.com/a/16829
- http://cs.stackexchange.com/a/10303/5323
- http://stackoverflow.com/a/10304235
- https://en.wikipedia.org/wiki/Hopcroft%E2%80%93Karp_algorithm
- https://en.wikipedia.org/wiki/Dilworth's_theorem
- https://en.wikipedia.org/wiki/K%C5%91nig%27s_theorem_(graph_theory) - Konig’s theorem proves an equivalence between a maximum matching and a minimum vertex cover in bipartite graphs.  A minimum vertex cover is the complement of a maximum independent set for any graph

jtorrents:
This pull request fixes #2127

When a disconnected graph is passed to `bipartite.sets` more than one
coloring solution is possible. This led to surprising results and hard
to diagnose bugs (see #2127) in algorithms that used `bipartite.sets`,
such as bipartite matching and friends. These algorithms now have a new
optional parameter that allows the user to specify the nodes in one of
the two bipartite sets, and thus allows to use disconnected graphs in
these algorithms without ambiguity.

I've added a new exception `AmbiguousSolution` in order to be more
explicit when we raise it. It might have more use cases beyond
`bipartite.sets`.

The changes in this pull request are backwards compatible for connected
graphs as the new parameter is optional. For disconnected graphs is not
backward compatible, but since the assignation of nodes to a bipartite
set depends on iteration order, that code was already subtly broken
as it would yield different results on different python versions.

### DIFF
--- a/networkx/algorithms/bipartite/basic.py
+++ b/networkx/algorithms/bipartite/basic.py
@@ -139,19 +139,39 @@ def is_bipartite_node_set(G,nodes):
     return True
 
 
-def sets(G):
+def sets(G, top_nodes=None):
     """Returns bipartite node sets of graph G.
 
-    Raises an exception if the graph is not bipartite.
+    Raises an exception if the graph is not bipartite or if the input
+    graph is disconnected and thus more than one valid solution exists.
 
     Parameters
     ----------
     G : NetworkX graph
 
+    top_nodes : container
+
+      Container with all nodes in one bipartite node set. If not supplied
+      it will be computed. But if more than one solution exists an exception
+      will be raised.
+
     Returns
     -------
     (X,Y) : two-tuple of sets
        One set of nodes for each part of the bipartite graph.
+
+    Raises
+    ------
+    AmbiguousSolution : Exception
+
+      Raised if the input bipartite graph is disconnected and no container
+      with all nodes in one bipartite set is provided. When determining
+      the nodes in each bipartite set more than one valid solution is
+      possible if the input graph is disconnected.
+
+    NetworkXError: Exception
+
+      Raised if the input graph is not bipartite.
 
     Examples
     --------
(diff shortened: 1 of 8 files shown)


networkx/algorithms/bipartite/basic.py
```diff
@@ -166,10 +186,22 @@ def sets(G):
...
+        is_connected = nx.is_connected
+    if top_nodes is not None:
+        X = set(top_nodes)
+        Y = set(G) - X
+    else:
+        if not is_connected(G):
+            msg = 'Disconnected graph: Ambiguous solution for bipartite sets.'
+            raise nx.AmbiguousSolution(msg)
+        c = color(G)
+        X = {n for n in c if c[n]} # c[n] == 1
```
jfinkels: Could be `{n for n, is_top in c.items() if is_top}`

jtorrents:
Thank you @jfinkels! I changed the dict comprehensions as per your suggestion. Using better variable names in them allows to safely remove the comments.

PR #1535, opened 2015-05-20
Adds functions computing structural hole measures

jfinkels:
This commit creates a new module, `networkx.algorithms.structuralholes`,
that provides functions for measuring structural holes in a graph.

I have cleaned and documented the functions created by @hagberg in issue #192. They are failing the tests provided by a previous contributor. Using Google to search for a name, I'm guessing the originator of this code was @drdee (please feel free to ignore this mention if you are not the original contributor). In any case, this pull request still needs some tests and needs to pass the existing tests, but at least now there is a pull request associated with the (documented) code.

Issue #429, opened 2012-07-20
Structural Holes (migrated from Trac #435)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/435
Reported 2010-09-21 by trac user alex, assigned to @hagberg.

Calculates Burt's structural holes values

dschult:
Fixed by #2378

Issue #192, opened 2012-07-20
Add structural holes measure (migrated from Trac #191)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/191
Reported 2008-11-24 by @hagberg, assigned to @hagberg.

Structural holes:
http://jung.sourceforge.net/doc/api/edu/uci/ics/jung/algorithms/metrics/StructuralHoles.html

dschult:
Fixed by #2378

PR #2378, opened 2017-03-05
 Adds structural hole measures

jtorrents:
This PR supersedes #1535, #429, and #192.

This commit fixes the implementation of constraint and effective size in
PR #1535. That implementation was using `nx.all_neighbors` to iterate
over all predecessors and successors of nodes. The problem was that if a
neighbor is both a predecessor and a successor is appears twice in
`nx.all_neighbors` and thus it was double counted.

Regarding constraint, PR #1535 was following the formula of constraint
posted at a very visible web page [1], which is wrong as it says that
the local constraint has to be calculated only on neighbors that are
both predecessors and successors of the focal node. Burt's paper [2]
at page 362 states that "I measure the constraint on each manager with
respect to the immediate network of discussion partners, composed of
anyone the manager cited as a discussion partner and anyone who cited
the manager.", which means that it should be computed for all neighbors
that are either successors or predecessors.

Regarding effective size, I implemented the original formula proposed by
Burt in [3], which was different that the one used in #1535. For the
case of unweighted and undirected graphs I implemented the simplified
formula proposed by Borgatti [4] which is faster.

After these changes the tests posted in #192 now pass. These test were
checked using UCINET, I rechecked them using IGraph and also computed them
by hand to make sure that they were right as I don't have access to
UCINET and effective size is not implemented in IGraph. I also added more
tests to cover all cases applicable (weighted, unweighted, directed,
and undirected).

I changed the API of the functions to compute the measures on all nodes
of the graph and accept an arbitrary edge weight name to compute the
weighted versions, following the API of other centrality measures in
NetworkX.

After profiling these measures, I sped-up the running time by
reimplementing the inner function `mutual_weight` as a succession of
`try` and `except` instead of the original generator expression, which
is ~40% faster. This inner function is called millions of times for not
so large graphs (thousands of nodes) and thus the speed up is
significant. On the other hand, I also removed the decorator
`not_implemented_for` of this function, which according to profiling
accounted for 50% of the total running time of the measures.

[1] http://www.centiserver.org/?q1=centrality&q2=Burts_constraint
[2] Burt, R. (2004). "Structural holes and good ideas". American Journal
of Sociology (110): 349–399.
[3] Burt, Ronald S. (1995). Structural Holes: The Social Structure of
Competition. Cambridge: Harvard University Press.
[4] Borgatti, Stephen (1997). "Structural Holes: Unpacking Burt's
Redundancy Measures". Connections. INSNA.


jtorrents:
This is a review of #1535, but instead of making comments there I thought it was easier to just do a new PR. @hagberg do you think it can be included in 2.0? This is a frequently asked feature (not at the bug tracker but to me directly by several colleagues) as `constraint` is frequently used in sociology empirical research.

jtorrents:
Thank you @jfinkels!

PR #2386, opened 2017-03-14
Check alternating paths using iterative DFS in to_vertex_cover.

### LINKED ISSUE #2384: bipartite.to_vertex_cover() gives incorrect results
This program:
```python
import networkx as nx

print('Using version ' + nx.__version__)

G = nx.Graph([(0, 3), (1, 3), (1, 4), (2, 3)])
print('Edges: ' + repr(list(G.edges())))
assert nx.is_bipartite(G)
assert nx.is_connected(G)
matching = nx.bipartite.maximum_matching(G)
print('Matching: ' + repr(matching))
for u, v in matching.items():
    assert matching[v] == u
vertex_cover = nx.bipartite.to_vertex_cover(G, matching)
print('Vertex cover: ' + repr(vertex_cover))
for u, v in G.edges():
    assert u in vertex_cover or v in vertex_cover
```

produces:
```
$ python2.7 vertex_cover.py
Using version 1.11
Edges: [(0, 3), (1, 3), (1, 4), (2, 3)]
Matching: {0: 3, 1: 4, 3: 0, 4: 1}
Vertex cover: set([0, 1])
  File "vertex_cover.py", line 16, in <module>
    assert u in vertex_cover or v in vertex_cover
AssertionError
$ python3.6 vertex_cover.py
Using version 1.11
Edges: [(0, 3), (3, 1), (3, 2), (1, 4)]
Matching: {0: 3, 1: 4, 3: 0, 4: 1}
Vertex cover: {0, 1}
  File "vertex_cover.py", line 16, in <module>
    assert u in vertex_cover or v in vertex_cover
AssertionError
```

The vertex cover is incorrect, as verified by the `assert`s.

Current git version (21114082) fails in the same way.

jtorrents:
This PR fixes #2384.

Reimplement iteratively the inner function `_alternative_dfs` in
`_is_connected_by_alternating_path`. This function checks if nodes
in the graph are linked to target nodes by alternating paths, that
is, paths that alternate between matched and unmatched edges.

This PR also delegates to `_connected_by_alternating_paths`
computing the matched and unmatched edges because these are the same
for a given maximum matching and we do not need to recompute them
each time that we check if a node is linked to targets by an
alternating path.

I also changed the implementation of the computation of matched and
unmatched edges in order to avoid assuming orderable nodes.

Added a test with the example given in #2384 and another test with
unorderable nodes.

### DIFF
--- a/networkx/algorithms/bipartite/matching.py
+++ b/networkx/algorithms/bipartite/matching.py
@@ -300,7 +300,8 @@ def eppstein_matching(G, top_nodes=None):
             recurse(v)
 
 
-def _is_connected_by_alternating_path(G, v, matching, targets):
+def _is_connected_by_alternating_path(G, v, matched_edges, unmatched_edges,
+                                      targets):
     """Returns True if and only if the vertex `v` is connected to one of
     the target vertices by an alternating path in `G`.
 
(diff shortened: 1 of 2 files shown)


jfinkels:
Awesome. Sorry for all my bugs :\

jtorrents:
@jfinkels writing code is the only way to introduce bugs, so keep those bugs coming ;)

PR #2388, opened 2017-03-15
Generalized algorithm for force-directed layout.

Kdezaphi:
Implement an algorithm which, with a set of forces given by a force-directed layout compute this set and converge on a minimum for the given graph.


Kdezaphi:
Closed to submit a pull-request with no commit about the non-ascii characters.

edouardklein:
The new pull request:

https://github.com/networkx/networkx/pull/2389

PR #2389, opened 2017-03-15
Generalized algorithm for force-directed layout.

Kdezaphi:
Implement an algorithm which with a set of forces given by a force-directed layout compute this set and converge on a minimum for the given graph.

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -3,6 +3,8 @@
 #    Dan Schult <dschult@colgate.edu>
 #    Pieter Swart <swart@lanl.gov>
 #    Richard Penney <rwpenney@users.sourceforge.net>
+#    Edouard Klein <rdklein.fr>
+#    Kevin Dezaphi
 #    All rights reserved.
 #    BSD license.
 #
@@ -35,7 +37,9 @@ __all__ = ['circular_layout',
            'shell_layout',
            'spring_layout',
            'spectral_layout',
-           'fruchterman_reingold_layout']
+           'fruchterman_reingold_layout',
+           'force_atlas_2_layout',
+           'force_sim_until_no_displacement']
 
 
 def _process_params(G, center, dim):
@@ -231,11 +235,13 @@ def shell_layout(G, nlist=None, scale=1, center=None, dim=2):
 def fruchterman_reingold_layout(G, k=None,
                                 pos=None,
                                 fixed=None,
-                                iterations=50,
+                                iterations=0,
                                 weight='weight',
                                 scale=1,
                                 center=None,
-                                dim=2):
+                                dim=2,
+                                displacement_min=1,
+                                distance=None):
     """Position nodes using Fruchterman-Reingold force-directed algorithm.
 
     Parameters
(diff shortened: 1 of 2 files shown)


dschult:
This PR proposes to deprecate the current force directed methods and replaced with a couple of options. Could you give a very short description of what these methods are, how they differ and when to choose which one? 

I'm not looking for much -- even just "we try each one and see which looks best (the eyeball norm)" would be fine. I'm not up on the literature here -- it looks like one of the methods follows a paper that's been implemented by Gephi. Perhaps the answer to my question is just that we should add a link to the wikipedia article on [Force Directed Graph Drawing](https://en.wikipedia.org/wiki/Force-directed_graph_drawing).

Also you've got ```generalized_fruchterman_reingold_layout```.  Is this the same algorithm in some special case? How does it relate to the existing function.  Do we need both codes? Can we add a test in some case that checks for the same result?

Thanks!!

edouardklein:
Just acknowledging the question. I'll post an answer asap.

edouardklein:
All force directed layout algorithms work on the same principle: nodes repel each other according to some law, and attract each other according to some other law.

In a loop, one computes the effect of those attraction-repulsion forces and move the nodes accordingly, and then recompute the forces, etc.

These law can depend on the edges value, can be anisotropic, can be anything, really.

Fruchterman Reingold (also called "Spring layout") is just one particular choice of one attraction force and one repulsion force. Force-Atlas 2 is another particular choice, but both work according to the same principle.

To date, Fruchterman Reingold is the only force-directed layout algorithm available in networkx.

This PR proposes a generalized version of the force-directed layout mechanism: as long as the user provides the function that computes the forces, the `force_directed` function will do the loop on the graph using this forces until the nodes come at a rest (or as close to a rest as the `min_displacement` variable specifies).

One very nice property of `force_directed` is that given that the forces are non-stupid (i.e. they allow for the equilibrium to exist by having a distance range where attraction and repulsion compensate each other)  we can prove the convergence of the algorithm, because we use a variable step-size parameter (`alpha`) that make it so the actual displacement get smaller and smaller as the forces decreases, so that we don't overshoot the equilibrium point. This is something I intend to publish someday, but I have not found the time just yet. Note however that the convergence may be to a local minimum, but I don't think there's a simple way around that.

In order not to break the API, we re-implemented Fruchterman Reingold using `force_directed`, this is `generalized_fruchterman_reingold_layout`. The function is very long because of the docstrings, the conversion to numpy arrays, etc. but the actual implementation is three lines long:
```
def fruchterman_reingold(Dis, delta_unit, Pos, _):
    f_rep = (k * k / Dis**2).reshape((nnodes, nnodes, 1)) * delta_unit
    f_att = (W * Dis / k).reshape((nnodes, nnodes, 1)) * delta_unit
    return np.sum(f_att - f_rep, axis=1)
```

We just compute the attraction and repulsion forces and add them together.

Adding new force-directed layouts is child's play now. In order to illustrate, we wrote Force-Atlas 2, which the the flagship layout algorithm of Gephi (the paper is tied to Gephi's Java implementation). The implementation is a bit longer than 3 lines because Force-Atlas 2 has a very ugly "Gravity force" that must be computed from the center of the graph, but is is quite readable nonetheless (the force function is `force_atlas_2`, and the boilerplate-included, user-facing API function is `force_atlas_2_layout`.

We are working on an anisotropic repulsion function, where nodes are repelled more if they are on the same horizontal line. This is in order to make the labels readable (nodes will slip a bit vertically and nicely arrange themselves so that labels don't overlap).

The tests are very difficult to write, because those algorithms are very sensitive to the initial conditions, and because the end result ("looking good") is hard to express in code. There were no such tests on the existing spring layout algorithm either. Maybe we can try something anyway, I'll think about it, but if you have any idea, I'm all ears. 

If the code need some cleanup, or if you need real tests before merging, I'd be happy to work with you to get this PR merged.

dschult:
Thank you for a very good summary and comments.

We recently merged yet another force directed layout function using the Kamada-Kawai forces (based on shortest paths) and minimizing with ```scipy.optimize.minimize```.  Clearly there are many force directed methods. :)  Also, it seems that there are different algorithms for minimization as well as different algorithms for computing forces. 

I tried to resolve the conflicts between this branch and master using the web-based editor and that wasn't as effective as I'd hoped, e.g. I notice that spring_layout is now defined twice.  Maybe you can take one of them out and commit, or revert my two commits and resolve the conflicts yourself. Sorry about that.

I think its fairly hard to write more complicated tests for these algorithms because of the approximate nature of the minimization. What you have looks good.

I do notice that you have two FIXME comments. Would you like to resolve them?  As far as a name for the function ```force_directed```, maybe you can create a name that captures the idea of your method for minimization. 

Thanks!

edouardklein:
Hi !

I refactored the diff into one commit. I removed the FIXMEs, changed the name of `force_directed` to `force_sim_until_no_displacement` and removed the previous `fruchterman_reingold_layout` function for two reasons:
- the new implementation with `force_sim_until_no_displacement` is as fast on big graphs (and even marginally faster on the tests I did) as the previous sparse version
- the results look better, see eg for a path graph old: https://i.imgur.com/SkwvfOl.png  vs new https://i.imgur.com/z5Q6GhI.png

I cant reproduce the doc build fail for Python 2.7 on my machine, but I suspect this PR is not responsible as there is no non-ASCII char in the diff, and the build process is choking on a non ASCII char.

Do you think this PR is acceptable as-is, or do you want me to try to make all checks pass ? If so I think I would need some help with that.

edouardklein:
@dschult, it would be easier for us to have this merged into the master branch, because as is we must maintain our fork and dependency management is getting cumbersome.

We are ready to work on this PR if you dont think is is ready for merging yet. 

Please advise :)

dschult:
Thanks for the polite nudge...

The only trouble I have with this change is that it removes the old method which users may have come to count on. I guess it is three functions with one public facing and the other two support (private) functions.  I am ready to merge if we:

- put the previous functions back in place.
- rename in some way that makes sense to make both available.
- explain in the docstrings for each that the other exists and may be helpful.

Do these suggestions make sense?

edouardklein:
Hi,

I apologize about the nudge and the delay of this answer. Your suggestions make perfect sense. I'll try to get this done before the end of 2018.

rossbar:
I will go ahead and close this one since it's been inactive for a while. If interested in continuing work here, please feel free to reopen after resolving the merge conflicts and addressing the comments from the first round of review!

Issue #2394, opened 2017-03-19
Error message generation is broken for linalg.graphmatrix.incidence_matrix


nadesai:
Found this while I was working on something else.

To duplicate, try the following:
```
>>> G = nx.Graph()
>>> G.add_nodes_from([0,1,2])
>>> G.add_edges_from([(0,1),(1,2)])
>>> nx.linalg.incidence_matrix(G,edgelist=[(3,2)])
NameError: global name 'NetworkXError' is not defined
```
The error should be
```
NetworkXError: node 3 or 2 in edgelist but not in nodelist
```

nadesai:
Issue is a typo in the error handling code for `incidence_matrix`. Made a fix in #2395

PR #2395, opened 2017-03-19
Fix typos in generating NXError in networkx.linalg.graphmatrix.incidence_matrix

nadesai:
Fixes https://github.com/networkx/networkx/issues/2394.


jtorrents:
Thanks for fixing this!

PR #2397, opened 2017-03-21
[Fixes #2342] remove calls to plt.hold(), deprecated in mpl2.0

### LINKED ISSUE #2342: Matplotlib2.0 compatibility
There are some deprecation warning with Matplotlib2.0rc2:

trigering code:
````python3
# Networks graph Example : #https://github.com/ipython/ipywidgets/blob/master/examples/Exploring%20Graphs.ipynb
%matplotlib inline
from ipywidgets import interact
import networkx as nx
# wrap a few graph generation functions so they have the same signature

def random_lobster(n, m, k, p):
    return nx.random_lobster(n, p, p / m)

def powerlaw_cluster(n, m, k, p):
    return nx.powerlaw_cluster_graph(n, m, p)

def erdos_renyi(n, m, k, p):
    return nx.erdos_renyi_graph(n, p)

def newman_watts_strogatz(n, m, k, p):
    return nx.newman_watts_strogatz_graph(n, k, p)

@interact(n=(2,30), m=(1,10), k=(1,10), p=(0.0, 1.0, 0.001),
        generator={'lobster': random_lobster,
                   'power law': powerlaw_cluster,
                   'Newman-Watts-Strogatz': newman_watts_strogatz,
                   u'Erdős-Rényi': erdos_renyi,
                   })
def plot_random_graph(n, m, k, p, generator):
    g = generator(n, m, k, p)
    nx.draw(g)
    plt.title(generator.__name__)
    plt.show()
````

deprecation messages:
````
matplotlib\__init__.py:914: UserWarning: axes.hold is deprecated. Please remove it from your matplotlibrc and/or style files.
````

````
networkx\drawing\nx_pylab.py:126: MatplotlibDeprecationWarning: Future behavior will be consistent with the long-time
    default: plot commands add elements without first
    clearing the Axes and/or Figure.
  b = plt.ishold()
````

MridulS:
hold() has been removed from matplotlib  https://github.com/matplotlib/matplotlib/issues/3070, it will be `True` by default.
should we introduce our own `hold` logic which basically calls `plt.cla()` and `plt.clf()` when `hold=False`

@hagberg

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -38,7 +38,7 @@ __all__ = ['draw',
            'draw_shell']
 
 
-def draw(G, pos=None, ax=None, hold=None, **kwds):
+def draw(G, pos=None, ax=None, **kwds):
     """Draw the graph G with Matplotlib.
 
     Draw the graph as a simple representation with no node
@@ -59,10 +59,6 @@ def draw(G, pos=None, ax=None, hold=None, **kwds):
 
     ax : Matplotlib Axes object, optional
        Draw the graph in specified Matplotlib axes.
-
-    hold : bool, optional
-       Set the Matplotlib hold state.  If True subsequent draw
-       commands will be added to the current axes.
 
     kwds : optional keywords
        See networkx.draw_networkx() for a description of optional keywords.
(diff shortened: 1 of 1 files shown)


farid-fari:
can confirm that his works as intended, cannot comment on side effects or the removed lines, however

Issue #2400, opened 2017-03-23
Counter-intuitive API design for degree centrality in bipartite module


ericmjl:
I have noticed something kind of counter-intuitive with the bipartite module.

In the [degree centrality docs](https://networkx.github.io/documentation/development/_modules/networkx/algorithms/bipartite/centrality.html#degree_centrality), the degree centrality computation requires passing in a list of nodes from one partition, but returns the degree centrality distribution over all nodes. Just wondering, why is the API design as such? Perhaps I'm lacking some historical background here.

I think it might be more intuitive if it were one of the following API designs, perhaps? Here are my thoughts:

- Pass in a container of nodes, and return only degree centrality scores for that container of nodes.
- Do not require passing in a container of nodes, and return degree centrality scores for all nodes in both partitions.
- Make container of nodes an optional keyword, and return DC scores for those nodes only if a container of nodes is passed in, otherwise return all DC scores.

What do others think? As I mentioned earlier, I might be lacking some historical perspective here, so please pardon me if I misunderstood the original API design.

jtorrents:
Hi @ericmjl, the API design is a result of the fact that we don't have a proper bipartite graph class in NetworkX, the convention (as explained in the module level docstrings of the bipartite package) is to use a node attribute to keep track of which node set each node belongs to.

As the Notes section of the docstring of `degree_centrality` explains, this measure is the degree of each node divided by the maximum degree possible for each node. In the unipartite case this is `n-1` (where `n` is the number of nodes in the graph) but for the bipartite case the maximum degree of a node in a bipartite node set is the number of nodes in the opposite node set.

Thus we need to know to which node set each node belongs to in order to properly compute centrality metrics (and others) in bipartite graphs. This can be computed using `bipartite.sets`, but the trouble comes in the case of disconnected graphs because more than one valid solution exists. To deal with this ambiguity we require the user to specify the nodes in one of the bipartite node sets as an argument for most bipartite functions to avoid returning wrong results depending on the concrete solution found in `bipartite.sets` (see #2375 for recent work on this). 

Do you think that we should update the docstrings for bipartite centrality functions to make this more clear?

ericmjl:
@jtorrents thanks for the explanation! Yes, that definitely clears up the historical perspective for me.

Having seen the explanation, I'm now just wondering whether it might be better to harness the 'bipartite' keyword argument rather than pass in a list of nodes? Here's what I'm thinking in terms of API/code changes:

```python
[...]

PR #2402, opened 2017-03-26
Improve bipartite documentation.

jtorrents:
This PR addresses #2400

Improve bipartite documentation by clarifying that the use of a node
attribute named `bipartite` with values 0 and 1 in order to
distinguish bipartite node sets is only a convention which is not
enforced in the source code. Also state explicitly that bipartite
functions do not check that the input graph is actually bipartite
nor than the container with a bipartite node set contains all nodes
from that bipartite node set.

Added links form all relevant function docstring to the module
level documentation where all this is explained.

Also added links to some custom exceptions to the generated
documentation that were missing.

### DIFF
--- a/doc/source/reference/exceptions.rst
+++ b/doc/source/reference/exceptions.rst
@@ -17,8 +17,16 @@ Exceptions
 
 .. autoclass:: networkx.NetworkXNoPath
 
+.. autoclass:: networkx.NetworkXNoCycle
+
 .. autoclass:: networkx.NodeNotFound
 
 .. autoclass:: networkx.NetworkXUnbounded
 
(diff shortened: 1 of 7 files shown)


jtorrents:
@ericmjl Please take a look at this changes in the documentation of the bipartite package. What do you think?

ericmjl:
@jtorrents thanks for the PR! Yes, I think the new documentation makes it clearer as to why the API is designed as such. 

I happen to be making a DataCamp course on network analysis, and for the sections on bipartite graphs, I can add in this information there too!

jtorrents:
@ericmjl Thanks for looking at this. It's great that you add this information in your course on network analysis!

PR #2406, opened 2017-03-28
Return collections from draw()

Mpaull:
Allows creation of matplotlib animations.


dschult:
This only returns the collections from ```draw_networkx``` and not from ```draw```.
The docs should be changed to reflect this change.

PR #2413, opened 2017-04-02
Fix all tests for 3.6

jtorrents:
This PR supersedes #2319 

Fixed all remaining doctests ordering issues for python 3.6 without using the `+SKIP` directive. I think that most docstrings examples are as clear as before, maybe with the exception of `networkx.readwrite.adjlist`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
-  - "3.6-dev"
+  - "3.6"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6-dev"
 
 cache:
   directories:
--- a/networkx/algorithms/centrality/flow_matrix.py
+++ b/networkx/algorithms/centrality/flow_matrix.py
@@ -16,9 +16,9 @@ def flow_matrix_row(G, weight='weight', dtype=float, solver='lu'):
     C = solvername[solver](L, dtype=dtype) # initialize solver
     w = C.w # w is the Laplacian matrix width
     # row-by-row flow matrix
-    for u,v,d in G.edges(data=True):
+    for u,v in sorted(sorted((u,v)) for u,v in G.edges()):
         B = np.zeros(w, dtype=dtype)
-        c = d.get(weight,1.0)
+        c = G[u][v].get(weight,1.0)
         B[u%w] = c
         B[v%w] = -c
         # get only the rows needed in the inverse laplacian 
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -144,14 +144,18 @@ class DiGraph(Graph):
     The fastest way to traverse all edges of a graph is via
     adjacency(), but the edges() method is often more convenient.
 
-    >>> for n,nbrsdict in G.adjacency():
-    ...     for nbr,eattr in nbrsdict.items():
+    >>> for n, nbrsdict in G.adjacency():
+    ...     for nbr, eattr in nbrsdict.items():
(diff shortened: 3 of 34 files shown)


jtorrents:
Still needs work.

.travis.yml
```diff
@@ -10,6 +10,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6-dev"
```
hagberg: Looks like travis has 3.6 so we can drop the -dev.

networkx/classes/digraph.py
```diff
@@ -144,14 +144,18 @@ class DiGraph(Graph):
...
-    ...     for nbr,eattr in nbrsdict.items():
+    >>> for n, nbrsdict in G.adjacency():
+    ...     for nbr, eattr in nbrsdict.items():
     ...        if 'weight' in eattr:
-    ...            (n,nbr,eattr['weight'])
-    (1, 2, 4)
-    (2, 3, 8)
-    >>> list(G.edges(data='weight'))
-    [(1, 2, 4), (2, 3, 8), (3, 4, None), (4, 5, None)]
+    ...            # Do something useful with the edges
```
hagberg: I like this compromise of "do something useful"

networkx/algorithms/centrality/flow_matrix.py
```diff
@@ -16,9 +16,9 @@ def flow_matrix_row(G, weight='weight', dtype=float, solver='lu'):
     C = solvername[solver](L, dtype=dtype) # initialize solver
     w = C.w # w is the Laplacian matrix width
     # row-by-row flow matrix
-    for u,v,d in G.edges(data=True):
+    for u,v in sorted(sorted((u,v)) for u,v in G.edges()):
```
hagberg: I probably shouldn't have hidden this bug fix in a larger PR.
But I don't mind leaving it here.

hagberg:
Looks like the doctests are failing because 

```
Collecting scikits.sparse
  Downloading scikits.sparse-0.2.zip (221kB)
    100% |████████████████████████████████| 225kB 4.2MB/s 
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-w8kvc_fh/scikits.sparse/setup.py", line 38, in <module>
        from Cython.Distutils import build_ext
    ModuleNotFoundError: No module named 'Cython'
```

jtorrents:
Cython should be installed by the time we install `scikits.sparse`. I'm not sure why only fails for 3.6. Maybe the `-dev` suffix was the problem?

jtorrents:
Now all tests pass for python 3.6 without optional dependences. Looking more closely, the problem for 3.6 miniconda seems to be:

```
UnsatisfiableError: The following specifications were found to be in conflict:

  - gdal -> libnetcdf 4.3.2 -> curl 7.38.0 -> openssl 1.0.1*
  - gdal -> numpy 1.9* -> python 2.7*
  - python 3.6*

Use "conda info <package>" to see the dependencies for each package.
CondaEnvironmentNotFoundError: Could not find environment: test-environment .
You can list all discoverable environments with `conda info --envs`.
```

But I still don't understand why it only fails for 3.6

hagberg:
Maybe a conda expert can help us?  

I suggest we merge this ASAP.

jtorrents:
Yes, that would be great. I'm not very familiar with conda. It seems that the problem is that it's trying to install an outdated gdal package that depends on python 2.7. Which causes an error, and the conda environment `test-environment` is not created and thus installing `scikits.sparse` fails.

I found this recent question on stack overfow: 

http://stackoverflow.com/questions/42571957/installing-gdal-on-ubuntu

But I don't know how to solve this in the `.travis.yml` configuration file.

hagberg:
I merged this even though it fails the conda install 3.6 tests.
We'll either figure that out ASAP or disable the tests.

PR #2416, opened 2017-04-02
Workaround for gdal python3.6 at travis and more doctests fixes

### LINKED ISSUE #2415: Travis with Python3.6 miniconda fails
Fails with a gdal conflict error.
Can a conda expert help us figure out how to fix this?

```
UnsatisfiableError: The following specifications were found to be in conflict:

  - gdal -> libnetcdf 4.3.2 -> curl 7.38.0 -> openssl 1.0.1*
  - gdal -> numpy 1.9* -> python 2.7*
  - python 3.6*

Use "conda info <package>" to see the dependencies for each package.
CondaEnvironmentNotFoundError: Could not find environment: test-environment .
You can list all discoverable environments with `conda info --envs`.
```

jtorrents:
This PR fixes #2415 (well not really, but at least it doesn't kill the build for python 3.6 with optional dependencies).

This workaround still does not install gdal for python 3.6 but it doesn't kill the build, and gdal installation works for all other python versions. It seems that the problem is that conda for python 3.6 installs the latest version of gdal (2.1.3) which is not compatible with the package libgdal-dev provided by Travis version of Ubuntu.

Also more doctests fixes for python 3.6 at subgraph centrality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,9 +98,10 @@ before_install:
 install:
   ### Install any prerequisites or dependencies necessary to run the build.
   - if [[ "${OPTIONAL_DEPS}" == miniconda ]]; then
-      DEPS="pyyaml numpy scipy matplotlib Cython pandas pip gdal";
+      DEPS="pyyaml numpy scipy matplotlib Cython pandas pip";
       conda create -n test-environment python=$TRAVIS_PYTHON_VERSION $DEPS ;
       source activate test-environment;
+      conda install gdal;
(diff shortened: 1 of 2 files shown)


jtorrents:
The failure in appveyor is from `test_asyn_lpa.test_two_communities` and unrelated to this PR.

jtorrents:
Hmm, the last commit did not trigger the Travis build, not sure why.

PR #2418, opened 2017-04-04
Fix closeness centrality for DiGraphs

bioglio:
Closeness centrality for a node u is calculated as C(u) = \frac{n - 1}{\sum_{v=1}^{n-1} d(v, u)}, that is the reciprocal of the sum of the shortest path distances from all n-1 other nodes to u, but in closeness_centrality method it is implemented as the reciprocal of the sum of the shortest path distances from u to all n-1 other nodes. For undirected graphs, using d(u,v) instead of d(v,u) is not a problem, but for directed graph it can produce totally different results. For this reason I have slightly modified the closeness centrality function, by using predecessors of a node instead of successors in case of undirected graph through the new method "single_target_shortest_path_length" added to the file _networkx/algorithms/shortest_paths/unweighted.py_: for the sake of completeness, I have also added the function "single_target_shortest_path" to the same file.

I really appreciate any feedback, and please feel free to modify any line of code ;)

### DIFF
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -357,6 +357,10 @@ def _single_shortest_path(adj, firstlevel, paths, cutoff, join):
             paths for starting nodes, e.g. {source: [source]}
         cutoff : int or float
             level at which we stop the process
+        join : function
+            function to construct a path from two partial paths. Requires two
+            list inputs `p1` and `p2`, and returns a list. Usually returns
+            `p1 + p2` (forward from source) or `p2 + p1` (backward from target)
     """
     level = 0                  # the current level
     nextlevel = firstlevel
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -13,6 +13,8 @@ __author__ = """Aric Hagberg (hagberg@lanl.gov)"""
 __all__ = ['bidirectional_shortest_path',
            'single_source_shortest_path',
            'single_source_shortest_path_length',
+           'single_target_shortest_path',
+           'single_target_shortest_path_length',
            'all_pairs_shortest_path',
            'all_pairs_shortest_path_length',
            'predecessor']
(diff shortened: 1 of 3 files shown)


hagberg:
Yes, good catch.

networkx/algorithms/shortest_paths/unweighted.py
```diff
@@ -75,6 +77,64 @@ def single_source_shortest_path_length(G,source,cutoff=None):
...
+    level = 0                  # the current level
+    nextlevel = {target:1}  # dict of nodes to check at next level
+
+    while nextlevel:
+        thislevel = nextlevel  # advance to next level
+        nextlevel = {}         # and start a new list (fringe)
+        for v in thislevel:
+            if v not in seen:
+                seen[v] = level # set the level of vertex v
+                if hasattr(G, 'pred'): # if G is a DiGraph..
```
hagberg: Moving this check out of the loop would improve the performance.

bioglio:
thank you very much for the comment! 
I have changed the condition for checking if the graph is directed, moving it out of the loop.

jfinkels:
Is it really necessary to add these new functions `single_target_shortest_path` that are essentially verbatim copies of existing functions? Can we just use `G.reverse()`?

bioglio:
Thank you for your feedback, jfinkels!
We can use `G.reverse()`, but in my opinion it is less efficient: in fact this method changes the original graph, or creates a new copy of the graph, using additional memory: in any case, to compute the reverse of a graph is time consuming.
About the new methods, I agree with your critics. I have created two new methods that are copies of existing ones just because their meaning is different: using different names for methods doing different things should improve clearness, in spite of code reuse. A different solution could be to modify directly `single_source_shortest_path` and `single_source_shortest_path_length` by adding a "reverse" argument, that in case of directed graph uses predecessors instead of neighbours: in this case we avoid the copy-and-paste of code, but users looking for that particular functionality should read the documentation carefully. I don't know which solution is better, but if you prefer to reduce the copy-and-paste of code I can implement the alternative solution.

jfinkels:
Perhaps you can use the `networkx.utils.reversed` context manager for this purpose: it calls `DiGraph.reverse(copy=False)` which seems to be memory-efficient.

dschult:
It looks like the paths for ```..._target_shortest_path``` are backwards.
I also unified the code by creating helper functions for ```_path_length``` and ```_path``` 
Then added tests and pep8.

@bioglio can you [add permission](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) for me to commit to this PR? It is supposed to appear as an option in the right sidebar of this pull request. Something like: ```allow edits from maintainers```
If  that's any trouble at all I will just open a new PR to finish this one.

bioglio:
@dschult thank you very much for your amazing job, the result is great!

PR #2420, opened 2017-04-10
Fluid communities algorithm implementation

FerranPares:
Hello there,

We are interested in adding the fluid communities algorithm [1] to networkx. It's a community detection algorithm competitive both in performance (close to multilevel) and in scalability (close to LPA).
So far we have a tested version of the code based on the LPA code originally found in  "networkx/networkx/algorithms/community/asyn_lpa.py".

Best regards,

[1] https://arxiv.org/pdf/1703.09307.pdf

### DIFF
--- a/doc/source/reference/release_2.0.rst
+++ b/doc/source/reference/release_2.0.rst
@@ -106,3 +106,8 @@ API changes
    Most of the shortest_path algorithms now raise a NodeNotFound exception
    when a source or a target are not present in the graph.
 
+* [`#2420 <https://github.com/networkx/networkx/pull/2420>`_]
+   New community detection algorithm provided. Fluid Communities is an asynchronous
+   algorithm based on the simple idea of fluids interacting in an environment, 
+   expanding and pushing each other. The algorithm is completly described in 
+   [`https://arxiv.org/pdf/1703.09307.pdf <https://arxiv.org/pdf/1703.09307.pdf>`_]. 
--- a/networkx/algorithms/community/__init__.py
+++ b/networkx/algorithms/community/__init__.py
@@ -1,4 +1,5 @@
 from networkx.algorithms.community.asyn_lpa import *
+from networkx.algorithms.community.asyn_fluidc import *
 from networkx.algorithms.community.centrality import *
 from networkx.algorithms.community.community_generators import *
 from networkx.algorithms.community.kclique import *
--- a/networkx/algorithms/community/asyn_fluidc.py
+++ b/networkx/algorithms/community/asyn_fluidc.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
-#    Copyright (C) 2015
+#    Copyright (C) 2017
 #    All rights reserved.
 #    BSD license.
+#    Author: Ferran Parés <ferran.pares@bsc.es>
 """Asynchronous Fluid Communities algorithm for community detection."""
 
 from collections import Counter
 import random
+from networkx import NetworkXError
+from networkx import is_connected
 from networkx.utils import groups
 from networkx.utils.decorators import not_implemented_for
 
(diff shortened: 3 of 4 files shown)


aryamccarthy:
_Totally_ off-topic, but how did you produce the stacked standard deviation/mean plots in your paper, @FerranPares ?

aryamccarthy:
And more on-topic: you should include tests for your code. I understand that it's difficult for an asynchronous algorithm (e.g. #2339 and #2341), but it's important.

FerranPares:
@aryamccarthy Before answering your question, we have already the final version of our paper available at arxiv ([https://arxiv.org/pdf/1703.09307.pdf](https://arxiv.org/pdf/1703.09307.pdf)). Regarding your question, I guess you refer to the plots showing the performance of several CD algorithms in terms of NMI. For that purpose, first of all we generate several graphs with different properties (combining number of vertices and mixing parameter) through a generative process known as LFR benchmark.  However, for each set of parameters (6 graph sizes and 25 mixing parameters) we generated 20 graphs with same properties (but they are different since the generative process is stochastic). Thanks to that, we can compute the mean and standard deviation performance among these 20 graphs.

FerranPares:
@aryamccarthy I will include some tests ASAP.

aryamccarthy:
@FerranPares Perhaps my question is simpler than you read it to be. My thesis was in a similar vein—using LFR benchmarks to appraise a CD algorithm—so I understand the benchmark process. I'm more curious about the matplotlib (or other plotting tool) code to plot standard deviation and means on one plot, without the axis numbers overlapping.

FerranPares:
@aryamccarthy I have used matplotlib tool. The code to plot everything in the same plot is using plt.subplot:
```python
fig, (ax0, ax1) = plt.subplots(2, 1, sharex=True, gridspec_kw = {'height_ratios':[1, 3]})
plt.subplots_adjust(wspace=0, hspace=0)
```
Furthermore, you can specify axis numbers you wish to display using:
```python
ax0.yaxis.set_ticks()
ax1.yaxis.set_ticks()
```
and avoid overlapping specifying the axis limits through:
```python
ax0.set_ylim()
ax1.set_ylim()
```

dschult:
These commits look good -- just need to correct the import of nx, I suspect.
You might like to add an ```# Author: Name <email>``` line, and update the copyright to 2015-2017.
The only other thing is to add a short description to ```doc/source/reference/release_2.0.rst``` but I can do that if you prefer.

Thanks very much!

FerranPares:
@dschult Ok, I just fixed those import problems. I apply the changes you suggest right now. About documentation, I can do a first version, so you just have to review it (I guess that is going to be easier).

FerranPares:
@dschult I have already included the authorship and changed the copyright year. Furthermore, I have added documentation related to this pull request at `doc/source/reference/release_2.0.rst` as a first version.
Regarding the unsuccessful checks, I'm having issues to ensure same behaviour of python random.seed() among all Python versions. I would try to solve it tomorrow, but if u have any suggestion about how to ensure same behaviour it will be very welcome.

dschult:
My attempt to make the test work for all versions also failed. @aryamccarthy do you have any ideas for how to do this?  I suppose we can just remove the ```test_five_clique_ring``` test.  What is best here?

FerranPares:
@dschult @aryamccarthy I have not been able to ensure random seed across all python versions. I would say to remove the `test_five_clique_ring` test. From my point of view this test is not crucial at all.

FerranPares:
Hello @dschult,
I just saw that FluidC algorithm does not appear in the documentation at the following link [http://networkx.readthedocs.io/en/latest/reference/algorithms/community.html](http://networkx.readthedocs.io/en/latest/reference/algorithms/community.html).
Guess that I miss to add FluidC documentation at ```networkx/doc/reference/algorithms/community.rst```.
Could we reopen the pull request just to add this? Since it is a minor change I don't really know if it really requires to reopen a pull request, maybe you can just add the following Fluid Communties documentation into the ```community.rst``` file:
```
Fluid Communities
-----------------
.. automodule:: networkx.algorithms.community.asyn_fluidc
.. autosummary::
   :toctree: generated/

   asyn_fluidc
```

dschult:
A new small PR would be great -- And if you include a reference to #2420 in the commit comments it will link to the original PR automatically.  Thanks for spotting this.

PR #2424, opened 2017-04-14
Adding support for windmill graphs to the community generators

DonQuixoteDeLaMancha:
Take care, sir," cried Sancho. "Those over there are not giants but windmills. Those things that seem to be their arms are sails which, when they are whirled around by the wind, turn the millstone."

### DIFF
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -473,24 +473,29 @@ def windmill_graph(n, k):
     selecting one point from each, and contracting all of the selected points.
     Alternatively, one could generate `n` cliques of size `k-1` and one node
     that is connected to all other nodes in the graph.
+
     Parameters
     ----------
     n : int
       Number of cliques
     k : int
       Size of cliques
+
     Returns
     -------
     G : NetworkX Graph
       windmill graph with n cliques of size k
+
     Raises
     ------
     NetworkXError
         If the number of cliques is less than two
         If the size of the cliques are less than two
+
     Examples
     --------
     >>> G = nx.windmill_graph(4, 5)
+
     Notes
     -----
     The node labeled `0` will be the node connected to all other nodes.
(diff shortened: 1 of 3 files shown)


mcognetta:
Whats up with Travis here?

networkx/generators/community.py
```diff
@@ -464,3 +464,47 @@ def ring_of_cliques(num_cliques, clique_size):
...
+
+def windmill_graph(n, k):
+    """Generate a windmill graph.
+    A windmill graph is a graph of `n` cliques each of size `k` that are all
+    joined at one node.
+    It can be thought of as taking a disjoint union of `n` cliques of size `k`,
+    selecting one point from each, and contracting all of the selected points.
+    Alternatively, one could generate `n` cliques of size `k-1` and one node
+    that is connected to all other nodes in the graph.
+    Parameters
```
jtorrents: You need to add a blank line between sections in the docstrings.

networkx/generators/community.py
```diff
@@ -464,3 +464,47 @@ def ring_of_cliques(num_cliques, clique_size):
...
+    Raises
+    ------
+    NetworkXError
+        If the number of cliques is less than two
+        If the size of the cliques are less than two
+    Examples
+    --------
+    >>> G = nx.windmill_graph(4, 5)
+    Notes
+    -----
```
jtorrents: The doctest fails in travis because there is no blank line between the `Examples` and `Notes` section and the docstring assumes that the output of `G = nx.windmill_graph(4, 5)` should be the rest of the docstring.

jtorrents:
@mcognetta, see my inline comments for why the doctest fail.

PS: I also like the single use @DonQuixoteDeLaMancha account for windmill graphs. I guess that now we know who is behind that account ;)

Issue #2315, opened 2016-11-18
Add SPQR tree for finding triconnected components and 2-vertex cuts


jfinkels:
NetworkX has a function for computing [biconnected components][1], maximal 2-vertex-connected subgraphs. It would be nice to have a function for computing [triconnected components][2], maximal 3-vertex-connected subgraphs, those subgraphs in which the removal of three vertices causes the graph to become disconnected. (Note: even though a subgraph may be 3-vertex-connected, the entire graph itself may only be 2-vertex-connected.) The [SPQR tree][2] data structure gives the algorithm for finding triconnected components, as well as an algorithm for finding each 2-vertex cut in the graph (that is, a pair of vertices whose removal disconnects the graph).

[1]: https://en.wikipedia.org/wiki/Biconnected_component
[2]: https://en.wikipedia.org/wiki/SPQR_tree

jtorrents:
That would be a great addition to NetworkX. The tricky thing might be to do a linear time implementation. I attach here a dense paper on the subject for reference. 
[triconnected_components.pdf](https://github.com/networkx/networkx/files/621161/triconnected_components.pdf)

JamesClough:
Is anybody else working on this? If not then I will have a go.

jtorrents:
Hi @JamesClough , I'm not sure if @jfinkels is working on this. A PR adding this feature will be very welcome. I'll help you in any way I can to get it merged. Thanks!

jfinkels:
I'm not working on it right now. Go for it!

PR #2430, opened 2017-04-21
Add SPQR tree for finding triconnected components

JamesClough:
As described in #2315 
This addition is reasonably complex and not yet finished but I will push incremental progress to this pull request as I go so that anyone interested in helping can do so, and so that mistakes I make can be seen earlier.
This commit contains just the function to calculate the palm tree of a graph, which is a precursor step for the SPQR tree but is perhaps useful in its own right (or for other algorithms).

### DIFF
--- a/networkx/algorithms/spqr.py
+++ b/networkx/algorithms/spqr.py
@@ -1,0 +1,92 @@
+"""
+Find SPQR tree of a biconnected graph
+"""
+import networkx as nx
+from operator import itemgetter
+
+__authors__ = "\n".join(['James Clough <james.clough91@gmail.com>'])
+#    Copyright (C) 2017 by
+#    James Clough <james.clough91@gmail.com>
+#    All rights reserved.
+#    BSD license.
+
+__all__ = ['palm_tree']
+
+def palm_tree(G, root):
+    """ Returns a Palm Tree of undirected simple graph G
+
+    P has the same vertex set as G, and the same edges although they are now
+    directed.
+    A palm tree consists of a directed spanning tree of G where all tree arcs
+    go from nodes with a lower number to a higher one, and `fronds' which are
+    edges from higher numbered nodes to lower numbered ones.
+    The tree arcs and fronds together make up all of the edges of the original
+    graph G.
+
+    Edges in P have the attribute 'edge_type' which is either 'tree_arc' or
+    'frond'.
+    Nodes in P have the attribute 'number' as described above.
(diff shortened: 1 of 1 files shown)


jfinkels:
This sounds similar to the "chain decomposition" I added in pull request #2284; what is the relationship between the chain decomposition and the palm tree? (I've been out of the mindset for a while so I don't remember.) Just trying to avoid duplicating code.

JamesClough:
They do certainly seem similar - I think if you already had the chain decomposition you could find a palm tree more easily. It's probably not worth taking that approach here though since in this case there are other things needed for the SPQR tree which are calculated in the same depth-first-search the palm tree uses.

jfinkels:
Can we extend the chain decomposition function so that it computes the
extra information you need?

JamesClough:
So to merge these would I guess involve having chains.py call `palm_tree` to do its DFS, instead of the `_dfs_cycle_forest` function in chains.py and then ignore the extra information its getting. 
But most of the logic in `palm_tree_dfs` worrying about these other attributes (like the lowpoint1, lowpoint2 functions) - so chains.py would be wasting quite a bit of time doing all of that.
I think that will turn out to be a lot more trouble than its worth. The only repeated code is really just the DFS which is very simple. But to combine everything would make chains.py quite a bit more complicated, especially as I'm trying to stick to the notation used in the algorithm in the paper linked in #2315. So I'd recommend keeping them separate.

As an aside though - is the recommended practice for this kind of function to define the dfs function (like `_dfs_cycle_forest` or  `palm_tree_dfs`) inside the main function that users would call from the outside? Or does it not matter too much?

sjackman:
Here's my poor-man's version of `triconnected_components`, in case anyone else stumbles on this issue.

```python
    def triconnected_components(g):
        "Return the triconnected components of the graph."
        components = []
        for component in nx.biconnected_components(g):
            if len(component) < 3:
                components.append(component)
                continue
            try:
                cuts = next(nx.all_node_cuts(
                    g.subgraph(component),
                    k=2, flow_func=nx.algorithms.flow.shortest_augmenting_path))
                if len(cuts) > 2:
                    components.append(component)
                    continue
                assert len(cuts) == 2
                subcomponents = list(nx.connected_components(g.subgraph(component - cuts)))
                if len(subcomponents) == 1:
                    components.append(component)
                    continue
                components += subcomponents
                components.append(cuts)
            except StopIteration:
                components.append(component)
        return components
```

PR #2456, opened 2017-05-17
Pass heuristic function full node object, not just names

tb7554:
Currently only the node names are passed to the heuristic function, which is quite limiting (for example, to calculate distance function in your example you must name the nodes by the coordinates (x,y)). If the node object was passed it would be possible to write heuristic functions which can use all node properties to return the result. 

I have included my proposal with a try statement so as not to interfere with legacy code.


dschult:
This does not pass the node object.  In fact it only passes the attribute dictionary for the node.
I think your use case is easy to implement from the nodes themselves. Pass the nodes in -- then have the function get any attributes needed by looking them up using G.node[n]['attr_name'].

PR #2458, opened 2017-05-19
Add graph view classes for nodes/edge/degrees

dschult:
This is just a clean up before implementing views.

### DIFF
--- a/doc/source/reference/release_2.0.rst
+++ b/doc/source/reference/release_2.0.rst
@@ -12,13 +12,14 @@ Please send comments and questions to the networkx-discuss [mailing list](http:/
 API changes
 -----------
 * Base Graph Class Changes
-  With the release of NetworkX 2.0 we are moving towards an iterator reporting API.
+  With the release of NetworkX 2.0 we are moving towards a view/iterator reporting API.
   We used to have two methods for the same property of the graph, one that returns a
-  list and one that returns an iterator. With 2.0 we have removed the one that returns
-  a list and renamed the iterator returning method as the original one. For example,
-  ``G.nodes()`` used to return a list and ``G.nodes_iter()`` used to return an iterator, now
-  ``G.nodes()`` returns an iterator and ``G.nodes_iter()`` is removed. Routines that used to
-  return a dict now return an iterator of (key,value) 2-tuples, so that dict(G.degree())
+  list and one that returns an iterator. With 2.0 we have replaced them with a view.
+  A view is a read-only object that is quick to create, automatically updated, and 
+  provides basic access like iteration, membership and set operations where appropriate.
+  For example, ``G.nodes()`` used to return a list and ``G.nodes_iter()`` an iterator.
+  Now ``G.nodes()`` returns a view and ``G.nodes_iter()`` is removed. ``G.degree()``
+  returns a view with (node, degree) iteration, so that dict(G.degree())
   returns a dict keyed by node with degree as value.
   The old behavior
 
@@ -32,10 +33,18 @@ API changes
 
     >>> G = nx.complete_graph(5)
     >>> G.nodes()
-    <dictionary-keyiterator at 0x10898f470>
+    NodesView([0, 1, 2, 3, 4])
     >>> list(G.nodes())
     [0, 1, 2, 3, 4]
 
+  New feature include lookup of node and edge data from the views, property
+  access without parentheses, and set operations.
+
+    >>> G.add_node(3, color='blue')
+    >>> G.nodes[3]
+    'blue'
+    >>> G.nodes & {3, 4, 5}
+    {3, 4}
 
   The following methods have changed:
     * Graph/MultiGraph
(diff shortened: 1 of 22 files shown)


dschult:
Does anyone have info about changes in python3.6 that would make node order important for to_numpy_matrix and friends?  I can get rid of the errors by adding a nodelist to the call.  But I want to understand what might be causing this "strange" ordering of nodes.  Reading the change log it looks like storage in dicts has been updated in python3.6 to save as much as 30% in dict storage. I'd love to know more about it.   Thanks

hagberg:
Significant changes in dictionaries in 3.6.  This is a decent discussion with pointers http://stackoverflow.com/questions/39980323/dictionaries-are-ordered-in-python-3-6

dschult:
@hagberg that was a very helpful post (stackoverflow rocks).
We just need to keep in mind that adding edges before adding nodes affects order of nodes.

Turns out this twist on a test also uncovered another bug in the tests for convert_numpy where we were testing empty graphs due to an edge iterator used twice and so being exhausted. 

Tests now pass for the views...  I've still got a lot of docs to change but what do people think of the structure here?

dschult:
Summary:
This PR adds views for nodes, edges and degree for all base graph classes. A view means a read-only object that is quick to create, automatically updated when graph changes, and provides basic access like ```n in V```, ```for n in V```, ```V[n]``` and set operations.

NodeView:  options: data, default  (make ```__iter__``` and ```__contains__``` act on (node, data) 2-tuples)

    NV=G.nodes()
    assert(2 in NV)
    for n in NV: print(n)
    datadict = NV[2] 
    NV & {1,2,3}
    
    NVdata=G.nodes(data=True)
    assert( (2,NV[2]) in NVdata)
    for n,dd in NVdata: print(dd['color'])
    assert(NVdata[2] == NV[2])    

EdgeView: options: nbunch; data, default; keys  

    EV=G.edges()
    assert((2,3) in EV)
    for u,v in EV: print((u,v))
    EV & {(1,2), (3,4)}
    
    EVdata=G.edges(data='color')
    assert((2,3,'blue') in EVdata)
    for u,v,c in EVdata: print("({},{}) has color:{}".format(u,v,c))
    EVdata & {(1,2,"blue"), (2,3,"red")}

    EVnbunch=G.edges(nbunch=2)
    assert((2,3) in EVbunch)
    assert((5,6) in EVbunch)   #  at the moment, nbunch is ignored in __contains__
    for u,v in EVbunch: assert(u==2 or v==2)
    
    EVmulti=MG.edges(keys=True)
    assert((2,3,0) in EVmulti)   #  keys==True  enforces 3-tuples for contains
    assert_raises((2,3) in EVmulti, ValueError)
    for u,v,k in EVmulti: print(u,v,k)

DegreeView:  options:  nbunch, weight (default=None)   

    DV=G.degree()
    assert(DV[2] == 4)
    s = sum(deg for n,deg in DV)
    
    DVweight = G.degree(weight="span")
    assert(DVweight[2] == 6)   #  default value for edges without "span" is 1
    s = sum(span for n,span in DVweight    
    
    DVnbunch = G.degree(nbunch=(2,3))
    assert(len(list(DVnbunch)) == 2)     # iteration only passes over nbunch

Note: NodeView with ```data is not None``` violates an implied contract that contains and getitem act on the same types of objects. 

Note: EdgeView doesn't implement __getitem__ yet. I am thinking about how best to do that for all the various cases.  Maybe just make getitem only work for 2-tuples in graph (3-tuples in multigraph). This would again violate contains and getitem agreement when ```data is not None```

Note: DegreeView can still lookup any node even if nbunch is specified. nbunch only affects iteration. DiDegreeView looks both directions, InDegreeView and OutDegreeView look in a directed way.
DegreeViews do not provide set operations.

dschult:
View Data Lookup:

Is it better to use ```G.edges()[(2,3)]```  or make G.edge a read-only lookup  ```G.edge[2,3]```?

Similarly, is it better to use ```G.nodes()[3]``` or make G.node a read-only lookup ```G.node[3]```?

In both cases we could move the read/write object e.g. ```G.node``` to ```G._node``` and replace it with a read-only Mapping.  In both cases the resulting dict would be read-write. 
So that ```G.node[3]["color"]="blue"``` still works, but ```G.node[3]="blue"``` is forbidden and leads to an error. 

For edges, I would prefer a 2/3-tuple argument like ```G.edge[2,3]["color"]="blue"``` leaving the adjacency style argument for ```G.adj[2][3]["color"]="blue"```

dschult:
Questions about set operations on undirected edges:

Directed edges map nicely onto 2-tuples, so the set operators work well. But undirected edges are not uniquely represented by 2-tuples. So the set operations become tricky.

    G = nx.path_graph(3)
    ev = G.edges()      # ev -> EdgeView([(0, 1), (1, 2)])
    some_edges = {(0, 1), (0, 2), (1, 0)}
    ev & some_edges     # should this be {(0,1)} or {(1,0)} or {(1,0), (0,1)}

The abc Set class implements __and__ as:   ```(value for value in other if value in self)```
which is in this case leads to both (1,0) and (0,1).  Fundamentally this occurs because ```some_edges``` duplicates an edge and sets aren't supposed to do that.

My question: Should edge set operations check ```other``` for duplicate representation of edge tuples?  

A related question: Should edge set operations distinguish between the directed edge (0,1) and the undirected edge (0,1)?  If so, I see two options: make set operations only work with other ```EdgeViews````, or change undirected edge representation to ```frozenset(u,v)```.

Ideas?

dschult:
I'm ready to squash and merge these view objects.
I'll continue with doc changes. But I think its time to get this much merged so the other issues like G._node can be explored.  Any comments before I merge?

PR #2459, opened 2017-05-20
[WIP] edge connectivity algorithms

Erotemic:
This PR is a work in progress. I've submitted it to gauge interest in the proposed functionality. If interest exists, I'll do the work to conform it to networkx standards and test it. 

Preface
----------

In my research I've needed to compute properties of graphs based on edge-connectivity. Namely, I needed to find the k-edge-connected components in a graph, and given a graph I needed to compute a k-edge-connected augmentation.

k-edge-connected components are subgraphs of G, where it is impossible to disconnect the subgraph if less than k edges are removed.  

k-edge-augmentation takes a graph G that is not k-edge-connected and finds a small set of edges that would make G k-edge-connected if they were added. 

My work is primarily concerned with the case where k=2. This special case is also referred to as bridge-connected components, however I want to be able to compute the solutions for other values of k. 

Work So Far
-----------------

Given my problem, I checked to see if NetworkX had any out of the box solutions. I found a few tools. In the `nx.algorithms.connectivity` module I found the `edge_connectivity` and `local_edge_connectivity` functions to be quite useful. However, most other functions in this module were related to node connectivity. I concluded that NetworkX did not have this functionality (correct me if I'm wrong about that). 

Therefore I implemented the edge connectivity algorithms I needed myself internal to my project. Now that the implementation is done, I want to submit it to NetworkX in case anyone else wants to use them. 

I'm not sure what the guidelines for contributing are, so I haven't put much work into this PR other than copy pasting my implementations into `nx.algorithms.connectivity` and doing some minor cleanup. I haven't done any tests or even tried to import the modules yet. I'm sure it will error, but before I spend time fixing things up and writing in depth tests I wanted to make sure that this is a desirable PR. 

Whats Inside
-----------------

I've checked in two modules each with several algorithms.

### module 1
The first module is `edge_kconnectivity' to be consistent with the existing module `kconnectivity'

`EdgeComponentAuxGraph` finds k-edge connected components for arbitrary k. It takes |V| iterations of max-flow, so its a little slow, but its a simple algorithm that produces the right answer.

`bridge_compoments` solves the problem efficiently for k=2. There's not much too the function. It calls `find_bridges` which essentially find edges that aren't in the chain decomposition, and then removes them. I've had a previous PR rejected because it was essentially a one liner, so I'm not sure that is also the case for this. I think its conceptually different enough to warrant its own function though. 

`k_edge_components` combines the previous functions, essentially calling `bridge_compoments` if k is 2 and `EdgeComponentAuxGraph` otherwise. If more efficient algorithms are added (I'm aware of one where k=3) they could be added to this function. 

### module 2
The second module is `edge_augmentation`

I only have efficient algorithms for the case where k=1 and k=2 (although in my project I just wrote a greedy algorithm for k>=3). Each function has weighted and unweighted versions. I'm not sure what the best way to handle the API for that is yet. My working solution takes in a list of available edges that can have optional weights defaulted to 1. Otherwise it is assumed that the complement of G is available and those edges have weights 1.

This module probably needs more work than just cosmetics and it might be better handled as a separate PR. 

When k=1, `one_edge_augmentation` basically solves an MST problem, but my implementation solves it without constructing the complement of the graph. The weighted version is done in `weighted_one_edge_augmentation`. 

When k=2, `bridge_augmentation` performs an algorithm described by Eswaran and Tarjan. In the unweighted case it is exact. For the weighted case the problem becomes NP-hard. I'm not sure what the policy on approximation algorithms is, but `weighted_bridge_augmentation` solves the problem within an approximation ratio of 2. 

Lastly, I was surprised that I couldn't find a NetworkX function to collapse a set of nodes into a single node in a meta graph. There is the condense function for SCCs, but I implemented `collapse` for arbitrary groupings. Maybe I missed a way to do it that already exists. Currently I plan to keep it in this module as a private function, but I think it might be worth exposing in the NetworkX API.

### DIFF
--- a/networkx/algorithms/connectivity/edge_augmentation.py
+++ b/networkx/algorithms/connectivity/edge_augmentation.py
@@ -1,0 +1,363 @@
+# -*- coding: utf-8 -*-
+import networkx as nx
+
+
+def one_edge_augmentation(G, avail=None):
+    """Finds minimum weight set of edges to connect G.
+
+    Adding these edges to G will make it connected.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+
+    avail : list
+        available edges, if None nx.complement(G) is assumed.
+        if each item is a (u, v), the problem is unweighted.
+        if each item is a (u, v, d), the problem is weighted.
+        with d[weight] corresponding to the weight.
+    """
+    if avail is not None:
+        for edge in weighted_one_edge_augmentation(G, avail):
+            yield edge
+    else:
+        ccs1 = list(nx.connected_components(G))
+        C = collapse(G, ccs1)
+        mapping = C.graph['mapping']
+        # When we are not constrained, we can just make a meta graph tree.
+        meta_nodes = list(C.nodes())
+        # build a path in the metagraph
+        meta_aug = list(zip(meta_nodes, meta_nodes[1:]))
+        # map that path to the original graph
+        inverse = ut.group_pairs(C.graph['mapping'].items())
+        for mu, mv in meta_aug:
+            yield (inverse[mu][0], inverse[mv][0])
+
+
+def weighted_one_edge_augmentation(G, avail):
+    """ this is the MST problem """
+    ccs1 = list(nx.connected_components(G))
+    C = collapse(G, ccs1)
+    mapping = C.graph['mapping']
+
+    avail_uv = [tup[0:2] for tup in avail]
+    avail_w = [1 if len(tup) == 2 else tup[-1][weight] for tup in avail]
+    meta_avail_uv = [(mapping[u], mapping[v]) for u, v in avail_uv]
+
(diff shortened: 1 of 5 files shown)


dschult:
Hi, I'm finally taking a look at this. It seems like the two modules should be split into different PRs. If you take the augmentation into a new PR, then I think all the test failures go away too. They are almost all import errors. 

If you are up for it, the new modules should be added to __init__.py appropriately and probably given .rst files in  doc/source/reference too.  

Thanks!

Erotemic:
Thanks for the response. I'll take some time to make the PRs nice and then submit them.

Erotemic:
@dschult the first part of this PR has been separated out and is ready for review in #2554

Erotemic:
I made an initial commit in PR #2572 for the second part of this PR. Therefore, I'm closing this PR.

PR #2466, opened 2017-06-06
added documentation for write_gexf viz attrs

salotz:
I added a sentence describing that networkx does indeed write
attributes like visualization parameters, "viz", correctly according
to the GEXF schema.

I added an example setting the size color and layout positions of a
node which was tested to work in Gephi 8.0.2.

There was a few questions on stackoverflow and some promoting
work-around behavior because it was not evident in the documentation.

I will note that beyond viz there are other behaviors that are not
documented and the user would not know they are available without
actually reading through the code.

### DIFF
--- a/networkx/readwrite/gexf.py
+++ b/networkx/readwrite/gexf.py
@@ -46,11 +46,6 @@ def write_gexf(G, path, encoding='utf-8', prettyprint=True, version='1.1draft'):
     schemas used for parameters which are not user defined,
     e.g. visualization 'viz' [2]_. See example for usage.
 
-    # visualization data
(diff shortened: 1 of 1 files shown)


dschult:
This looks good.  Can you move the example to just below the other example in the same docstring? Your language same "see the example" so people should be able to find it in the "Examples" section. Also it will allow the code to work because G will exist.

Thanks for this!

salotz:
Oops! Ya stupid mistake. Glad you like it, its my first pull request.

dschult:
Ack -- now the errors show a typo:  an extra closing brace {

PR #2473, opened 2017-06-14
find_cycle.  Fixes #2323 and #2324 by fixing #2439

### LINKED ISSUE #2323: find_cycle() doesn't find cycle in cyclic digraph
```python
print("version: {}".format(nx.__version__))
dg = nx.DiGraph()
dg.add_edge(1, 0)
dg.add_edge(2, 0)
dg.add_edge(1, 2)
dg.add_edge(2, 1)
find_cycle(dg)
```

```
version: 1.11

  File "<ipython-input-127-ea87764febbd>", line 7, in <module>
    find_cycle(dg)

  File "xxxx\AppData\Local\Continuum\Anaconda\lib\site-packages\networkx\algorithms\cycles.py", line 453, in find_cycle
    raise nx.exception.NetworkXNoCycle('No cycle found.')
```

I believe the problem lies in that when `head` is a previously explored node all paths from the current `start_node` are ignored instead of just pruning from `head`

```python
elif head in explored:
     # Then we've already explored it. No loop is possible.
     break
```

Removing this condition fixes the bug as far as I can tell given the graphs I am working with. However, a more efficient solution that does proper pruning would be more efficient especially for larger graphs.

### LINKED ISSUE #2439: Find_cycle returns an edge in a DiGraph
Here is an example:
```
>>> import networkx as nx
>>> G = nx.DiGraph()
>>> G.add_edge(1, 2)
>>> G.add_edge(2, 0)
>>> G.add_edge(3, 1)
>>> G.add_edge(3, 2)
>>> nx.find_cycle(G)
[(3, 2)]
```

dschult:
Fixes #2439 as well as #2323 and #2324.  The problem was when a node is discovered that had already been used as a source to no avail. Old version added it as a potential head to be explored. It should just ignore that edge.

### DIFF
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -1,14 +1,18 @@
-"""
-========================
-Cycle finding algorithms
-========================
-"""
 #    Copyright (C) 2010-2012 by
 #    Aric Hagberg <hagberg@lanl.gov>
 #    Dan Schult <dschult@colgate.edu>
 #    Pieter Swart <swart@lanl.gov>
 #    All rights reserved.
 #    BSD license.
+#
+# Authors: Jon Olav Vik <jonovik@gmail.com>,
+#          Dan Schult <dschult@colgate.edu>
+#          Aric Hagberg <hagberg@lanl.gov>
+"""
+========================
+Cycle finding algorithms
+========================
+"""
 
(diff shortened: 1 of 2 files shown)


chebee7i:
@dschult Thanks! :) These days, time is not what it once was for me. :-/

PR #2483, opened 2017-06-23
Remove confusing sentence from tutorial

### LINKED ISSUE #2437: Tutorial incorrectly states that shortest_path is not valid for MultiGraphs
Referring to MultiGraphs and MultiDiGraphs the tutorial states that "...many algorithms are not well defined on such graphs. Shortest path is one example."
https://github.com/networkx/networkx/blob/v1.11/doc/source/tutorial/tutorial.rst#multigraphs

However, this is not mentioned in any of the documentation for the shortest path functions, and with a bit of digging it appears that in fact multigraphs are handled in the code: https://github.com/networkx/networkx/blob/8b9c2677461c385c2e6d86938c97d31703d629fb/networkx/algorithms/shortest_paths/weighted.py#L85

Seems like another example should be picked for the tutorial to avoid confusing me.

jarrodmillman:
Let me know if you want me to change anything, but I suspect this should be completely non-controversial.

### DIFF
--- a/doc/source/tutorial/tutorial.rst
+++ b/doc/source/tutorial/tutorial.rst
@@ -309,8 +309,8 @@ between any pair of nodes.  The :class:`MultiGraph` and
 :class:`MultiDiGraph`
 classes allow you to add the same edge twice, possibly with different
 edge data.  This can be powerful for some applications, but many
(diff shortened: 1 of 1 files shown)


jarrodmillman:
@dschult this is trivial fix and ready for merging.

Issue #1108, opened 2014-04-18
Connected Components for DiGraphs


chebee7i:
It looks like we are missing an algorithm for calculating connected components for directed graphs.  This is strictly different from weakly and strongly connected components. So sayeth Wikipedia (http://en.wikipedia.org/wiki/Connected_graph):

> A directed graph is called weakly connected if replacing all of its directed edges with undirected edges produces a connected (undirected) graph. It is connected if it contains a directed path from u to v or a directed path from v to u for every pair of vertices u, v. It is strongly connected or strong if it contains a directed path from u to v and a directed path from v to u for every pair of vertices u, v. The strong components are the maximal strongly connected subgraphs.

So:  strongly => connected => weakly

```
1. weakly connected but neither connected nor strongly connected
```

![nx_txnq_d](https://cloud.githubusercontent.com/assets/326005/2738441/bcc0cabe-c68e-11e3-9436-ff360452d578.png)

```
2. connected (and also weakly connected) but not strongly connected
```

![nx_fbdqot](https://cloud.githubusercontent.com/assets/326005/2738505/a17d9088-c68f-11e3-9e9d-3fde159a4676.png)

```
3. strongly connected (and also weakly connected and connected)
```

![nx_0vubdm](https://cloud.githubusercontent.com/assets/326005/2738535/305e8528-c690-11e3-8130-5bc40fb48498.png)

Is it correct that we no longer have a partition?  In the following, it seems like there are two components: (A,B,C) and (B,C,D).

![nx_azz9bo](https://cloud.githubusercontent.com/assets/326005/2738686/7c79ad62-c694-11e3-85e5-9eeba6e7a62a.png)

Aside, I don't have time to work on this right now, but if someone had an itch for it...

[Drawings made using [nxpd](https://github.com/chebee7i/nxpd). Look for nxd3 this summer.]

PR #2484, opened 2017-06-23
Implement basic connectivity algorithms for directed graphs

jlumpe:
* Changed functions in `networkx.algorithms.components.connected.py` to work with directed graphs and updated docstrings.
* Adapted tests for undirected graphs to also test directed ones.
* Cleaned up test code a bit, fixed broken test for `connected_component_subgraphs` that contained code that looks like it belonged to `test_weakly_connected.py`.

### DIFF
--- a/networkx/algorithms/components/connected.py
+++ b/networkx/algorithms/components/connected.py
@@ -11,7 +11,6 @@
 #          Christopher Ellison
 """Connected components."""
 import networkx as nx
-from networkx.utils.decorators import not_implemented_for
 from ...utils import arbitrary_element
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


dschult:
Take a look at #1108

If that takes care of the issue great -- if it doesn't then can you say how this is different?
Also, it may be that some strategically placed documentation would help people a lot here. Suggestions are very welcome.

jlumpe:
Thanks for the link @dschult, I hadn't found that in a search. It seems I confused the definition of connected and weakly connected for directed graphs, so this PR is actually incorrect. I'll look into what I can contribute to the documentation.

PR #2488, opened 2017-06-26
Ensure _relabel_copy() respects node order of OrderedGraphs

### LINKED ISSUE #2482: _relabel_copy and OrderedGraph
The current behavior of `_relabel_copy()` is somewhat unintuitive for OrderedGraphs, where the nodes in the newly created graph are ordered according to the edge iterator of the original graph, not the node iterator. I think this would be fixed by placing `H.add_nodes_from(mapping.get(n, n) for n in G)` at the beginning of the function.

jarrodmillman:
This fixes #2482 as suggested. @dschult 

I added a test, which failed with the previous code and passes with this commit.  Please let me know if I should provide additional tests.

I also made some minor pep8 modifications to the functions I touched. I noticed that both `networkx/relabel.py` and `networkx/tests/test_relabel.py` have additional pep8 violations.  When I edit a file (such as these two) for some other purpose, is it acceptable to take the opportunity to make those files pep8 compliant?  I don't think it is worth making a big deal about pep8 compliance, but I am happy to make things more standard as I go.  For instance, I am happy to make additional pep8 style fixes to this PR before you merge it.

If you don't need any additional tests and would prefer that I resist making additional pep8 style changes outside the functions I've already edited, then this is ready to merge (assuming the CI system doesn't complain).

### DIFF
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -150,19 +150,21 @@ def _relabel_inplace(G, mapping):
         G.add_edges_from(new_edges)
     return G
 
+
 def _relabel_copy(G, mapping):
     H = G.__class__()
+    H.add_nodes_from(mapping.get(n, n) for n in G)
+    H._node.update(dict((mapping.get(n, n), d.copy())
+                        for n, d in G.node.items()))
     if G.name:
         H.name = "(%s)" % G.name
     if G.is_multigraph():
-        H.add_edges_from( (mapping.get(n1, n1),mapping.get(n2, n2),k,d.copy())
-                          for (n1,n2,k,d) in G.edges(keys=True, data=True))
+        H.add_edges_from((mapping.get(n1, n1), mapping.get(n2, n2), k, d.copy())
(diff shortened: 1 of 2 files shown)


dschult:
Let's go ahead and make the relabel files pep8 compliant.  Our general philosophy is to improve pep8 stuff only when we touch code for other reasons. Sometimes that means only part of a file and sometimes it means the whole file. The key is to make it so the pep8 changes don't swamp the actual changes and they won't here.  I don't change the tests for pep8 very often because it feels like I make the readability worse by adding so many line breaks to keep the code within 80 chars per line. But that's personal preference and depends on the test code your looking at. I'll leave that to you.
Thanks!

jarrodmillman:
This is ready to merge.

I rebased on master and made a separate commit for pep8 changes outside of the functions I modified.  I didn't make any new code wraps and took out an existing one in `networkx/tests/test_relabel.py` that I thought was unhelpful in terms of readability.

jarrodmillman:
@dschult Ready for merge.

Issue #2442, opened 2017-05-02
Reverse breaks DiGraphs with default-hashed objects


zneak:
The NetworkX documentation [states that graph nodes can be any hashable object](https://networkx.github.io/documentation/networkx-1.10/reference/introduction.html#networkx-basics), and the glossary [defines hashable](https://networkx.github.io/documentation/networkx-1.10/reference/glossary.html#term-hashable) per the Python definition (that is, implements `__hash__` and a comparison method). It goes out of its way to state that instances of user-defined classes are hashable (and unique) by default:

> All of Python’s immutable built-in objects are hashable, while no mutable containers (such as lists or dictionaries) are. **Objects which are instances of user-defined classes are hashable by default; they all compare unequal, and their hash value is their id().**

That would seem to suggest that NetworkX supports graphs made of default-hashed objects (and nodes are considered the same if they're reference-equal). However, I found that the `reverse` method doesn't respect this property. Example:

```python
import networkx as nx

class Unique(object):
    pass

u = Unique()

g = nx.DiGraph()
g.add_edge("A", u)
assert "A" in g
assert u in g # passes
g.nodes() # ['A', <__main__.Unique object at 0x102e0fe10>]
g.edges() # [('A', <__main__.Unique object at 0x102e0fe10>)]

rev = g.reverse()
assert "A" in rev
assert u in rev # fails ಠ_ಠ
rev.nodes() # ['A', <__main__.Unique object at 0x102e0fd10>]
rev.edges() # [(<__main__.Unique object at 0x102e0fe10>, 'A')]
# Unique address is different in nodes() and edges(): `fd10` vs `fe10` ಠ_ಠ
```

What seems to happen is that `reverse` "deep copies" the graph (this is where it calls `Unique.__new__`, at least), and then seemingly doesn't keep track of objects that it has already copied.

hagberg:
See #2282.

jfinkels:
I think this is a slightly different issue, one with DiGraph.reverse. The test case here should be something like

```python
    def test_reverse_hashable(self):
        class Foo(object): pass
        x = Foo()
        y = Foo()
        G = nx.DiGraph()
        G.add_edge(x, y)
        self.assertCountEqual(G.nodes(), G.reverse().nodes())
        self.assertCountEqual(map(reversed, G.edges()), H.edges())
```
Is the contract for `DiGraph.reverse` supposed to be that the nodes are the exact same objects?

zneak:
[...]

PR #2489, opened 2017-06-27
Ensure `digraph.reverse` respects hashable objects

jarrodmillman:
I more or less implemented the suggested test and fix from #2442.  I changed the test from this

```
self.assertCountEqual(G.nodes(), G.reverse().nodes())
self.assertCountEqual(map(reversed, G.edges()), G.reverse().edges())
```

to

```
assert_items_equal(G.nodes(), G.reverse().nodes())
assert (y, x) in G.reverse().edges()
```

Let me know if you have a better idea.

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -1124,7 +1124,8 @@ class DiGraph(Graph):
             H.add_edges_from((v, u, deepcopy(d)) for u, v, d
                              in self.edges(data=True))
             H.graph = deepcopy(self.graph)
-            H._node = deepcopy(self._node)
+            for n in self._node:
+                H._node[n] = deepcopy(self._node[n])
         else:
             self._pred, self._succ = self._succ, self._pred
             self._adj = self._succ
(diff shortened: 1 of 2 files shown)


dschult:
It looks like self.assertCountEqual isn't available in python2.  Your assert_count_equal looks good. 
for the tests, can we use 

    assert_equal([(y, x)], list(G.reverse().edges()))
    assert_count_equal(G.nodes(), G.reverse().nodes())

The current code should pass the edges test and fail the nodes.  Your fix should pass both.  
It's basically the same as your proposal, but tests the edges for equality instead of one-way inclusion. 
If you have other ideas that is fine; I'm just making suggestions.

hagberg:
Do any of the functions in https://github.com/networkx/networkx/blob/master/networkx/testing/utils.py
help? (assert_nodes_equal)

jarrodmillman:
@dschult Ready to merge.

dschult:
Thanks!!

PR #2490, opened 2017-06-27
Fix call to `from_pandas_dataframe` from `to_networkx_graph`

### LINKED ISSUE #2111: Invalid call to from_pandas_dataframe function?
While looking for a problem in my branch, I noticed what looks like an invalid chunk of code (direct link to the line [here](https://github.com/networkx/networkx/blob/d7d906e1d16ef331da0bc1d149953e7532155acc/networkx/convert.py#L133-L134)):

```
    try:
        import pandas as pd
        if isinstance(data, pd.DataFrame):
            try:
                return nx.from_pandas_dataframe(data, create_using=create_using)
            except:
                msg = "Input is not a correct Pandas DataFrame."
                raise nx.NetworkXError(msg)
```

`from_pandas_dataframe` requires, at the very least, _three_ arguments -- with `create_from` being an optional keyword argument. Because the `try: except:` block is untyped, this will always report a `NetworkXError` no matter what.

jarrodmillman:
Fixes #2111.

To make this clean, I decided to make `from_pandas_dataframe` only
require one argument by changing the signature from

```
def from_pandas_dataframe(df, source, target, edge_attr=None,
        create_using=None):
```

to

```
def from_pandas_dataframe(df, source='source', target='target', edge_attr=None,
        create_using=None):
```

I added additional tests and expanded a doctest.

I have a few minor cleanups I want to make before this is merged, but I would like to have feedback on the changes I made to the call signature (or anything else) first.

### DIFF
--- a/networkx/convert.py
+++ b/networkx/convert.py
@@ -67,8 +67,9 @@ def to_networkx_graph(data,create_using=None,multigraph_input=False):
        Current known types are:
          any NetworkX graph
          dict-of-dicts
-         dist-of-lists
+         dict-of-lists
          list of edges
+         Pandas DataFrame (row per edge)
          numpy matrix
          numpy ndarray
          scipy sparse matrix
--- a/networkx/convert.py
+++ b/networkx/convert.py
@@ -132,7 +132,7 @@ def to_networkx_graph(data,create_using=None,multigraph_input=False):
         import pandas as pd
         if isinstance(data, pd.DataFrame):
             try:
-                return nx.from_pandas_dataframe(data, create_using=create_using)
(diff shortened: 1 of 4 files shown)


jarrodmillman:
@dschult @hagberg To fix #2111 I changed `from_pandas_dataframe` so that it only requires one argument.  I thought it was cleaner to put this here than to build assumptions into `to_networkx_graph`.  Before finalizing this PR, I want to make sure this seems reasonable to you.

dschult:
Yes, that seems like a good approach to me.

jarrodmillman:
@dschult This is ready to merge.

PR #2497, opened 2017-07-07
Ensure label in gml output is quoted

### LINKED ISSUE #2362: label in gml output should be quoted
The output of the `write_gml` function currently (1.11) does not quote the label values, but I do think it should according to the [GML spec.](http://www.fim.uni-passau.de/fileadmin/files/lehrstuhl/brandenburg/projekte/gml/gml-technical-report.pdf). Because of this, tools like Cytoscape, jhive and gml2gv fails to read the gml file.

A node from my output:
```
  node [
    id 0
    label 1203
  ]
```
that should be 

```
  node [
    id 0
    label "1203"
  ]
```

jarrodmillman:
Attempting to fix #2362. 

@dschult  I am not confident that my approach is correct.  Would you take a quick look and let me know if this seems like a reasonable start.  Should I add `and key != 'label'` to the blocks of code handling dicts and lists in `stringize` as well?  I wasn't sure how we should handle labels that were more complex than just ints, longs, and floats.

If this seems like the correct approach, the problem I am running into is that the `literal_stringizer` function takes the unicode string
```
u'This is "quoted" and this is a copyright: \xa9'
``` 
and converts it to the text string
```
'u\'This is "quoted" and this is a copyright: \\xa9\''
```

The function `escape` handles the unicode string correctly producing
```
'This is &#34;quoted&#34; and this is a copyright: &#169;'
```
But the text string becomes
```
"u'This is &#34;quoted&#34; and this is a copyright: \\xa9'"
```

I left some print statements for debugging purposes in for now, but will remove them before finalizing the pull request.
Any thoughts?

### DIFF
--- a/networkx/readwrite/gml.py
+++ b/networkx/readwrite/gml.py
@@ -565,7 +565,7 @@ def generate_gml(G, stringizer=None):
     Notes
     -----
     Graph attributes named 'directed', 'multigraph', 'node' or
-    'edge',node attributes named 'id' or 'label', edge attributes
+    'edge', node attributes named 'id' or 'label', edge attributes
     named 'source' or 'target' (or 'key' if `G` is a multigraph)
     are ignored because these attribute names are used to encode the graph
     structure.
@@ -580,9 +580,9 @@ def generate_gml(G, stringizer=None):
         if not isinstance(key, str):
             key = str(key)
         if key not in ignored_keys:
-            if isinstance(value, (int, long)):
+            if isinstance(value, (int, long)) and key != 'label':
                 yield indent + key + ' ' + str(value)
-            elif isinstance(value, float):
+            elif isinstance(value, float) and key != 'label':
                 text = repr(value).upper()
                 # GML requires that a real literal contain a decimal point, but
                 # repr may not output a decimal point when the mantissa is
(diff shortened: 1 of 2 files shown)


dschult:
It seems to work in Python3.6   I get the error only for python2.7 though I haven't tested in all the versions. I'm assuming it is a python2 vs python3 issue. 

In the test you are processing all strings through ```literal_stringizer``` instead of letting str and unicode be unprocessed. I guess this fixes the int node name ```0``` to ```"0"``` but now the test fails on the attribute 'demo' having a unicode value because ```literal_stringizer``` in python2.7 doesn't convert the unicode symbol to HTML correctly.  In the old tests, strings were not processed by ```literal_stringizer```. So this wasn't a problem.

We might need a separate section just for labels.  Instead of calling literal_stringizer on all strings AND on labels, you would just call it on the labels. I think that will make this test work but is kind of a hack because its not clear how to handle labels that are unicode strings. Perhaps just send them through without processing and apply ```literal_stringizer``` to the numeric types.  You could hardcode the stringizer to use or make literal_stringizer the default.

Perhaps we should add a test where a node is a unicode string.  

Does this make any sense?

dschult:
OK... clearer head -- hopefully clearer answer...

The problem arises due to using literal_stringizer on the strings that don't need it.  Instead of excluding 'label' from the numeric handling routines and relying on the stringizer to fix them, I added code to handle the numeric labels specially. Then we don't have to send all the strings through the stringizer just to get the numerics through the stringizer.

I also added a test for a node that is a string with unicode present.

These changes passed python2.7 and python3.6 on my machine. We'll see what Travis and Appveyor say.

jarrodmillman:
Sounds good.  If this passes all the tests, (obviously) feel free to merge.

PR #2498, opened 2017-07-07
Update geometric networks with new models

cole-maclean:
Contribution to add the Soft Random Geometric Graphs model by Penrose [1] and Thresholded Random Geometric Graphs to the geometric.py network generator module. Generator function is called "soft_random_geometric_graph" and "thresholded_random_geometric_graph".

Some notes:
1. I had to refactor the _fast and _slow__construct_edges functions to return a list of edges instead of directly adding the edges to the graph within those functions so that other models can take advantage of the KDTree implementation. These functions are now called _fast_edges/slow_edges and return a list of possible edge node pairs without adding them to the network. The edge addition is now done internal to the network model (ie. within the random_geometric_graph function)
2. I've implemented the model so an arbitrary probability density function, called p_dist, can be supplied for the connection probability as a function of distance parameter. This allows users to use arbitrary pdfs, for example from scipy.stats library of implemented distributions, and you can pass the .pdf method as the p_dist parameter. The variable name p_dist seems ok to me, but we can change it if there's something better. Also, it would be nice if we can refactor the other models that have a similar probability parameter (ie. waxman and geometric_threshold_graphs) to be able accept the same arbitrary pdfs for expansibility and consistency of the models, but I haven't implemented that here. 
3. I don't think the ReadTheDocs is updated correctly for my code, I think I'm missing something here for how to properly build the docs with the new generator model. **Edit: Figured it out, working now**

Left TODO:
1. Testing - I haven't written any tests yet for the new model. **Edit: Tests Added**
2. Check Documentation - check how to get this working for new model **Edit: Got it, working now**
3. Arbitrary pdf for waxman and GTG model - I'll wait for your review of the proof of concept implemented in the SRGG model. **Edit: Ok, I went ahead and did this for the GTG model. I've kept the alpha parameter for backwards compatibility, but provided the ability to input a custom probability distribution.** 
4. Implement another new geometric model - Thresholded Random Geometric Graphs. **Edit: Implemented**
5. Jupyter notebook tutorial highlight different uses for geometric models. **Edit See [PR 6](https://github.com/networkx/notebooks/pull/6) in the notebooks repo.**

Last note, this is my first pull request ever, sorry in advance for any hiccups!

[1] https://arxiv.org/pdf/1311.3897.pdf


networkx/generators/geometric.py
```diff
@@ -156,9 +163,148 @@ def random_geometric_graph(n, radius, dim=2, pos=None, p=2):
...
+        dist = (sum(abs(a - b) ** p for a, b in zip(u_pos, v_pos)))**(1/p)
+        #Check if dist is <= radius parameter. This check is redundant if scipy
+        #is availible and _fast_edges routine is used, but provides the check incase
+        #scipy is not availible and all edge combinations need to be checked
+        if dist <= radius:
+            return random.random() < p_dist(dist)
+        else:
+            return False
+
+    if _is_scipy_available:
```
cole-maclean: If scipy is available, we use the KDTree implementation to find the possible edges, and use that to add these possible edges with probability obtained from the supplied pdf. Note that this requires another iteration over the possible edge list and recomputing their distances, but my intuition is that this method will scale better than running a single iteration over all possible edge combinations. Might be a good idea to do some performance testing for the two possibilities though.

networkx/generators/geometric.py
```diff
@@ -279,14 +432,19 @@ def geographical_threshold_graph(n, theta, alpha=2, dim=2, pos=None,
...
 
+    #if p_dist not supplied, use default r^alpha
+    if p_dist == None:
+        def p_dist(r):
+            return r**-alpha
+
     # Returns ``True`` if and only if the nodes whose attributes are
     # ``du`` and ``dv`` should be joined, according to the threshold
     # condition.
     def should_join(pair):
```
cole-maclean: I had to flip around the logic here to incorporate the custom p_dist function, so the default function is now r^-alpha instead of r^alpha. This seems more consistent with your original GTG paper @hagberg , but there's probably a reason you had the logic the way it was, so want to be sure to double check with you that this is ok.

networkx/generators/geometric.py
```diff
@@ -279,14 +432,19 @@ def geographical_threshold_graph(n, theta, alpha=2, dim=2, pos=None,
...
         u, v = pair
         u_pos, v_pos = pos[u], pos[v]
         u_weight, v_weight = weight[u], weight[v]
-        return theta * metric(u_pos, v_pos) ** alpha <= u_weight + v_weight
+        return (u_weight + v_weight)*p_dist(metric(u_pos, v_pos)) >= theta
```
cole-maclean: I had to flip around the logic here to incorporate the custom p_dist function, so the default function is now r^-alpha instead of r^alpha. This seems more consistent with your original GTG paper @hagberg , but there's probably a reason you had the logic the way it was, so want to be sure to double check with you that this is ok.

networkx/generators/geometric.py
```diff
@@ -279,14 +432,19 @@ def geographical_threshold_graph(n, theta, alpha=2, dim=2, pos=None,
     nx.set_node_attributes(G, 'weight', weight)
     nx.set_node_attributes(G, 'pos', pos)
 
+    #if p_dist not supplied, use default r^alpha
+    if p_dist == None:
+        def p_dist(r):
+            return r**-alpha
+
```
cole-maclean: I've kept the alpha parameter for this default function to allow for backward comp. It would be cleaning to just remove the alpha parameter all together and have a single default function, but not sure a way to do that without breaking backwards comp.

cole-maclean:
Ok, I think this is ready for review. @hagberg can you tag this with needs-review and then you guys can get around to it whenever you have a chance? Thanks! -Cole

networkx/generators/geometric.py
```diff
@@ -156,9 +163,148 @@ def random_geometric_graph(n, radius, dim=2, pos=None, p=2):
...
+            return random.random() < p_dist(dist)
+        else:
+            return False
+
+    if _is_scipy_available:
+        edges = _fast_edges(G, radius, p)
+        G.add_edges_from(filter(should_join, edges))
     else:
-        _slow_construct_edges(G, radius, p)
+        G.add_edges_from(filter(should_join, combinations(G, 2)))
```
cole-maclean: If scipy is available, we use the KDTree implementation to find the possible edges, and use that to add these possible edges with probability obtained from the supplied pdf. Note that this requires another iteration over the possible edge list and recomputing their distances, but my intuition is that this method will scale better than running a single iteration over all possible edge combinations. Might be a good idea to do some performance testing for the two possibilities though.

networkx/generators/geometric.py
hagberg: Maybe the best thing to do is *break* backwards compatibility here completely.  That way we can clearly document the alpha -> -alpha change by showing how to update existing code.

networkx/generators/geometric.py
hagberg: I don't have any good justification of why it was r^alpha.  It is fine like this.

jarrodmillman:
Please document the backwards-incompatible change in the ``API Changes`` section of ``doc/release/release_dev.rst`` by adding a new item with the heading:
```
* [`#2498 <https://github.com/networkx/networkx/pull/2498>`_]
  <insert text here>
```

jarrodmillman:
You may want to rebase your branch on the upstream master and resolve the file conflicts.

Assuming you named your remotes as follows:
```
$ git remote -v
origin  git@github.com:cole-maclean/networkx.git (fetch)
origin  git@github.com:cole-maclean/networkx.git (push)
upstream        git@github.com:networkx/networkx.git (fetch)
upstream        git@github.com:networkx/networkx.git (push)
```
(If you are using different names than ``origin`` and ``upstream``, then you will need to modify the following commands accordingly.)

To rebase on the current master, you will need to do the following:
```
$ git checkout master 
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
$ git pull upstream master
<some output>
$ git checkout update-geometric-networks
Switched to branch 'update-geometric-networks'
$ git rebase master
```
At this point, you will need to resolve any conflicts and then finish the rebase.  The command ``git status`` should provide helpful tips about how to proceed.  You may find this page helpful:
  http://networkx.readthedocs.io/en/latest/developer/gitwash/development_workflow.html#some-other-things-you-might-want-to-do

Feel free to ask me questions, if this isn't clear.  If you are reluctant to change history, let me know and I can cherry-pick your commits in a new PR (that way you will get the correct attribution in the commit history).

Assuming you rebase, then once you are done you will need to force push as you deliberately changed history:
```
git push -f origin update-geometric-networks
```

cole-maclean:
Awesome, thanks for taking a look guys, I'll clean up the suggestions and ping you again when I've updated.

hagberg:
We are close to a 2.0 beta release (later this week) so if you update before then we can include it in the beta.  Else we'll put it in when it is ready and mark it for 2.1.

cole-maclean:
I'll be without a laptop for about a week, so we'll have to push it to 2.1. Congrats on getting to 2.0.

hagberg:
Great, we'll finish up when you get back to your computer.

cole-maclean:
Ok, I think I've managed to rebase and deprecate the alpha parameter for the GTG model including documenting the change in API Changes as well as a note in the function doc string, and updated the tests for it. Looks like the tests are passing, so I think this should be good to go unless you guys see any other gotchyas. The GTG model is especially versatile now with this change, so hopefully users find some cool use-cases!

PR #2506, opened 2017-07-09
How to install

jarrodmillman:
@hagberg , @dschult I would like to update the INSTALL file.  So far, I moved it to the top of the repository.  If you clone the repository or view it on GitHub, you will look at the root directory first.  Any one looking at the online documentation won't notice a difference.

I would also like to lightly edit the INSTALL text.  For example, 
```
http://scipy.org/Download
```
is from the old Moin site and is now a redirect to
```
https://scipy.org/install.html
```

I would also update some of the pip stuff.  Should I go ahead and draft something for you to comment on?


hagberg:
I'm not sure why it wasn't at the root level anyway...so yes, this is a good start.  Please do update this to reflect current practices, improve clarity, etc.

jarrodmillman:
Ready to merge assuming the changes are OK.

Issue #2465, opened 2017-06-06
All pairs dijkstra returning both paths and lengths


prcastro:
Today Networkx only have `all_pairs_dijkstra_path` and `all_pairs_dijkstra_path_length`, but there is no way to get both paths and lengths at the same time (as we have in `single_source_dijkstra`)

dschult:
This should be pretty straightforward to add. It's very similar to the methods you mention only using ```single_source_dijkstra``` instead of the path and length varieties of these functions. 

Note that if you have the paths it is fast to find the lengths, so maybe you don't need both.

In any case, would anyone put forward a PR for this?

markhkim:
Here it is! https://github.com/networkx/networkx/pull/2510

dschult:
implemented in #2510 and #2612

PR #2510, opened 2017-07-12
Added all_pairs_dijkstra, solving #2465

markhkim:
* Finds shortest weighted paths and lengths between all nodes.
* Returns an iterator of dictionaries, keyed by source nodes.
* Each dictionary essentially contains the value of
  single_source_dijkstra computed at the source node.
  The tuple that single_source_dijkstra returns is unpacked
  and saved as values to two keys: 'distance' and 'path'.

### DIFF
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -31,6 +31,7 @@ __all__ = ['dijkstra_path',
            'multi_source_dijkstra',
            'multi_source_dijkstra_path',
            'multi_source_dijkstra_path_length',
+           'all_pairs_dijkstra',
            'all_pairs_dijkstra_path',
            'all_pairs_dijkstra_path_length',
            'dijkstra_predecessor_and_distance',
@@ -396,11 +397,12 @@ def single_source_dijkstra(G, source, target=None, cutoff=None,
 
     Returns
     -------
-    distance, path : pair of dictionaries, or numeric and list
-       If target is None, returns a tuple of two dictionaries keyed by node.
-       The first dictionary stores distance from one of the source nodes.
-       The second stores the path from one of the sources to that node.
-       If target is not None, returns a tuple of (distance, path) where
+    distance, path : pair of dictionaries, or numeric and list.
+       If target is None, paths and lengths to all nodes are computed.
+       The return value is a tuple of two dictionaries keyed by target nodes.
+       The first dictionary stores distance to each target node.
+       The second stores the path to each target node.
+       If target is not None, returns a tuple (distance, path), where
        distance is the distance from source to target and path is a list
        representing the path from source to target.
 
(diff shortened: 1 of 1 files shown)


dschult:
Thanks for this PR!

At the moment, it doesn't give any way to figure out which source is attached to each dict. Also, I think we don't need the keys 'distance' and 'path' if we just return 2-tuples where the first entry is always distance and the second always path.

I thought about yielding the 3-tuple  (n, distance, path) but I think it is better to yield (n, (distance, path)) because then someone could pipe it directly to ```dict``` and create a dict keyed by source to a 2-tuple of dicts keyed by target.  Perhaps the spirit of iterators over data storage would say distance and path should themselves be iterators rather than dicts. I'm open to new ideas about what should be returned.

markhkim:
`(n, (distance, path))` seems reasonable, yes. On the other hand, `all_pairs_dijkstra_path` returns a `dict` keyed by source, whereas `all_pairs_dijkstra_path` yields (source, `dict`). Is there a reason for this?

As for iterators vs. data storage, wouldn't a typical use case of dijkstra involve directly examining a particular shortest path, or a particular distance? It would seem reasonable to have a subscriptable return, rather than having to go through a generator. Plus, `single_source_dijkstra` and `multi_source_dijkstra` currently do not return generators. Opting for generators would mean rewriting the entire dijkstra portion of the code---not that it would be a difficult edit, per se.

dschult:
Good questions.  It has been a long time since I went through all the different shortest_path functions looking for consistency.  Which two do you see returning a dict and yielding a ```(source,dict)``` tuple? (you say ```all_pairs_dijkstra_path``` for both)

I think a typical use of all_pairs dijkstra is to go through each source looking at the distances/paths to each node from that source. That's the idea behind yielding for all_pairs.  Also these structures can be large and yielding provides good memory savings in the case where only one source is viewed at a time.  The output can of course be made into a ```dict``` using ```dict(all_pairs_dijkstra...)```. The single_source functions tend to return dicts because a single source doesn't create so large outputs. In general we are coming from code written to return ```dict``` and trying to modernize by yielding instead where-ever it makes sense to do so.

markhkim:
Oh, sorry! I meant `all_dijkstra_path_length` for yield, and `all_dijkstra_path` for `dict`. If you'd like, I could edit `multi_source_dijkstra` to return generators.  As it stands now, `all_pairs_dijkstra` depends on `single_source_dijkstra`, which, in turn, depends on `multi_source_dijkstra`.

dschult:
I think the single_source and multi_source versions should return a dict (they are presumably small enough to manipulate) while all_pairs/all_ (functions that collect the data for each source) should yield a (source, info) tuple.

markhkim:
I see. Are you suggesting that all_ variants should be reimplemented, independent of the single_source / multi_source versions?

dschult:
No...   They should just call the single/multi over and over again. So yielding is easy:

    for s in G:
        yield (s, multi_source_...(G, [s], ...))     #  you could do single_source_... if you prefer

This is done, for example in ```all_pairs_dijkstra_path_length```.   Some will need to be changed from:

    return {n: path(G, n, cutoff=cutoff, weight=weight) for n in G}   # from all_pairs_dijkstra_path

Some of the ```single_source``` functions currently create the dict and then make an iterator from it which is silly really, so let's have the single_source and multi_source return dicts while the all_pairs yield 2-tuples.

The next confusing thing is that for those functions with an optional ```target```, the output should be different when ```target``` is specified.  So ```single_source_dijkstra(G,0)``` should return two dicts while ```single_source_dijkstra(G,0,1)``` should return a length and a path.  This should already be true so you just need to check  that it is true for all those routines.


Here is what I think needs doing from the ```weighted.py``` module.  Your mileage may vary. :)

Should return a dict instead of current treatment

    multi_source_dijkstra_path_length
    single_source_bellman_ford_path_length

Should yield ```(source, stuff)``` 2-tuples

    all_pairs_dijkstra_path
    NEW:  all_pairs_dijkstra
    all_pairs_bellman_ford_path

Similarly, from ```unweighted.py```:  
Should yield ```(source, stuff)``` 2-tuples

    all_pairs_shortest_path

Should yield dict of what they yield: (Note: this leaves ```_single_shortest_path_length``` as an iterator. But it is a private function and it is nice to have it written this way. We're putting the others into dicts for consistency. Perhaps we should make all of the single_source be iterators, but not for now.

    single_source_shortest_path_length
    single_target_shortest_path_length

Do you agree?  
Anyway, it might be best to change the tests first and then make sure the code passes the tests.  This is turning into more than I thought it would.

markhkim:
Sounds reasonable, yes. Which tests do you speak of?

dschult:
The tests are in ```tests/test_weighted.py``` and ```tests/test_unweighted.py```. If the tests expect iterators and get dicts or vice versa we will know about it. My comment was just that it might make sense to change the tests to what we want and then make the code match that.

markhkim:
Understood. I'll take a look.

dschult:
@markhkim Have you made any progress on this PR?
Any chance of progress in the next few days? 
If not, I will take a stab at it.  Thanks!

markhkim:
hey! sorry, I've been extremely busy. But yes, now I have some time to take a stab at it. Perhaps we could divvy up the work?

dschult:
That would be cool -- what would you like to take on?

markhkim:
Since I was working on all_pairs_dijkstra, I could start by tackling `all_pairs_dijkstra_path`, `all_pairs_dijkstra`, `all_pairs_bellman_ford_path`. How does that sound?

dschult:
Sounds great!...   Thanks!

dschult:
merged via #2612

PR #1011, opened 2013-11-16
Issue 762 shortest path hide nodes edges

francis-liberty:
I add the monkey patching context manager in a separate file named ```restrictions.py```. I'm glad to hear your code reviews. About testing, I add some in the ```test_generics.py```, because I think it duplicated if I also test them in ```test_weight.py``` or ```test_unweight.py```, etc. Let me know if you have concerns about adding more tests. ;)

Issue #793, opened 2012-11-26
Feature request: n-best paths


dan-blanchard:
As far as I can tell, there's currently no way with networkx to get the n best/shortest paths between two nodes in networkx. I've tried to implement it myself, but it has proved difficult to do efficiently (and in such a way that I could make it allow for ties like all_shortest_paths currently does).

Is there a simple/efficient way to do this I'm just not aware of?

Wind4Greg:
A while ago, for teaching and R&D purposes, I implemented a version of Yen's K-shortest path algorithm in Python/NetworkX.  The code has been posted with the rest of the code for a (data/optical) network course at http://grotto-networking.com/netDesignCourse.html#Code with Sphinx generated documentation at http://grotto-networking.com/files/NetDesignCourse/code_doc/Utilities.html#module-Utilities.YenKShortestPaths.  I recently heard from a biophysics PhD student that has used this code on networks with several thousand nodes and links to find hundreds of shortest paths. 

If there is sufficient interest I can try to modify these to meet NetworkX coding/style/API standards, but would need a bit of guidance from the NetworkX developers.

Cheers
Greg B.

aparamon:
Hello Greg!

It would be great if mine or yours version could be finally integrated into
NextworkX :-)

Regarding your code, I propose that instead of separate methods
findFirstShortestPath, getNextShortestPath, object YenKShortestPaths uses
standard Python iterator protocol (e.g. is generator, see
https://wiki.python.org/moin/Generators).

Best wishes,
Andrey Paramonov

Wind4Greg:
Good idea Andrey, definitely a more Python way to handle the situation.  I'll do an update and post the code with some examples to my website this week.
Cheers
Greg B.

Wind4Greg:
I converted the interface to the K shortest path algorithm (Yen's) to use the iterator pattern, cleaned up all PEP 8 errors (except camel case), Created two different test examples on three different networks and zipped it all to http://www.grotto-networking.com/files/KShortestPaths.zip. 
Cheers
Greg B.

jtorrents:
[...]

PR #762, opened 2012-08-29
Shortest path algorithms with restrictions

aparamon:
Many interesting algorithms e.g. Yen's k-shortest simple paths, White's node-independent paths require a procedure for finding shortest path with restrictions:
- ignore a subset of graph nodes
- ignore certain graph edges

It is proposed to introduce optional ignore_nodes and ignore_edges parameters for all shortest path procedures.

Here is the initial version (only supporting shortest_path, unweighted case).

Issue #335, opened 2012-07-20
Consider views for graphs (migrated from Trac #338)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/338
Reported 2010-04-14 by @dschult, assigned to @dschult.

It would be nice to have a fast way to process parts of a graph.
A view could be faster than creating the induced subgraph
because it might not have to create the whole graph structure.
Let's see if that is really true.

dschult:
This is implemented in #2523

PR #2512, opened 2017-07-12
GraphView with hidden_nodes and hidden_edges

dschult:
This implements a restricted view on graphs where ```hidden_nodes``` and ```hidden_edges``` do not appear in the resulting graph.  This implementation is read-only though you could change the original graph and the restricted view would update accordingly. At this point you can't update the hidden_nodes and hidden_edges without creating a new view. I might add that feature if I can come up with a good way to do it. 

Related tickets include:   
#762
#1011 
#793
#335
and there are probably many other places where this feature has been discussed.

### DIFF
--- a/networkx/classes/__init__.py
+++ b/networkx/classes/__init__.py
@@ -3,6 +3,7 @@ from .digraph import DiGraph
 from .multigraph import MultiGraph
 from .multidigraph import MultiDiGraph
 from .views import *
+from .graphview import *
 from .ordered import *
 
 from .function import *
--- a/networkx/classes/graphview.py
+++ b/networkx/classes/graphview.py
@@ -1,0 +1,159 @@
+#    Copyright (C) 2004-2017 by
+#    Aric Hagberg <hagberg@lanl.gov>
+#    Dan Schult <dschult@colgate.edu>
+#    Pieter Swart <swart@lanl.gov>
+#    All rights reserved.
+#    BSD license.
+#
+# Author:  Aric Hagberg (hagberg@lanl.gov),
+#          Pieter Swart (swart@lanl.gov),
+#          Dan Schult(dschult@colgate.edu)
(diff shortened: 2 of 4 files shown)


hagberg:
Looks good - doesn't this feature belong as a (more general) subgraph view?  That is, a subgraph where you specify the edges and/or nodes you want to view.  Depending on the application it could be easier to specify inclusion or exclusion.

dschult:
I wondered about that too.  Should they be the same object? (specifying either include or exclude nodes and either include or exclude edges)  Or does it make more sense to have one set of objects for exclude and one for include?

Would the name ```SubGraphView`` be better?  also ```excluded_nodes``` vs ```hidden_nodes```?

Finally, we could have the user provide a ```node_filter``` function that determines which nodes get through and another for edges?  Lots of ways to do all this...

dschult:
I've been looking at how best to make a SubGraphView interface. Sometimes we want to exclude specific nodes. Other times we want to include only specific nodes.  It'd be nice to be able to do either with a single interface.  It would not be hard to have the user supply one function for nodes and another for edges. If the function returns True then that node/edge is in the subgraph. Otherwise it isn't. This puts the onus on the user to come up with a function and storage mechanism even for simple cases. Maybe we can supply some helper functions/objects to make that process easier.  I'm thinking of the filter function inputs as (Graph, node) and (Graph, edge-tuple) with bool return value.

I found [a description of filters](https://support.google.com/analytics/answer/1034832?hl=en) which might provide a basis for how to make a helper object.  Users provide ```includeonly_filters``` and ```exclude_filters``` for each of nodes and edges. If one is not provided that condition is ignored. If the user provides a ```set``` instead of a ```function```, we treat that ```set``` as the included/excluded values.

Does this approach seem simple enough to pursue?

dschult:
I prefer #2523 to this proposal.

PR #2516, opened 2017-07-15
Doc ordered graph classes

jarrodmillman:
When I was working on PR #2504, I almost missed the fact that `nx.OrderedMultiGraph` existed and started to implement my own multigraph subclass as suggested in the `nx.MultiGraph` docstring.  This PR documents the existence of the ordered variants so (hopefully) others won't (almost) overlook their existence as well.

I was thinking of removing the examples of creating a ordered graph subclass for the various base graph classes, but wanted feedback first.  (@dschult,  @hagberg)  Instead we could add a line like:
```
Please see `ordered.py` for examples of creating graph subclasses by overwriting
the base class `dict` with a dictionary-like object.
```

At the very least, we should add a line something like:
```
This example is provided as part of NetworkX via nx.OrderedGraph.
```
to each of the base graphs' docstrings.  This is only done in `graph.py` currently.

Also, in `doc/source/reference/api_1.10.rst` it states
```
* [`#1314 <https://github.com/networkx/networkx/pull/1314>`_]
  Allow overwriting of base class dict with dict-like:
  OrderedGraph, ThinGraph, LogGraph, etc.
```
I couldn't figure out what a `LogGraph` is.  Is that a typo? Or is there some obvious meaning I am missing?  Perhaps we could add an example somewhere in the codebase or remove `LogGraph` from the api reference file.


hagberg:
I'm not sure I know what LogGraph is.  Perhaps it was referring to https://github.com/networkx/networkx/blob/master/examples/subclass/printgraph.py ?

I think it is fine to remove the ordered graph examples as you suggest.

dschult:
Yes, LogGraph was a reference to a graph that logged each mutation. That's what PrintGraph does. It's certainly fine to remove the confusing reference.

ThinGraph only appears as you see it in the docstrings. But OrderedGraph eventually evolved to being included (along with Di/Multi versions) as you point out.  I think removing the OrderedGraph examples and pointing readers to the ```ordered.py``` code would be good. It might even be time to add docs to the ```OrderedGraph``` classes. I think another Issue asks for it to improve the subgraph and reverse behavior for ```OrderedDiGraph``` so that code may expand soon.  Finally, it should be noted somewhere that ```OrderedGraph``` give a consistent order for reporting of nodes and edges, but for edges, the order is not necessarily the order that the edges were added.  The dict's are ordered, not the edges.

dschult:
OK... I looked through these changes and while there are a lot due to pep8, I've looked at the others and think there are two more steps needed.

- remove the examples for OrderedGraph and friends and replace with pointers to the actual classes.

- change the language to not claim that OrderedGraph and friends maintain the order of adding edges.  While it is true that the order of node reporting agrees with node adding, that is not true for edges. The adjacency dicts are ordered, and edges will be reported in a consistent order.  But the edge order is not typically the order that edges were added. (that's because we pass through the adjacency dicts to get the edge order, the edge reporting exhausted one nodes neighbors before moving to the next node.

Thanks for all this!

jarrodmillman:
This is ready to merge.

PR #2522, opened 2017-07-18
Add {to, from}_numpy_array functions, resolves #2479

### LINKED ISSUE #2479: Add {to, from}_numpy_array functions
Should we add a `to_numpy_array` function to generate an `np.array` from a graph and a `from_numpy_array` function for the reverse? These would be similar to the existing `{to, from}_numpy_matrix` functions.

**Motivation:**
* Times have changed since #895.
* `np.matrix` is as good as deprecated, especially after the introduction of the `@` operator for matrix multiplication of arrays in Python 3.5.
* `np.matrix` doesn't always produce the same results as `np.array` and can lead to confusion.

harryscholes:
* to_numpy_array returns the graph adjacency matrix as a NumPy array

* from_numpy_array returns a graph from NumPy array

* Adapted tests for to_numpy_matrix and from_numpy_matrix, which all pass

Resolves #2479

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -37,7 +37,8 @@ __author__ = """\n""".join(['Aric Hagberg <aric.hagberg@gmail.com>',
 __all__ = ['from_numpy_matrix', 'to_numpy_matrix',
            'from_pandas_dataframe', 'to_pandas_dataframe',
            'to_numpy_recarray',
-           'from_scipy_sparse_matrix', 'to_scipy_sparse_matrix']
+           'from_scipy_sparse_matrix', 'to_scipy_sparse_matrix',
+           'from_numpy_array', 'to_numpy_array']
 
 
 def to_pandas_dataframe(G, nodelist=None, dtype=None, order=None,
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for this!

Is the code the same as for to_numpy_matrix, only at the end you convert it to array?  If so, you should just call the existing function, convert the result, and return that.  But I think the argument was that ```numpy.matrix``` was going the way of the dinosaur, in which case perhaps we shouldn't rely on it here. Would it be easy to change this function so it doesn't use the numpy.matrix class?

Also, could you take a look at the docstrings and make sure it says "matrix" and "array" appropriately?  In my quick first-read it seemed like it said matrix more often than it should given that we are working with arrays.

harryscholes:
Thanks for your comments @dschult! Firstly, the code for `to_numpy_matrix` and `from_numpy_matrix` is fine, it's just that the `np.matrix` class is not :)

To address your first concerns about keeping these additions lightweight:

* `to_numpy_matrix` actually treats the adjacency matrix as `np.array` until the last line of the function where it converts it to `np.matrix`. So, how about we move most of this code into `to_numpy_array`, which returns `np.arry`. Then all `to_numpy_matrix` would need to do is call `to_numpy_array` and convert the result to `np.matrix`. Something like:

```
def to_numpy_array(G, ...):
    # most of the code currently in to_numpy_matrix
    return M  # np.array

def to_numpy_matrixG, ...):
    M = to_numpy_array(G, ...)
    M = np.asmatrix(M, dtype=dtype)  # final line in the current to_numpy_matrix
    return M  # np.matrix
```

* `from_numpy_matrix` doesn't check the type of the adjacency matrix passed to it and already actually works on `np.array`. As such, `from_numpy_matrix` could simply do:

```
def from_numpy_array(A, parallel_edges=False, create_using=None):
    return from_numpy_matrix(A, parallel_edges=parallel_edges, create_using=create_using)
```

Do you agree with these ideas of reducing the weight of this PR?

Regarding the docstrings: I was trying to distinguish between the *adjacency matrix* of a graph, and the *array* type that the function accepts/returns. I think it would be confusing for users for the docs to refer to *adjacency arrays*. What do you reckon?

dschult:
Yes, I like the idea of making ```to_numpy_array``` be the old function without the conversion and then ```to_numpy_matrix``` adding the conversion t the output of the array function.

For ```from_numpy_array``` should we just monkeypatch: ```to_numpy_array = to_numpy_matrix```?  I guess that makes the docstring identical too... which is OK by me.

On the docstrings, I understand and agree that ```adjacency array``` isn't a recognized term. So long as it is clear that ```matrix``` means ```adjacency matrix``` and not ```numpy matrix``` its great. And that might just mean that the first time you use the term you give the adjective.  I'll leave that judgement up to you.

Thanks!

dschult:
The fails on the Travis site are just the docstring tests looking for an array instead of a matrix or vice versa.  [You can test the docstrings on your machine using ```nodetests --with-doctest```.]

harryscholes:
Changes added:
* `to_numpy_array` now has the meat of the code from `to_numpy_matrix`
* `to_numpy_matrix` calls `to_numpy_array` and converts the result to a NumPy `matrix`
* `from_numpy_array` calls `from_numpy_matrix`. I gave it its own function and docstring (my personal opinion is that it's best to have proper docs for the function, rather than relying on `from_numpy_matrix`)

All tests, including doctests, now pass.

dschult:
Looks great -- Thanks very much!!

PR #2523, opened 2017-07-18
SubGraph filter on nodes/edges

dschult:
Here SubGraph means a view restricting (or enhancing) a graph. 
Two filters are possible: ```node_filter``` and ```edge_filter```
```node_filter``` takes a ```node``` argument and returns True to keep the node.
```edge_filter``` takes two nodes as arguments (or for multigraph,
an additional key argument) and returns True to keep the edge.

This seems cleaner than the ```GraphView``` of #2512 and can hide nodes easily by building a filter function: ```lambda n: n not in hidden_nodes```, but can be much more involved.

    def fish_nodes(n):
        return 'fish' in G.nodes[n]
    def red_edges(u,v):
        return 'red' == G.edges[u, v]['color']
    SG = nx.SubGraph(G, node_filter=fish_nodes, edge_filter=red_edges)

Related tickets include:
#2512 
#762
#1011
#793
#335
and there are probably many other places where this feature has been discussed.

### DIFF
--- a/networkx/__init__.py
+++ b/networkx/__init__.py
@@ -98,9 +98,9 @@ from networkx.exception import *
 import networkx.external
 import networkx.utils
 
+import networkx.classes.filters
 import networkx.classes
 from networkx.classes import *
-
 
 import networkx.convert
 from networkx.convert import *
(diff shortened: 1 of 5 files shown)


dschult:
I'm going to merge this.  I think we should consider replacing the ```G.subgraph``` method but that should be a different pull request.  Initial tests suggest it doesn't save much time.  It should save memory but I haven't looked at it yet.

PR #2525, opened 2017-07-19
Add local_bridges to bridges.py

dschult:
Resolves #2273

Includes ideas about local_bridges from #2273 and adds tests.
Bridges were also discussed in #903
@erotemic Perhaps parts of bridges.py could be used in (or improved by) #2459


Erotemic:
I think I can definitely reuse part of this functionality. Thanks for the ping. I'm probably going to work on my PR a bit next week.

PR #2528, opened 2017-07-21
Small-world analysis

romain-fontugne:
This implements algorithms for quantifying the "small-wordness" of a graph. There is commonly two metrics used to do that, sigma and omega. Both are documented on the wikipedia page on small-world network: https://en.wikipedia.org/wiki/Small-world_network


dschult:
It looks like your tests are taking too long for the Travis-CI system.
I also notice that they are testing random graphs and that the test is an inequality.

Perhaps it would be better to create a small deterministic graph that you know the sigma and omega values for. Then you are testing equality on a known input. It might be a better indicator of future bugs introduced by mistake.

Pegdwende649:
Is there now any possibility to compute the sigma and/or omega value for the multigraphs?

PR #2530, opened 2017-07-21
Ensure `make html` doesn't fail build on exit

jarrodmillman:
Mark `parallel_betweenness` as a known failure


jarrodmillman:
I should have set up readthedocs for testing already, but I didn't.  I just set it up and will wait to see that this fixes the doc build failure before I merge it.

PR #2536, opened 2017-07-22
Remove deprecated code

jarrodmillman:
This removes deprecated code marked for removal in 2.0 or code that was deprecated more than 4 years ago.  I also marked 2.1 as the removal release for ``bellman_ford``.

I still need to add a note to ``doc/release/release_2.0.rst`` and to ``doc/news.rst``.

### DIFF
--- a/doc/reference/algorithms/minors.rst
+++ b/doc/reference/algorithms/minors.rst
@@ -10,4 +10,3 @@ Minors
    contracted_nodes
    identified_nodes
    quotient_graph
-   blockmodel
--- a/doc/reference/utils.rst
+++ b/doc/reference/utils.rst
@@ -38,7 +38,6 @@ Random Sequence Generators
 
    pareto_sequence
    powerlaw_sequence
-   uniform_sequence
    cumulative_distribution
    discrete_sequence
    zipf_sequence
--- a/doc/reference/utils.rst
+++ b/doc/reference/utils.rst
@@ -36,7 +36,6 @@ Random Sequence Generators
 .. autosummary::
    :toctree: generated/
 
-   pareto_sequence
    powerlaw_sequence
    cumulative_distribution
    discrete_sequence
--- a/doc/reference/utils.rst
+++ b/doc/reference/utils.rst
@@ -36,7 +36,6 @@ Random Sequence Generators
 .. autosummary::
    :toctree: generated/
 
-   create_degree_sequence
    pareto_sequence
    powerlaw_sequence
    uniform_sequence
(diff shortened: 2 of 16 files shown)


jarrodmillman:
The deprecated ``create_degree_sequence`` was used in several tests, which now fail.  The old wrapper function body is:
```
    tries=0
    max_deg=n
    while tries < max_tries:
        trialseq=sfunction(n,**kwds)
        # round to integer values in the range [0,max_deg]
        seq=[min(max_deg, max( int(round(s)),0 )) for s in trialseq]
        # if graphical return, else throw away and try again
        if nx.is_valid_degree_sequence(seq):
            return seq
        tries+=1
    raise nx.NetworkXError(\
          "Exceeded max (%d) attempts at a valid sequence."%max_tries)
```
Presumably, the failures are happening now because we are no longer trying to produce a valid degree sequence several times until we get one.  Should we add the multiple tries to each generator?  Perhaps using a ``_create_degree_sequence`` function internally?  Suggestions?

hagberg:
I'd prefer to completely remove the degree sequence generators.   Except maybe discrete_sequence which is a little nontrivial and perhaps useful.  Would that break things in many places?  

The tests in kcutsets and kcomponents that use create_degree_sequence aren't very compelling.  They could be removed (or replaced) with something else.

jarrodmillman:
I removed the  degree sequence generators, which were trivial to remove.  There are a few that will take some discussion (these are just my first thoughts, but I wanted feedback before I spent much time on thinking about this):

1. ``powerlaw_sequence`` is used ``random_powerlaw_tree_sequence`` in ``networkx/generator/random_graphs.py``.  But it uses ``tries`` in a funny (and incorrect) way.  Thoughts?

    Note:  There is a ``_powerlaw_sequence`` in ``networkx/algorithms/community/community_generators.py``.

2. Should we move ``discrete_sequence`` anywhere?  Maybe somewhere in ``networkx/generators``?

3. I would use ``from numpy.random import choice`` or limit this functionality to Python 3.6+ and use [random.choices](https://docs.python.org/dev/library/random.html#random.choices) for

    - ``random_weighted_sample`` used by ``havel_hakimi_graph`` in ``networkx/generators/degree_seq.py`` and
    - ``weighted_choice`` used by ``random_k_out_graph`` in ``networkx/generators/directed.py``

4. use ``np.random.zipf`` for

    - ``zipf_rv`` in ``networkx/algorithms/community/community_generators.py``.

5. and maybe something in ``scipy.stats`` for

    - ``cumulative_distribution`` used by ``double_edge_swap`` in ``networkx/algorithms/swap.py``.

PR #2539, opened 2017-07-24
Document release process

jarrodmillman:
Am I missing anything obvious?  We can fine-tune this as we do the 2.0 release.

(I don't mention making a 2.0 branch as I don't think it is necessary.   Particularly, if we are releasing 2.1 soon after the 2.0 release.)

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -1,16 +1,44 @@
-*****************************************
-Version 2.0 release notes and API changes
-*****************************************
+Announcement: networkx 2.0
+==========================
 
-This page includes more detailed release information and API changes from
-NetworkX 1.10 to NetworkX 2.0.
+We're happy to announce the release of networkx 2.0!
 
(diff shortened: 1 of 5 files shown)


doc/release/release_dev.rst
```diff
@@ -1,16 +1,44 @@
...
+Highlights
+----------
+
+This release is the result of X of work, with over X pull requests by
+X contributors. Highlights include:
+
+- big items ...
+
+There is a `migration guide for people moving from 1.X to 2.0
+<http://networkx.readthedocs.org/en/latest/release/migration_guide_from_1.x_to_2.0.html>`_.
```
MridulS: We need to update this doc. This page was written when `G.nodes()`, `G.edges()`, `G.degree()`, .... were changed to return iterators and now these functions return views.

jarrodmillman: @MridulS Would you be willing to create a PR to update:
```
doc/release/migration_guide_from_1.x_to_2.0.rst
doc/release/release_dev.rst
```
with this and any other changes, which should be noted?

PR #2545, opened 2017-07-25
Drop 3.3 support

jarrodmillman:
While all our tests currently work with Python 3.3, I propose we officially drop support for it before the 2.0 release.

``numpy``, ``scipy``, ``matplotlib``, ``pandas``, and many other packages have already dropped Python 3.3 support in their previous releases.

3.3.6 is from 2014 and I doubt there will be another release as its end-of-life is 2017-09-29.

Python 3.3 use was negligible a year ago (and has presumably decreased in popularity as packages have dropped support for it):
- https://github.com/pypa/pip/issues/3796
- http://www.randalolson.com/2016/09/03/python-2-7-still-reigns-supreme-in-pip-installs/

Our CI checks will be faster and we will have to worry about less versions of python for bug reports.

Does this seem reasonable?


dschult:
This sounds good to me.

jarrodmillman:
@dschult , @hagberg This is ready to merge.

I should point out that no recent version of pypy worked before.  I am no longer testing on PyPy3 2.4.0, which did pass.  However, PyPy3 2.4.0 is old:
- https://morepypy.blogspot.com/2014/10/pypy3-240-released.html
and it was intended as a drop-in replacement for Python 3.2.5, which we already dropped support for.  It also isn't provided as part of trusty, which is the new default ubuntu image for travis-ci.

This also includes other minor updates and changes.

PR #2549, opened 2017-07-25
WIP: Enable doctests for timingclasses module

jarrodmillman:
Enabling doctests, which were skipped per fdc906e4.


jarrodmillman:
@dschult Is there any reason for me to hold-off fixing the doctests for ``timingclasses``?  I assume the API has now stabilized.

dschult:
The timingclasses are there to allow comparison between the old API and the new API. I envision them being removed as soon as v2.0 is put in place.  So I don't think you should put time into them unless you think you might use them for some other purpose later.

PR #2554, opened 2017-07-26
[ENH] Added functions for k-edge-connected components/subgraphs

Erotemic:
(Rewriten on 7-31-2017 to be more concise)

This is the first component of my original PR from #2459 which adds functionality for extracting k-edge-connected components (k-edge-ccs) and k-edge-connected subgraphs (k-edge-subgraphs).

Originally this PR was only k-edge-ccs, but I extened it to include k-edge-subgraphs after realizing the subtle difference (and sorting out the naming conventions) between the two.

### Definitions 

**K-Edge-Connected Component**: A k-edge-cc is a maximal set of nodes that is k-edge-connected *in the original graph*

**K-Edge-Connected Subgraph**: A k-edge-subgraph is a maximal set of nodes from G, whose subgraph is a k-edge-connected *subgraph*

### API
This PR exposes two main functions: `k_edge_components(G, k)` and  `k_edge_subgraphs(G, k)`., which both take a graph and a positive integer k and yields the nodes in the corresponding components/subgraphs.  

### Implementation

Based on the value of `k` and whether or not `G` is directed or undirected, these functions will call different underlying algorithms, attempting to use the most efficient ones available. 

When `k=1` both the k-edge-ccs and the k-edge-subgraphs are equivalent, and I reuse the `nx.connected_compoments` and `nx.strongly_connected_components` algorithms for both of them. 

When `k=2`, k-edge-ccs and the k-edge-subgraphs are only equivalent for undirected graphs, which I've written an efficient algorithm called `bridge_components` to find. 

In all other cases I call general algorithms, which will result in the correct answer, but they may not be the most efficient (for k=3 there are certainly more efficient algorithms known). 

Depending on the value of k and whether G is directed or undirected, different algorithms are chosen. So far, I've only implemented a very efficient algorithm for the case where G is undirected and k=2. Both cases where k=1 are already taken care of by existing networkx algorithms, so I re-use those. Otherwise, I call a general algorithm, which is slow but handles all cases. 

### Algorithms

For k-edge-ccs, I've implemented the algorithm in ["A simple algorithm for finding all k-edge-connected components"](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0136264), which is based on AuxillaryGraphs. This algorithm can handle any positive integer value of `k` and works on undirected and directed graphs. 

For k-edge-subgraphs, I've implemented the algorithm in ["Finding Maximal k-Edge-Connected Subgraphs from a Large Graph"](https://openproceedings.org/2012/conf/edbt/ZhouLYLCL12.pdf), which is based on recursive subdivisions of global min-cuts of a graph. 

For 2-edge-ccs on undirected graphs (also known as bridge-components), I've implemented a simple algorithm by Tarjan, where all bridges are found and removed, and then the remaining CCs are returned. 

### Note on the naming convention
When I started this PR I was confused about the naming convention of k-edge-ccs and k-edge-subgraphs. After reviewing several papers in the literature I've concluded that the definitions given above are correct. However, I have found an inconsistency in the  [wolfram library](http://reference.wolfram.com/language/ref/KEdgeConnectedComponents.html), where there k-edge-cc algorithm seems to actually be returning k-edge-subgraphs. I'm not sure why the wolfram function is named the way it is. 

### Summary of tasks 

- [x] Implement k-edge-ccs algorithms 
- [x] Implement k-edge-subgraphs algorithms
- [x] Add tests for k-edge-ccs
- [x] Add tests for k-edge-subgraphs 
- [x] Add "See Also" docstring blocks to relevant locations
- [x] Add .rst documentation for the new module. 

I will removing the debugging code after at least one reviewer has OK-ed the general API.


networkx/algorithms/connectivity/edge_kcomponents.py
```diff
@@ -0,0 +1,570 @@
...
+    if G.is_directed():
+        for cc in nx.strongly_connected_components(H):
+            yield cc
+    else:
+        for cc in nx.connected_components(H):
+            yield cc
+
+
+def general_k_edge_subgraphs(G, k):
+    """
```
Erotemic: This function should probably be renamed or removed from `__all__`, but I'm sure what the best name for this function is. Currently I have it in `__all__` because I want to use it in the tests. I can't seem to access it otherwise (is there a way I can get to the function even if its not in `__all__`?).

networkx/algorithms/connectivity/tests/test_edge_kcompoments.py
```diff
@@ -0,0 +1,574 @@
...
+        _check_edge_connectivity(G)
+
+
+def test_karate_directed():
+    G = nx.karate_club_graph().to_directed()
+    _check_edge_connectivity(G)
+
+
+def timing():
+    seed = 3894723670
```
Erotemic: This is timing code using my own libraries. It will be removed once the rest of the PR is ok-ed.

networkx/algorithms/connectivity/tests/test_edge_kcompoments.py
```diff
@@ -0,0 +1,574 @@
...
+
+if __name__ == '__main__':
+    r"""
+    CommandLine:
+        python ~/code/networkx/networkx/algorithms/connectivity/tests/test_edge_kcompoments.py
+
+        nosetests ~/code/networkx/networkx/algorithms/connectivity --with-doctest --verbose
+        nosetests ~/code/networkx/networkx/algorithms/connectivity/tests/test_edge_kcompoments.py --with-doctest --verbose
+    """
+    # TODO: remove after development is complete
```
Erotemic: testing code that will be removed after the main structure of the PR is OKed.

networkx/algorithms/bridges.py
```diff
@@ -58,7 +59,13 @@ def bridges(G, root=None):
...
     nodes in the graph and *m* is the number of edges.
 
+    References
+    ----------
+    .. [1] https://en.wikipedia.org/wiki/Bridge_(graph_theory)#Bridge-Finding_with_Chain_Decompositions
     """
+    # This is an implementation of the algorithm described in _[1]. Bridges can
+    # be found using chain decomposition. An edge e in G is a bridge if and
+    # only if e is not contained in any chain.
     chains = nx.chain_decomposition(G, root=root)
```
Erotemic: I had my own implementation of this algorithm, but in lieu of #2525 I used that implementation. I merged in the documentation from my own function because I think its helpful. Not sure if that's appropriate for this PR.

networkx/algorithms/connectivity/edge_kcomponents.py
dschult: Yes - you can:
```from nx.algorithms.connectivity.edge_kcomponents import general_k_edge_subgraphs```
That magic is done via the ```.__init_.py``` files in each directory which make the directory a module for others to import.  When you don't put a function in ```__all__```, then you have to import it specifically to get it, but you can still get it.

networkx/algorithms/bridges.py
dschult: I would go ahead an include this comment under a docstring section called "Notes".

dschult:
The general API looks good.  You should check the github wiki page for style of the header for the module. (copyright and author in comments now instead of author variable -- then docs then imports.

I'm not sure about names for ```general_k_edge_subgraphs```. You would probably know the terms better than I.  I have also not looked at the references nor check for accuracy. 

This looks on the right track.

Erotemic:
I prefixed it with general because the algorithm works for arbitrary k. A better name would be `k_edge_subgraphs`, but I'm using that name for the main function, which will only call `general_k_edge_subgraphs`, if there is not a more efficient implementation given the special case of k (e.g. for k=2, it calls `bridge_components`).  However, I don't like the general prefix. Another option might be `any_k_edge_subgraphs`, or maybe just prefixing with an underscore and calling it `_k_edge_subgraphs`. I'm open to suggestions on this.

Erotemic:
@dschult I read through the github wiki and I believe I've made the appropriate changes. I also rebased and squashed the PR into 4 commits. 

Is it ok for the module docstring to be multiline or should that just be a single line? I didn't see anything that says otherwise, but all the modules I looked at had very short module docstrings. I made mine a little longer because I think its useful to have the definition of the structures it computes instead of just saying it computes them.

dschult:
Including more information in the module docstring is encouraged. :)  

It looks like you still have some profiling code in the testing module (```if __name__ == ```...)

Thanks!

Erotemic:
fixed that.

dschult:
This looks ready for merge as far as i can tell.  Are you planning more for this PR?

There is one comment just before the ```construct``` method about not_implemented_for('multi'). I'm fine with that being there as a comment, but just thought I would verify that there is some issue with actually using the decorator.

Erotemic:
Actually, I'll get the code coverage up to 100% before we merge. I should be able to do this today or tomorrow.

Erotemic:
Its a good thing I checked the coverage. I had a `k<1` instead of a `k<=1`, which make the slow algorithm run instead of the fast one. It should be up to 100% now for the new modules.

PR #2558, opened 2017-07-27
Create separate functions for df as edge-lists and adjacency matrices

### LINKED ISSUE #2464: to_pandas_dataframe is not the inverse of from_pandas_dataframe
The function `from_pandas_dataframe` assumes that the data frame has edge-list like structures, but `to_pandas_dataframe` generates an adjacency matrix. 

Personally, I think the edge-list like structures are more useful because:
(1) Otherwise these two functions are just (wrapped) duplications of `numpy_matrix` related functions
(2) The edge-list like data frames are convenient for other interfaces such as `py2cytoscape`. 

Some related commits:
[52eb3d3](https://github.com/networkx/networkx/commit/52eb3d38ff78e4002a424ae5e8976998afdaf1a2): `from_pandas_dataframe` and `to_pandas_dataframe` were added. At that time both functions were about adjacency matrix.  
[f0c22dd](https://github.com/networkx/networkx/commit/f0c22dd3305ff236860c651d8a08fbabcb8b0a77): `from_pandas_dataframe` was changed to current behavior, but it seems that `to_pandas_dataframe` was forgotten to be changed accordingly.

jarrodmillman:
Fixes #2464.

Some related commits:
52eb3d3: from_pandas_dataframe and to_pandas_dataframe were added to work with adjacency matrix.
f0c22dd: from_pandas_dataframe was changed to work with edge-lists, but to_pandas_dataframe was not changed.

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -1,3 +1,9 @@
+#    Copyright (C) 2006-2017 by
+#    Aric Hagberg <hagberg@lanl.gov>
+#    Dan Schult <dschult@colgate.edu>
+#    Pieter Swart <swart@lanl.gov>
+#    All rights reserved.
+#    BSD license.
 """Functions to convert NetworkX graphs to and from numpy/scipy matrices.
 
 The preferred way of converting data to a NetworkX graph is through the
@@ -8,32 +14,25 @@ Examples
 --------
 Create a 10 node random graph from a numpy matrix
 
->>> import numpy
->>> a = numpy.reshape(numpy.random.random_integers(0,1,size=100),(10,10))
+>>> import numpy as np
+>>> a = np.reshape(np.random.random_integers(0, 1, size=100), (10, 10))
 >>> D = nx.DiGraph(a)
 
 or equivalently
 
->>> D = nx.to_networkx_graph(a,create_using=nx.DiGraph())
+>>> D = nx.to_networkx_graph(a, create_using=nx.DiGraph())
 
 See Also
 --------
 nx_agraph, nx_pydot
 """
-#    Copyright (C) 2006-2014 by
-#    Aric Hagberg <hagberg@lanl.gov>
-#    Dan Schult <dschult@colgate.edu>
-#    Pieter Swart <swart@lanl.gov>
-#    All rights reserved.
-#    BSD license.
+
 import warnings as _warnings
 import itertools
 import networkx as nx
 from networkx.convert import _prep_create_using
 from networkx.utils import not_implemented_for
-__author__ = """\n""".join(['Aric Hagberg <aric.hagberg@gmail.com>',
-                            'Pieter Swart (swart@lanl.gov)',
-                            'Dan Schult(dschult@colgate.edu)'])
+
 __all__ = ['from_numpy_matrix', 'to_numpy_matrix',
            'from_pandas_dataframe', 'to_pandas_dataframe',
            'from_pandas_adjacency', 'to_pandas_adjacency',
@@ -141,7 +140,7 @@ def to_pandas_adjacency(G, nodelist=None, dtype=None, order=None,
 
 
 def from_pandas_adjacency(df, create_using=None):
-    """Return a graph from Pandas DataFrame.
+    r"""Return a graph from Pandas DataFrame.
 
     The Pandas DataFrame is interpreted as an adjacency matrix for the graph.
 
@@ -170,7 +169,6 @@ def from_pandas_adjacency(df, create_using=None):
     --------
     Simple integer weights on edges:
 
-    >>> import numpy as np
     >>> import pandas as pd
     >>> df = pd.DataFrame([[1, 1], [2, 1]])
     >>> df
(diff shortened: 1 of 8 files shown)


networkx/tests/test_convert_pandas.py
```diff
@@ -34,40 +34,69 @@ def __init__(self, ):
     def assert_equal(self, G1, G2):
```
jarrodmillman: Should we use ``assert_graphs_equal`` instead?

dschult: ```is_isomorphic``` is much harder to test than what our ```assert_graphs_equal``` tests.
It looks like the current code doesn't actually need the is_isomorphic though.  
Those graphs really are the same -- not just isomorphic.  If that's true, then yes -- please use ```assert_graphs_equal```.

networkx/convert_matrix.py
```diff
@@ -129,8 +143,123 @@ def to_pandas_dataframe(G, nodelist=None, dtype=None, order=None,
...
+
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> df = pd.DataFrame([[1, 1], [2, 1]])
+    >>> df
+       0  1
+    0  1  1
+    1  2  1
+    >>> G = nx.from_pandas_adjacency(df)
+    >>> nx.info(G)
```
jarrodmillman: I tried using ``print(nx.info(G))``to make the output nicer.  But I was getting an error presumably due to spacing issues.  Thoughts?

networkx/convert_matrix.py
```diff
@@ -129,8 +143,123 @@ def to_pandas_dataframe(G, nodelist=None, dtype=None, order=None,
...
+
+    """
+    edgelist = nx.to_edgelist(G)
+    source = [s for s, t, d in edgelist]
+    target = [t for s, t, d in edgelist]
+    all_keys = set().union(*(d.keys() for s, t, d in edgelist))
+    edge_attr = {k: [d.get(k, float("nan")) for s, t, d in edgelist] for k in all_keys}
+    edgelistdict = {'source': source, 'target': target}
+    edgelistdict.update(edge_attr)
+    return pd.DataFrame(edgelistdict)
```
jarrodmillman: This is just a sketch.  I wanted to get some feedback before I put too much effort into this.  Does this seem like a reasonable approach?  If so, I will implement this more fully for more substantial feedback.

networkx/convert_matrix.py
```diff
@@ -20,29 +26,37 @@
...
+    """
+    msg = "to_pandas_dataframe is deprecated and will be removed" \
+        "in 2.1, use to_pandas_adjacency instead."
+    _warnings.warn(msg, DeprecationWarning)
+
+    return to_pandas_adjacency(G, nodelist=None, dtype=None, order=None,
+                               multigraph_weight=sum, weight='weight', nonedge=0.0)
+
+
+def to_pandas_adjacency(G, nodelist=None, dtype=None, order=None,
```
jarrodmillman: TODO: document ``dtype`` and ``order``.

networkx/convert_matrix.py
dschult: Yup -- I think this is what is needed.
You should just use ```G.edges(data=True)``` instead of ```nx.to_edgelist(G)```

jarrodmillman:
@dschult Would you review this?  In particular, here is how I checked that both ``to``/``from`` pairs would provide a roundtrip:
```
    def test_roundtrip(self):

        # edgelist
        Gtrue = nx.Graph([(1, 1), (1, 2)])
        df = nx.to_pandas_edgelist(Gtrue)
        G = nx.from_pandas_edgelist(df)
        assert_graphs_equal(Gtrue, G)

        # adjacency
        Gtrue = nx.Graph(({1: {1: {'weight': 1}, 2: {'weight': 1}}, 2: {1: {'weight': 1}}}))
        df = nx.to_pandas_adjacency(Gtrue, dtype=int)
        G = nx.from_pandas_adjacency(df)
        assert_graphs_equal(Gtrue, G)
```

Note that if I did
```
>>> Gtrue = nx.Graph([(1, 1), (1, 2)])
>>> df = nx.to_pandas_adjacency(Gtrue, dtype=int)
>>> G = nx.from_pandas_adjacency(df)
```
then
```
>>> Gtrue.edges(data=True)
EdgeDataView([(1, 1, {}), (1, 2, {})])
>>> G.edges(data=True)
EdgeDataView([(1, 1, {'weight': 1}), (1, 2, {'weight': 1})])
```

I am not sure that is what we want.

Other than that I don't have an other plans for the code.  You can ignore the fact that Travis is failing (for now) because it is due to conda not supporting Python 3.4 recently.  So the newest version of pandas they have is quite old.  I am working on a PR related to this, which I think will solve the problem.  So I will wait for that to be sorted out before I finalize this.  But I don't anticipate changing the API.  So let me know if this seems ready to merge once I sort out the old pandas in conda on python 3.4 issue.

dschult:
Hmmm...  Why does ```to_pandas_edgelist``` add the edge attribute ```weight```?
I would expect that 

    >>> Gtrue = nx.Graph([(1, 1), (1, 2)])
    >>> df = nx.to_pandas_edgelist(Gtrue, dtype=int)
    
would create a pandas frame without any ```weight``` column.  Just ```source``` and ```target```.  I understand that this is not the API that was in place before this PR, but I think it would be better.  Ideally you could specify a list of edge attributes that would each create a column in the dataframe. If the list is empty (or maybe None?) then no edge attribute columns are written. This is how we read the dataframes so we should write them that way too.  I would be fine with that as a separate PR though.

I think the way you are testing the round trip is good.

PR #2559, opened 2017-07-28
Graphml: consistent I/O for key and data attributes of edges in multi-graphs

mhsekhavat:
Before this commit, in order to write a multigraph to qraphml file, both `key` and `data` items of  edges(`G.edges(keys=True, data=True).items()`) were stored as `<data>` children of `<edge>` elements. This caused difference in edges of multigraphs with key and data attributes, before and after consequent write and reads of graphml. `test_preserve_multi_edge_data` was added to `test_graphml.py` to show this inconsistency.
This commit changes graphml reader and writer behavior to:

* read and write `key` of networkx edges only in `id` attribute of `<edge id="...">` elements of graphml.
* read and write `data` of networkx edges only in `<data></data>` elements of graphml.

Only in non-multigraphs, `id` attribute of `<edge id="...">` elements will be read into `data` of networkx edges, because we have no `key` for edges.

### DIFF
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -794,15 +794,16 @@ class GraphMLReader(GraphML):
         # attribute is specified
         edge_id = edge_element.get("id")
         if edge_id:
-            self.edge_ids[source,target] = edge_id  # will be set as edge data if graph is non-multigraph
+            # self.edge_ids is used by `make_graph` method for non-multigraphs
+            self.edge_ids[source, target] = edge_id
             try:
                 edge_id = self.edge_key_type(edge_id)
-            except ValueError: # Could not convert.
+            except ValueError:  # Could not convert.
                 pass
         else:
             edge_id = data.get('key')
 
-        if G.has_edge(source,target):
+        if G.has_edge(source, target):
             # mark this as a multigraph
             self.multigraph = True
 
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -181,7 +181,7 @@ def generate_graphml(G, encoding='utf-8', prettyprint=True):
 
 
 @open_file(0, mode='rb')
-def read_graphml(path, node_type=str):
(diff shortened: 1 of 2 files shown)


mhsekhavat:
I have no idea why all tests with `OPTIONAL_DEPS=miniconda` fail and all tests with `OPTIONAL_DEPS=no` pass!

dschult:
The miniconda includes numpy/scipy, etc.  With them turned off, lots of tests get skipped.
To find out what the actual error was, click on one of the red (bad) reports from the Travis page. It iwll give you the complete output of the nosetests command.  (You can run that on your own machine if you install nose). Search that output for "fail" or "error" and you should find out what the error was.  Usually its something silly and then you submit another commit.  :)

mhsekhavat:
Thanks for your hint @dschult . I amended the commit and fixed the problem.

mhsekhavat:
@dschult Would you please take a look and give me a feedback about this PR? 
https://github.com/networkx/networkx/pull/2559/files#diff-d5131b2d01be42b3d27bc0a554ce7072R340
Shows a test that will fail without this PR.

dschult:
This looks pretty good.  The tests is clear and the way you've solved the issue makes the code slightly easier to read than what was there before too.   Could you check for PEP8 issues? The typical ones are spaces after commas and operators and line length of <80 chars.  You can do ```pep8 graphml.py``` at the command line to check a file once the pep8 program is installed (I use ```pip install pep8```). I suppose you can read the [PEP8 guidelines](https://www.python.org/dev/peps/pep-0008/) and check the code yourself too, but I wouldn't advise it. :)   If that's too much for you now I will get to cleaning it up at some point fairly soon.   Thanks again!

mhsekhavat:
Thanks for your review @dschult . I pip installed pep8 and fixed errors of two files `graphml.py` and `test_graphml.py` which my patch has touched.

dschult:
Thanks very much!

PR #2561, opened 2017-07-28
Some changes to reduce the really long parts of tests

dschult:
The most impact is to only check gl1 and wlm3 using 1 of the flow_func algorithms. They are the slow graphs for testing. The other graphs still use all flow_func algorithms.  We could also cut down on check every pair of nodes to ensure a correct answer, but I didn't get into that.  This seems to cut about 30-40 secs off the time to run nosetests on my machine.


dschult:
@jtorrents I have reduced the testing of flow functions for graphs ```gl1``` and ```wlm3```.
By only running them with one algorithm it saves about 20% of my testing time and we still test the other algorithms with other graphs.  Is this change going to lead to problems?

networkx/algorithms/connectivity/tests/test_cuts.py
```diff
@@ -34,7 +34,7 @@ def _generate_no_biconnected(max_attempts=50):
 def test_articulation_points():
     Ggen = _generate_no_biconnected()
     for flow_func in flow_funcs:
-        for i in range(3):
+        for i in range(1):
```
jarrodmillman: Did you mean to leave this loop here?

networkx/algorithms/connectivity/tests/test_kcutsets.py
```diff
@@ -181,7 +182,7 @@ def _generate_no_biconnected(max_attempts=50):
 
 def test_articulation_points():
     Ggen = _generate_no_biconnected()
-    for i in range(2):
+    for i in range(1):
```
jarrodmillman: Ditto.

networkx/algorithms/connectivity/tests/test_cuts.py
dschult: Thanks for the review!!
I did intend to leave the loop there so it would be easy for someone to change back to multiple random graphs. But perhaps that is misguided on my part.

What do you think?  I could put a comment that the 1 could be increased for testing multiple realizations.

jarrodmillman: Adding the comment seems reasonable.

Issue #1582, opened 2015-06-10
to_agraph failing on node attribute name `n`


jdrudolph:
I have some graph where I add the node attribute `n`

```
G = nx.DiGraph([(1,2)])
G.node[1]['n'] = 1
nx.to_agraph(G)

networkx/drawing/nx_agraph.py:146
TypeError: add_node() got multiple values for argument 'n'
```

Is this behavior wanted? For semantic reasons I would enjoy having 'n' as a node attribute. I am aware that `G.add_node(3, n=10)` has the same behavior. This is the part of `to_agraph` where the error occurs:

```
def to_agraph(N)
    ...
    # add nodes
    for n,nodedata in N.nodes(data=True):
        A.add_node(n,**nodedata)
    ...
```

I would suggest the following change: Allow for a keyword argument to define the attributes one wants to export.

```
def to_agraph(N, node_attr=None):
    ...
    # add nodes
    for n,nodedata in N.nodes(data=True):
        A.add_node(n,**{k: v for k,v in nodedata.values() if node_attr is None or k in node_attr})
    ...
```

The argument would have to be propagated upwards to `nx.draw_networkx` etc. but would allow for an easy attributes filter.

chebee7i:
This is definitely a bug. `n` should be an acceptable node attribute.

chebee7i:
The issue, of course, is that we have to use some name for the parameter to `add_node`. We could change it to `def add_node(self, _n, attr_dict=None, **attr)` or you can pass it in as: `G.add_node(3, attr_dict={'n': 10})`. Not sure if we should change or you should.

jdrudolph:
@chebee7i This is why I'm asking here before I decide to change my personal networkx branch or my project.

Currently it is at least consistent that both `nx.Graph.add_node` and `pygraphviz.AGraph.add_node` do not allow 'n' to be passed in as keyword argument. The latter does not allow for a `attr_dict` though. Should I propose the change in `pygraphviz` instead?

chebee7i:
I'm not sure. 

I just opened #1583 which would change the `n` in `add_node` to a `u` (but `u` as an attribute name would still be problematic). This is mostly a consistency issue, but it would address your problem (at least for networkx). As for pygraphviz, you should definitely open an issue there and link to it here. I think adding an `attr_dict` makes sense, independent of whether `n` is changed to `u` throughout the code.

jdrudolph:
I opened an issue in pygraphviz/pygraphviz#57

[...]

PR #2562, opened 2017-07-28
Help ensure **kwargs doesn't clobber by using specific argument names

### LINKED ISSUE #1583: Normalize node variable in function signatures.
For methods/functions like `G.add_node` and `G.add_edge`, we allow arbitrary attributes to be passed in at the same time. The name choice for the arguments essentially excludes them from being keyword arguments at call time.

For example, `G.add_node` does not allow passing attributes with names `n` or `attr_dict`, while `G.add_edge` does not allow attributes with names `u` or `v`, or `attr_dict`.

We probably should mention somewhere that `n`, `u`, `v`, `attr_dict` are "special" attribute names that cannot be specified in some function signatures, but can still be added via other routes.

I'm wondering if we should normalize usage and always use `u` whenever we are using `n`. This would make `u`, `v`, and `attr_dict` the only special attribute names. This frees us `n` which is perhaps a more common choice for an attribute name having to deal with "the number of".

dschult:
ekey is the new name for the edge key. That is the change with the biggest backward compatibility impact.  Code that says ```G.add_edge(0, 1, key=3)``` will not do what it used to do for a multigraph.

Just think of multiedges as 3-tuples (u, v, ekey) and don't use the keyword for the 3rd argument.

Other changes like changing ```n``` to ```node_to_add``` will have minimal impact because not many people used those names anyway.

Fixes #1583 
Addresses #1582 but doesn't solve it completely without changes on pygraphviz side (which were turned down when proposed almost 2 years ago). Perhaps this change should be turned down too. :)

Feedback welcome.

### DIFF
--- a/networkx/algorithms/bipartite/projection.py
+++ b/networkx/algorithms/bipartite/projection.py
@@ -117,7 +117,7 @@ def projected_graph(B, nodes, multigraph=False):
                     links = set(B[u]) & set(B[n])
                 for l in links:
                     if not G.has_edge(u, n, l):
-                        G.add_edge(u, n, key=l)
+                        G.add_edge(u, n, ekey=l)
         else:
             G.add_edges_from((u, n) for n in nbrs2)
     return G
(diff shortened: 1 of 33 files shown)


dschult:
I'm going to close this and move the associated issues to v2.1.  This attempt to change the function signatures introduces too many subtle bugs in existing code.  Will try again after the release of v2.0.

Issue #2547, opened 2017-07-25
Convert subgraph/reverse/to_directed/to_undirected to readonly views


dschult:
Before I do this, I would like some feedback.

We have tried to make ```subgraph/reverse/to_directed/to_undirected``` fast and light-weight by doing a shallow copy of the graph structure and not copying attribute dicts.  We have run into problems with expectations for how shallow the copies are.  I lump these methods together because they all create a new graph from the original. I don't include ```copy``` because that already either does a ```deepcopy``` or a ```G.subgraph``` and I think it should do whatever subgraph does.

The returned objects are currently graphs in their own right -- somewhat independent of the original.

**I propose to make these return read-only views of the graph.** That implies that they are fast to create and lightweight, but that they are not independent of the original graph. All graph mutations would be reflected in the view and the view itself would not allow mutations.

Users can create an independent graph from the view using ```.copy()``` or ```nx.Graph(subgraph)```. These are not a shallow copy -- all attribute dicts become new dicts requiring memory.  Of course a ThinGraph could be used as a shallow copy (not including attribute dicts -- see docstring for Graph -- and maybe we would want to include this along with ```ordered.py``` as useful subclasses of the base classes).  With ThinGraph they would need to look up any attributes in the original graph. 

What do people think?  
- readonly views that update with the graph
- easy to make into a mutable graph with ```Graph(subgraph)```
- some reduction in availability of shallow copy versions until we can figure out the best way to do that.

dschult:
#2585 provides an example of how this might be done.

Issue #2048, opened 2016-03-28
Shouldn't OrderedGraph.subgraph() preserve edge order?


baharev:
My expectation was that when running:

``` python
from __future__ import print_function
from collections import OrderedDict
from networkx import DiGraph

class OrderedGraph(DiGraph):
    node_dict_factory = OrderedDict
    adjlist_dict_factory = OrderedDict

g = OrderedGraph()
g.add_nodes_from([1, 2, 3])
# Note the relative order: 2 and 1 below!
g.add_edges_from([(2, 3), (1, 3)])
print(list(g.pred[3]))
g_sub = g.subgraph([1, 2, 3])
print(list(g_sub.pred[3]))
```

to see:

```
[2, 1]
[2, 1]
```

that is, the relative order of the predecessors will be the same in the subgraph but this is clearly not the case, as I get

```
[2, 1]
[1, 2]
```

In graph `g`, the relative order of nodes 1 and 2 is 1, 2 in the nodes, but 2, 1 in the predecessors; most likely this is the reason why I see this. I checked with b6b362a8cbd09e58ce44723747e4cdc2fed5f4e8.

**What is the intended behavior? Or am I using the `OrderedGraph` wrong?**

I am currently implementing reverse-mode automatic differentiation where the argument order of an operation, such as subtraction or division, has to be kept (`z=x-y` but in the subgraph I get`z=y-x` which is a disaster for me). Maybe I misunderstood what the `OrderedGraph` is meant to be used for.

hagberg:
I'd say that is unintended.  The DiGraph.subgraph function intentionally uses a shortcut in assigning the successors and predecessors which can permute the order of predecessors (as you have found).   This implementation of subgraph is slower but should preserve the order.

```python
    def subgraph(self, nbunch):
        """Return the subgraph induced on nodes in nbunch.

        The induced subgraph of the graph contains the nodes in nbunch
        and the edges between those nodes.

        Parameters
        ----------
        nbunch : list, iterable
            A container of nodes which will be iterated through once.

        Returns
        -------
        G : Graph
            A subgraph of the graph with the same edge attributes.

        Notes
        -----
        The graph, edge or node attributes just point to the original graph.
        So changes to the node or edge structure will not be reflected in
        the original graph while changes to the attributes will.

        To create a subgraph with its own copy of the edge/node attributes use:
        nx.Graph(G.subgraph(nbunch))

        If edge attributes are containers, a deep copy can be obtained using:
[...]

PR #2565, opened 2017-07-28
WIP Add reversed view and to_(un)directed view

dschult:
Part of #2547, this PR implements views for the reverse view of a directed graph.
It also provides an undirected view of a directed graph in to_undirected and similar for to_directed.

Actually to_directed is the hard case as it requires the view to combine G.succ and G.pred without actually combining the data structures.


dschult:
Views are returned for ```G.subgraph, G.edge_subgraph, G.reverse, G.to_directed, G.to_undirected``` in the spirit of returning iterators/views instead of large objects. This is a fairly large change in terms of code. The API changes in the sense that the views are readonly. I still need to update the docs.  I think this is the last major base class API change for v2.0. (famous last words)

dschult:
Fixes #2547 
Fixes #2048 (OrderedGraph order for subgraphs)

PR #2553, opened 2017-07-26
Refactor set_node_attributes() and set_edge_attributes()

Michael-E-Rose:
This is related to #2343:

- `set_node_attributes()` and `set_edge_attributes()` no longer throw KeyErrors when a key in the passed dict is not present in the graph
- `set_node_attributes()` and `set_edge_attributes()` both accept dictionary of dictionaries, conditional on parameters "values" and "name"
- parameter "name" now has a default value (`None`), and therefore parameter "name" now has to come last
- harmonized, corrected and extended documentation for `set_node_attributes()` and `set_edge_attributes()`

PR #2568, opened 2017-07-31
[DOC] Added blurb about set_edge/node_attributes

Erotemic:
I noticed that after PR #2553 the new set_edge_attributes and get_edge_attribute API broke my codebase. 
I didn't see any documentation about the new API in the migration from 1.x.x to 2.0 doc, so after finding a fix, I added a write-up describing it.

### DIFF
--- a/doc/release/migration_guide_from_1.x_to_2.0.rst
+++ b/doc/release/migration_guide_from_1.x_to_2.0.rst
@@ -108,3 +108,26 @@ Let's have a look on them.
 	[1]
 
 The same changes apply to MultiGraphs and MultiDiGraphs.
+
+
+The order of arguments to `set_edge_attributes` and `set_node_attributes` has
+changed.  The position of `name` and `values` has been swapped, and `name` now
+defaults to `None`.  The previous call signature of `(graph, name, value)` has
+been changed to `(graph, value, name=None)`. The new style allows for `name` to
+be ommitted in favor of passing a dictionary of dictionaries to `values`.
+
+A simple method for migrating existing code to the new version is to explicitly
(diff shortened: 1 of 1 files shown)


dschult:
Oops... sorry about that. I am guilty of merging changes faster than I'm documenting them. I am planning to shift to doc mode soon. 

Meanwhile:  would it be better to revert to the old argument ordering and make BOTH ```name``` and ```values``` have default ```None```?  If ```values is None```, we could raise an error, or better yet, check if ```name``` is a dict and raise an error explaining to use the keywords ```values=```.   Hmmm...  This doesn't sound very smooth either.    gnash -- gnash....   :)

Erotemic:
@dschult I think we all are guilty of that sometimes. For me, if its not a docstring I almost never remember to do it. 

On the other subject: 

I think if you are willing to have this function break backwards compatibility (which seems fine as the rest of networkx 2.0 does), the new signature works well. 

I really like that we can now specify dicts of dicts. It was very cumbersome to specify multiple properties previously. So that functionality is really nice. I also did like having the name be on the left. It seems more natural that way and mirrors the key/value idiom. However, the new signature still sort of allows for this (e.g. `nx.set_node_attributes(G, {'name': val})`). 

Other solutions I thought of might be to use `kwargs` or a third argument, but I ended up not liking them enough to describe them.

dschult:
I agree with all you say...   I always get the order of ```hasattr(obj, name)``` backwards. 
But luckily getting the order wrong results in an obvious error -- not a secret behind the scene bug. As long as we're moving to v2.0 people will expect some changes.

Thanks for the PR!

PR #2572, opened 2017-08-01
Added function for finding a k-edge-augmentation

Erotemic:
This is the second part of PR #2459. Currently this is just an initial commit that lays out the basic components of the new module. 

After PR #2554 is working and merged, I'll continue work on this part. 

### Definition
* **k-edge-augmentation**: given a graph G that is not k-edge-connected a k-edge-augmentation is a small set of edges that would make G k-edge-connected if they were added.

-------
From the other PR. 

The second module is edge_augmentation

I only have efficient algorithms for the case where k=1 and k=2 (although in my project I just wrote a greedy algorithm for k>=3). Each function has weighted and unweighted versions. I'm not sure what the best way to handle the API for that is yet. My working solution takes in a list of available edges that can have optional weights defaulted to 1. Otherwise it is assumed that the complement of G is available and those edges have weights 1.

This module probably needs more work than just cosmetics and it might be better handled as a separate PR.

When k=1, one_edge_augmentation basically solves an MST problem, but my implementation solves it without constructing the complement of the graph. The weighted version is done in weighted_one_edge_augmentation.

When k=2, bridge_augmentation performs an algorithm described by Eswaran and Tarjan. In the unweighted case it is exact. For the weighted case the problem becomes NP-hard. I'm not sure what the policy on approximation algorithms is, but weighted_bridge_augmentation solves the problem within an approximation ratio of 2.

Lastly, I was surprised that I couldn't find a NetworkX function to collapse a set of nodes into a single node in a meta graph. There is the condense function for SCCs, but I implemented collapse for arbitrary groupings. Maybe I missed a way to do it that already exists. Currently I plan to keep it in this module as a private function, but I think it might be worth exposing in the NetworkX API.

-----

### Tasks 

- [x] get PR #2554 reviewed, finished, and merged
- [x] finalize API
- [x] cleanup implementations for k=1 and k=2 (and ensure they are working)
- [x] add comprehensives tests
- [x] cleanup documentation and docstrings
- [x] add examples to docstrings
- [x] whatsnew / release notes
- [x] add an algorithm for a partial-augmentation
- [x] add a greedy "approximation" algorithm for general k


Erotemic:
(I realized forgot to add a whatsnew entry for #2554 so I did it in this PR). 

@dschult  I've finished the main development on this PR and it is ready for review. 

* I'm confident in the correctness of `unconstrained_one_edge_augmentation`, `weighted_one_edge_augmentation`, `unconstrained_bridge_augmentation`, and `weighted_bridge_augmentation`. 
* I've profiled the code and done some speed optimizations. 
* I came up with a method for finding a good "partial" augmentation (see `partial_k_edge_augmentation`) in the case where 
  a real augmentation is not feasible. Its fairly easy to show the correctness of this method, and I have tests that ensure this. 

Here are some potential issues / things I wasn't sure about:

* Perhaps the  `is_k_edge_connected` and `is_locally_k_edge_connected` algorithms should be in a different module?
* Perhaps the `collapse` / `complement_edges` function should be moved to a different module?
* There are several acceptable formats that `avail` can be specified as. Currently it can either be None, a list of 2-tuples, a list of 3-tuples of a dictionary. (I'm also tempted another way where you could specify it as another nx.Graph, which might be less efficient but it might also be more convenient). However, it might be better to have a single way of specifying `avail`. Personally, I like having the format of this parameter be flexible. If having multiple ways to specify it is ok, then I'd like to add the nx.Graph option as well. 
* The function `greedy_k_edge_augmentation` is something I came up with. I can gaurentee it produces a feasible solution if one exists, but I haven't done the work to figure out if it provides any approximation ratio. Its a simple method, so perhaps there's an associated publication that has this information, but I don't know it. I could understand not wanting to include it in NetworkX, in which case I would remove it and raise a `NotImplementedError` if `k>2` was given to `k_edge_augmentation`
* There is a random component to my greedy algorithm. I expose the seed parameter in the algorithm itself, but not in the main `k_edge_augmentation`, where I always seed it with 0 to ensure the function is deterministic. I'm not sure if this is consistent with networkx standards for handling randomization.
* There is one place that I rely on the order of the undirected edges in `_lightest_meta_edges`, which could cause an issue if a graph contained nodes of mixed types. However, this is only applied to edges in a collapsed metagraph, so those nodes should always be guaranteed to be integers.

Erotemic:
I fixed the test failures which were due to python 2 compatibility issues. To fix one test I had to address an issue where `random.shuffle` works differently between python2 and python3 even if the random number generator is seeded. I simply placed this `compat_shuffle` method in the same file, but is there a better place for this? I don't see any python2/3 compatibility module, nor do I see `six` as a dependency. It might be nice to import zip and range for all those python2 folks, but on the other hand its 2017 and python2 is 7 years old.

dschult:
The lowest_common_ancestor code has been merged in #2597 
Rebase to use it.

Erotemic:
I've rebased and used the new LCA algorithm. It makes computation much faster because I'm doing multiple LCA computations in the same graph. 

For reference here is how I compared timing: 

```python

from networkx.algorithms.connectivity.edge_augmentation import *
from networkx.algorithms.connectivity.edge_augmentation import (
    _unpack_available_edges, _lightest_meta_edges, _minimum_rooted_branching,
    collapse, complement_edges, it, nx
)

# Generate a graph that isn't 2-edge-connected
# G = nx.gnp_random_graph(7, .5, seed=0)
# G = nx.gnp_random_graph(7, .3, seed=0)
G = nx.gnp_random_graph(100, .02, seed=0)

# Make sure G is connected
G.add_edges_from(nx.k_edge_augmentation(G, k=1))

bridge_ccs = nx.connectivity.bridge_components(G)
C = collapse(G, bridge_ccs)
# Use the meta graph to shrink avail to a small feasible subset
mapping = C.graph['mapping']
weight = None
avail = list(complement_edges(G))
avail_uv, avail_w = _unpack_available_edges(avail, weight=weight, G=G)
# Choose the minimum weight feasible edge in each group
meta_to_wuv = {
    (mu, mv): (w, uv)
    for (mu, mv), uv, w in _lightest_meta_edges(mapping, avail_uv, avail_w)
}


root = next(n for n in C.nodes() if C.degree(n) == 1)
TR = nx.dfs_tree(C, root)


def _lowest_common_anscestor(T, u, v, root):
    # Find a least common anscestors
    v_branch = nx.ancestors(T, v).union({v})
    u_branch = nx.ancestors(T, u).union({u})
    common = v_branch & u_branch
    if len(common) == 0:
        lca = None
    else:
        lca = max(
            (nx.shortest_path_length(T, root, c), c)
            for c in common
        )[1]
    return lca


import ubelt
for timer in ubelt.Timerit(5, 'new method'):
    D = nx.reverse(TR).copy()
    with timer:
        lca_lookup = dict(nx.tree_all_pairs_lowest_common_ancestor(
            TR, root=root, pairs=meta_to_wuv.keys()))
        for (mu, mv), (w, uv) in meta_to_wuv.items():
            lca = lca_lookup[(mu, mv)]

            if lca == mu:
                # If u is an ancestor of v in TR, then add edge u->v to D
                D.add_edge(lca, mv, weight=w, generator=uv)
            elif lca == mv:
                # If v is an ancestor of u in TR, then add edge v->u to D
                D.add_edge(lca, mu, weight=w, generator=uv)
            else:
                # If neither u nor v is a ancestor of the other in TR
                # let t = lca(TR, u, v) and add edges t->u and t->v
                # Track the original edge that GENERATED these edges.
                D.add_edge(lca, mu, weight=w, generator=uv)
                D.add_edge(lca, mv, weight=w, generator=uv)
D1 = D.copy()


import ubelt
for timer in ubelt.Timerit(5, 'old method'):
    D = nx.reverse(TR).copy()
    with timer:
        for (mu, mv), (w, uv) in meta_to_wuv.items():
            lca = _lowest_common_anscestor(TR, mu, mv, root)

            if lca == mu:
                # If u is an ancestor of v in TR, then add edge u->v to D
                D.add_edge(lca, mv, weight=w, generator=uv)
            elif lca == mv:
                # If v is an ancestor of u in TR, then add edge v->u to D
                D.add_edge(lca, mu, weight=w, generator=uv)
            else:
                # If neither u nor v is a ancestor of the other in TR
                # let t = lca(TR, u, v) and add edges t->u and t->v
                # Track the original edge that GENERATED these edges.
                D.add_edge(lca, mu, weight=w, generator=uv)
                D.add_edge(lca, mv, weight=w, generator=uv)
D2 = D.copy()

passed = (D1.edges() == D2.edges())
print('passed = {!r}'.format(passed))
```

The results of the 100 node gnp graph are 20x as fast in this subroutine. 

```
Timing complete for: new method, 5 loops, best of 3
    time per loop : 10.64 ms seconds
Timing complete for: old method, 5 loops, best of 3
    time per loop : 195.8 ms seconds
```

The smaller 7 node case twice as fast in this subroutine.

Erotemic:
@dschult I'm finished with this PR, any chance it could be reviewed and merged?

dschult:
@erotemic please go through the docstrings and add ```Parameters``` sections. I think it would also be helpful to add a summary sentence for what each function does -- along with definitions. You have some text at the module level to define your terms, but even that doesn't define k-edge-connected. We need definitions at the function level too -- remember that in the online docs, each function has their own page without the viewer being able to see the module level docs.  I'm not saying these should be long -- they shouldn't. But one or two sentence definitions of what the function does would be helpful.

Erotemic:
@dschult I added lots more documentation as well as refining existing documentation. 

There were a few helper functions I didn't put full blown parameter sections in for. I ensured that those functions were all prefixed with an underscore. If you'd prefer the helpers to have parameter sections let me know and I'll put them in.

dschult:
Looks great!   Thanks very much!

PR #2574, opened 2017-08-01
WIP: Add guide to setting up a development environment

harryscholes:
This PR adds a guide to setting up a development environment for newbies.

**Motivation:**
 - I recently contributed #2522 to NetworkX (my first open source contribution)
 - Learning how to set up the development environment took longer than adding the commits
 - This guide should make it easier for relative novices to contribute

The guide uses Anaconda's virtual environments to create a clean environment and then installs a git cloned copy of NetworkX along with its required packages.

### DIFF
--- a/CONTRIBUTE.rst
+++ b/CONTRIBUTE.rst
@@ -33,6 +33,9 @@ Developer overview
 
       git checkout -b bugfix-for-issue-1480
 
+   * For more information on how to set up a development environment, refer to
+     the :doc:`setup guide <setup>` for developers
+
    * Commit locally as you progress (``git add`` and ``git commit``)
 
 3. To submit your contribution:
--- a/doc/developer/index.rst
+++ b/doc/developer/index.rst
@@ -7,4 +7,5 @@ Developer Guide
    :maxdepth: 2
 
    contribute
+   setup
    gitwash/index
--- a/doc/developer/setup.rst
+++ b/doc/developer/setup.rst
@@ -1,0 +1,60 @@
+.. _setup:
+
+.. highlight:: bash
+
+===============================================
+ Guide to setting up a development environment
+===============================================
+
+This guide explains how to set up a clean development environment using
+`Anaconda <https://anaconda.org/>`__ virtual environments and ``virtualenv``
+environments.
+
+Anaconda
+========
+
+First, ensure that Anaconda is installed, then fork and clone NetworkX and check
+out a development branch. Now create a new Conda environment, in this case
+called ``networkx_dev``::
+
+   conda create -n networkx_dev python=3
+
+If you use Jupyter notebooks to run code during development, instead run::
+
+   conda create -n networkx_dev python=3 jupyter nb_conda
+
+Once the environment is created, activate it::
+
+   source activate networkx_dev
+
+Install an editable version of NetworkX and its required packages::
+
+   pip install -e .
+   pip install -r requirements.txt
+
+N.B. ``gdal`` may need to be installed before installing the required packages
+e.g.::
+
+================ ============================================================
+Debian / Ubuntu  ``sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable``
+                 ``sudo apt-get update``
+                 ``sudo apt-get install gdal``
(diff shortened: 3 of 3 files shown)


dschult:
Nice -- thanks...   

How portable is this? For example, you use ```brew``` to install ```gdal```?  Will that work for everyone?  
Does it need to happen when the ```source activate``` is active or can it be done any time?

Can you specify what is generic and what is special for each OS?

Finally, I think the ```Guide``` link is an [internal link and can be marked up](http://docutils.sourceforge.net/docs/user/rst/quickref.html#internal-hyperlink-targets) as such.

jarrodmillman:
This might be better to add here:
- https://github.com/networkx/networkx/wiki
or here (this would be a new file):
- doc/developer/setup.rst

How you setup your environment is fairly developer and OS specific.  (For example, I don't use conda or brew at all.)  I would keep the `CONTRIBUTE.rst` short and generic.  Of my two suggestions, I would slightly prefer putting it in the developer docs.

If you want to include the text about notebooks, you should probably make it clear that the notebook is just for playing around with and that notebooks aren't meant to be added to the source repository.

harryscholes:
Thanks for your comments guys.

@dschult 
* I will change the instructions for installing `gdal` and try to make the whole guide more OS-agnostic.

@jarrodmillman 
* I think `docs/developer/setup.rst` would be a good home.
* I know that Conda isn't for everyone, but it can at least give people ideas about how to install an editable copy of NetworkX. Out of interest, do you use `virtualenv`? Do you think giving a guide using `virtualenv` would be more useful?

jarrodmillman:
Yes, I use ``virtualenv``.  It probably make sense to include instructions for both.  If you want to add ``virtualenv`` instructions, I am happy to review them.  I can also provide some text (later today), if you wish (just let me know).

Also, since I use Fedora, I use ``dnf`` instead of brew.  We should have a little table for things like this.  For example, here is a table for installing ``git``:
  http://networkx.readthedocs.io/en/latest/developer/gitwash/git_install.html

Another suggestion for the name is ``docs/developer/environment.rst``.  I am not sure which I prefer.  So I will leave it to you.

harryscholes:
@jarrodmillman if you could provide the `virtualenv` instructions that would be very handy as I have never properly used it. I'll migrate the guide into `docs/developer/setup.rst` and commit again.

CONTRIBUTE.rst
```diff
@@ -33,6 +33,9 @@ Developer overview
 
       git checkout -b bugfix-for-issue-1480
 
+   * For more information on how to set up a development environment, refer to
+     the `guide <https://github.com/networkx/networkx/blob/master/CONTRIBUTE.rst#guide-to-setting-up-a-development-environment>`__ below
```
jarrodmillman: ... refer to the :doc:`developer guide <setup>`.

CONTRIBUTE.rst
```diff
@@ -150,6 +153,48 @@ you may be interested in more fundamental comments about design.
 When you think the pull request is ready to merge, change the title (using the
 *Edit* button) to remove the ``WIP:``.
 
+Guide to setting up a development environment
```
jarrodmillman: Please remove this section.

jarrodmillman:
@harryscholes No problem, I will add the ``virtualenv`` section later tonight or early tomorrow.  In the meantime, would you be able to address my inline comments and squash your commits into one commit?  Thanks.

harryscholes:
Inline comments have been corrected. The commits have been squished into a single commit.

jarrodmillman:
Thanks!  I am looking into something related to ``gdal`` and will finalize and merge this after that.

dschult:
@jarrodmillman 
Are we ready to merge this Developer Setup guide?
Any remaining tasks?

jarrodmillman:
I will take a look at this while I am looking into #3030

dschult:
@jarrodmillman could you see if this PR is ready to merge? 
It says conflicts exist, but I can't find any...

jarrodmillman:
@dschult I'll take another look and merge it before the 2.4rc1 (after adding venv instructions).

PR #2576, opened 2017-08-03
Update harmonic centrality

bioglio:
I propose a new version of harmonic centrality that exploits the method nx.single_target_shortest_path_length (form pull request #2418) to avoid reversing the graph. 
Moreover, I made its signature similar to the other centrality algorithms, by adding "normalized" and "reverse", and by replacing "nbunch" (a container of nodes) with "u" (a single node).


bioglio:
I forgot to check the tests before committing.. I'm working on these issues

dschult:
I'm going to close this. The reverse operation is a pretty light-weight view and shouldn't be avoided. I'd like to keep the reverse keyword out of the shortest_path routines too. So I'm closing this PR.  If there are other parts that I missed open it again.

PR #2577, opened 2017-08-03
Use codecov instead of coveralls

jarrodmillman:
This is ready to merge.

Codecov has a much nicer interface and better features than coveralls.  ``scipy``, ``scikit-image``, ``scikit-learn``, and ``pandas`` have already switched.  This PR also deletes cruft and has stricter coverage reporting.

The coverage is now 89% (mostly) because I changed
```
branch = False
```
to
```
branch = True
```
in ``.coveragerc``.

The function
```
def <name>(): 
    if <somecase>:
        <dosomething>
    return <something>
```
has two execution paths through ``if <somecase>:``.  Our tests should cover both paths.  Now we track whether they do.

For example, line 96 on the [``convert.py`` page](https://codecov.io/gh/networkx/networkx/src/0eef781b4e77cc8e02d03d58902437d816ac38bf/networkx/convert.py#L96) is yellow because our tests don't cover all possible paths through this line.

You can read more about branch coverage [here](http://coverage.readthedocs.io/en/latest/branch.html#branch).

I will submit additional PRs to improve test coverage by adding additional test over the new two weeks.  For example, getting annoyed with ``coveralls`` reporting while working on [this](https://github.com/networkx/networkx/pull/2558/commits/33b0fddf7920a80baba90e15f4f8cd32a5e54ec8#diff-2cf221e677cf8fd2435587a6fd0f7971) was what motivated this PR.


jarrodmillman:
And
```
codecov/project — No report found to compare against 
```
failed because ``codecov`` has never reported on the ``master`` branch.  Once this is merged, it will and so it won't fail like this again.

dschult:
I haven't used coveralls very much -- my loss... not anything against coveralls.

Does it post the results to the conversation as it did in this conversation? Or is that something you put here to show us what is possible?  I think it might crowd out some conversations to have each push result in a report...  especially if they stay in the conversation when we squash commits.

I like the fact that it handles ```if```` statements better by noting whether they test both branches -- especially when one branch doesn't have code to run, so is effectively invisible.

jarrodmillman:
You can configure whether ``codecov`` posts reports as a comment.  I just added a [.codecov.yml](https://github.com/networkx/networkx/pull/2577/files#diff-1f7b4b85ed8525c5239f741431a72872) file that has the line ``comment: off``, which disables this.  The comment report on this PR already existed and this change doesn't remove existing comments.  But going forward it won't create any more while ``.codecov.yml`` has the ``comment: off`` line.

We can change other aspects of the reporting now or in the future using the ``.codecov.yml`` file.  For more options, see [codecov-yaml](https://docs.codecov.io/v4.3.6/docs/codecov-yaml).

Issue #2581, opened 2017-08-04
approximation.kcomponents seems to depend on order of nodes


dschult:
I'm finding that the test for the approximation k_components routine called ```test_karate_1``` no longer works with the subgraph views replacing the current subgraph implementation. At first I suspected an error in the view code. But it looks like it could be just a dependence of the approximation method to the order of the nodes reported.  Replacing the subgraph calls with subgraph view calls makes the ```k_number``` for nodes 24 and 25 show as 2 instead of 3. All other nodes are correct. In tracing it down, it looks like nodes 32 and 33 are placed before many of the other nodes when the views report nodes. The nodes reported appear to be correct, just in a different order. This may be affecting the approx k_number calculation.

Is this possible? reasonable?    

If this is the problem, should the test check for two possible answers, one of which is wrong but almost right?  Are there better ways to handle this test?

PR #2585, opened 2017-08-05
One attempt to meld graphviews with base classes

dschult:
There are many choices for depth of copies with copy, to_(un)directed, subgraph, reverse. 

Here is one suggestion that still needs rebasing and two tests fail from approxmate kcomponents (see #2581). But it is meant as a start of a conversation.

**Some choices made**

G.copy() - as_view=False     No longer deepcopy...like dict.copy()
G.copy(as_view=True) - a View of the graph
G.to_directed() - as_view=False    Still a deepcopy of data attributes
G.to_directed(as_view=True)    Directed View of the graph
G.to_undirected() - as_view=False    Still a deepcopy of data attributes
G.to_undirected(as_view=True)    Directed View of the graph
G.subgraph(nodes)  Return subgraph view. For new graph use G.subgraph(nodes).copy()
G.edge_subgraph(edges)  Return a subgraph view.
G.reverse()   Return a reversed view. For new graph use G.reverse().copy()

Graphviews are subclasses of graphs, so use of __class__() no longer works when G is a view.

New method:
G.fresh_copy()   Return a new null graph with class of (possibly underlying) graph


dschult:
G.copy:

   - v1.11:  always deepcopy
   - master: with_data=True* -> deepcopy
           with_data=False -> copy of graph structure, but using same datadicts
   - pr2585: as_view=True -> readonly view
           as_view=False* -> shallow copy of graph structure including datadicts

Thoughts: G.copy() should probably look like dict.copy() for uniformity with python.
That is, it provides a shallow copy of data attributes.
If user wants a deepcopy, they can use ```deepcopy```. But then they get a deepcopy of
views too when G is a view. If they treat a view as a graph that might be surprising.

to_directed/to_undirected:

   - v1.11, master, pr2585:   all provide a directed version of the graph with deepcopy
                            of all datadicts.

reverse:   (directed only)

   - v1.11, master:  copy=True -> deepcopy of all datadicts
                   copy=False -> an inplace swap of G.succ/G.pred
   - pr2585:         copy=True -> deepcopy of all datadicts
                   copy=False -> a view with swapped G.succ/G.pred

subgraph and edge_subgraph:

   - v1.11, master:  new graph structure but using old datadicts
   - pr2585:         always a view. to get a new graph use G.subgraph(nodes).copy() 

Thoughts: Using nx.Graph(G.subgraph(nodes)) seems fine until you don't know what type
of graph G is. G.fresh_copy() helps a little, but you still have to add nodes and edges manually.

dschult:
The errors in approximation k-connected were related to AntiGraph and I pushed a fix. 
This attempt to meld graphviews into graph classes should now pass tests.

dschult:
Closed in favor of #2593

Issue #2546, opened 2017-07-25
`networkx.version` shadows any other module named `version` if imported first


benkuhn:
Steps to reproduce:

```
$ pip freeze | grep networkx
networkx==1.11
$ touch version.py
$ python -c 'import version; print(version)'
<module 'version' from '/Users/ben/scratch/version.py'>
$ python -c 'import networkx; import version; print(version)'
<module 'version' from '/Users/ben/.virtualenvs/personal/lib/python3.6/site-packages/networkx/version.py'>
```

Reading the code, it looks like the `release` module is adding the networkx package to `sys.path`, importing version and deleting it again?

Erotemic:
Perhaps this may be fixed by moving some of the logic in release.py to setup.py using some static parsing? In one of my modules, my setup.py looks in the __init__.py file and statically parses the value of `__version__`. Here is that code adapted to networkx:

```python
def parse_version():
    """ Statically parse the version number from version.py """
    from os.path import dirname, join
    import ast
    init_fpath = join(dirname(__file__), 'networkx', 'version.py')
    with open(init_fpath) as file_:
        sourcecode = file_.read()
    pt = ast.parse(sourcecode)
    class VersionVisitor(ast.NodeVisitor):
        def visit_Assign(self, node):
            for target in node.targets:
                if target.id == 'version':
                    self.version = node.value.s
    visitor = VersionVisitor()
    visitor.visit(pt)
    return visitor.version
```

The above function assume you are running from `setup.py`. 

If the release.py code isn't actively being used in the networkx module, this may be a good alternative.

dschult:
I think the issue with shadow version modules is that release.py adds and removes the main networkx directory to sys.path.   I don't think it is a problem to do this in setup.py. And I don't think it is necessary to do it in release.py.  If I am right, then the fix is simply to remove the lines in release.py where we add basedir to sys.path and then pop it. This happens twice in release.py.

I've been able to recreate the symptoms on all versions using networkx packages installed with pip. I then change the release.py file in the site-packages directory. I remove the sys.path changes before and after importing version. I do this in two places in the file.   After making those changes, the import test passes -- no shadow module appears.

PR #2587, opened 2017-08-05
Use relative imports for version and release.

ProgVal:
Closes GH-2546.


ProgVal:
Nevermind, it messes up the import order when running from the setup.py.

PR #2596, opened 2017-08-09
Travis refactor

jarrodmillman:
I have done the following:

1. stop running known failing ci for pypy and ironpython
2. use ``virtualenv``/``pip`` instead of ``conda``
3. moved body of ``before_install`` (``script``) to ``tools/travis/before_install.sh`` (``tools/travis/script.sh``)

I don't think there is much value in running the failing builds.  It takes longer, isn't informative, and makes the build recipes more complicated than they would otherwise be.  Given how long ago these recipes were added, I would likely start from scratch (i.e., I would start by looking at the upstream documentation instead of just modifying the existing code) if I was trying to fix them.   But I haven't seriously looked into whether it would be difficult to get them passing again.  Does this seem reasonable?  If so, I would like to remove the remaining old pypy and ironpython code.  If we want to add these build targets back, I could take another look at it after the 2.0 release.  (If this doesn't seem reasonable, I can just put them back now and leave them failing.)

It is more important for us to test ``virtualenv``/``pip`` than ``conda``.  Pip has become much better, since the conda recipe was added (3 years ago).  For example, the conda recipe starts with a note reading "Workaround for travis-ci/travis-ci#2683" and then has some arcane looking statements.  For ``pip``, here is the entire recipe:

```
if [[ "${OPTIONAL_DEPS}" == pip ]]; then

  # create new empty venv
  virtualenv -p python ~/venv
  source ~/venv/bin/activate

  # needed to build Python binding for GDAL
  export CPLUS_INCLUDE_PATH=/usr/include/gdal
  export C_INCLUDE_PATH=/usr/include/gdal

  # install required packages
  pip install --upgrade pip
  pip install --retries 3 $PIP_FLAGS -r requirements.txt

  # show what's installed
  pip list
fi
```
I have also run into problems with conda not supporting older versions of Python (e.g., PR #2558 fails because conda doesn't provide new packages for Python 3.4 and the version of pandas is several years old).  Using ``virtualenv``/``pip``  solves this issue as it has support for older versions of Python.

Finally, I moved some of the conditional logic to separate bash scripts.  This doesn't change the functionality and makes the logic of ``.travis.yml`` easier to see.

I have more I plan to do, but I wanted to get some feedback before continuing.  Here are the things I would like to do:

1. completely remove the pypy and ironpython recipes (for now)
2. add an osx build
3. further cleanup code and simplify logic

Thoughts?


jarrodmillman:
@hagberg , @dschult --- I would like to finalize this soon, but would like some feedback first.

hagberg:
I agree with the changes.

I do think it is wise to consider keeping (maybe removing and re-adding later) the pypy and ipy testing.  Those implementations found a lot of output ordering issues with our tests since the dictionary implementations are better.  I think they could have even uncovered a bug or two in some algorithms.

Issue #2590, opened 2017-08-08
Add fluid community to docs


dschult:
See #2420

dschult:
Thanks very much @FerranPares !
closed by #2599

PR #2599, opened 2017-08-10
Reference documentation of fluid communities added

FerranPares:
Solving Issue [#2590](https://github.com/networkx/networkx/issues/2590).


FerranPares:
@dschult Seems that everything should be properly referenced now!

PR #2612, opened 2017-08-14
shortest_path return types as discussed for #2510

dschult:
This is a partial solution to #2510.  This commit only handles the following:

Changed to return a dict:
- multi_source_dijkstra_path_length
- single_source_bellman_ford_path_length

Changed to yield 2-tuples:
- all_pairs_shortest_path

changed to return dicts:
- single_source_shortest_path_length
- single_target_shortest_path_length

The following are in #2510:
Still need these to yield 2-tuples:
- all_pairs_dijkstra_path
- all_pairs_bellman_ford_path

Need to add this (and yield 2-tuple):
- all_pairs_dijkstra

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ NetworkX
    :target: https://ci.appveyor.com/project/dschult/networkx-pqott
 
 .. image:: https://readthedocs.org/projects/networkx/badge/?version=latest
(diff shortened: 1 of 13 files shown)


dschult:
I've added/changed the all_pairs_ methods, and folded in the commits from #2510.  So this now
Fixes #2510 
And in turn also
Fixes #2465

PR #2618, opened 2017-08-16
This seems to fix the shadow import for version

### LINKED ISSUE #2546: `networkx.version` shadows any other module named `version` if imported first

dschult:
It works when I make this change to release.py in
already installed v1.11 site-packages. It makes sense
that it would fix it for all future releases too.

Don't add the current directory to sys.path
and then removing it again in release.py. Just import version.

We can still play with sys.path in setup.py.  That gets run 
from the base directory. But release.py runs in the same directory as version.py
So this is the only change needed.

I'm writing all this out because I know very little about this code and I'm hoping to either justify it to myself, or give someone else enough info to tell me I'm wrong (or right).

Fixes #2546

### DIFF
--- a/networkx/release.py
+++ b/networkx/release.py
@@ -101,9 +101,9 @@ vcs_info = %(vcs_info)r
             # This is *good*, and the most likely place users will be when
             # running setup.py. We do not want to overwrite version.py.
             # Grab the version so that setup can use it.
-            sys.path.insert(0, basedir)
+            #sys.path.insert(0, basedir)
             from version import version
-            del sys.path[0]
+            #del sys.path[0]
         else:
             # This is *bad*.  It means the user might have a tarball that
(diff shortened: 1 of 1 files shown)


dschult:
I verified that this allows an install using ```python setup.py install``` that does not have the shadowing behavior.  I'm going to merge this fix.

PR #2614, opened 2017-08-15
WIP: Install optional deps for osx build

jarrodmillman:
This PR enables testing OSX with the optional requirements (see ``requirements/extras.txt``) installed.

@dschult Do you have any thoughts about how to fix the resulting error:

```
======================================================================

FAIL: test_agraph.TestAGraph.testMultiDirected

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/networkx/drawing/tests/test_agraph.py", line 69, in testMultiDirected

    self.agraph_checks(nx.MultiDiGraph())

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/networkx/drawing/tests/test_agraph.py", line 36, in agraph_checks

    self.assert_equal(G, H)

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/networkx/drawing/tests/test_agraph.py", line 29, in assert_equal

    assert_edges_equal(G1.edges(), G2.edges())

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/networkx/testing/utils.py", line 43, in assert_edges_equal

    assert_equal(c1, c2)

AssertionError: 4 != 3

    '4 != 3' = '%s != %s' % _common_shorten_repr(4, 3)

    '4 != 3' = self._formatMessage('4 != 3', '4 != 3')

>>  raise self.failureException('4 != 3')

    

======================================================================

FAIL: test_agraph.TestAGraph.testMultiUndirected

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/networkx/drawing/tests/test_agraph.py", line 66, in testMultiUndirected

    self.agraph_checks(nx.MultiGraph())

  File "/Users/travis/build/networkx/networkx/venv/lib/python3.6/site-packages/networkx/drawing/tests/test_agraph.py", line 36, in agraph_checks

    self.assert_equal(G, H)

[...]

PR #2626, opened 2017-08-19
Add miniconda osx test environment

dschult:
Since Graphviz 2.40.1 from homebrew seems to have trouble with multiedges and it is not obvious how to get an older version of graphviz via homebrew, I put together this alternative where miniconda is used for the optional deps version of osx tests.

related to #2614


jarrodmillman:
This seems like a good solution for now.

PR #2632, opened 2017-08-22
Make parameter nodes effective in effective_size()

### LINKED ISSUE #2631: Parameter 'nodes' is not effective in nx.effective_size()
https://github.com/networkx/networkx/blob/ca839ad24da9ec72abf920eaa33d3fce0a888f4a/networkx/algorithms/structuralholes.py#L144-L162

Parameter variable `nodes` is not used anymore after assigning the default value, which is `G`, at line 145. As another function `nx.constraint()` does in the same file, it seems that for-loops in this code snippet should iterate over `nodes`, not `G`.

kyukyukyu:
* Fixes #2631.

### DIFF
--- a/networkx/algorithms/structuralholes.py
+++ b/networkx/algorithms/structuralholes.py
@@ -145,7 +145,7 @@ def effective_size(G, nodes=None, weight=None):
         nodes = G
     # Use Borgatti's simplified formula for unweighted and undirected graphs
     if not G.is_directed() and weight is None:
-        for v in G:
+        for v in nodes:
             # Effective size is not defined for isolated nodes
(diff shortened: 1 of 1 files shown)


dschult:
Thank you for this.  It looks like the doc_string for this function could also be improved to make this more clear. Would you like to take a stab at it? If not, I'll do that.

kyukyukyu:
Yes, I think that the description for this parameter should be clarified, too. I'll take a look at it, and update the description not only in `effective_size()` but also in `constraint()`, where parameter `nodes` has the same meaning.

dschult:
Thanks

PR #2633, opened 2017-08-22
Deprecate node

jarrodmillman:
This is ready to merge.

### DIFF
--- a/networkx/classes/graph.py
+++ b/networkx/classes/graph.py
@@ -711,6 +711,9 @@ class Graph(object):
         self.__dict__['nodes'] = nodes
         return nodes
 
+    # for backwards compatibility with 1.x, will be removed for 3.x
+    node = nodes
+
     def number_of_nodes(self):
         """Return the number of nodes in the graph.
 
--- a/networkx/classes/tests/test_graph.py
+++ b/networkx/classes/tests/test_graph.py
@@ -9,6 +9,12 @@ import networkx as nx
 from networkx.testing.utils import *
 
 
+def test_deprecated():
+    # for backwards compatibility with 1.x, will be removed for 3.x
+    G = nx.complete_graph(3)
+    assert_equal(G.node, {0: {}, 1: {}, 2: {}})
+
+
 class BaseGraphTester(object):
     """ Tests for data-structure independent graph class features."""
 


jarrodmillman:
With the changes in a44cf71 and https://github.com/scikit-image/scikit-image/pull/2766/files, I believe ``scikit-image`` will pass its tests with networkx 1.X and 2.X.

I made two changes to ``scikit-image``.

1. The original implementation of the ``copy`` method was
```
        g = super(RAG, self).copy()
        g.max_id = self.max_id
        return g
```
had the following behaviour
```
>>> from skimage.future import graph
>>> G = graph.rag.RAG()
>>> G
<skimage.future.graph.rag.RAG at 0x7fbc15273690>
>>> G.copy()
<networkx.classes.graph.Graph at 0x7fbc15273dd0>
```
After changing it to
```
        return deepcopy(self)
```
it behaves like this
```
>>> from skimage.future import graph
>>> G = graph.rag.RAG()
>>> G
<skimage.future.graph.rag.RAG at 0x7f92a619bbd0>
>>> G.copy()
<skimage.future.graph.rag.RAG at 0x7f92a3c9fa10>
```

2. I also added this ``subgraph`` method
```
    def subgraph(self, nbunch):
        """Return the subgraph induced on nodes in nbunch.

        For compatibility with older versions of networkx.  Versions <= 2.0
        ....
        """
        if nx.__version__[0] == '1':
            return super(RAG, self).subgraph(nbunch)
        else:
            bunch = self.nbunch_iter(nbunch)
            # create new graph and copy subgraph into it
            H = self.__class__()
            # copy node and attribute dictionaries
            for n in bunch:
                H.node[n] = self.node[n]
            # namespace shortcuts for speed
            H_adj = H._adj
            self_adj = self.adj
            # add nodes and edges (undirected method)
            for n in H.node:
                Hnbrs = {}
                H_adj[n] = Hnbrs
                for nbr, d in self_adj[n].items():
                    if nbr in H_adj:
                        # add both representations of edge: n-nbr and nbr-n
                        Hnbrs[nbr] = d
                        H_adj[nbr][n] = d
            H.graph = self.graph
            return H
```

@dschult Does this seem like a reasonable approach?

dschult:
This seems to be set up correctly and tests fine.
I don't get any deprecation warning, but that is probably my setup somehow.

I would prefer to make the ```_iter``` routines return an iterator instead of a list.
Also, I have a slight preference to have the properties return ```self.nodes``` instead of ```self._node``` and ```self.adj``` instead of ```self._adj```.

The biggest issue I have with this is that after v2.1 we still have no way to allow people to easily produce code that works for both v1.x and v2.  So the reason for putting it back in with deprecated status will still be there with v2.1.  Don't we need to make a long term decision here?

Here's my proposal:
1) We leave out the ```edges_iter``` and ```nodes_iter``` -- All  they need to do is change their code to ```edges``` and ```nodes``` and it will work in both v1.x and v2.x

2) We put back the ```G.node``` simply duplicating the code for ```G.nodes```.  That gives two names for the same thing -- small pain for backward compatibility.  No deprecation...  We commit to having this available for the long haul --  maybe we can only advertise one of them in the docs.  We'll eventually remove ```G.node``` but not until v1.x is basically no longer used.  Maybe v3.x?

3) We should not provide ```G.edge``` any more.  Have them change their code to ```G.adj``` which is what ```G.edge``` used to point to anyway.

4) Don't provide ```neighbors_iter```, ```adjacency_iter```, ```adjacency_list```, etc.  They should change their code to ```neighbors```, ```adjacency```, etc.

There's still another Issue #2634 about add_path, etc. that should be addressed too. But I think that is a different sort of issue.  So to summarize, my proposal is to make this PR only include code for the ```node``` property -- identical to the ```nodes``` property without the doc_string and without any deprecation notice.  What do you think?

dschult:
Oops... I forgot to address your first point about ```rag.copy```.

Instead of switching to ```deepcopy```, we should leave the ```rag.copy``` code the same but include a new method ```rag.fresh_copy()``` which is just ```return rag()```.  The ```Graph.copy``` method uses that to set up the new object to hold the graph. 

I guess we should add docs to our discussion on subclassing that includes this wart.  We have to expect a lot of subclasses of Graph now that views are subclasses. ```fresh_copy``` was the best solution I came up with.  Maybe there's some OO magic that could make it so we can just use ```G.__class__``` like we used to, but I don't think so.

jarrodmillman:
I am not able to use ``G.edges[1, 2]`` for ``G.edge[1][2]`` with ``networkx-1.11``:
```
>>> import networkx as nx
>>> nx.__version__
'1.11'
>>> G = nx.Graph()
>>> G.add_edge(1, 2, {'weight': 20})
>>> G.edge[1][2]['weight']
20
>>> G.edges[1, 2]['weight']
TypeError: 'instancemethod' object has no attribute '__getitem__'
```

dschult:
You need to replace ```G.edge``` with ```G.adj```   then it works for both.
So:  ```G.node``` equates to ```G.nodes```   but ```G.edge``` should switch to ```G.adj``` (which is what it actually is in v1.x).  

As you point out: the interface for ```G.edges[u, v]``` is different from ```G.edge[u][v]```  and they are NOT interchangeable.

dschult:
Said another way -- 

In code that works for both v1.x and v2.x we should have ```G.node```, ```G.nodes()```, ```G.adj``` and ```G.edges()```. 

We should not have ```G.edge```, ```G.*_iter()```.

jarrodmillman:
In https://github.com/scikit-image/scikit-image/pull/2766/files, I added this method to RAG:
```
    def subgraph(self, nbunch):
        """Return the subgraph induced on nodes in nbunch.

        For compatibility with older versions of networkx.  Versions <= 2.0
        ....
        """
        if nx.__version__[0] == '1':
            return super(RAG, self).subgraph(nbunch)
        else:
            bunch = self.nbunch_iter(nbunch)
            # create new graph and copy subgraph into it
            H = self.__class__()
            # copy node and attribute dictionaries
            for n in bunch:
                H.node[n] = self.node[n]
            # namespace shortcuts for speed
            H_adj = H._adj
            self_adj = self.adj
            # add nodes and edges (undirected method)
            for n in H.node:
                Hnbrs = {}
                H_adj[n] = Hnbrs
                for nbr, d in self_adj[n].items():
                    if nbr in H_adj:
                        # add both representations of edge: n-nbr and nbr-n
                        Hnbrs[nbr] = d
                        H_adj[nbr][n] = d
            H.graph = self.graph
            return H
```

The ``else`` body is just cut and paste from networkx-1.11.  I also looked at the subgraph method here:
  http://networkx.readthedocs.io/en/latest/auto_examples/subclass/plot_printgraph.html#sphx-glr-auto-examples-subclass-plot-printgraph-py

The printgraph example looks like it may need updating as it has the following untested lines:

```
            # copy node and graph attr dicts
            H.node = dict((n, deepcopy(d))
                          for (n, d) in self.node.items() if n in H)
```
@dschult Do you have a better suggestion for what to do about subgraph?

jarrodmillman:
Also, I am not sure ``G.node`` should equate to ``G.nodes`` in v2.  Changing ``G.node`` to ``G.nodes`` in downstream code seems perfectly reasonable to me.

However, I am not sure how to handle ``G.node`` in downstream projects, which are assigning to it.  For example, take a look at ``def _rename_node(graph, node_id, copy_id)`` in 
  https://github.com/scikit-image/scikit-image/blob/master/skimage/future/graph/graph_merge.py
What is the recommended way to modify this code so that it runs on v1 and v2.0b1?

We also need to update the printgraph example (see my comment above).

If we can fix these, then I think we can still remove ``node`` from v2 and just tell folks to modify their code so that it runs on both v1 and v2.

dschult:
I put a comment in https://github.com/scikit-image/scikit-image/pull/2766/files#r134671549 about subgraph.  I think we don't need an updated subgraph method for scikit-image. We can remove it completely and just let it use the inherited version. It is true that the v1.x subgraph creates a read-write graph while v2 creates a read-all/write-attributes view.  RAG actually relies heavily on the fact that both v1.x and v2 subgraph's attributes ARE the original graph's attribute dicts.  So changes to the subgraph attributes change the original graph.  The RAG class never changes the graph structure of the subgraphs. In my tests, everything works for RAG if we leave ```subgraph()``` as inheriting from the base graph class.

In v1.11 the docs say that people who want to change the subgraph (but don't want to change the original graph) should use ```nx.Graph(G.subgraph(nbunch))``` or ```G.subgraph(nbunch).copy()```.  Both of those work for v2 too. If anyone wants to change the graph structure of subgraphs of RAG outside of the scikit-image code, they will need to use one of these methods.  But the scikit-image code itself should be fine.  (I think it is highly unlikely that someone will want to change the graph structure of subgraphs of RAG. And they would probably already be using one of the two methods above to copy the attributes to the new graph anyway so would not have issues.) 

If someone wants to do: ```H=RAG.subgraph(nodes);H.add_edge(e)```, they will need to change their code (in either v2 or v1.x) to ```H=RAG.subgraph(nodes).copy();H.add_edge(e)```.  We need to keep G.subgraph as a shallow copy for it to work for scikit-image. A view does that.

Finally, I don't think the code from v1.11 that you put in actually works for v2.0. It uses ```self.__class__()``` instead of ```self.fresh_copy()``` to create the new graph structure to fill -- so won't work when ```G``` is a subgraph/reverse/graphview.  Also the code assigns to ```H.node[n]``` when in v2.0 there is no ```H.node```.

Can you try it without that method in RAG and see what you think?

dschult:
> However, I am not sure how to handle G.node in downstream projects, which are assigning to it. For example, take a look at def _rename_node(graph, node_id, copy_id) in
https://github.com/scikit-image/scikit-image/blob/master/skimage/future/graph/graph_merge.py
What is the recommended way to modify this code so that it runs on v1 and v2.0b1?

Line 45 of the linked code assigns to ```graph.node``` when it doesn't need to.  This is an example of why we made views read-only.  To make this work for v1.x and v2.x, the code should replace

    graph.node[copy_id] = graph.node[node_id] 

with

    graph.node[copy_id].update(graph.node[node_id])

Then the attrdict for node ```copy_id``` which was just created in line 44 is used to hold the node attributes.  In this case, they don't want/need the two attribute dicts to be the same object.  If someone really wants to make two nodes have the same attribute dicts they are breaking our data structure and they will have to muck with making their code work for both via their own trickery.

dschult:
The printgraph example should be OK for v1.x and v2.x with a change of ```self.node.items()``` to ```self.nodes(data=True)```.   I'm guessing the example code is older than the ```data=True``` functionality.

dschult:
Q: Can we remove ```G.node``` (as we have done) and still have people create code that works for both v1 and v2?

The iterator parts should be fine:  ```G.node.items()``` can be ```G.nodes(data=True)``` in both v2 and v1.
Getting and setting node attributes are harder.

Old code v1:

    G.node[n]['weight'] = 1.2     # v2.x => no attribute "node"
    G.node[n]['weight'] == 1.2

New code v2:

    G.nodes[n]['weight'] = 1.2     # v1.x => 'instancemethod' object has no attribute '__getitem__'
    G.nodes[n]['weight'] == 1.2

What can we do to create a single code idiom that does this kind of operation in both v2 and v1? I guess the first can be ```G.add_node(n, weight=1.2)```.  There is no "get_node" method in v1.x other than ```G.node[n]```.  You could use: 

    [data for u, data in G.nodes(data=True) if u == n][0]['weight'] == 1.2

but that makes me cringe.

===========================

Notice that this is not an issue with ```G.edge```.  We can always replace ```G.edge``` with ```G.adj``` and it will work in both v1 and v2.  That's because ```G.edge``` was ```G.adj``` in v1.


So, unless we can come up with a way to write node attribute code that works in both v2 and v1, we will need to put ```G.node``` back in.  I think it should be a property that returns the same thing as ```G.nodes``` instead of ```G._node```.  This issue will persist to v2.1 unless people stop using v1.x.  I'm fine with committing to make ```G.node``` remain in the code base until people stop supporting v1.x.

jarrodmillman:
@dschult I think I made all the changes you suggested.  I am still getting test failures.  Here is the way I am testing:
 
```
$ PYTHONPATH=. pytest --doctest-modules  skimage/future/
=============================================================================================== test session starts ================================================================================================
platform linux2 -- Python 2.7.13, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /home/jarrod/src/scikit-image/skimage/future, inifile:
collected 17 items                                                                                                                                                                                                  

skimage/future/manual_segmentation.py ..
skimage/future/graph/graph_cut.py
```

The problem I am seeing is a doctest, so you need to add the ``--doctest-modules`` flag to see the failure.  You should see the above output and it "should" hang at that point.

dschult:
It looks like the test is actually passing.  It is taking FOREVER to run though.

I've run it with v1.11 and v2.0b1 and the ```cut_normalized()``` function has a doctest that takes much longer in v2 than v1.  It does eventually pass though and the sequence of subgraph sizes for all its cuts are the same in both versions.

I hypothesize that it is the recursive subgraph that's killing the speed. I'm running it now and it looks like once we get above 20 recursive subgraphs the speed slows to a crawl.  Every single node that is reported must be checked 20 times with full function lookups and calls.

I recall that the networkx.algorithm.approximation.ramsey code used recursion and subgraphs too, and I ended up not using views there because of slow speed from too many views-of-views-of-views.

So:  
1) I can work on a way for cut_normalized to not use views.
2) We may want to add an optional keyword on subgraph that makes a shallow copy of the graph -- shallow meaning that it keeps the same node/edge data dicts as the original graph, but isn't a nestable view.  

Perhaps there is a way to have the subgraph views detect nesting and bypass it with a single view that involves multiple filters. I'll think about that too.

Any other ideas?

dschult:
A change to the subgraph method that eliminates nested subgraphs made the skimage doctest go from 43:11 wall time to 0:05.8 wall time.  I'll try to put together a pull request tonight.

jarrodmillman:
Thanks!!

jarrodmillman:
@dschult This is ready to merge.  It mainly add this line to ``networkx/classes/graph.py``:
```
    # for backwards compatibility with 1.x, will be removed for 3.x
    node = nodes
```

We haven't discussed a 3.x release, but I thought we will probably drop Python 2.7 support in a few years (it will reach its EOL in 2020).  Perhaps we could release 3.x when that happens.  We don't have to make major changes, but I figure there isn't much of a cost in keeping the above line until some future 3.x release.  I don't think we need to make a big deal about it being deprecated, but we should avoid using it in the code and documentation.  In a few years, we will have several releases of 2.x out and we can assume downstream projects will eventually drop 1.x support.

If that seems reasonable, feel free to merge this.

PR #2635, opened 2017-08-26
Unravel subgraph chains

dschult:
Chains of views are view-of-view-of-view constructs. For subgraphs, if they get long (about 15 views) simple node/edge reporting slows to a crawl.

There could be times when we want the flexibility provided by short chains of subgraphs, but many common cases used in recursive algorithms should avoid chains. 

I have changes the base class method ```subgraph``` to short-cut chains. That is, when the subgraph method of a subgraph view is called, it constructs a subgraph view of the original graph rather than of itself. 

For people who want chains, the function ```nx.induced_subgraph(G, nodes)``` does not short-cut. I also considered ```edge_subgraph``` and ```restricted_view```, but short-cuts are harder to figure out from introspection. In the end I added documentation to describe options. Let the user choose which way they want to go.

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -1211,19 +1211,19 @@ class DiGraph(Graph):
 
         Returns
         -------
-        G : Graph
-            A subgraph of the graph with the same edge attributes.
+        G : SubGraph View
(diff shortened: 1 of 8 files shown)


dschult:
I think this is ready for merging (assuming tests pass).

Issue #2634, opened 2017-08-23
Question re: API change for add_path/star/cycle


JDWarner:
Looking through the 2.0 dev docs, I noticed these methods have been removed from the `Graph` namespace in favor of the base `networkx` namespace.  Trying to understand the justification I've perused the discussion in #1562, #1883, and #1970.  What I see is a light stylistic discussion, not a deep dive into something that was really problematic before to justify it.  I don't understand is why this was done.

If there is a huge push to make the entire package functional, a la NumPy, and there are some Python purists who live by 'one best method' around here I can see why this was pursued.  But there are many other Graph-specific methods, and my feel by using NetworkX for several years has been that the overall style was - appropriately - more OO than functional.   When you are using a graph and you want this functionality, some of it is exposed via methods - indeed, all of these functions are barely more than wrappers around the still-public `add_edges_from` - but these functions are no longer available as Graph methods.  They are, thus, less discoverable to early or novice users.  A partial paradigm shift greatly weakens the stylistic argument.

Also, from the Pythonic ideal of maximally readable code, the new method is inferior:

``` python
nx.add_path(G, [1, 2], kwarg=something)  # Add path from G to 1 and/or 2... oh, that's not it...
G.add_path([1, 2], kwarg=something)  # In graph G, add a path from 1 to 2.
```

This is a hard-breaking API change, with no path forward for supporting dual 1.x and 2.0 branches.  When there is good reasons to do this, I fully understand.  But not only does this cause headaches, but the new API _is less intuitive and more cumbersome_ than having these as graph methods!  In addition to the inferior semantics above, every time these methods are called, they now require an additional argument (the graph) and more characters than before to contribute to repetitive stress injury.

Finally, it's trivial to wrap these and expose the functionality in both places here in the package.  It is not anywhere near so trivial to do this from the user's or maintainer's standpoint.  In such cases, when reasonable it's probably best to think of the user.

__IMHO, these methods should be returned to Graph and wrapped in the main namespace.__

dschult:
Thanks for this feedback. Your careful description and analysis is very helpful.

The self-loop methods also were moved from the graph class to functions in the main namespace. I'm guessing they have smaller impact, but potentially similar arguments. Still, they are reporting methods, while ```add_path```, etc are mutating methods/functions.  It's hard to know where the line should be drawn. Simple interfaces are nice: as simple as possible but not simpler.  :)

PR #2639, opened 2017-08-27
Add deprecated methods for moved base class methods.

dschult:
This eases creating code that works for both v1 and v2.
It doesn't address the concerns of #2634 but it eases
the transition to v2.

What do people think about adding back the moved methods?

### DIFF
--- a/networkx/classes/graph.py
+++ b/networkx/classes/graph.py
@@ -18,6 +18,7 @@ Self-loops are allowed but multiple edges are not (see MultiGraph).
 For directed graphs see DiGraph and MultiDiGraph.
 """
 from __future__ import division
+import warnings
 from copy import deepcopy
 from collections import Mapping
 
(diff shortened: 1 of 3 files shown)


networkx/classes/graph.py
```diff
@@ -714,6 +715,39 @@ def nodes(self):
...
+        msg = "number_of_selfloops is deprecated." \
+              "Use nx.number_of_selfloops instead."
+        warnings.warn(msg, DeprecationWarning)
+        return nx.number_of_selfloops(self)
+
+    def selfloop_edges(self, data=False, keys=False, default=None):
+        msg = "selfloop_edges is deprecated. Use nx.selfloop_edges instead."
+        warnings.warn(msg, DeprecationWarning)
+        return nx.selfloop_edges(self, data=False, keys=False, default=None)
+    # Done with backward compatibility methods for 1.x
```
jarrodmillman: Did you mean to put the comment at the end (and not the beginning) of the block of backward compatible methods?

jarrodmillman: Nevermind.

PR #2642, opened 2017-08-28
Fix some examples for python 3 compatibility

MSeifert04:
Some of the examples threw errors at me, when I tried to build the documentation locally.

There are still two more failures, one related to `mailbox.UnixMailbox` and one regarding a not-pickable function in `advances/plot_parallel_betweenness`.

I also changed the intersphinx links to `https` to avoid these warnings:

```
loading intersphinx inventory from http://docs.scipy.org/doc/numpy/objects.inv...
intersphinx inventory has moved: http://docs.scipy.org/doc/numpy/objects.inv -> https://docs.scipy.org/doc/numpy/objects.inv
loading intersphinx inventory from http://docs.python.org/objects.inv...
intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/2/objects.inv
```

Also is there some documentation on how one should build the documentation? It was a bit of trial&error to get that working and I'm not sure I'm building them as "supposed".

### DIFF
--- a/examples/drawing/plot_random_geometric_graph.py
+++ b/examples/drawing/plot_random_geometric_graph.py
@@ -28,9 +28,9 @@ p = dict(nx.single_source_shortest_path_length(G, ncenter))
 
 plt.figure(figsize=(8, 8))
 nx.draw_networkx_edges(G, pos, nodelist=[ncenter], alpha=0.4)
-nx.draw_networkx_nodes(G, pos, nodelist=p.keys(),
+nx.draw_networkx_nodes(G, pos, nodelist=list(p.keys()),
                        node_size=80,
-                       node_color=p.values(),
+                       node_color=list(p.values()),
                        cmap=plt.cm.Reds_r)
 
 plt.xlim(-0.05, 1.05)
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -199,8 +199,8 @@ latex_documents = [('reference/index', 'networkx_reference.tex',
 latex_appendices = ['tutorial']
 
 # Intersphinx mapping
-intersphinx_mapping = {'http://docs.python.org/': None,
-                       'http://docs.scipy.org/doc/numpy/': None,
+intersphinx_mapping = {'https://docs.python.org/': None,
+                       'https://docs.scipy.org/doc/numpy/': None,
                        }
 
 # The reST default role (used for this markup: `text`) to use for all
--- a/examples/drawing/plot_sampson.py
+++ b/examples/drawing/plot_sampson.py
@@ -19,15 +19,18 @@ Shows how to read data from a zip file and plot multiple frames.
 #    BSD license.
 
 import zipfile
-import cStringIO
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import BytesIO as StringIO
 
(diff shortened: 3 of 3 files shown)


jarrodmillman:
Thanks!

To the build the docs, you should run ``make html`` in the ``doc`` directory.  Depending on what you've installed, you may need to run ``pip install -r requirements.txt`` from the ``doc`` directory first.

I have a few things I want to change about this process, so I will make a PR with those changes.

dschult:
This looks good -- Thanks!

MSeifert04:
> To the build the docs, you should run make html in the doc directory.

Ah, I probably should've said that I'm a windows user (please don't laugh).

examples/drawing/plot_random_geometric_graph.py
```diff
@@ -28,9 +28,9 @@
 
 plt.figure(figsize=(8, 8))
 nx.draw_networkx_edges(G, pos, nodelist=[ncenter], alpha=0.4)
-nx.draw_networkx_nodes(G, pos, nodelist=p.keys(),
+nx.draw_networkx_nodes(G, pos, nodelist=list(p.keys()),
```
dschult: Did you get an error using ```p.keys()``` instead of ```list(p.keys())```?
As far as I can see from the code, an iterator with a length should be sufficient and ```p.keys()``` should provide this. I know the documentation says it should be a list, but perhaps it is better to change the docstring of ```draw_networkx_nodes``` than the code. But if you got an error using ```p.keys()``` then I need to investigate further.

MSeifert04: Yes, it's

```
WARNING: C:\Users\Admin\GIT\networkx\examples\drawing\plot_random_geometric_graph.py failed to execute correctly:Traceback (most recent call last):
  File "C:\Programming\Anaconda\envs\networkxsource\lib\site-packages\sphinx_gallery\gen_rst.py", line 450, in execute_code_block
    exec(code_block, example_globals)
  File "<string>", line 27, in <module>
  File "C:\Programming\Anaconda\envs\networkxsource\lib\site-packages\networkx-2.0rc1.dev20170828015156-py3.5.egg\networkx\drawing\nx_pylab.py", line 399, in draw_networkx_nodes
    label=label)
  File "C:\Programming\Anaconda\envs\networkxsource\lib\site-packages\matplotlib\__init__.py", line 1897, in inner
    return func(ax, *args, **kwargs)
  File "C:\Programming\Anaconda\envs\networkxsource\lib\site-packages\matplotlib\axes\_axes.py", line 3982, in scatter
    c_array = np.asanyarray(c, dtype=float)
  File "C:\Programming\Anaconda\envs\networkxsource\lib\site-packages\numpy\core\numeric.py", line 583, in asanyarray
    return array(a, dtype, copy=False, order=order, subok=True)
TypeError: float() argument must be a string or a number, not 'dict_values'
```

The problem is that a `KeysView` can't be used as iterable for `np.array`:

```
>>> np.array({1:3, 2:4}.keys())
array(dict_keys([1, 2]), dtype=object)
```

dschult: I understand now --  Thank you.

PR #2647, opened 2017-09-03
Refactor travis tests and deploy docs with travis

### LINKED ISSUE #2646: Readthedocs pain
Readthedocs (RTD) is a pain to work with and keeps having timeout errors.  I started to look into whether we can build the docs on our own and push them to RTD instead of having it built on the site.  It would also make more sense to have the doc build process as part of our CI process, rather than only checked after the fact.

Has there been any discussion about moving away from RTD before (at least the build process)?  If so, was there a reason not to move?  I assume it is too late to move back to hosting the docs on github, but I thought I'd check since it might be easier to do.

jarrodmillman:
See
- https://travis-ci.org/networkx/networkx/builds/271948102
- https://networkx.github.io/dev-docs/

This PR accomplishes two main goals:

1. Travis-ci now tests that our documentation builds without error.
2. Travis-ci can now push the built docs to GitHub pages when a
   build is triggered on the master branch.

I also changed the travis-ci tests so that we only install packages
when necessary and only generate a coverage report when it will
be uploaded via codecov.
This substantial speeds up the tests so that we are able to test more
things now (e.g., whether the docs build).
Generating the coverage reports, in particular, was costly (i.e., it roughly
doubled or tripled the time it takes the test suite to run).
I no longer test extra dependencies on 3.4 and 3.5 (since I suspect they
aren't often as useful and I suspect that most folks on Python 3 are using
3.6 anyway), but I can easily add them back
(though they will slightly increase the testing time).

To deploy the built docs on GitHub pages, I generated an ed25519
deploy key using ``ssh-keygen``.
I added the ``deploy-key.pub`` contents to the ``networkx/dev-docs`` repo’s
settings under ``Settings -> Deploy Keys``.
I encrypted the private deploy key with the travis-ci symmetric key
using ``travis encrypt-file deploy-key``.
You can see how I use it in ``tools/travis/deploy_docs.sh``.

I haven't been able to directly test this without triggering a
build on the master branch (however, I could temporarily have
it deploy from the PR by changing the conditional at the top
of ``tools/travis/deploy_docs.sh``).
If we merge this, we should leave ReadTheDocs working
until we are certain this pushes the docs to GitHub pages
and that we like the results.
(We will need to update a few links on
http://networkx.github.io/documentation/ and
https://networkx.github.io/ as well.)
I tested it by hand and you can see the results here:
  https://networkx.github.io/dev-docs/
Once we verify this PR works as I intended, we can shut down the ReadTheDocs site
and clean up a few things I left behind (e.g., ``doc/requirements.txt``).

Initially I looked into pushing to the ``gh-pages`` branch of this repo:
  http://networkx.github.io/documentation/
But that repo is huge and takes a long time to clone.
We can add the stable 2.0 docs (and any other missing builds) there.
Alternatively, we could create a new repo (e.g., networkx/docs)
for 1.11 and future docs.
This would leave the networkx/documentation repo as it is.

(There were other options, I looked into and that we could still use.
Travis-ci has builtin  support: https://docs.travis-ci.com/user/deployment/pages/
However, they use a personal ``GITHUB_TOKEN``, which has
access to all of the repositories that that person has access to
(and it isn't as flexible as the approach I used). 
If we want to use that method, I would recommend creating a new
machine user account on GitHub that only has access to one repo.
However, that seems like overkill to me.
There is also this project (developed by the sympy folks):
https://drdoctr.github.io/doctr/
That seems like a big project for the little bit of work
necessary to accomplish what we want and it isn't as flexible
as the current approach.)

This isn't exactly what I did, but this blog post describes
something similar to what I did:
https://djw8605.github.io/2017/02/08/deploying-docs-on-github-with-travisci/
If this PR seems reasonable, I will write up something similar
except with the steps I followed.
(That could be useful as documentation for us and might be
helpful for other projects.)
Setting up the keys is a one-time deal, which we won't need to
bother with again.
The key pair I generated will only allow travis-ci (or someone
with the travis-ci key) to push to ``networkx/dev-docs``.
We don't need to keep a copy of the private key, so I could
delete it once we verify that this works.

If we go with this method, then we can control the documentation
build by modifying ``tools/travis/deploy_docs.sh``.
It is a pretty straightforward bash script and it should
be obvious how you could do whatever you want
(e.g., we could use the various text editing command-line
tools on Linux to correct or modify anything we want,
we could grab additional files from anywhere on the web
and copy them into any directory we want, etc.).

I've spent some time reorganizing the commits into somewhat
reasonable chunks of work, so you may want to review the
changes commit by commit.
However, I suspect looking at the full changes will be
reasonable as well.

This is a fairly big change to our doc system and one
we probably don't want to immediately change again.
So I would be happy to get a lot of tough feedback.

Thoughts?

Fix #2646.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ doc/networkx-documentation.zip
 doc/networkx_reference.pdf
 doc/networkx_tutorial.pdf
 doc/build
+doc/doc_build
 .coverage
 *.class
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,42 +5,70 @@ sudo: false
 
 language: python
 
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-
 cache:
   directories:
     - $HOME/.cache/pip
 
-addons:
-  apt:
-    packages:
-    - libgdal-dev
-    - graphviz
-
-env:
-  matrix:
-    - OPTIONAL_DEPS=pip
-    - OPTIONAL_DEPS=no
-
 matrix:
   include:
+    - os: linux
+      python: 2.7
+      env:
+      - OPTIONAL_DEPS=1
+      - MINIMUM_REQUIREMENTS=1
+      - REPORT_COVERAGE=1
+      addons:
+        apt:
+          packages:
+          - libgdal-dev
+          - graphviz
+    - os: linux
+      python: 2.7
+      env:
+      - OPTIONAL_DEPS=1
+      addons:
+        apt:
+          packages:
+          - libgdal-dev
+          - graphviz
+    - os: linux
+      python: 3.6
+      env: OPTIONAL_DEPS=1
+      addons:
+        apt:
+          packages:
+          - libgdal-dev
+          - graphviz
+    - os: linux
+      python: 3.6
+      env:
+      - OPTIONAL_DEPS=1
+      - MINIMUM_REQUIREMENTS=1
+      addons:
+        apt:
+          packages:
+          - libgdal-dev
+          - graphviz
+    - os: osx
+      language: generic
+      env:
+      - TRAVIS_PYTHON_VERSION=3.6.0
+      - OPTIONAL_DEPS=1
+      - OSX_PKG_ENV=miniconda
     - os: osx
       language: generic
       env: TRAVIS_PYTHON_VERSION=3.6
-    - os: osx
-      language: generic
-      env: TRAVIS_PYTHON_VERSION=3.6.0 OPTIONAL_DEPS=pip OSX_PKG_ENV=miniconda
+    - python: 2.7
+    - python: 3.4
+    - python: 3.5
+    - python: 3.6
 
 before_install:
   # prepare the system to install prerequisites or dependencies
   - source tools/travis/before_install.sh
   - uname -a
   - printenv
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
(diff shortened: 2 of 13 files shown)


jarrodmillman:
@dschult , @hagberg Please take a look and let me know what you think.

dschult:
This looks quite promising. I think we should go ahead and merge it and try it out.

If I understand correctly you are suggesting that we manually upload the tagged repository docs as part of the release process. This updates the dev version of the docs.

When I first looked at this I got lost in trying to figure out how it differs from e.g. SciPy. They have a documentation repository too with automated travis-ci push to github. I could understand how this proposal works more easily than I could understand how their's works. 

I'm fine with merging.

chebee7i:
It seems like you did some tweaking, in part, to get under the allowed build time that TravisCI allows. How close are we to that? Is this going to be an issue that comes back to bite?

One cautionary note is that we were originally under time with RTD, then they changed their policy and we were over their new limit.

dschult:
I have redirected the landing page of all versions from ReadTheDocs to the NetworkX landing page.
I was not able to redirect each doc page to its corresponding docpage on github because RTD doesn't have ```.htaccess``` available. 

So:   people looking for the docs of a specific function will get a 404 error.

I did some test searches and for whatever reason it seems like Google wants to show the networkx-1.10 docs on the github site ahead of the ReadTheDocs stable docs. I've seen this before too.  Hopefully the bots will be able to find the latest and stable versions with the new config.

PR #2657, opened 2017-09-09
Fixing typo

gvkalra:
Section: Convenient access to all edges


jarrodmillman:
This is already fixed on master, which will be released as 2.0 soon.  We will not be making another 1.x release.

PR #2660, opened 2017-09-10
Use isinstance in is_string_like

jarrodmillman:
See https://github.com/matplotlib/matplotlib/pull/8011


dschult:
I understand the change matplotlib is making toward ```isinstance``` over duck-typing (the duck-typing doesn't work with e.g. ```open```). But instead of spreading lots of basestring cruft throughout our codebase, could we continue to use ```is_string_like``` with the basestring cruft built into it. That would contain the py2/py3 checking to ```utils/misc.py```.  [eventually py2 checking will need to be deprecated too -- can we contain the impact?]

PR #2672, opened 2017-09-17
rm arg `strict` from function `networkx.drawing.nx_pydot.to_pydot`

johnyf:
This keyword argument is unused, because the value is overwritten in the function's body.

### DIFF
--- a/networkx/drawing/nx_pydot.py
+++ b/networkx/drawing/nx_pydot.py
@@ -179,7 +179,7 @@ def from_pydot(P):
     return N
 
 
-def to_pydot(N, strict=True):
+def to_pydot(N):
     """Return a pydot graph from a NetworkX graph N.
 
     Parameters


johnyf:
The argument `strict` appears to have been introduced when the function `to_pydot` was introduced, in https://github.com/networkx/networkx/commit/18e6bcf0a6d66e71709177ccc2dca7ef16402456#diff-d3baa32798b7fb11c991d5c546cb4b30R140. At that point `strict` was used.

The argument `strict` appears to have stopped being used by https://github.com/networkx/networkx/commit/60329692839c4455e3a62a372e8fdc99f26a40d4#diff-d3baa32798b7fb11c991d5c546cb4b30R185, which assigns to it unconditionally, thus overwriting the argument's value. That this change occurred at some point suggests that it may have been a better design choice than passing `strict` as an argument. The [documentation of GraphViz](http://www.graphviz.org/doc/info/lang.html) quoted below gives more reasons:

     A graph may also be described as strict. This forbids the creation of multi-edges, i.e.,
    there can be at most one edge with a given tail node and head node in the directed case.
    For undirected graphs, there can be at most one edge connected to the same two nodes.
    Subsequent edge statements using the same two nodes will identify the edge with the
    previously defined one and apply any attributes given in the edge statement.
    For example, the graph
    
    strict graph { 
      a -- b
      a -- b
      b -- a [color=blue]
    } 

    will have a single edge connecting nodes a and b, whose color is blue.

Based on this behavior, it seems to me that:

1. passing `strict=False` for a graph that has no self-loops will have no observable effect, and

2. passing `strict=True` for a graph that does have self-loops will cause confusing behavior, because attributes of different multi-edges with same source and target nodes will overwrite each other in an arbitrary order that depends on the order in which edges get dumped to `dot` (at least this is my understanding from reading the documentation above).

So it seems that defining `strict` from the graph itself would lead to the least surprising behavior. Somewhat relevant also to: https://github.com/erocarrera/pydot/issues/15

Regarding API uniformity, `strict` is assigned within the function [`networkx.drawing.nx_agraph.to_agraph`](https://github.com/networkx/networkx/blob/0d93bd37e3931acbd4f3921ca8debcf634b0a341/networkx/drawing/nx_agraph.py#L142), but is not exposed as an argument.

dschult:
Thanks very much -- this is helpful!

Issue #2381, opened 2017-03-08
Calling nx.fiedler_vector() on certain graphs hangs indefinitely


hinnefe2:
In certain cases making a call to `nx.fiedler_vector` will hang indefinitely. 

Minimum working example (for me this will hang indefinitely)
```
import networkx as nx
G = nx.Graph()
G.add_path([1,2,3,4])
G.add_edges_from([(5,3), (6,2)])
nx.fiedler_vector(G)
```

Some observations:
1. Calling the function with `method='lanczos'` works fine, it's only the default behavior that sometimes hangs.

2. A cursory look through the code / stack trace when interrupting suggests that the `while True:` loop at line 73 in `networkx/linalg/algebraicconnectivity.py` might be the culprit.

hagberg:
Seems like a bug somewhere in the vicinity of where you suggest.   Another report which could be the same issue http://stackoverflow.com/questions/43382204/networkx-cant-calculate-algebraic-connectivity/43383130#43383130

tabac:
I have the same issue. 

In both my case and the example above, the `while` loop hangs because `beta` becomes `inf` [here](https://github.com/networkx/networkx/blob/master/networkx/linalg/algebraicconnectivity.py#L81) and as a consequence _"everything else"_ becomes `nan` which results in the failure of the exit condition [here](https://github.com/networkx/networkx/blob/master/networkx/linalg/algebraicconnectivity.py#L78).

I am afraid I am not very familiar with the Preconditioned Conjugate Gradient Method. However, I have to say that something like:

```python                                  
from math import isinf
...
    t = ddot(r, z)
    if isinf(t):
        return x
    beta = t
...
```

looks reasonable to me.

dschult:
Line 270 in ```_find_fiedler_function``` needs to add postprocessing of the regular expression:

    if method is None:
        method = 'pcg'

When method is None (the default), it causes ```q``` to be treated as if chol/lu methods were being used instead of pcg. But in the tracemin function 'pcg' is used. So the method is a cross between the two. The major impact I see is that 'pcg' is always set to find the "largest" two eigenvectors -- fiedler vector being the second one and the first being a constant vector. This makes 'pcg' look for more than two vectors.  The 'chol' and 'lu' methods find more than two eigenvectors. 

[...]

PR #2681, opened 2017-09-24
Fix bug and add checks for non-convergent fiedler_vector

dschult:
Addresses #2381 

There was a bug (default method didn't set method to 'pcg') that made the algorithm not converge consistently for any network which required more than one iteration to converge.  Even with the bug fixed, the algorithm doesn't always converge. The initial conditions are random, so restarting could work. I put in a check for non-convergence that simply raises a NetworkXError. The user would have to check this to restart with new random conditions.

It would be nice if we could make one of the other methods the default, but they have optional dependencies. Perhaps we would check if those optional dependencies are present and if so switch the default.

To be clear: the non-convergence is not a fault of the algorithm. That is apparently guaranteed to converge. But up to roundoff error the matrix W becomes singular -- specifically the second column becomes entirely zeros up to round-off.

Suggestions?

### DIFF
--- a/networkx/linalg/algebraicconnectivity.py
+++ b/networkx/linalg/algebraicconnectivity.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
-"""
-Algebraic connectivity and Fiedler vectors of undirected graphs.
-"""
-
-__author__ = """ysitu <ysitu@users.noreply.github.com>"""
 # Copyright (C) 2014 ysitu <ysitu@users.noreply.github.com>
 # All rights reserved.
 # BSD license.
+#
+# Author: ysitu <ysitu@users.noreply.github.com>
+"""
+Algebraic connectivity and Fiedler vectors of undirected graphs.
+"""
 
 from functools import partial
 import networkx as nx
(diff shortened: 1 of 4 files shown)


dschult:
The Fiedler vector pcg method in rare (but sometimes simple #2381) cases
projected away too much energy from b in Ax=b so that x was within
roundoff error and the solver never finished (hung). Despite my efforts
to restore the energy, add noise in orthogonal directions and project
either more or less often, the system would hang on unlucky initial
conditions. (Though I did get it to only hang 5 times out of 1 million
seeds.)

In the end I took a hint from a comment line that the PCG method was
using a different but mathematically equivalent method to that of
Cholesky and LU. Rewriting the PCG method to use the same method
as the Chol and LU methods did the trick. Many fewer projections might 
also save time, certainly makes for simpler code, and the
vectors are not close to round-off size. Trials with the simple example graph
and initial conditions from seeds 1:1000000 show no signs of hanging.

I believe we can now remove the restriction in the tests that the seed
be the same each time we test.  Is there a reason to keep the seeds
the same other than avoiding an infinite loop?

dschult:
Preserving seed in tests for alg_conn has been removed. No longer necessary.
This is ready to merge (once tests pass).

PR #2682, opened 2017-09-26
[WIP] Spectral Clustering in networkx

### LINKED ISSUE #2401: Spectral Clustering
Would this repo invite PRs with spectral clustering techniques? Is there a reason there aren't any yet?

alexlenail:
Closes #2401

Work in progress.

### DIFF
--- a/examples/algorithms/rcm.py
+++ b/examples/algorithms/rcm.py
@@ -14,7 +14,10 @@ reduces the matrix bandwidth.
 # BSD License
 import networkx as nx
 from networkx.utils import reverse_cuthill_mckee_ordering
-import numpy as np
+try:
+    import numpy as np
+except ImportError:
+    raise ImportError('rcm requires NumPy ', 'http://scipy.org/')
 
 # build low-bandwidth numpy matrix
 G = nx.grid_2d_graph(3, 3)
--- a/examples/algorithms/rcm.py
+++ b/examples/algorithms/rcm.py
@@ -14,10 +14,7 @@ reduces the matrix bandwidth.
 # BSD License
 import networkx as nx
 from networkx.utils import reverse_cuthill_mckee_ordering
-try:
-    import numpy as np
-except ImportError:
-    raise ImportError('rcm requires NumPy ', 'http://scipy.org/')
+import numpy as np
 
 # build low-bandwidth numpy matrix
 G = nx.grid_2d_graph(3, 3)
--- a/networkx/linalg/spectrum.py
+++ b/networkx/linalg/spectrum.py
@@ -113,11 +113,20 @@ def spectral_clustering(G, k):
     G : Graph
        A NetworkX Graph or DiGraph
 
+    k : int
+       The number of desired output clusters.
+
     Returns
     -------
     clustering : NumPy array
       The spectral clustering
 
+    Notes
+    -----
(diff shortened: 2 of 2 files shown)


zfrenchee:
@dschult please advise on whether this is going in the right direction =)

examples/algorithms/rcm.py
```diff
@@ -14,7 +14,10 @@
 # BSD License
 import networkx as nx
 from networkx.utils import reverse_cuthill_mckee_ordering
-import numpy as np
+try:
```
jarrodmillman: For the examples, I would just import things and let them fail.   I don't feel strongly about this, but it seems like the examples should be what people normally would write in a script and I don't think the should normally try to import.

jarrodmillman: The modules in ``networkx`` (as opposed to ``examples``) should have numpy imports in a try statement, though.

zfrenchee: Done.

dschult:
Thanks for the PR. You've got a reasonable start. 
I agree with @jarrodmillman that examples should just import and let the viewer install before trying. 
In the codebase we protect the imports.

My concern with how it looks so far is that it might just be wrappers around code from sklearn. If that's the case then I'm not sure it makes sense to include it in the package as functions. Maybe examples would be sufficient? Are there other functions you see adding to this?

smopucilowski:
I don't think introducing a new dependency on scikit just for spectral clustering is wise, especially when calculating eigenvalues and eigenvectors is part of numpy; see http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.165.9323

zfrenchee:
@dschult @smopucilowski I suppose it depends on what role we think networkx should serve. I've always assumed networkx intends to be a "swiss army knife" for working with graphs in python. If that's correct, then spectral clustering is one of the fundamental tools I would want in that swiss army knife. I would want spectral clustering implemented in networkx such that I wouldn't need to  go and understand the sklearn API and figure out de novo how to plug networkx and sklearn together. Figuring out those two lines of code in this PR took a couple hours, which I don't want to save the next person who comes along. 

As for why to use sklearn, I think having fewer implementations of spectral clustering floating around the python data science ecosystem is better.

dschult:
@zfrenchee I am fine with a swiss-army knife approach. 

It is important to save people from having to spend hours figuring out how to interface two APIs.
Let's work to do that. One way is to create new functions that create a new interface to sklearn but hide the algorithmic ideas. Another way would be to create an example that shows people how to connect the API of NetworkX to the API of sklearn. Which is more effective?

In some sense it doesn't matter. Either way, we need a paragraph describing what spectral clustering is. And we need a parameter definition for k, perhaps along with a quick description of how to choose k. Then we need a simple graph G to demonstrate the input and verify the output. (This last sentence would be a docstring test if this is a function, or the heart of the example if this is an example.)

Why don't we keep developing this with a function interface and see if we can build an effective interface that bridges the two.

zfrenchee:
@dschult those are good points. I'd like to see this method included, but at this point, I don't have terribly strong feelings about it, since I've updated the sklearn docs slightly with an example, and my stackoverflow question seems to be getting enough attention.  Happy to proceed as you see fit.

Issue #2661, opened 2017-09-11
Probably wrong implementation of nx.effective_size() for directed graph


kyukyukyu:
https://github.com/networkx/networkx/blob/0ab38ce6c46ce678bda23c1d26494683ef93e40f/networkx/algorithms/structuralholes.py#L157-L163

As you see, a node is tested if it is isolated by the number of **outgoing** edges (`G[v]`) while the effective size of a node is computed by summing up the redundancy of the node and all the neighbor nodes, which are **both ingoing and outgoing** edges. (`nx.all_neighbors(G, v)`) I think these targets should be matched.

dschult:
This makes sense to me. It looks like the input-check for isolated nodes was written without directed edges being considered. Would you like to submit your fix as a pull request?

Thanks!

kyukyukyu:
Okay. Since the current unit test for this module does not cover the case where there are neighbors for a node but no outgoing edges, I think a new test case should be added. I'll try to do so within a week.

dschult:
Any progress on a test case here?

kyukyukyu:
Unfortunately, I couldn't take time to create new test cases for this change. May I open a PR with [WIP] prefix first?

dschult:
Yes -- that is often done.  Thanks

PR #2695, opened 2017-10-03
[WIP] Fix nx.effective_size() for nodes w/o outgoing ties

kyukyukyu:
This PR is to fix the bug discussed in #2661. However, new test cases should be added that test against a directed graph which contains some nodes with no outgoing ties.


hagberg:
Needs test.

Issue #2503, opened 2017-07-08
bug when creating a plot with a legend if edges have arrowheads


joelmiller:
Currently, it appears to me that when you plot an edge in a directed graph with an arrowhead and assign a label, the legend in the plot will show both the edge and the arrowhead with whatever label.  

Looking at the source code, it appears that the error is in the command `arrow_collection.set_label(label)`.  I don't think this serves any purpose other than to ensure that the arrowhead is labelled but the edge itself is already labelled from the earlier command `edge_collection.set_label(label)`

See https://stackoverflow.com/q/44966874/2966723 for an example of this.

viraj96:
Hi, I am willing to submit a PR if this is still open and no one is assigned to it. All there needs to be changed is removing `arrow_collection.set_label(label)`?

dschult:
Go for it!   Yes -- I think that's it -- try it to make sure of course...

dschult:
Fixed in #2503

PR #2696, opened 2017-10-03
Fixes bug #2503 by removing arrow labels

viraj96:
Tested it in my virtual environment. 
Fixes bug #2503 .


dschult:
Thanks very much!!

PR #2703, opened 2017-10-08
Fix dag_longest_path bug

yelly:
Fix a bug in dag_longest_path that raises an exception when an empty
graph is given.

### DIFF
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -489,6 +489,8 @@ def dag_longest_path(G, weight='weight', default_weight=1):
     --------
     dag_longest_path_length
     """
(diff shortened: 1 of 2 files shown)


networkx/algorithms/dag.py
```diff
@@ -489,6 +489,8 @@ def dag_longest_path(G, weight='weight', default_weight=1):
     --------
     dag_longest_path_length
     """
+    if not G:
```
jarrodmillman: It would be great if you added a check to make sure that this works as intended (and that way if we accidentally change it later, we will know).

yelly: No problem, done.

PR #2706, opened 2017-10-10
Adding rich-core function and test 

iaciac:
Extraction of the rich-core of complex networks ad in Ma A and Mondragón RJ (PLoS One, 2015).

### DIFF
--- a/networkx/algorithms/rich_core.py
+++ b/networkx/algorithms/rich_core.py
@@ -10,6 +10,7 @@ as defined by Ma A and Mondragón RJ (2015) in Rich-cores in networks. PLoS One 
 
 
 import networkx as nx
+from collections import OrderedDict
 from networkx.utils import not_implemented_for
 from __future__ import division
 
@@ -47,12 +48,18 @@ def extract_rich_core(G, weight='weight'):
     Returns
     -------
     sigmas: dictionary
-        Dictionary of nodes with the $\sigma$ as the value,
-        representing the strength of the node after rescaling
-        the weights in units of the minimal weight.
+        Ordered dictionary (collection.OrderedDict) of nodes with the
+        $\sigma$ as the value, representing the strength of the node after
+        rescaling the weights in units of the minimal weight.
         
     node_max_sigma: node
         The core boundary node.
+        
+    Note
+    ----
+    Since the sigmas dictionary is ordered, the values can be directly
+    plotted as plt.plot(sigmas.values()) to obtain the plot of Figure 3
+    of [1]_.
       
     References
     ----------
@@ -68,13 +75,13 @@ def extract_rich_core(G, weight='weight'):
     
     #Normalising by the minimal weight (adding a new attribute)
     for u, v, wt in G.edges.data(weight):
-        G[u][v][norm_weight]=wt/minw
+        G[u][v]['norm_weight']=wt/minw
     
     #Ranking the nodes in units of the minimal weight (by normalised strength)
-    norm_strength = G.degree(weight=norm_weight)
-    ranked_nodes = sorted(norm_strength, key=norm_strength.get, reverse=True)
+    norm_strength = G.degree(weight='norm_weight')
+    ranked_nodes = [n[0] for n in sorted(norm_strength, key=lambda x: x[1], reverse=True)]
     
-    sigmas={}
+    sigmas=OrderedDict()
     for i in ranked_nodes:
         sigma_i=0
         for j in G.neighbors(i):
--- a/networkx/algorithms/rich_core.py
+++ b/networkx/algorithms/rich_core.py
@@ -10,8 +10,8 @@ as defined by Ma A and Mondragón RJ (2015) in Rich-cores in networks. PLoS One 
 
 
 import networkx as nx
-from networkx.utils import accumulate
 from networkx.utils import not_implemented_for
+from __future__ import division
 
 __all__ = ['extract_rich_core']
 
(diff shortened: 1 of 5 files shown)


dschult:
Looks like you need to add an import of richcore into the ```__init__.py``` file in the algorithms directory.
Also, then your test can simply call ```nx.extract_rich_core```

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,82 @@
...
+        "Rich-cores in networks".
+        PLoS One 10(3):e0119678.
+        
+        """
+    
+    
+    G = G.to_undirected()
+    
+    #Looking for the minimal weight
+    weights = [e[2][weight] for e in G.edges_iter(data=True)]
```
dschult: ```G.edges_iter``` won't work with NetworkX V2.0.   
Also, you can avoid creating a list of all weights.
Use:   ```minw = min(wt for u, v, wt in G.edges.data(weight))```

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,82 @@
...
+    minw = min(weights)
+    
+    #Normalising by the minimal weight
+    for e in G.edges_iter(data=True):
```
dschult: Perhaps:     ```for u, v, wt in G.edges.data(weight):```

Also, you are mutating the edge weight attribute in this function. A warning should be added to the doc_string or avoided by creating another attribute to store those values.  Or even better -- just divide the ```sigma_i``` by ```minw``` when you store it.  Then the original graph doesn't change and I think your calculations and sorting is the same.  You also avoide this loop that way.

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,82 @@
...
+    for i in ranked_nodes:
+        sigma_i=0
+        for j in G.neighbors(i):
+            if strength[j]>strength[i]:
+                sigma_i += strength[j]
+            else: continue
+        sigmas.append(sigma_i)
+    
+    r_star = sigmas.index(max(sigmas))
+
```
dschult: You should consider a different return data structure.
Of course, it may depend what you expect people to do with the results.
 
A dict from nodes to sigma values is our usual way to return node properties.
This loses the order of rankings as well as ```r_star```. 
Perhaps ```r_star``` could be replaced by the node with maximal sigma -- since at this point it is a index of that node.  Or you could just give the user the dict of sigma values and let the user find the node with max sigma value.

networkx/algorithms/tests/test_richcore.py
```diff
@@ -0,0 +1,16 @@
...
+    G=nx.Graph()
+    G.add_edge('a','b',weight=0.6)
+    G.add_edge('a','c',weight=0.2)
+    G.add_edge('c','d',weight=0.1)
+    G.add_edge('c','e',weight=0.7)
+    G.add_edge('c','f',weight=0.9)
+    G.add_edge('a','d',weight=0.3)
+
+    sigmas, ranked_nodes, r_star = nx.richcore.extract_rich_core(G, weight='weight')
+    assert_equal(sigmas, [0, 19.0, 19.0, 19.0, 10.999999999999998, 30.0])
```
dschult: tests with floating point numbers are tricky to get exactly equal. 
You might consider ```from nose.utils import assert_almost_equal``` for the tests.

There is also a nice way to make sure division works for floating point for python2 and 3. Use this at the top of the rich_core.py module:  ```from __future__ import division```     Then you don't need to use ```1.*``` when you divide.

dschult:
I think the sigma values you return are actually the sigma^+ values, not the sigmas.
Which should be returned?

It also looks like the normalization of weights is not needed (or can be done after all computations).
Calculating strength, rank and boundary node can be done without scaling edge weights and the results for rank and boundary node are the same. Scaling strength by minw can be done after computations if desired, but is not necessary.  Am I understanding this correctly?

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+    plotted as plt.plot(sigmas.values()) to obtain the plot of Figure 3
+    of [1]_.
+      
+    References
+    ----------
+    .. [1] Ma A and Mondragón RJ (2015),
+       "Rich-cores in networks".
+       PLoS One 10(3):e0119678.
+    """
+    if nx.is_directed(G):
```
jfinkels: This is not needed, since there is decorator that checks it.

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+        raise Exception('rich_core is not implemented for directed graphs.')    
+   
+    #Looking for the minimal weight
+    minw = min(wt for u, v, wt in G.edges.data(weight))
+    
+    #Normalising by the minimal weight (adding a new attribute)
+    for u, v, wt in G.edges.data(weight):
+        G[u][v]['norm_weight']=wt/minw
```
jfinkels: Let's add a note in the documentation that this function modifies the graph.

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+#
+# Authors: Iacopo Iacopini (i.iacopini@qmul.ac.uk)
+"""Function for extracting the rich-core of a network,
+as defined by Ma A and Mondragón RJ (2015) in Rich-cores in networks. PLoS One 10(3):e0119678."""
+
+
+import networkx as nx
+from collections import OrderedDict
+from networkx.utils import not_implemented_for
+from __future__ import division
```
jfinkels: Must be before any other imports.

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+    
+    The *rich-core* of a network comes from the theoretical coupling the
+    underlying principle of a *rich-club* with the escape time of a
+    random walker. A $\sigma_i$ is associated to each node $i$ of the
+    network, representing the strength of the node after rescaling the
+    weights in units of the minimal weight:
+    
+    .. math::
+    
+        \sigma_i = \sum_j \frac{w_ij}{w_{min}}
```
jfinkels: I think it should be `w_{ij}`.

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+from networkx.utils import not_implemented_for
+from __future__ import division
+
+__all__ = ['extract_rich_core']
+
+
+@not_implemented_for('directed')
+@not_implemented_for('multigraph')
+def extract_rich_core(G, weight='weight'):
+    r"""Returns the core/periphery structure of a the weighted undirected
```
jfinkels: The first sentence should fit in one line.

jfinkels: typo: `of a the`

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+    norm_strength = G.degree(weight='norm_weight')
+    ranked_nodes = [n[0] for n in sorted(norm_strength, key=lambda x: x[1], reverse=True)]
+    
+    sigmas=OrderedDict()
+    for i in ranked_nodes:
+        sigma_i=0
+        for j in G.neighbors(i):
+            if norm_strength[j]>norm_strength[i]:
+                sigma_i += norm_strength[j]
+            else: continue
```
jfinkels: This line is unnecessary.

networkx/algorithms/rich_core.py
```diff
@@ -0,0 +1,96 @@
...
+    
+    #Normalising by the minimal weight (adding a new attribute)
+    for u, v, wt in G.edges.data(weight):
+        G[u][v]['norm_weight']=wt/minw
+    
+    #Ranking the nodes in units of the minimal weight (by normalised strength)
+    norm_strength = G.degree(weight='norm_weight')
+    ranked_nodes = [n[0] for n in sorted(norm_strength, key=lambda x: x[1], reverse=True)]
+    
+    sigmas=OrderedDict()
```
jfinkels: In this line and others there is missing whitespace around the operator.

dschult:
Needs tests and updates to reflect reviews above.

MridulS:
Thanks for this @iaciac!

This PR hasn't been touched in a while and will need some effort to make it work with the `main` branch, it's py3.10+ and a bunch of things have changed in networkx over the last 5 years. I'll go ahead and close this PR, feel free to open a new one if you have the bandwidth. Thanks again!

Issue #2427, opened 2017-04-18
shortest_simple_paths() missing some valid paths


nasribo:
I'm using the shortest_simple_paths function on a DiGraph to calculate the top-k paths for my network. I found a few edge cases where the results are missing some valid paths.

I am using networkx 1.10 for Python 2.7.

Please see my simple example below.

```python
#In this example, I want to find all shortest simple paths
#from IN to OUT. A is an intermediate node, B is an extra irrelevant node.
#
#Graph looks like this:
#
#[B]<--->[IN]----------->[OUT]
#         |                ^
#         |                |
#         +----->[A]-------+
#
#There are two feasible paths:
#  - First path: ['IN', 'A', 'OUT'] with cost 30
#  - Second path: ['IN', 'OUT'] with cost 60
#For some reason, second path isn't detected unless B is removed.

#Run the code below.
import networkx as nx

#Create our full graph.
G = nx.DiGraph()
G.add_edge('IN', 'A', {"weight":10})
G.add_edge('A', 'OUT', {"weight":20})
G.add_edge('IN', 'OUT', {"weight":60})
G.add_edge('IN', 'B', {"weight":40})
G.add_edge('B', 'IN', {"weight":50})

g = nx.shortest_simple_paths(G, 'IN', 'OUT', "weight")
print [p for p in g]
#Expected: [['IN', 'A', 'OUT'], ['IN', 'OUT']]
#Observed (BUG): [['IN', 'A', 'OUT']]

#Try again without B.
G = nx.DiGraph()
G.add_edge('IN', 'A', {"weight":10})
G.add_edge('A', 'OUT', {"weight":20})
G.add_edge('IN', 'OUT', {"weight":60})
g = nx.shortest_simple_paths(G, 'IN', 'OUT', "weight")
print [p for p in g]
#Expected: [['IN', 'A', 'OUT'], ['IN', 'OUT']]
#Observed (OK): [['IN', 'A', 'OUT'], ['IN', 'OUT']]
```

I made a small change to the function to fix it (for my network at least). I don't know why it works, and I don't know if it will break anything for other networks. Please see possible bug fix below.

```python
#Attempted bug fix. Please check. I don't know why this works and if it breaks anything.
#Note that my graph has no reflexive edges (v->v).
while True:
	if not prev_path:
		length, path = shortest_path_func(G, source, target, weight=weight)
		listB.push(length, path)
	else:
		ignore_nodes = set()
		ignore_edges = set()
[...]

PR #2712, opened 2017-10-14
Fix for issue #2427

md0000:
Bug was in deviation finding part of algorithm.  Instead of ignoring first (i - 1) vertices in path, algorithm ignored i-th vertex in path.

### DIFF
--- a/networkx/algorithms/simple_paths.py
+++ b/networkx/algorithms/simple_paths.py
@@ -333,7 +333,6 @@ def shortest_simple_paths(G, source, target, weight=None):
                 for path in listA:
                     if path[:i] == root:
                         ignore_edges.add((path[i - 1], path[i]))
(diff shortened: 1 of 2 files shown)


dschult:
It looks like all those vertices are already added to ```ignored_nodes``` in previous iterations (except the last). Perhaps a solution like adding ```root[-1]``` at the end of the loop would work the same way without having to add the entire path each iteration. 

Also, can you include a test that shows the error and then ensures that it works with the fix?

md0000:
Changed fix and added tests.

Also, there was a slight bug in _bidirectional_dijkstra and _bidirectional_shortest_path which sometimes cloaked bug in shortest_simple_paths. Fixed it and added some tests.

dschult:
This looks good to me.  Anything else you are planning for this PR?

md0000:
No, thats all.

PR #2713, opened 2017-10-15
Fix for inter_community_edges

### LINKED ISSUE #2600: inter_community_non_edges ignore directionality
Hi,

I think the function:

nx.algorithms.community.quality.inter_community_non_edges()

does not work properly for directed graph. It always return the non-edge of a undirected graph, basically halving the number of edges. This mean that the performance function (nx.algorithms.community.performance) will never by higher than 50% for a directed graph.

I'm using version '2.0.dev_20170801111157', python 3.5.1

Best,

Nicolas

md0000:
inter_community_edges sometimes returned wrong answer when working with digraphs. This patch should also fix #2600.


dschult:
Thanks!!

Issue #2578, opened 2017-08-03
Problem with transitive_closure not copying data


Dorah3D:
Hello everyone!

It seems that with some of the new development versions posterior to May, 2017 the transitive_closure function of DAG does not transfers the node and edge data to the new graph. Perhaps his part of the code:

`TC = nx.DiGraph()
TC.add_nodes_from(G.nodes())
TC.add_edges_from(G.edges())`

might need the "data = True" statement as argument.

dschult:
As far as I can tell this method never transferred the edge data to the new graph.  Can you verify that you see the problem with the dev version, but that you don't know whether the v1.11 release has the problem?  

It also looks like tis code will not work with multigraphs.

It might be that transferring edge data to the new graph should be done with ```set_edge_attributes``` or similar. But if its easy to do in the routine, we should probably put it in the code. The first attempt should replace the cited lines of code with ```TC = G.copy()```

PR #2718, opened 2017-10-22
Copy graph in transitive closure algorithm.

aweltsch:
Hi,
I implemented a small patch to fix issue #2578. I used the solution propsed there and added appropriate tests.
Andre


dschult:
This looks great -- Thanks!!

Issue #2451, opened 2017-05-11
nx_shp does not handle NoneType objects


kcavagnolo:
### Expected Behavior
In the case when a shapefile has not been rigorously created, there may be missing geometries which Python sees as NoneType. These NoneType instances are caught and an exception raised for graceful failure, or a keyword to nx_shp allows the function to skip the bad geometries and continue on.

### Actual Behavior
NoneTypes raise an uncaught AttributeError:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-33-2307d2480ccb> in <module>()
      1 # load shapefile into a digraph
----> 2 g = nx.read_shp(shapefile, simplify=False)

/usr/local/anaconda/lib/python3.5/site-packages/networkx/readwrite/nx_shp.py in read_shp(path, simplify)
     74             attributes = dict(zip(fields, flddata))
     75             attributes["ShpName"] = lyr.GetName()
---> 76             if g.GetGeometryType() == 1:  # point
     77                 net.add_node((g.GetPoint_2D(0)), attributes)
     78             if g.GetGeometryType() == 2:  # linestring

AttributeError: 'NoneType' object has no attribute 'GetGeometryType'
```

### Suggested Fix
Add a try-except above line 76 and pass the error.

PR #2721, opened 2017-10-25
nx_shp fixes

### LINKED ISSUE #2432: Bug in write_shp when some edges miss attribute(s)
When writing shapefiles with v1.11, `write_shp` uses the last valid values for attributes an edge doesn't have. The reason is that the values of missing `key`s in the dictionary `attributes` are not updated in the edge loop. A work-around is to make sure all attributes are set before using `write_shp` or to remove edges with missing attributes but this requires a separate edge loop. So, it would be nice to have a fix integrated in `write_shp`. Thanks!

md0000:
Hi, This PR fixes some minor bugs in nx_shp:
- Updates read_shp to raise RuntimeError when file cant be open or read.
- Adds new keyword `strict` to read_shp and implements suggestion #2451 , read_shp now raises NetworkXError when feature lacks geometry and `strict=True`. Conversely `strict=False` can be used to silently ignore features without geometry.
- Adds small change to write_shp, should fix #2432 .
- Some new tests which check this updates.

Update:
Some builds were failing because of no ogr module. Fixed testcase which was causing problems.

### DIFF
--- a/networkx/readwrite/nx_shp.py
+++ b/networkx/readwrite/nx_shp.py
@@ -23,7 +23,7 @@ __author__ = """Ben Reilly (benwreilly@gmail.com)"""
 __all__ = ['read_shp', 'write_shp']
 
 
-def read_shp(path, simplify=True, geom_attrs=True):
+def read_shp(path, simplify=True, geom_attrs=True, strict=True):
     """Generates a networkx.DiGraph from shapefiles. Point geometries are
     translated into nodes, lines into edges. Coordinate tuples are used as
     keys. Attributes are preserved, line geometries are simplified into start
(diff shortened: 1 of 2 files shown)


dschult:
For the part that fixes #2451 it looks from the OP that the code raises an exception when g is NoneType because it can't find ```g.getGeometryType```. Does this fix avoid that exception?  It looks like this fix correctly handles the case when ```g.getGeometryType``` returns an invalid type, but would still fail when g is NoneType.

That said, I don't fully understand the OP example -- and you say this implements the fix suggested in #2451 so can you help me understand what is going on with that Issues?  Thanks!

md0000:
Yes this fix avoid that exception (AttributeError described in #2451). The problem is that `f.geometry()` under certain circumstances returns None, so g is None. Check for this possibility is on line 101. I hope this explanation will be helpful.

Also I added one testcase for this possibility (class TestMissingGeometry).

dschult:
Yes -- Thank you!  I was getting mixed around with the strict argument's intent.  I understand much better now. This was helpful.

Also, thank you for the PR! :)

PR #2722, opened 2017-10-26
removed reference to create_using from union docs

mcognetta:
The doc string for union had a reference to "create_using" which is not an actual parameter.


dschult:
Thanks!

PR #2731, opened 2017-11-02
Correct when we raise NetworkXNotImplemented

gfyoung:
`algorithms/components/connected.py` was littered with incorrect descriptions that we raise `NetworkXNotImplemented` for undirected graphs, when we meant to say directed.


dschult:
Thanks!

PR #2739, opened 2017-11-06
The graph6 generator fails when relabeling

hagberg:
When relabeling for consecutive nodes 0,..,n-1 use the new (relabeled) graph to test for edges.


hagberg:
Fixed in https://github.com/networkx/networkx/commit/5c22e401bdf5fdff19147bd1328a934f8d3205b5

PR #2742, opened 2017-11-08
Fix 1568: Pydot - GraphViz visualization for complex objects

### LINKED ISSUE #1568: Graphviz layout breaks if two nodes have the same string representation
If two nodes in a graph have the same string representation (i.e, `str(u) == str(v)`), computing a graphviz layout is broken. For example:

```
import networkx as nx

class Node(object):
    def __str__(self):
        return 'Foo'

u = Node()
v = Node()
w = Node()

g = nx.DiGraph()
g.add_edges_from([(u,v), (v,w)])

print nx.graphviz_layout(g)
```

This outputs:

```
{
    <__main__.Node object at 0x1045b8dd0>: (0.0, 0.0), 
    <__main__.Node object at 0x1045b8e50>: (0.0, 0.0), 
    <__main__.Node object at 0x1004e2150>: (0.0, 0.0)
}
```

Notice that each of the three nodes is (undesirably) mapped to the same location. If the `__str__` method is removed from the `Node` class definition and the example is run again, we get a more sensible result:

```
{
    <__main__.Node object at 0x1004b4d90>: (-37.382, 60.934), 
    <__main__.Node object at 0x1046c0d50>: (6.6387, 3.6105), 
    <__main__.Node object at 0x1046c0dd0>: (30.743, -64.544)
}
```

The problem can be attributed to the fact that `networkx.graphviz_layout` uses `pydot` to compute the layout. In converting the input graph to a `pydot` graph, the nodes must be converted to strings. This occurs on the [following line](https://github.com/networkx/networkx/blob/master/networkx/drawing/nx_pydot.py#L212) of the `to_pydot` function.

It appears that the issue also occurs when using `networkx.drawing.nx_agraph.pygraphviz_layout` to compute the layout for a similar reason.

The straightforward fix, I think, is to relabel the nodes of the graph given to `networkx.drawing.nx_pydot.pydot_layout` so that they have unique integer indices, then proceed as normal. We must then invert the keys of the resulting layout dictionary before returning it to the user so that they are the original node objects.

Because this must also be done for `networkx.drawing.nx_agraph.pygraphviz_layout`, I think the best fix might be to implement a decorator which performs these transformations, and apply it to both functions. I'd be happy to submit a pull request in the next few days if this sounds like a reasonable approach.

thammegowda:
As reported in #1568, GraphViz breaks down when the nodes are complex python objects (especially when string representation of two nodes in the graph are same). See comments https://github.com/networkx/networkx/issues/1568#issuecomment-341931155 
## After this fix:
screen shot 2017-11-07 at 5 27 21 pm

## Without this fix:
screen shot 2017-11-04 at 3 42 53 pm

- Closes #1568 
- Please merge this, or let me know if any changes are to be made. This is crucial for my use, otherwise I will have to keep track of 2 versions of networkx.
- Tested on Python 3.6

### DIFF
--- a/networkx/drawing/nx_pydot.py
+++ b/networkx/drawing/nx_pydot.py
@@ -23,7 +23,7 @@ DOT Language:  http://www.graphviz.org/doc/info/lang.html
 #    All rights reserved.
 #    BSD license.
 from locale import getpreferredencoding
-from networkx.utils import open_file, make_str
+from networkx.utils import open_file, make_str, make_id
 from pkg_resources import parse_version
 import networkx as nx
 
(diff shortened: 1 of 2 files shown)


dschult:
This feels like a good (hack) fix for this case, but using ```id``` to identify nodes will potentially create problems from one run to the next if you do anything else with the dot file. It also doesn't pass tests without a way to read in that structure (ad I'm not sure we want to always read in that structure). 

So, Hopefully people will find this PR by searching the repo if they need it. But I'm not going to merge it.  Thanks for documenting your ideas and submitting the PR!  :)

PR #2745, opened 2017-11-09
returning information on edges for simple path on a multigraph

vicmancr:
This is a simple implementation to show edges information on simple paths for a multidigraph. There is an important difference: it is returning now two lists, one with path as nodes and one with path as edges. I have combined them in a global list, such that it has the form [list_of_nodes, list_of_edges].

I am open to suggestions.

### DIFF
--- a/networkx/algorithms/simple_paths.py
+++ b/networkx/algorithms/simple_paths.py
@@ -171,7 +171,7 @@ def all_simple_paths(G, source, target, cutoff=None):
     if cutoff is None:
         cutoff = len(G) - 1
     if G.is_multigraph():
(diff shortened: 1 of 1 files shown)


dschult:
Returning edge data for simple paths should be implemented by the user as post-processing of the simple path. We want our functions to be simple and efficient and the upside of avoiding looking up the edge data from the graph doesn't make up for the downside of building and returning data lists even when data is not wanted.

Try (untested):

    [[(u, v, G[u][v]['data_attr']) for u, v in nx.utils.pairwise(path)] for path in nx.all_simple_paths(G)]

PR #2748, opened 2017-11-12
BF: propagate `kwds` to MPL call

arokem:
See: https://github.com/nipy/nitime/issues/153, https://github.com/nipy/nitime/pull/157

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -397,7 +397,8 @@ def draw_networkx_nodes(G, pos,
                                  vmax=vmax,
                                  alpha=alpha,
                                  linewidths=linewidths,
-                                 label=label)
+                                 label=label,
+                                 **kwds)
 
     node_collection.set_zorder(2)
     return node_collection


hagberg:
Yeah, we've been there before (2010)
(adding kwds) https://github.com/networkx/networkx/commit/0a8d31dfe027f8ea0edd24026f458ed9d9161b3f#diff-6acdfd281af96e48405fe5f51cad122f
(then removing) https://github.com/networkx/networkx/commit/3843ffe3b428344ae2f173217f9ba3057ae581b4#diff-6acdfd281af96e48405fe5f51cad122f

If there is a particular keyword/s we might add to help you I think that is the safest approach.

arokem:
Thanks! We need the `edge_color` key word argument, I believe. Either way, if you don't want to use the `**kwds`, you should probably also remove it in the function signature?

https://github.com/networkx/networkx/pull/2748/files#diff-6acdfd281af96e48405fe5f51cad122fL287

hagberg:
Yes, *kwds should be removed in `draw_networkx_nodes` too.
Let's add `edge_color` as a keyword. 

Since the `draw_networkx_nodes` function returns the node collection you can set the edge color that way too
```python
In [1]: import networkx as nx

In [2]: G = nx.Graph()

In [3]: G.add_edge(1,2)

In [4]: pos = nx.spring_layout(G)

In [5]: n = nx.draw_networkx_nodes(G,pos)

In [6]: n.set_edgecolor('k')
```

dschult:
I think the ```**kwds``` argument trouble only occurs when you have one NetworkX function calling multiple functions with the same kwds dict.  For example, that means ```nx.draw_networkx``` should not use ```**kwds``` because those arguments could be meant for nodes, or edges, or labels.

But ```nx.draw_networkx_nodes``` can use kwds because it only calls ```ax.scatter``` and we can document that any kwds not specifically provided will be passed to matplotlib's ```scatter```.

```nx.draw_networkx_edges``` calls ```LineCollection``` twice -- once for edges and once for arrows. Perhaps arrow drawing should be separated from edge drawing. Or we can provide the same ```kwds``` to both?

```nx.draw_networkx_labels``` and ```nx.draw_edge_labels``` simple pass kwds to ```ax.text```.

I think using ```**kwds``` whenever it is clear what function they get delivered to will be easier in the long run than trying to keep up with the matplotlib argument changes. If we don't think matplotlib will change their arguments anytime soon, we can simply provide all their arguments explicitly.  But even then it seems cleaner to use kwds and document that we are passing those to a specific matplotlib tool.

dschult:
The reason this doesn't pass tests is that ```**kwds``` cannot be removed even though it LOOKS like it is not used. In fact it is used in ```draw_networkx_nodes``` to collect all the keywords for edges and labels that aren't intended for nodes. The function ```draw_networkx``` uses the same dict for attributes for all three types of drawing.  

Our use of ```**kwds``` here is not documented and is confusing. 

Best solution may be to pass keywords through to mpl for ```_nodes```, ```_edges```, ```_labels``` but to process them in ```draw_networkx``` -- allowing only select attributes to be passed along.

dschult:
I'm closing this in light of #2788 adding edgecolors for node borders and the  ongoing work in #2757
I envision a reconsideration of ```**kwds``` for these routines that somehow protects/separates the keywords meant for each part of the drawing.

PR #2753, opened 2017-11-16
Add directed graphs support for jit_graph reading

Dark-Storm:
Just added the support for getting directed graphs from a previously generated JIT file.

### DIFF
--- a/networkx/readwrite/json_graph/jit.py
+++ b/networkx/readwrite/json_graph/jit.py
@@ -39,7 +39,7 @@ from networkx.utils.decorators import not_implemented_for
 
 __all__ = ['jit_graph', 'jit_data']
 
-def jit_graph(data):
+def jit_graph(data, directed=False):
     """Read a graph from JIT JSON.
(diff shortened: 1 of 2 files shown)


dschult:
Instead of a ```directed``` argument, should this function have a ```create_using``` argument so people can make ```MultiGraph``` and ```MultiDiGraph``` too?

Dark-Storm:
The jit_data function is not implemented for MultiGraph. But that's an idea. I'll check this
```
@not_implemented_for('multigraph')
def jit_data(G, indent=None):
```

dschult:
```jit_data``` is writing the data while ```jit_graph``` is reading it.  So I would be fine with having only ```jit_graph``` support ```create_using```. That API also makes it easier for people to use their home-grown graph classes with this function...  Probably a small market, but I think its the better API.
Thoughts?

    def jit_graph(data, create_using=None):
        ...
        if create_using is None:
             G = nx.Graph()

PR #2759, opened 2017-11-19
Fix #2726: ensure add_path to add the first node

### LINKED ISSUE #2726: Add a node when add_path involves no edge.
`add_path` is convenient for adding edges without adding nodes explicitly.
Consider a case where there exists no edge to be added:

```
G = nx.Graph()
nx.add_path(G, [0])
```
If one wants to test the neighbors of a node in a graph, `G.neighbors(0)` will throw networkx.exception.NetworkXError: The node 0 is not in the graph.

Obviously, one can check whether the argument for `add_path` is a singleton and explicitly do

```
path = [0]  # or something else
G = nx.Graph()
if len(path) == 1:
    G.add_node(path[0]) 

nx.add_path(G, path)
G.neighbors(0)
```

Should adding a zero-edge path with a single node add a node in the graph if the node isn’t already in the graph?
This also applies to `add_star`.

sanghack81:
Fixes #2726 

If the given path (list of nodes) is of length one, `add_path` will add the first node.

### DIFF
--- a/networkx/classes/function.py
+++ b/networkx/classes/function.py
@@ -258,7 +258,13 @@ def add_path(G, nodes, **attr):
     >>> nx.add_path(G, [0, 1, 2, 3])
     >>> nx.add_path(G, [10, 11, 12], weight=7)
     """
-    G.add_edges_from(pairwise(nodes), **attr)
+    nlist = iter(nodes)
+    try:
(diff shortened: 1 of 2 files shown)


hagberg:
Thanks (especially for the tests).

Issue #2163, opened 2016-06-14
DiGraph arrowheads not colored properly with numeric edge_color


ChrisMyers:
In draw_networkx_edges, the edge_color argument can be passed in as a numeric list/array which gets mapped to colors via edge_cmap, edge_vmin, edge_vmax.  Even though the edges are colored correctly, the mapped edge_colors do not get attached correctly to the arrow heads drawn for a DiGraph.  (Note: arrowheads are colored properly if a non-numeric edge_color list is passed.)  Even though the arrow colors are set to the edge colors (arrow_colors=edge_colors), for the numeric mapping case, edge_colors=None and the colors are instead set directly in the edge_collection.  No such similar mapping is carried out for the arrow_collection.  One can replicate the same "if edge_colors is None" logic inside the arrow drawing block (setting instead the array for the arrow_collection), but perhaps there is a cleaner way to organize this so as not to duplicate that logic.

hagberg:
That code is patches on hacks and really does need a cleaner overall approach (and proper arrow drawing).  But maybe another patch like you suggest is the best approach at this point.

hagberg:
We can add this feature to #2760 which is making "proper" arrows.

Issue #1199, opened 2014-06-24
Feature request: alpha argument in draw_networkx_nodes (or edges) can take a list


hdmetor:
In that way it would be possible to change each one separately. Exactly the same behavior as in node_color and node_size

PR #2760, opened 2017-11-19
Digraph Arrows to fix #2757

### LINKED ISSUE #2757: Improvement to arrowheads of a DiGraph edges.
I was wondering if there was  planing on changing the arrowheads of the edges of directed graphs to be more arrow like (triangular shape instead of thicker line). For instance see [this question](https://stackoverflow.com/questions/47213650/changing-arrowhead-type-in-networkx) on stackoverflow.

rodogi:
This pull request targets to replace the thicker lines used to draw arcs (directed edges) used in the nx_pylab.py file with arrow heads.

The main idea is to replace the `matplotlib.collections.LineCollection` with a list of `matplotlib.patches.FancyArrowPatch`.

The implementation is a bit slower than the original, but adds some clarity when zooming on a a network, e.g. consider the following the network with N nodes and N*2 edges:

```
import networkx as nx


G = nx.Digraph()
G = nx.generators.directed.random_uniform_k_out_graph(N, 2)
pos = nx.layout.spring_layout(G)
nx.draw_networkx(G, pos=pos, edge_color='g', arrowstyle='-|>')
plt.show()
```

For **N=100**, this normally draws:
regular_100

With this pull request:
arrows_100

For **N=1000**

Zooming on the network gives:
regular_1000_zoom

With this pull request:
arrows_1000_zoom

I added the keyword `arrowstyle` to be able to use all the styles for `matplotlib.patches.FancyArrowPatch`, however, the drawing is chaotic for some of them.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -513,7 +513,7 @@ def draw_networkx_edges(G, pos,
     >>> arcs = nx.draw_networkx_edges(G, pos=nx.spring(layout(G)))
     >>> alphas = [0.3, 0.4, 0.5]
     >>> for i, arc in enumerate(arcs):  # change alpha values of arcs
-            arc.set_alpha(alphas[i])
+    ...     arc.set_alpha(alphas[i])
 
     Also see the NetworkX drawing examples at
     https://networkx.github.io/documentation/latest/auto_examples/index.html
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -517,10 +517,9 @@ def draw_networkx_edges(G, pos,
         import matplotlib
         import matplotlib.pyplot as plt
         import matplotlib.cbook as cb
-        from matplotlib.colors import colorConverter, Colormap
+        from matplotlib.colors import colorConverter, Colormap, Normalize
         from matplotlib.collections import LineCollection
         from matplotlib.patches import FancyArrowPatch
-        from matplotlib.markers import MarkerStyle
         import numpy as np
     except ImportError:
         raise ImportError("Matplotlib required for draw()")
@@ -573,7 +572,7 @@ def draw_networkx_edges(G, pos,
             raise ValueError(
                 'edge_color must be a single color or list of exactly m colors where m is the number or edges')
 
-    if not (G.is_directed() or arrows):
+    if (not G.is_directed() or not arrows):
         edge_collection = LineCollection(edge_pos,
                                          colors=edge_colors,
                                          linewidths=lw,
@@ -607,9 +606,9 @@ def draw_networkx_edges(G, pos,
     arrow_collection = None
 
     if G.is_directed() and arrows:
-        # Note: Waiting for someone to implement arrow to edge of marker.
-        # Meanwhile, this works well for polygons with more than 4 sides and
-        # circle.
+        # Note: Waiting for someone to implement arrow to intersection with
+        # marker.  Meanwhile, this works well for polygons with more than 4
+        # sides and circle.
 
         def to_marker_edge(marker_size, marker):
             if marker in "s^>v<d":  # `large` markers need extra space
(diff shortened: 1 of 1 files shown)


hagberg:
Thanks, that is a good start.  Can you figure out how to get the arrows to scale correctly with node size?  For circular nodes this shouldn't be too complicated.  For other shapes this will be harder since you probably want to find the closest polygon vertex and route the arrow there.  

```
G = nx.Graph([(1,2)])
nx.draw(G,node_size=10)
nx.draw(G,node_size=1000)
````

figure_1
fig2

rodogi:
This solves the problem for marker=='o':

figure_1 5
figure_2 5
 

For the rest of the markers it gets trickier as we need to calculate the distance from the center of the marker to the intersection point between the edge and the marker.  It is doable but it will certainly take me a bit of time.

hagberg:
Also see #2163 for arrow colors fixes

hagberg:
And #1199 for alpha-values setting

rodogi:
Two keyword arguments to define the style and size of the arrow are included (`arrowstyle`, and `arrowsize`, respectively). There's the possibility to use an array of numbers for the edge colors and have the respective color for the arrow:

net1

The arrows will never overlap the marker, however for some markers (namely the square, diamond and triangles) the arrow might be clearly distanced for the marker. However, this doesn't seem to affect (too much) the quality of the drawings.


net2

For the rest of the markers, which it works well (polygons with more than 4 sides).
The alphas as a list is not yet included.

dschult:
I've been wondering about what should be returned with ```_edges```... 
We could return an edge_collection when arrows are not requested
and an ```(edge_collection, arrow_collection)``` tuple if arrows are requested.

Another approach would be to make two functions -- one for edges and one for arrows.
Each returns the collection it creates.  Each gets its own ```**kwds```.  Is there much
overlap between keywords for line and fancyarrow?

rodogi:
@dschult We could try to use a `mpatches.FancyArrow` instead of `mpatches.FancyArrowPatch` to draw the arrows. Patch `mpatches.FancyArrow` has no bug when used with `mcollections.PatchCollections`. This way we could maybe return `edge_collection` when non-directed and `arrow_collection` when directed?

Making a second function is a good idea too depending on the number of flexibility in terms `**kwds` for `mpatches.FancyArrow` intended. If the intention is to leave a couple of arguments to play with the arrow length, width, and shape, maybe only one function is okay?

dschult:
I think for now let's try to get the features in with one function and a few arrow-specific arguments. I guess I didn't understand that for the arrows case we don't have a line collection at all...  Guess I need to look at the code more. :}  It seems like ```FancyArrowPatch``` has some features that help relative to ```FancyArrow```, so I'll leave the choice of which to use up to you. And if the bug with ```patchcollections``` is likely to be fixed then let's not make the choice depend on that.```

rodogi:
Yes actually a line collection is returned when the graph is either non-directed or the arrows are disabled. For the case when the graph is directed and arrows are enabled we could use a patch collection for the arrows. The thing is that `FancyArrowPatch` has been reported to not work with `PatchCollection` since 2013 See [here](https://github.com/matplotlib/matplotlib/issues/2341).

Probably the best idea is to leave a couple of arrow-specific arguments dealing with the head of the arrow and implement `FancyArrow` rather than `FancyArrowPatch` to return a patch collection when arrows are drawn and line collection otherwise...?

dschult:
I'm fine with returning None for the time being.
Any container of the patches would be fine with me also.
What do you think?

rodogi:
Maybe a list of `FancyArrowPatch` instances adding an example in the documentation of how to manipulate the looking of the arrows using the list?

dschult:
Yes -- that is a good idea.
If the example is to change the alpha values of the arrows individually, then I think that would take care of the last unaddressed target for this implementation.  Are there any others?

hagberg:
This is a really great contribution.
I suggest calling it good for this PR with maybe the addition of making sure our desire for arrow collections to enable this more efficiently is noted over at the matplotlib site.

rodogi:
I'm glad to be of some help to such an amazing package, keep the good work.

Should I include a plot example in :examples:drawing: varying for instance alpha values, or just include it in the documentation of `draw_networkx_edges`/`draw_networkx`?

dschult:
I think the examples folder would be for something bigger while the doc_string works better for a small example. If you can create a 2-3 edge network with different alphas for each arrow, I'm guessing that would be short enough to put in the doc_string.  I'd prefer it there. Then as a separate PR we could create an example for the examples folder that highlights the arrows features more generally.

Does this make sense? Is there a better way?

rodogi:
Yes -- it makes sense. I can eventually make a pull request to add a plot with the arrows in the example directory. I added an example to change the alpha values in the doc-string.

I also added a list in the returns section of the doc-string, it is not clear to me what is the proper syntax used when two different types are returned depending on the use of arrows or not. So I might need to change it, I currently added:

     
    Returns
    -------
    matplotlib.collection.LineCollection
        `LineCollection` of the edges

    list of matplotlib.patches.FancyArrowPatch
        `FancyArrowPatch` instances of the directed edges

    Depending whether the drawing includes arrows or not.

Should I change it?

Issue #1764, opened 2015-09-07
Graph generators change default random seed


AgostinoSturaro:
I noticed that, when passing graph generators a `seed` parameter, they change the _default_ seed for `random`. This affects affecting all subsequent calls to `random` in the module.
That is, if a single generator is passed a `seed` parameter, all other generators are affected.
[Most](https://github.com/networkx/networkx/blob/v1.10/networkx/generators/random_graphs.py#L47) [generators](https://github.com/networkx/networkx/blob/v1.10/networkx/generators/random_graphs.py#L602) seem to be doing this. The code in question is

```
if seed is not None:
    random.seed(seed)
```

Here's a discussion [on SO](http://stackoverflow.com/q/12368996/1600770). There is a simple solution. Creating an instance of `random` inside each function would make the seed change self-contained and prevent it from affecting other calls. Something like this would do

```
if seed is not None:
    my_random = random.Random(seed)
else:
    my_random = random
...
# use my_random for calls
```

The same trick can be used if a NumPy random generator is needed

```
my_random = np.random.RandomState(seed)
```

This could be especially useful to add a `seed` parameter to the layout functions, that currently lack one, and do not provide a way to arrange nodes deterministically (the same way in two different calls). see for instance [this function](https://github.com/networkx/networkx/blob/v1.9.1/networkx/drawing/layout.py#L55).

NeilGirdhar:
We shouldn't be taking a seed parameter at all.  We should be accepting an instance of `random.Random` or if None using `random`.  This would allow running simulation for example with a random number generator and getting repeatable results.  You don't want to have to come up with seeds for every time the rng is used.  You want to seed it once and pass it to ever function that needs an rng.

chebee7i:
Related: #1550

jeffzemla:
This bug just cost me three days of my time, so I second using `my_random=random.Random(seed)`. 

Personally I would prefer to pass it a random seed rather the an instance of random.Random, but I can see the merit in that as well.

NeilGirdhar:
@jeffzemla The problem with passing in seeds is that you might have many other pieces of code that use random numbers.  Are you supposed to seed all of those random number generators at the start of your simulation?  When you want to restore the simulation to a particular point, you need to save all, say 5, seeds?  That is really poor design.

That's the reason that all of the scipy and numpy functions that take random numbers accept a random number generator.  It is the de facto Python way of introducing randomness, and in my opinion, it's good design.

[...]

Issue #1550, opened 2015-05-27
Interface for randomness


chebee7i:
In various places, we make use of random number generators. Due to our pure Python tilt, we've tended to use the `random` module. Some functions accept seeds as input, others (I have not checked) might not provide that option at all.

This briefly came up in https://github.com/networkx/networkx/issues/1341#issuecomment-76645971, and maybe it's time to dig in a bit further.

My opinion is that we should not use the `random` module. While it has satisfied our needs thus far, it does not allow enough flexibility since 1) there is always a global state and 2) there aren't many distributions provided. The ability to have different random number generator instances requires that we move away from using the `random` module.

Instead, I think any function that requires randomness should have a hard dependency on NumPy and that we should standardize on NumPy's `RandomState` API. This isn't that much of a burden since most installations can easily obtain NumPy, and a good deal of NetworkX now relies on NumPy anyway. This is also what _most_ Python packages do if they require randomness. See for example, [scikit-learn-#random-numbers](http://scikit-learn.org/stable/developers/#random-numbers). The idea is that functions/classes/methods will have a parameter, `prng` or `random_state` (as in `sklearn`) or `seed`.  We then use that to retrieve a random number generator using something identical to [check_random_state](http://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html#sklearn.utils.check_random_state) (but with a better name). Then the function uses that random number generator. This will make code more robust and more easily tested.
#1523 will be affected by this decision.

hagberg:
It's a reasonable idea.  A couple thoughts -
- The dependence on Numpy is all  (or mostly) through "linear algebra" approaches to graph problems so this would add a general dependency in other places.
- Without making it overly complicated would it be possible to still use the built-in Python random generator?   That would be nice for e.g. pypy or other cases when Numpy is unavailable.

ysitu:
Actually, you can instantiate the `random.Random` class. The way it is documented makes it look as if it were undocumented, but it is indeed documented.

chebee7i:
Ah ok, then that should be doable!

jfinkels:
When this change is made, a reminder should also be documented somewhere that randomness should not be used for cryptographic security (unless NetworkX explicitly includes a secure random number generator).

NeilGirdhar:
[...]

PR #2766, opened 2017-11-22
Add seed keyword argument to random_layout and spring_layout

harryscholes:
This PR adds functionality that allows reproducible node layouts to be generated by `random_layout` and `spring_layout` via the `seed` keyword argument.

---

Re-running code that employs stochastic layout algorithms to generate figures positions nodes differently each time:

```
import networkx as nx, matplotlib.pyplot as plt

G = nx.erdos_renyi_graph(10, 0.4, seed=1)

fig, axes = plt.subplots(3, 3)
axes = axes.flatten()

for ax in axes:
    pos = nx.spring_layout(G)
    nx.draw_networkx(G, pos, ax=ax, node_size=150, font_color='w')
    ax.axis('off')
```

random

This PR adds the `state` keyword argument to the `random_layout` and `spring_layout` functions to set the seed of `numpy.random.seed`. Setting `seed` allows reproducible node layouts to be generated:

```
fig, axes = plt.subplots(3, 3)
axes = axes.flatten()

for ax in axes:
    pos = nx.spring_layout(G, random_seed=1)
    nx.draw_networkx(G, pos, ax=ax, node_size=150, font_color='w')
    ax.axis('off')
```
seeded

This feature will be very useful when generating figures for presentations/papers.


ghoresh11:
Great idea! This will be very useful

dschult:
This looks good.  But could you take a look at #1764 and #1550 where it is argued that we shouldn't be doing this? What is the right approach to handling random number generators and numpy.random in particular?

harryscholes:
@dschult thanks for these comments. As I have currently implemented this, the random state is set globally and leaks out of the scope of `spring_layout` and `random_layout`.

I have fixed this by providing a `random_state` argument that accepts an instance of a NumPy random number generator, an `int`, or `None`. If an `int` is provided, this will be used to instantiate `np.random.RandomState`. If `None` or `np.random`, this will be used to instantiate an unseeded `np.random.RandomState`.

```python
if random_state is None or random_state is np.random:
    rng = np.random.mtrand._rand
elif isinstance(random_state, np.random.RandomState):
    rng = random_state
else:
    rng = np.random.RandomState(random_state)
```
This is heavily inspired by how sklearn [does this](https://github.com/scikit-learn/scikit-learn/blob/a24c8b464d094d2c468a16ea9f8bf8d42d949f84/sklearn/utils/validation.py#L617).

Example :

```python
In [6]: nx.spring_layout(G, seed=1)
   ...: np.random.rand(5)
   ...:
Out[6]: array([ 0.84351284,  0.68102688,  0.11044101,  0.92682998,  0.81743538])

In [7]: nx.spring_layout(G, seed=1)
   ...: np.random.rand(5)
   ...:
Out[7]: array([ 0.72168598,  0.08398381,  0.82544323,  0.49460946,  0.62301562])
```

The same approach could be taken to fix #1764.

dschult:
Very nice!   
This could be helpful elsewhere I'm sure... in addition to #1764 but certainly there.

It'd be nice, (for backward compatibility even for private functions) to leave the existing arguments in the same order and add the random state argument to the end. But if there are other reasons to add it earlier I'm open to that.  Also it should probably have a default -- perhaps of None -- even in private functions.

Does it make sense to pull together some helper code into the nx.utils to help with random states? I suppose it could even be a decorator for functions similar to the one we use to open files. But perhaps I am getting too ambitious and we should merge this first and see if we like it.  I'm fine with either approach.

harryscholes:
Updates:
 - Put `random_seed` at the end of the function arguments
 - Created `check_random_state` function in `nx.utils.misc` that returns instances of `np.random.RandomState`
 - Add `@random_state()` decorator, which calls `check_random_state` on the value of `random_state` function arguments
 - Decorate `random_layout` and `fruchterman_reingold_layout`

harryscholes:
Fixed lots of conflicts due to #2776 and #2681

dschult:
Don't worry about the travis test failure in mac osx. They changed their build environment.

Minor things:   
- don't wrap the imports in try/except. (too much overhead and they don't cause errors if not in the module namespace.
- For sphinx, the docstring should start with a single line description (that means <80 chars too). Then a blank line and then (if reasonable) a paragraph defining what the terms mean in the single line description.   Blank lines separate sections. There are competing styles for docstrings, but take a look at [this one](http://www.sphinx-doc.org/en/stable/ext/example_numpy.html) from numpy and sphinx.

This is great stuff!  Thanks for chasing the bugs too!

harryscholes:
Cheers Dan – I've incorporated your minor points and added tests for the `@random_state` decorator. All the tests are now passing.

dschult:
OK... Is this all that is planned for this PR?  
If so, I'll take one more check through it and then merge.
:)

harryscholes:
Your changes look good 😄

Yep I think this is all for this PR. In the future we can use these findings to address #1764 and related issues to replace `random.seed` with `random.Random` instances.

Issue #2762, opened 2017-11-21
networkx.algorithms.simple_paths.all_simple_paths function cutoff parameter and returned paths


manoskav:
Hello,

I would like to point 2 things about the all_simple_paths function:
1) By the definition of simple paths, the function should not return cyclic paths, that is paths from one node to itself.
2) Please update the documentation to state explicitly that the function **by default** searches for paths with cutoff=(number of nodes in graph -1). You must inspect the source code in order to find that, which may lead, as in my case, to a couple of hours of debugging.

Thanks for your time and sorry if this issue is unimportant to be posted here.

hagberg:
The intent of all_simple_paths is to produce paths without cycles. The function doesn't check if the source = target which can produce cycles in that special case (that probably is a bug).  Is that the issue you are asking about?

manoskav:
Yes, that is my point.

hagberg:
This is a simple fix - if the user asks for source == target just return an empty list.

PR #2770, opened 2017-11-25
all_simple_paths should not return cycles. Fix issue #2762

ForFer:
Following instruction on fix for issue #2762


pelegm:
I think that the right thing to do here is different; if source and target are the same, then a unique path should be yielded, the path [source] (consisting of a single vertex and no edges).

Issue #2529, opened 2017-07-21
Blockmodel with incomplete/partial partitions


pckroon:
Hi all,

This is part question, part easy improvement suggestion.
Currently, if I want to make a blockmodel of my graph I have to assign every node to a partition:

```
G = nx.path_graph(6)
partition = [[0, 1, 2], [3, 4, 5]]
M = nx.blockmodel(G, partition)
```

Is it desirable to be able to provide incomplete partitions?

```
G = nx.path_graph(6)
partition = [[0, 1], [4, 5]]
M = nx.blockmodel(G, partition)
```

This would generate a disjoint graph, but I don't think this is really an issue (but I'm not well versed in graph theory). 
Currently, in version 1.11, if I try this it gives me a `KeyError` which can be easily fixed/circumvented by either a check or a `try`/`except` in `networkx/algorithms/block.py` around lines 100 and 101.

What's your thoughts on this? I'd be happy to make a PR.

mddddd:
I would be happy to contribute PR for this issue but have some questions. As I see `blockmodel` was replaced by more general `quotient_graph`, taking this into acount is this issue still valid? (some math oriented people may argue that there is no quotient graph without proper partition, so this should be error)

If it is still valid then what is the expected behaviour when partition is incomplete? I see 2 possibilities:
- Return a quotient graph of a subgraph induced by nodes from partition.
- Update partition with singletons made from nodes not present in original partiton.

what do you think?

pckroon:
Hey,
thanks for the reply :)

`quotient_graph` seems to be capable of doing what I want, but it is a little cumbersome to have to define a function and filter out the singleton nodes afterwards. E.g:
```python
G = nx.Graph()

G.add_edges_from([(0, 1), (1, 2), (2, 3), (3, 4)])

partitions = [(0, 1), (3, 4)]

def quot(n1, n2):
    for partition in partitions:
        if n1 in partition and n2 in partition:
            return True
    return False

H = nx.quotient_graph(G, quot)

remove = set()
for node in H:
    if len(node) == 1:
        remove.add(node)
H.remove_nodes_from(remove)
```

I don't now enough about the math to be able to say anything about "proper" partitions and such, and what the ramifications are. 
What I expected as behaviour is option 1, so I need to filter out singleton nodes. 

[...]

PR #2771, opened 2017-11-25
Add support for incomplete partitions in quotient_graph.

mddddd:
Hi, this PR adds support for incomplete partitions as discussed in #2529 .

### DIFF
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -220,7 +220,27 @@ def quotient_graph(G, partition, edge_relation=None, node_data=None,
     # the blocks of the partition on the nodes of G induced by the
     # equivalence relation.
     if callable(partition):
+        # equivalence_classes always return partition of whole G.
         partition = equivalence_classes(G, partition)
+        return _quotient_graph(G, partition, edge_relation, node_data,
+                               edge_data, relabel, create_using)
+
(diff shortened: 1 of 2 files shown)


networkx/algorithms/minors.py
```diff
@@ -220,7 +220,27 @@ def quotient_graph(G, partition, edge_relation=None, node_data=None,
...
+
+    # If the user provided partition as a collection of sets. Then we
+    # need to check if partition covers all of G nodes. If the answer
+    # is 'No' then we need to prepare suitable subgraph view.
+    partition_nodes = set().union(*partition)
+    if len(partition_nodes) != len(G):
+        if create_using is None:
+            # This hack is necessary because we will
+            # be passing subgraph instance.
+            create_using = G
```
dschult: Instead of hacking create_using, could you fix the create_using treatment in _quotient_graph()?
We don't use ```type(create_using)()``` because it doesn't work with graph views (including subgraphs).

Please change line 247 in your file from

    H = type(create_using)() if create_using is not None else type(G)()

to 

    H = G.fresh_copy() if create_using is None else create_using.fresh_copy()

Thanks for helping me catch bugs!   :)

mddddd:
No problem :). Done.

PR #2779, opened 2017-11-30
Change how sparse6 tests filenames

### LINKED ISSUE #2777: Appveyor is not giving permission for write_sparse6 tests
I think the trouble is when we open a file and then pass the name of that file into ```write_sparse6```. The decorator tries to open the file again. That works on *nix systems, but not Windows. I'm not sure what changed last week to turn this on. But I'm hoping that passing in the file object instead of its name will do the trick.

dschult:
Use tempfile only to get a valid filename.
The contextmanager closes that file, making it
openable even in Windows. keep Appveyor from giving error.

Fixes #2777


dschult:
Now it looks like a Travis Error (unrelated to the Appveyor error)  is occurring for building docs on MacOS.

Issue #2787, opened 2017-12-06
Changing node-border color in draw_networkx_nodes


Zsailer:
I'd like to change the color of node borders using `draw_networkx_nodes`. 

While the function allows you to change the widths of these borders, it does not allow you to change the color. Instead, it sets the border color to the node face color. 

I'd like to add a keyword argument like `edgecolors` to `draw_networkx_nodes` that allows one to change these border colors. Something like this:

```python
nx.draw_networkx_nodes(G, pos={0:[0,0]}, 
                       node_size=5000, 
                       linewidths=5, 
                       edgecolors='k')
```

![node-border-color](https://user-images.githubusercontent.com/2791223/33676101-a0035f16-da69-11e7-8ac2-1804b09d22d9.png)

dschult:
Fixed in #2788

PR #2788, opened 2017-12-06
Control node-border color in draw_networkx_nodes

Zsailer:
Addresses issue #2787

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -283,6 +283,7 @@ def draw_networkx_nodes(G, pos,
                         vmax=None,
                         ax=None,
                         linewidths=None,
+                        edgecolors=None,
                         label=None,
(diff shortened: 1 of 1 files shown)


dschult:
See #2748 
Maybe this is the right short term fix for that issue as well...

Zsailer:
Yes, #2748 also works! :)

Though, I think `edgecolors` is a common enough action that it might be worth explicitly adding it to the list of kwargs in `draw_networkx_nodes`. 

Otherwise, a user has to know that networkx uses matplotlib's `scatter` function. Then, they have to look at their docs for the correct argument name.

Zsailer:
@dschult Oops, I misinterpreted what your comment! 😄  Sorry!

I agree. I think this might be a better short term solution while the ``**kwargs`` debate carries on (though I have some bias in this idea 😉 ).

PR #2790, opened 2017-12-11
Deleted a duplicated test_random_graph in bipartite.tests.test_genera…

atombrella:
…tors

I've seen more of such duplicated methods in the test-suite, the problem is that they're not identical like this one.

### DIFF
--- a/networkx/algorithms/approximation/tests/test_kcomponents.py
+++ b/networkx/algorithms/approximation/tests/test_kcomponents.py
@@ -251,7 +251,7 @@ class TestAntiGraph:
             assert_raises(nx.NetworkXError, A.neighbors, node)
             assert_raises(nx.NetworkXError, G.neighbors, node)
 
-    def test_degree(self):
+    def test_degree_thingraph(self):
         for G, A in self.GA:
             node = list(G.nodes())[0]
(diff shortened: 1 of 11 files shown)


dschult:
Ack -- Thanks for finding this.  What a mess...

It looks like the first removal ```test_random_graph``` should not be a removal, but a change of name for the second version of that function to ```test_random_digraph```

Thanks!

atombrella:
It's a great learning resource, the Gomory Hu implementation reads right out of the pseudo-code in Gusfield's article, so it's nice to contribute something, even if it's just cleaning up. I ran the analyzer in PyCharm (which also finds typos in the documentation of parameters for some functions) on the code. I'll list some of them (I've closed the session since), and I could try to clean up a bit of this. 

* test_function.test_neighbors is also declared twice
* `BaseMultiDiGraphTester` has all of `test_edges` and `test_out_edges` and `test_in_edges` declared twice.
* `readwrite.tests.test_edgelist.test_edgelist_digraph` but there's already a `test_edgelist_graph` in that class.

atombrella:
@dschult My branch is open for maintainers. I found these things with the inspection in Pycharm.

Issue #2793, opened 2017-12-12
(Digraph) object has no attribute '_adj' (with more info)


stablum:
I'm duplicating the ticket https://github.com/networkx/networkx/issues/2741 because I cannot re-open it myself.

While trying `copy.deepcopy` of a graph, I am encountering the following exception:
```
File "/home/francesco/.local/lib/python3.6/site-packages/networkx/classes/reportviews.py", line 894, in __setstate__
    self._adjdict = G._succ if hasattr(G, "succ") else G._adj
AttributeError: 'TrackGraph' object has no attribute '_adj'
```

where TrackGraph is a subclass of DiGraph:
```
class TrackGraph(nx.DiGraph):
    ...
```

Version is the following:

```
>>> nx.__version__
'2.0'
```

I cannot provide the full code because it's a big project, and I could not isolate the bug in a simpler case.
My hypothesis is that the order in which the unplicking of the attributes happens is relevant.
Apparently one of these attributes is an `OutEdgeView`. I cannot see which attribute because python's pickle is notoriously uninformative during debugging.

Anyways, I think that that attribute is being set (and hence `__setstate__` triggered) before _adj or other attributes are set. In that case the `OutEdgeView.__setstate__` is faced with an almost empty object, without _adj.

Here is a demonstration:
```
import pickle

class Bar(object):
    def __init__(self,foo,value):
        self.value = value
        self.foo = foo
    
    def __setstate__(self,state):
        self.value = state['value']
        print("__setstate__ value:",self.value)
        self.foo = state['foo']
        if 'a' in dir(self.foo):
            print(self.foo.a)
        else:
            print("self.foo does not have 'a'")
        if 'b' in dir(self.foo):
            print(self.foo.b)
        else:
            print("self.foo does not have 'b'")

    def __str__(self):
        return "Bar, value={}".format(self.value)

class Foo(object):
    def __init__(self):
        self.a = Bar(self,42)
        self.b = Bar(self,23)

def main():
    foo1 = Foo()
    with open('/tmp/foo','wb') as f:
        pickle.dump(foo1,f)
    print("just pickled. Now let's unpickle")
    with open('/tmp/foo','rb') as f:
        foo2 = pickle.load(f)
    print("just unpickled.",foo2)
    print("foo2.a",foo2.a)
    print("foo2.b",foo2.b)
    print("foo2.a.foo",foo2.a.foo)
    print("foo2.b.foo",foo2.b.foo)

if __name__=="__main__":
    main()
```
[...]

PR #2795, opened 2017-12-12
Attempt to tackle the unpickling problem. Ticket #2793  (2nd, fixed pull request)

stablum:
Tackling the unpickling problem in which the graph is being used before being completely unpickled. Instance variables have been replaced with @property getters. Ticket #2793


stablum:
sorry, still not the right pull request. closing.

PR #2796, opened 2017-12-12
Attempt to tackle the unpickling problem. Ticket #2793 (3rd, fixed pull request)

stablum:
attempt to tackle the unpickling problem in which the graph is being used before being completely unpickled. Instance variables have been replaced with @property getters. Ticket #2793


stablum:
I do not have enough knowledge of networkx to understand why those tests are failing. Maybe this pull request can be an inspiration to others to fix the bug.

dschult:
The underlying trouble is hopefully fixed by #3011
I'm going to close this but it could be reopened for more discussion if needed.

Issue #2799, opened 2017-12-15
complete_bipartite_graph does not check parameters right


dimpase:
Using complete_bipartite_graph (from version 2.0) from [Sagemath](http://www.sagemath.org/) shows to a bug in `complete_bipartite_graph()` - this won't happen if it did check parameters for being `int`
```
sage: import networkx
sage: networkx.complete_bipartite_graph(2,3).edges()
EdgeView([(0, 0), (0, 1), (0, 2), (1, 1), (1, 2)])
sage: networkx.complete_bipartite_graph(int(2),int(3)).edges()
EdgeView([(0, 2), (0, 3), (0, 4), (1, 2), (1, 3), (1, 4)])
```
It is due to Sagemath having its own integers, which are converted into something wrong by networkx (which is not so strange, as certainly it does not have to know anything about this type, but the fact that it happily outputs a graph which is not complete bipartite makes one wonder whether something is not right). 
```
sage: type(2)
<type 'sage.rings.integer.Integer'>
sage: type(int(2))
<type 'int'>
```

Note that in networkx 1.11 this issue does not arise, so this is a regression.

dimpase:
The fix is actually easy: instead of `isinstance(n2,int)` in line 67 of `algorithms/bipartite/generators.py` one should do
```
import numbers
if isinstance(n2,numbers.Integral):
```
to allow for using an `Integral` type other than `int` - which I think is a good idea generally.
I will do a pull request do change this.

PR #2800, opened 2017-12-15
use a generic Integral type for parameters check

dimpase:
fixes https://github.com/networkx/networkx/issues/2799
By right, the parameters to functions like complete_bipartite() need not be int, as long as they come from a correctly 
defined Integral type. This allows direct usage of Sagemath's Integers, for instance.

### DIFF
--- a/networkx/algorithms/bipartite/generators.py
+++ b/networkx/algorithms/bipartite/generators.py
@@ -10,6 +10,7 @@ Generators and functions for bipartite graphs.
 #    All rights reserved.
 #    BSD license.
 import math
+import numbers
 import random
 import networkx
 from functools import reduce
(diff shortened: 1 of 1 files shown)


dimpase:
Just [one of the many hosts on travis ci](https://travis-ci.org/networkx/networkx/jobs/316878165) reported errors, and they are all formatting, like these:
```
----------------------------------------------------------------------
File "/home/travis/venv/lib/python2.7/site-packages/networkx/convert_matrix.py", line 1053, in networkx.convert_matrix.to_numpy_array
Failed example:
    A
Expected:
    array([[ 1.]])
Got:
    array([[1.]])
```

dimpase:
as exactly the same Travis CI test fails for https://github.com/networkx/networkx/pull/2798, I assume it's something ongoing and not coming from this pull (or that) pull request.

hagberg:
As you pointed out those errors look like they come from a change in output text format for numpy arrays and matrices.  Your PR is good - thanks!

dimpase:
Thanks. I just noticed that the same pattern ought to be fixed in few more places, e.g. in `lollipop_graph()`.  You'll get another PR on this.

dimpase:
much less exotic Integrals may be found in numpy, e.g.
```
In [5]: isinstance(np.int8(7),int)
Out[5]: False
In [6]: isinstance(np.int8(7),numbers.Integral)
Out[6]: True
```
So it's easy to write tests with them, assuming numpy is a "standard" package.

Issue #2709, opened 2017-10-13
wrong normalization in betweenness_centrality(endpoints=True)


leonardomaccari:
I would expect that the normalized centrality should be in the [0,1] range, but when including the endpoints this does not happen. 

```
g = nx.star_graph(2)
nx.betweenness_centrality(g)
{0: 1.0, 1: 0.0, 2: 0.0}
nx.betweenness_centrality(g, endpoints=True)
{0: 3.0, 1: 2.0, 2: 2.0}
```
The documentation (and code) says that the normalization multiplies by 2/((n-1)(n-2)). I spent some time to understand this number, and i tracked down the following: 

In the original paper from Freeman (A set of measures of centrality based on betweenness, pag 38) he introduces a normalization by dividing per (n(n-1)*1/2)-(n-1) where:
 n(n-1)*1/2 : number of shortest paths in the undirected graph
 (n-1) : number of shortest paths ending/beginning in the node for which we compute centrality

this leads to division by (n*n - 3n + 2)*1/2 == (n-1)(n-2)*1/2 but, this implicitly removes the node we compute centrality on from the enpoints, as Freeman says. When centrality is computed with the endpoints, this term should be removed, and normalization should be applied simply with the potential number of paths: n(n-1)*1/2  

In general, normalization could be done with whatever constant, but:
* I think most people expect betweenness to be in [0,1]
* This was noted in the edge_betweenness_centrality() and corrected in #611, so right now edge and node centrality is normalized differently, which is pretty awkward.

So I suggest normalization is changed when betweenness is computed including the endpoints.

hagberg:
Thanks for tracking this down.   I agree with you that it is reasonable to normalize by the potential number of paths.  This should be a pretty easy fix.

manoelhortaribeiro:
Can I give it a try?

hagberg:
Yes, go for it.

bakerwho:
I'm interested in contributing and this does seem like it might have a simple fix. However, it looks @regstrtn has already tried a fix that has failed the Travis CI build. Any way I can help?

PR #2804, opened 2017-12-20
Changed scaling factor in betweenness.py (#2709)

regstrtn:
This is my first ever attempt at fixing a bug. The issue #2709 came across as simpler than the rest. 
I changed the _rescale function to take endpoints as an input parameter. 
If endpoints is true: 
scale = 1/(n*(n-1))
else:
scale = 1/((n-1)*(n-2))
Please review

### DIFF
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -120,7 +120,7 @@ def betweenness_centrality(G, k=None, normalized=True, weight=None,
             betweenness = _accumulate_basic(betweenness, S, P, sigma, s)
     # rescaling
     betweenness = _rescale(betweenness, len(G), normalized=normalized,
-                           directed=G.is_directed(), k=k)
+                           directed=G.is_directed(), k=k, endpoints=endpoints)
     return betweenness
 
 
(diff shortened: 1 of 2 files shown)


networkx/algorithms/centrality/betweenness.py
```diff
@@ -320,12 +320,15 @@ def _accumulate_edges(betweenness, S, P, sigma, s):
...
 
-def _rescale(betweenness, n, normalized, directed=False, k=None):
+def _rescale(betweenness, n, normalized, directed=False, k=None, endpoints=False):
     if normalized:
         if n <= 2:
             scale = None  # no normalization b=0 for all nodes
         else:
-            scale = 1.0 / ((n - 1) * (n - 2))
+            if endpoints:
+                scale = 1.0 / ((n) * (n - 1)) # If endpoints are included, change the scale factor to include these nodes
```
jfinkels: Let's move the comment to the line directly above, so that this line doesn't become too long (i.e., 80 characters or more).

jfinkels: Let's place the line `from __future__ import division` at the head of the file but after the module-level docstring (i.e. just above the line `from heapq import heappush, heappop`). See https://www.python.org/dev/peps/pep-0238/. I like to remove the `.0` in this case because it is an unnecessary artifact, but it's more of a matter of preference, so you don't have to do it if you prefer to minimize changes here.

The same comment applies to the line in the `else` branch as well.

networkx/algorithms/centrality/betweenness.py
```diff
@@ -320,12 +320,15 @@ def _accumulate_edges(betweenness, S, P, sigma, s):
     return betweenness
 
 
-def _rescale(betweenness, n, normalized, directed=False, k=None):
+def _rescale(betweenness, n, normalized, directed=False, k=None, endpoints=False):
     if normalized:
         if n <= 2:
```
jfinkels: Does the `n == 2` case matter when `endpoints` is `True`?

dschult:
This looks close to ready.  Can you make your changes consistent with PEP8 style guides. Search for PEP8 for the guidelines. You can install a pep8 checking program using ```pip install pep8``` to check your code.

The other nice features would be to add a test of the rescaling, and to check make the check for ```n==2``` case provide non-zero scaled values when ```endpoints is True```.

Thanks!

PR #2805, opened 2017-12-21
V1.11

djsvanwyk:


dschult:
We do NOT want to merge commits from v1.11 into master.
There is a history here where the branches diverged so that changes to the v1.11 branch were not compatible with v2.0. If there is something in particular you want to include in master, a separate PR introducing that feature to master is appropriate.

PR #2810, opened 2018-01-02
WIP: Legacy array printing for NumPy 1.14+

jarrodmillman:
https://github.com/numpy/numpy/blob/master/doc/release/1.14.0-notes.rst

This fixes a bunch of failures like:
```
File "/home/travis/venv/lib/python2.7/site-packages/networkx/convert_matrix.py", line 1053, in networkx.convert_matrix.to_numpy_array
Failed example:
    A
Expected:
    array([[ 1.]])
Got:
    array([[1.]])
```

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -1048,6 +1048,10 @@ def to_numpy_array(G, nodelist=None, dtype=None, order=None,
     resulting NumPy array can be modified as follows:
 
     >>> import numpy as np
+    >>> try:
+    ...    np.set_printoptions(legacy="1.13")
+    ... except TypeError:
+    ...    pass
     >>> G = nx.Graph([(1, 1)])
     >>> A = nx.to_numpy_array(G)
     >>> A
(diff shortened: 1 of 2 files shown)


jarrodmillman:
I first tried to put the following
```
# FIXME remove next 3 lines once 1.14 is minimum NumPy requirement
# and fix printing doctests to use new format.
from distutils.version import LooseVersion
if LooseVersion(np.__version__) >  LooseVersion("1.13.3"):
    np.set_printoptions(legacy="1.13")
```
after every
```
import numpy as np
```
But decided that since the only issues were with printing in doctests that using
```
# doctest: +NORMALIZE_WHITESPACE
```
after each printing of an array was a cleaner solution.

Unfortunately, I am still getting errors.  I believe these errors are not related to the changes in how NumPy prints arrays, but are due to changes between pydot 1.2.3 and pydot 1.2.4.  If that seems correct, I would like to merge this and then fix the pydot 1.2.4 issue.  I haven't looked into what it takes to fix the 1.2.4 (vs. 1.2.3 of pydot).

dschult:
This PR allowed all tests to pass for many days. But now it has changed again so that the error shows when the documentation build process tests the doctests.  In Travis it is the first config that produces the error in e.g. #2816 . Not sure why the legacy print options would not work in this environment.

PR #2814, opened 2018-01-06
Combine generator modules and tweak docs

dschult:
tree.py appended to trees.py
inverse_line.py appended to line.py

fixed algorithms.rst section for edge_augmentation


yaph:
I tried to call the ``line_graph`` method on a ``multigraph`` object which caused the exception that this method is not implemented for this type. I found that the `@not_implemented_for('multigraph')` was added in commit https://github.com/networkx/networkx/pull/2814/commits/0e215715bd295f76f14c648b1dae458fcfebe093. The commit message does not indicate this change and the doc string of the ``line_graph`` method says that it can be called with a multigraph object. The function works for multigraphs when the decorator is removed. Is this is a deliberate change?

dschult:
It looks like that was a mistake (or really, a poor review of an old commit which did not take into account changes to the code).  The short term fix is to call ```nx.generators.line._lg_undirected``` but a simple PR removing line 24 would allow the correct behavior.

PR #1259, opened 2014-10-02
Label propagation

aitoralmeida:
Added the label-propagation community detection as explained in #617 using Tyler Rush's code. Modified it to comply with the structure proposed in #951 (using yield and returning sets instead of lists or frozensets). Some minor fixes (now python supports using max() with sets) and style changes.

PR #2821, opened 2018-01-09
Label Propagation Community Detection

dschult:
Contributed by @aitoralmeida in #1259 
This resolves conflicts and updates docs and pep8


dschult:
The label_propagation_community module I'm trying to merge is failing tests, but only for the 32-bit version of python 3.4.4. Any hints on best way to test that locally on my 64 bit mac?

dschult:
The Label Propagation algorithm is not unique as it depends on the order that the ```colors``` are traversed. The article cited talks about this.  Thus we cannot expect a single outcome for the test of label_propagation_communities. I guess we will have to provide two ground truth answers and check if the result is one of those two. 

Note that this issue didn't occur in this special case in python2.7 because the order of dicts was deterministic and the same each test.  

@aitoralmeida do you have any better ideas for testing ```label_propagation_communities```? The biggest connected component sometimes shows up as a single community even though it is a barbell topology.

PR #2822, opened 2018-01-09
adding periodic moore neighbourhood generator

Florian-Roessler:
Adding to lattice generators one that produces a periodic moore neighbourhood
  
Feedback welcome, can work on a non-periodic addition too


Florian-Roessler:
sorry this was not the right branch.

Issue #2832, opened 2018-01-17
pydot read_dot doesn't strip quotes from node/edge/graph attributes


alubbock:
Some programs output dot files with quoted attributes, which are not stripped of their quotes when read using pydot (but are correctly stripped of quotes using pygraphviz).

Example:

```python
dot_data = StringIO("""
digraph G{
    "A_0" [shape="ellipse" style=filled];
}    
""")
G = nx.nx_pydot.read_dot(dot_data)
assert G.nodes(data=True)['A_0']['shape'] == 'ellipse'
```

PR #2833, opened 2018-01-17
pydot read_dot now strips quotes from attributes

alubbock:
.dot file node/edge/graph attributes which are double-quoted
are not stripped of their quotes when using `read_dot` via
pydot, but are stripped of their quotes when using pygraphviz.
This commit adds quote-stripping to `read_dot` when using pydot.

Fixes: #2832

### DIFF
--- a/networkx/drawing/nx_pydot.py
+++ b/networkx/drawing/nx_pydot.py
@@ -138,13 +138,14 @@ def from_pydot(P):
         n = p.get_name().strip('"')
         if n in ('node', 'graph', 'edge'):
             continue
-        N.add_node(n, **p.get_attributes())
+        N.add_node(n, **{k: v.strip('"')
+                         for k, v in p.get_attributes().items()})
 
     # add edges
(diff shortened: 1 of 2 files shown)


johnyf:
This modification relates to parsing with `pydot`. It may be a `pydot` issue (specifically an issue in the [module `dot_parser`](https://github.com/erocarrera/pydot/blob/d6ac9e9244d1a882103422ac2b35ceef96f5dfe3/dot_parser.py)), instead of a `networkx` issue.

dschult:
This strips ```"```
Do we also need to worry about ```'```?

@johnyf I agree -- It should be related to https://github.com/erocarrera/pydot/issues/66 or https://github.com/erocarrera/pydot/issues/72  over on pydot's repository. Let's get it fixed on pydot so that upstream code like ours doesn't have to worry about it. :)

johnyf:
Indeed, this issue is likely related to those `pydot` issues; I will try to address it within `pydot`.

PR #2841, opened 2018-01-22
Fixed strongly_connected_components [Issue 2831]

ArseniyAntonov:
Since we are using iterative method we spend too much time on
rechecking graph vertices that are already have been checked.
Removing vertices that already detected as strongly connected, from the
copy of the graph dramatically speeds up this method.

### DIFF
--- a/networkx/algorithms/components/strongly_connected.py
+++ b/networkx/algorithms/components/strongly_connected.py
@@ -84,8 +84,11 @@ def strongly_connected_components(G):
     lowlink = {}
     scc_found = {}
     scc_queue = []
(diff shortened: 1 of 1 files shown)


jfinkels:
While the running time is reduced, this implementation uses a good deal more memory than the previous implementation (a copy of the entire graph, plus the storage for an additional list of nodes). For large graphs this might be significant. It's probably worth updating the documentation to note the memory usage.

ArseniyAntonov:
@jfinkels please review, if you interested.

dschult:
See #2994

Issue #2710, opened 2017-10-13
local efficiency bug?


vlado-s:
Hi all,
I wrote my own routines for calculation of efficiency in weighted graphs. While testing them for un-weigted graph (all w=1) I discovered a difference in the values for local efficiency (global if fine). I believe there is a logical bug in the nx.local_efficiency(G) source which is:

return sum(global_efficiency(nx.ego_graph(G, v)) for v in G) / len(G)

however should be:

return sum(global_efficiency(nx.ego_graph(G, v, center=False)) for v in G) / len(G)

I conclude so based on the original paper by Latora & Marchiori (DOI: 10.1103/PhysRevLett.87.198701), where they state (page 2): "Since _E_ is also defined for a disconnected graph we can characterize the local properties of **G** by evaluating for each vertex _i_ the efficiency of **G**_i_ , the subgraph of the neighbors of _i_." Thus _i_ should not be in the ego_graph(G). How do you see it, am I right? Please let me know. Thank's a lot!
Cheers,
Vlado Sladek

dschult:
I interpret the paper as you do. It also matches how Wikipedia defines local efficiency.
Would you like to create a pull request with this change and a test that it is correct?
Or would you prefer that I do that.

dschult:
Take a look at #2233 where this (mis)behavior was implemented apparently to avoid an error raised when the center node is not present.

vlado-s:
Hi, thanks' for your response.
What exactly does the creation of a pull include? What should the testing include? I tested it on one example where I coded it both from scratch & with the modification as written above, and it is then identical to all digits that are displayed with print. Sorry for my ignorance, but are you from the NetworkX developers - i.e. is it easier for you to make the pull? In principle all that needed to be done is what I wrote in the first message, so no big coding is involved. Obviously this works for unweighted graphs (which is what's in NetworkX 2 anyway).

As to the issue https://github.com/networkx/networkx/issues/2233 : seems like the creation with G.subgraph(G[v]) is somewhat cumbersome. If _v_ is a list, then it would be sufficient to write G.subgraph(v). That is not explained there. But later on _v_ is used as a node index in the nx.ego_graph(G, v) command. I used the G.subgraph(list_of_v_neighbours) as well as ego_graph(G, v, center=False) and both yield identical results. The former was my initial idea, just because I had the neighbour list stored from previous steps anyway. Of course _list_of_v_neighbours_ did not include _v_.

dschult:
[...]

PR #2843, opened 2018-01-23
Fixes issue #2710

bakerwho:
Local efficiency should be computed on the subgraphs induced by *neighbors* of each node (ref: **Efficient Behavior of Small-World Networks**, Latora et al) - however, the node itself should not be included in this subgraph. Hence, `center = False`.

Thanks to @dschult and Vlado S.

### DIFF
--- a/networkx/algorithms/efficiency.py
+++ b/networkx/algorithms/efficiency.py
@@ -147,4 +147,4 @@ def local_efficiency(G):
 
     """
     # TODO This summation can be trivially parallelized.
-    return sum(global_efficiency(nx.ego_graph(G, v)) for v in G) / len(G)
+    return sum(global_efficiency(nx.ego_graph(G, v, center = False)) for v in G) / len(G)
--- a/networkx/algorithms/efficiency.py
+++ b/networkx/algorithms/efficiency.py
@@ -147,5 +147,5 @@ def local_efficiency(G):
 
     """
     # TODO This summation can be trivially parallelized.
-    efficiency_list = [global_efficiency(G.subgraph(G[v])) for v in G]
+    efficiency_list = (global_efficiency(G.subgraph(G[v])) for v in G)
     return sum(efficiency_list) / len(G)
--- a/networkx/algorithms/efficiency.py
+++ b/networkx/algorithms/efficiency.py
@@ -147,4 +147,5 @@ def local_efficiency(G):
 
     """
     # TODO This summation can be trivially parallelized.
-    return sum(global_efficiency(nx.ego_graph(G, v, center = False)) for v in G) / len(G)
+    efficiency_list = [global_efficiency(G.subgraph(G[v])) for v in G]
+    return sum(efficiency_list) / len(G)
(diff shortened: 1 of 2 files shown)


MridulS:
@bakerwho the tests are going to fail. Can you update the tests too?

bakerwho:
Fixed the two errors that were being raised in [test_efficiency.py](https://github.com/networkx/networkx/blob/master/networkx/algorithms/tests/test_efficiency.py)
 with new commits.

In `test_local_efficiency_complete_graph()`, the case for a 2-node graph was incorrect, as the resulting induced subgraphs would have 0 nodes (hence, 0 efficiency).

In `test_using_ego_graph()`, I have changed the comment to reference issue #2710, besides updating the expected value of the computed efficiency.

Should be good to go?

networkx/algorithms/efficiency.py
```diff
@@ -147,4 +147,4 @@ def local_efficiency(G):
 
     """
     # TODO This summation can be trivially parallelized.
-    return sum(global_efficiency(nx.ego_graph(G, v)) for v in G) / len(G)
+    return sum(global_efficiency(nx.ego_graph(G, v, center = False)) for v in G) / len(G)
```
dschult: ```ego_graph``` is a lot of overhead for this simple case. Could it be replaced with:

    return sum(global_efficiency(G.subgraph(G[v])) / len(G)

networkx/algorithms/tests/test_efficiency.py
```diff
@@ -53,15 +53,15 @@ def test_global_efficiency_complete_graph(self):
 
     def test_local_efficiency_complete_graph(self):
         """
-        Test that the local efficiency for a complete graph should be one.
+        Test that the local efficiency for a complete graph with at least 3 nodes should be one. For a graph with only 2 nodes, the induced subgraph has no edges.
```
dschult: This line should be wrapped so that [the PEP8 style guide](https://www.python.org/dev/peps/pep-0008/) of <80 characters/line  holds. There may be other pep8 considerations. 

You can read the style guide and/or install the pep8 style checker using ```pip install pep8```

networkx/algorithms/efficiency.py
bakerwho: Yep, makes sense.

networkx/algorithms/tests/test_efficiency.py
bakerwho: Changed.

networkx/algorithms/efficiency.py
```diff
@@ -147,4 +147,5 @@ def local_efficiency(G):
 
     """
     # TODO This summation can be trivially parallelized.
-    return sum(global_efficiency(nx.ego_graph(G, v)) for v in G) / len(G)
+    efficiency_list = [global_efficiency(G.subgraph(G[v])) for v in G]
```
dschult: This doesn't need to be a list comprehension.
It could be a generator expression and use less memory [see generator expressions](https://docs.python.org/3.6/howto/functional.html#generator-expressions-and-list-comprehensions).

Just replace ```[ ]``` with ```( )``` to make it an iterator instead of a list.

bakerwho: Making the change, thanks!

PR #2857, opened 2018-01-30
Added clarifying comment about relabelling nodes in situations where …

lewisrobbins:
…the order of nodes is important

### DIFF
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -192,7 +192,9 @@ def convert_node_labels_to_integers(G, first_label=0, ordering="default",
     -----
     Node and edge attribute data are copied to the new (relabeled) graph.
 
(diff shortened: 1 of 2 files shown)


networkx/relabel.py
```diff
@@ -192,6 +192,8 @@ def convert_node_labels_to_integers(G, first_label=0, ordering="default",
     -----
     Node and edge attribute data are copied to the new (relabeled) graph.
 
+    There is no guarantee that the relabelling of nodes to integers will give the same two integers for two (even identical graphs). Use the `ordering` argument to try to preserve the order.
```
dschult: Please make this follow PEP8 style guidelines by wrapping at <80 chars.

Thanks!

PR #2861, opened 2018-01-31
Replaced use of fractions.gcd with math.gcd.

atombrella:
`generate_unique_node` and `defaultdict` are unused imports.


dschult:
This adds a dependence on ```six``` and we'd like to keep dependencies to a minimum.
Also, its not clear that math.gcd is better than fractions.gcd.  I'd rather not add code that depends on python version.

atombrella:
`fractions.gcd` is deprecated on Python 3 in favor of `math.gcd`. You can do a big clean-up in 2020.

PR #2870, opened 2018-02-06
Floyd Warshall rebuild path

msramalho:
In this PR I have implemented a simple function to reconstruct the path from a node `source` to a node `target` with the result of executing: `floyd_warshall_predecessor_and_distance`

The code is being properly tested:
 . Expected behavior
 . Abnormal behavior: `source == target`
 . KeyError exception (this test is optional)

### DIFF
--- a/networkx/algorithms/shortest_paths/dense.py
+++ b/networkx/algorithms/shortest_paths/dense.py
@@ -14,6 +14,7 @@ import networkx as nx
 
 __all__ = ['floyd_warshall',
            'floyd_warshall_predecessor_and_distance',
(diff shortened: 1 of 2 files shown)


dschult:
This looks good.  Any other work to be done here?

Also, are there other places in the code that could use this (or a similar) utility?

msramalho:
It could, perhaps, be included in the `floyd_warshall_predecessor_and_distance` description, for one.

As for other places... all I can think of is completing any Floyd-Warshall examples that do not demonstrate how to get the path from a to b after running this algorithm.

PR #2873, opened 2018-02-12
Curved lines in loops for digraphs

ferdavid1:
added the boolean parameter 'draw_loop' to nx.draw() that uses simple_cycles() to find cycles in a directed graph, then makes curves the edges that make up the cycle, for aesthetic purposes. first PR, let me know how I can improve! 
Example:
circ

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -688,41 +688,46 @@ def draw_networkx_edges(G, pos,
               ax.add_patch(arrow)
             else:
               cycle_edges = list(nx.simple_cycles(G))
-              loop_points = [(x,y) for x,y in cycle_edges]
-              for x in loop_points:
-                to_remove = list(G.edges())
-                half_cycle = to_remove.index(x)
-                second_half = to_remove.index(x[::-1])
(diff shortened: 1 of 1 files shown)


ferdavid1:
Not sure why codecov/patch is failing, didn't change anything other than nx_pylab.py

dschult:
That just means that the tests only covered 88% of the code (well...  88% of the code was actually run while doing the tests).  The drawing routines often have lower coverage than the rest of the codebase. I guess you could add tests for "corner cases" to make sure more code is reached during the tests. But I wouldn't worry about it too much.  We can review and merge even with the coverage goals not passing.

ferdavid1:
@dschult that makes sense, thank you!

ferdavid1:
@dschult excess files removed from this one too.

dschult:
The drawloop PR now has the Betti number change in it.

ferdavid1:
@dschult thanks, fixed part 2.

dschult:
This needs tests and thoughts of how universal the choice of angles on the loop work.

dschult:
Does #3307 make this obsolete/unnecessary?  If not, open it up again.

PR #2874, opened 2018-02-13
Added dynamic graph attribute field `start` into `node element` in gexf utils.

vanam:
Hi,
I am sending a pull request which adds dynamic graph attribute field `start`/`end` into `node` element in gexf utils and tests that.

(Based on https://github.com/networkx/networkx/pull/1797)


vanam:
I don't understand why the tests are failing on CI... It looks like it runs different (old) code...

PR #2877, opened 2018-02-15
Removed the import star pattern.

atombrella:
This is just a few files. If it's well-received, I don't mind working more on this.

### DIFF
--- a/networkx/algorithms/tests/test_cluster.py
+++ b/networkx/algorithms/tests/test_cluster.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
-from nose.tools import *
+from nose.tools import assert_equal, assert_almost_equal
+
 import networkx as nx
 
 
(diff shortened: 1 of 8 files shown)


networkx/algorithms/tests/test_cluster.py
```diff
@@ -130,17 +131,17 @@ def test_k5(self):
         G.remove_edge(1, 2)
         assert_equal(nx.transitivity(G), 0.875)
 
-    # def test_clustering_transitivity(self):
```
atombrella: I wonder why this was commented out.

atombrella:
I see both styles used throughout the project.

dschult:
Let's try to focus on network algorithms rather than code style. It's fine to fix style if you are working on that code anyway, but we don't have the reviewer time to check these kinds of fixes and the potential problems they might create.  Thanks for your efforts.

PR #2878, opened 2018-02-16
WIP: Added edge_bunches iterator (also changed dicts titled 'seen' to sets)

wardbradt:


dschult:
Isn't ```edge_bunches``` just ```EdgeDataView```.  

That is, can't you get this with ```G.edges.data()```  or ```G.edges(data=True)```

PR #2879, opened 2018-02-18
link DOIs to preferred resolver

katrinleinweber:
Hello!

The DOI foundation recommends [this new, secure resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). This PR implements that recommendation by changing all static DOI links to `https://doi.org/`, see diff.

If you agree and merge, I'll also update other repos, website, docu, etc. in the same manner.

Cheers :-)


dschult:
Thanks!

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #2883, opened 2018-02-20
Fix #2880: Allow line_graph to apply to multigraph

### LINKED ISSUE #2880: Allow line_graph to apply to multigraph
The code is written for multigraphs and graphs, but recently put an errant restriction on multigraphs.
Line 24 of line.py

See #2814 

Short term fix is to call ```nx.generators.line._lg_undirected```

yaph:
Removed not_implemented_for decorator from line_graph function. See #2814 and #2880.


dschult:
Thank you!!

yaph:
Happy I could contribute!

PR #2888, opened 2018-02-23
Implementing Betti numbers of Graphs

ferdavid1:
Function found in algorithms.components.connected.py
In Topological Graph Theory:
the 0th betti number = # of edges
the 1st betti number = # of edges - # of nodes + # of connected components

function betti(Graph object): returns (0th, 1st) betti numbers

### DIFF
--- a/networkx/algorithms/components/connected.py
+++ b/networkx/algorithms/components/connected.py
@@ -21,6 +21,7 @@ __all__ = [
     'connected_component_subgraphs',
     'is_connected',
     'node_connected_component',
+    'betti'
 ]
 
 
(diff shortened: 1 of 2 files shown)


dschult:
This PR has many many changes that are not at all related to what you propose in the description.
I think your other recent PR has many of the same changes as well.  This makes it very hard to review the code. If you want to make spelling corrections, put those in their own PR and describe what it is. Then a PR like this one can involve few changes in a few files and will be much more likely to get reviewed.   Also, Thanks for the contributions!

ferdavid1:
@dschult I was wondering about that. I'll figure out where those changes came from and reduce the difference. Thank you!

ferdavid1:
@dschult fixed.

dschult:
I'm going to close this pull request because:
- once someone knows the Betti numbers for graph theory, the computation is easily constructed.
- For someone who doesn't know the Betti numbers for graph theory, there isn't enough other connected functionality in networkx for this to fit inside a context which leads to the Betti Numbers.

Please comment if these ideas are not correct.  And Thanks for the Pull Request!

PR #2892, opened 2018-03-06
Updated migration guide and release notes.

NeilGirdhar:
Updated migration guide and release notes for topological sort changes.

### DIFF
--- a/doc/release/migration_guide_from_1.x_to_2.0.rst
+++ b/doc/release/migration_guide_from_1.x_to_2.0.rst
@@ -15,12 +15,12 @@ work with both NetworkX v1.x and v2.0.
 We have made some major changes to the methods in the Multi/Di/Graph classes.
 The methods changed are explained with examples below.
 
-With the release of NetworkX 2.0 we are moving to a view/iterable reporting API.
+With the release of NetworkX 2.0 we are moving to a view/iterator reporting API.
 We have changed many methods from reporting lists or dicts to iterating over
 the information. Most of the changes in this regard are in the base classes.
 Methods that used to return containers now return views (inspired from
 `dictionary views <https://docs.python.org/3/library/stdtypes.html#dict-views>`_
-in Python) and methods that returned an iterable have been removed.
+in Python) and methods that returned an iterator have been removed.
 The methods which create new graphs have changed in the depth of data copying.
 ``G.subgraph``/``edge_subgraph``/``reverse``/``to_directed``/``to_undirected``
 are affected.  Many now have options for view creation instead of copying data.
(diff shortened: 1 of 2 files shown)


doc/release/migration_guide_from_1.x_to_2.0.rst
```diff
@@ -15,12 +15,12 @@ work with both NetworkX v1.x and v2.0.
...
 
-With the release of NetworkX 2.0 we are moving to a view/iterator reporting API.
-We have moved many methods from reporting lists or dicts to iterating over
+With the release of NetworkX 2.0 we are moving to a view/iterable reporting API.
+We have changed many methods from reporting lists or dicts to iterating over
 the information. Most of the changes in this regard are in the base classes.
 Methods that used to return containers now return views (inspired from
 `dictionary views <https://docs.python.org/3/library/stdtypes.html#dict-views>`_
-in Python) and methods that returned iterators have been removed.
+in Python) and methods that returned an iterable have been removed.
```
dschult: Why change ```iterator``` to ```iterable```?  Lots of things are iterable... A list is iterable....  
The API used to return an iterator, e.g. ```G.nodes_iter()``` similar to the old ```dict.iterkeys()```.
I think we should leave iterator in most of these places, but I didn't check carefully. Maybe it should be replaced in some of these spots.

doc/release/migration_guide_from_1.x_to_2.0.rst
```diff
@@ -237,6 +237,18 @@ because it may be a different graph type (directed/undirected) than the view.
...
 
+``topolgical_sort``  no longer accepts ``reverse`` or ``nbunch`` arguments.
+If ``nbunch`` was a single node source, then the same effect can now be achieved
+using the ``subgraph`` operator:
+
+    >>> nx.topological_sort(G.subgraph(nx.descendants(G, nbunch)))
+    
+To achieve a reverse topological sort, the output should be converted to a list:
+
+    >>>> reversed(list(nx.topological_sort(G)
```
dschult: missing parens at end?

dschult:
Thanks!

doc/release/migration_guide_from_1.x_to_2.0.rst
NeilGirdhar: By that logic, you might as well say that you returned generators since that's the most specific ABC.  I just figured that most people would only expect a method like `nodes_iter()` to be iterable.  They wouldn't expect it to necessarily be an iterator.  But okay, we can revert this.

doc/release/migration_guide_from_1.x_to_2.0.rst
NeilGirdhar: yes, thanks, I'll fix.

PR #2905, opened 2018-03-15
Add stochastic block models to generators

jg-you:
This is fairly straightforward PR, I've implemented the classical "binomial" SBM in `generators/community.py`.

There's a sparse switch that will create edges by jumping ahead by a random number of  non-edges,
drawn from a geometric distribution. This is the default. In the dense case, I use a Bernoulli test on each edges.

Note that with this PR,  `random_partition_graph` could be refactored as a call to `stochastic_block_model` (as well as `gnp_random_graph` and other Erdos Renyi type graphs).

### DIFF
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -223,64 +223,14 @@ def random_partition_graph(sizes, p_in, p_out, seed=None, directed=False):
     if not 0.0 <= p_out <= 1.0:
         raise nx.NetworkXError("p_out must be in [0,1]")
 
-    if directed:
-        G = nx.DiGraph()
-    else:
-        G = nx.Graph()
-    G.graph['partition'] = []
-    n = sum(sizes)
-    G.add_nodes_from(range(n))
-    # start with len(sizes) groups of gnp random graphs with parameter p_in
-    # graphs are unioned together with node labels starting at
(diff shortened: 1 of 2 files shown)


dschult:
This PR adds Stochastic Block Model generator. 

It needs to treat ```seed``` as a rng-or-int to match the new way to handle seeds in other generators.
It needs tests (start with all the exceptions that should be raised -- then some sort of smoke test to make sure it actually creates something reasonable -- and maybe a couple with seed set so it is deterministic and we check that it creates the right thing.)
You should rebase to avoid the conflicts which came because another method was added in this file.

dschult:
I'm going to leave the seed argument as it is until we figure out choice of random_state in #1764.
Other functions in this module will have to be changed in the same way as this new function.
I think this is ready for merging.  Comments?

dschult:
Yes -- can you take a shot at that?  Thanks--

jg-you:
The CI failure comes from float precision in the examples. How is this best addressed?

dschult:
You can handle floats in a doctest (code in the docstring) using ```round(expression, 5)``` where the number of digits can be changed from 5 to whatever you want. There is one example in networkx at networkx/algorithms/bipartite/cluster.py  bascially ```print(round(v[1]['density'], 5))```

jg-you:
All done.

PR #2907, opened 2018-03-19
Fixing typo in bridges module

Yomguithereal:
Just fixing a tiny typo.


dschult:
That ```iff``` stands for "if and only if", which is an abbreviation common in math theorems and proofs.
Could you replace the ```iff``` with ```if and only if```?  Thanks!

Yomguithereal:
Oh. I did not know that. Good to know :). I will update the PR then if you feel it still make sense.

Yomguithereal:
Done. I broke the line also not to overcome 80 chars.

dschult:
Excellent -- Thanks!

PR #2912, opened 2018-03-21
Avoid loosing the order of the nodes in subraphs

### LINKED ISSUE #2911: `OrderedGraph.subgraph` does not maintain the order of the nodes
A subgraph built from a `OrderedGraph` should keep the order of the nodes, yet nodes in the subgraph are neither in the order of the initial graph, nor in the order of the selection. The issue can be seen from the following snippet:

```python
graph = nx.OrderedGraph()
nodes = list(range(10))
random.shuffle(nodes)
graph.add_nodes_from(nodes)  # key order is (7, 2, 1, 9, 0, 8, 6, 4, 3, 5)

# We create a selection in the same order as the initial graph keys
to_keep = [key for key in graph if key % 2 == 0]  # [2, 0, 8, 6, 4]
subgraph = graph.subgraph(to_keep)  # (0, 2, 4, 6, 8)

# We create a selection in a different order
subgraph = graph.subgraph([5, 3, 1])  # (1, 3, 5)
```

From what I see, the issue is due to `Graph.subgraph` passing the selection to `nx.filters.show_nodes` that transforms it to a set. The nodes in the subgraph are then in the order of the set; as sets do not preserve the order, the order of the initial graph if not preserved.

jbarnoud:
* Make sure the tests catch the issue. The order of the nodes was
  already tested for subgraphs of OrderGraph. However, the nodes in the
  reference were sorted integer; as the faulty behavior was happened to
  be sorting the keys, the issue was not reported.
* Fix `classes.filters.show_nodes` to keep the order of the selection.

Fixes #2911

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -117,6 +117,7 @@ is partially historical, and now, mostly arbitrary.
 - Lewis Robbins
 - Mads Jensen, Github: `atombrella <https://github.com/atombrella>`_
 - Edward L. Platt, `elplatt <https://elplatt.com>`_
+- Jonathan Barnoud, `jbarnoud <https://github.com/jbarnoud>`_
 
 Support
 -------
--- a/networkx/classes/filters.py
+++ b/networkx/classes/filters.py
@@ -53,7 +53,15 @@ def hide_multiedges(edges):
 # write show_nodes as a class to make SubGraph pickleable
 class show_nodes(object):
     def __init__(self, nodes):
-        self.nodes = set(nodes)
+        # Remove duplicates in the selection,
+        # but keep the order in case it matters.
+        seen = set()
+        unique_nodes = []
+        for node in nodes:
+            if node not in seen:
+                unique_nodes.append(node)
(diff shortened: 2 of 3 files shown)


dschult:
Doesn't making ```show_nodes``` a tuple increase search time for ```n in show_nodes``` dramatically?
Most of the SubGraphViews performance is exactly these type of lookups.  In there a way to store the nodes as a set for lookup, and yet as an ordered list for reporting?  I know some recent Python versions have offered that kind of set/dict structure. I wonder if someone has backported it so that it is available for older versions of Python...

jbarnoud:
I did not think about performances, indeed. There are implementations for an ordered set on pypi. Most of them are implementing a snippet from Raymond Hettinger: <http://code.activestate.com/recipes/576694/>, <http://code.activestate.com/recipes/576696/>.

Adding a dependency for that may be overkill, but the snippet is licensed under the MIT license, so it should be possible to distribute it somewhere in networkx as a utility.

dschult:
See #2985

PR #2919, opened 2018-04-02
Fixed key error triggered when the labels dict contains nodes not in pos dict

austinlostinboston:
The fix required only a single line (networkx.drawing.nx_pylab.py - line 793.

Before, all labels would be looped over and assumed to be in pos (the position dict created during layout).
```python
for n, label in labels.items():
    (x, y) = pos[n] ## Throws keyerror if n is not in pos
```
I believe that this is an unnecessary assumption that places an unintended restriction on the labels dict. The graph should make no requirements about what's in or not in the labels dictionary. It should just simply add labels that match nodes in the graph. Therefore, I added a key check before the lookup in pos.

```python
for n, label in labels.items():
    if n in pos: ## This checks that the current label node, is in pos before looking it up
        (x, y) = pos[n]
```

I don't believe this will impact runtime since it only add an O(1) lookup to the process.


austinlostinboston:
Also, I apologize about all of the .pyc files that somehow came with the commit. I checked the .gitignore, and .pyc files are specificed at the top of the list. Also, I made sure to only add the file that I changed before committing. Do you have any ideas why these were created? If this messes  up the pull request, feel free to reject it and I can try again.

dschult:
Thanks for this -- it looks fine for inclusion, but the .pyc files will mess up the repository.
Can you try reverting and recommitting (--force will be needed) to your ```issue-2917-label-keyerror``` branch? Or start the PR over again?
Thanks!

dschult:
I'm finally looking at this again. Sorry for the delay.

Eliminating the assignment of ```(x,y)``` is insufficient to avoid the error: x and y are needed later.  Instead, we also need to avoid creating any label for nodes not in pos because we don't have a position for the label.

This means we silently leave out labels that are not in ```pos```. I think that could lead to surprises where labels that should be shown are not -- without any error message. The current behavior avoids surprises because the KeyError clearly flags the node as not in the dict ```pos```.  I prefer to avoid surprises by raising KeyError on potential mismatches of dicts.

If you want to use an uber-dict of labels with more keys than are in the Graph, I suggest the following and will put this in the docstring if it is deemed OK for these functions (edge_labels has the same problem).

    nx.draw_networkx_labels(G, pos, {k:v for k,v in labels.items() if k in pos})

PR #2936, opened 2018-04-11
Check if attribute is str like in writing to Pajek Fixes #2933

### LINKED ISSUE #2933: Write pajek error
When rewriting a network who has been read from pajek there is an error:
'float' object has no attribute 'strip'
apparently coordinates (x,y) don't get translated back to string
(Networkx 2.1)

wuhaochen:
Fixes #2933

### DIFF
--- a/networkx/readwrite/pajek.py
+++ b/networkx/readwrite/pajek.py
@@ -63,7 +63,7 @@ def generate_pajek(G):
         shape = na.get('shape', 'ellipse')
         s = ' '.join(map(make_qstr, (id, n, x, y, shape)))
         for k, v in na.items():
-            if v.strip() != '':
+            if is_string_like(v) and v.strip() != '':
                 s += ' %s %s' % (make_qstr(k), make_qstr(v))
         yield s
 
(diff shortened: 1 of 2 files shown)


wuhaochen:
There is a caveat in this version. It doesn't support writing non-string attribute to pajek, which is made consistent with the read function since it reads all attributes as strings. Alternatively we may convert non-string attributes to strings and write it to the files, but then when it is read back, attributes may not be the same.

I'm not sure how types are supported in pajek format at this time. We might need a parsing module in the read function if types are supposed to be supported.

rmbw51:
it happens with the "coordinates". actually the function seems to attempt a conversion to string, but x & y don't get converted...

dschult:
I'm +1 for the fix with only writing attributes that are strings -- option 1)

I'm not sure what the comment about "coordinates" meant -- does it make one option better than another?  Anybody else have input on which option is preferred?

A simple example of what this is fixing is in #2945

rmbw51:
yes, agree on only writing attributes that are strings. 

coordinates: in Pajek the x and y are used to draw the network. they are the coordinates of the nodes in the drawing window, so typically useless when writing from Networkx (in Pajek one would redraw...)

PR #2939, opened 2018-04-13
Add barycenter algorithm

wkschwartz:
Adds a `barycenter` function which returns a subgraph induced by vertices minimizing the sum of distances to all other vertices.

I had written this for another project, so there will be some stylistic things you'll want to clean up. Also I've never used nosetest so instead of putting my tests in with the other tests for `distance_measures`, I just copied and pasted my Python `unittest`-based module.

Thanks for any feedback, though I'd especially love it if someone else could take care of converting to nosetest.


dschult:
It looks like you've got an off-by-factor-of-two error in the tests (or in the code).

wkschwartz:
The math was right. The code was wrong. 61e1bdc371 fixes this. All three tests should pass now.

dschult:
Looks like a unittest issue with different versions of unittest. 
Might be better to switch to the nose paradigm which you asked for help with. I'll take a look to see how easy that would be. (probably easy).

dschult:
I have some questions about barycenters generally that relate to this code:

I can't find this approach to barycenter online. Is there another term that is used for this calculation that has a larger online presence? Is there a literature that uses it? Is it covered by more than West's book?

Why is the barycenter here the subgraph induced by the nodes instead of just the nodes? For example, the center of a graph is the set of nodes with some property. Would it make sense to have this function return the set of nodes.   The user would certainly be able to convert to a subgraph. (As is, the user can easily get the nodes from the subgraph too, but at the cost of creating the subgraph they might not need.)

Thanks -- and sorry for the delay in getting to this.

wkschwartz:
Sorry for the delay!

I have only seen _barycenter_ in West; he also calls it the _median_. The term _barycentricity_ is mine as far as I know, and purely for the convenience of having a name for the objective function. I'm not married to the term. For the fact that the barycenter of a tree is a single vertex of an edge, West cites
> Jordan C., Sur les assemblages de lignes. _J. Reine Angew. Math._ **70** (1869), 185-190.

but I don't read French, so I haven't bothered to follow that up. I can't say I've done much further literature review.

If the barycenter were defined to be a set of nodes rather than the subgraph induced by those nodes, then Jordan theorem wouldn't hold.

On page 72 of West (defn. 2.1.12), the _center_ of a graph is defined as the subgraph induced by the vertices of minimum eccentricity. In R. Diestel's _Graph Theory_, 4th ed., ch. 1, _central_ is defined for vertices but not for induced subgraphs.

I guess my point here is that whether the barycenter is a set of vertices or a subgraph induced by a set of vertices is a matter of convenience. We could even continue taking the definition to refer to an induced subgraph and return the inducing set of nodes anyway purely as an optimization.

I've set "allow commits from maintainers", so feel free to make the changes yourself. As you can see, I can be slow at these things. (However, I'm about to push a couple of small changes I noticed just now.)

wkschwartz:
I rebased onto master and cleaned up a bunch of docs. 

One last thing that occurs to me: `eccentricity` takes an argument of all pair shortest path lengths. Should `barycenter` also take this, for uniformity of the API?

dschult:
If ```barycenter``` can use a dict of shortest path lengths effectively (which it looks like it could) then it'd be cool to unify the API and reduce time for those who already computed that dict.

This looks good. I've shifted it to milestone v2.3 and will work to get it merged soon.
Thanks!

wkschwartz:
I had some free time and decided just to finish this. I've done everything we've discussed, including switching to returning nodes (unifying the API with `center` and `periphery`) and accepting the `sp` argument (unifying the API with `eccentricity`).

I moved the tests from their own file to `test_distance_measures.py`. I'll leave it to you to switch the test code from unittest to nose — I don't know the nose package and am deathly afraid of the `import *` at the top of the file. For what it's worth,
```
pip install nose
python -m nose networkx/algorithms/tests/test_distance_measures.py
```
successfully ran 17 tests.

wkschwartz:
I can't figure out why Travis is failing. It's in the doctests, apparently unrelated to my PR. The same error shows up in the previous build [here](https://travis-ci.org/networkx/networkx/jobs/461613413#L5490) without failure as in the current failing one [here](https://travis-ci.org/networkx/networkx/jobs/461614401#L5496).

Anyway, it's evidently not failing because of anything in this PR.

dschult:
It is failing because of different printing styles in numpy across versions. We've had that trouble before. Probably Travis changed the version of numpy.  Anyway, as you say -- it is not part of this PR.

PR #2944, opened 2018-04-16
Fix compatibility issue with new version of scipy.

wuhaochen:
From Scipy 1.1.0, `asformat` will throw a ValueError instead of an AttributeError if the format if not recognized. For backward compatibility, both are captured now.


dschult:
Thank you

Issue #2935, opened 2018-04-11
Treewidth Heuristic


Johannes-S:
We are a group of 8 students working on a project to improve graph libraries.
We first plan on implementing heuristics for an upper bound on the treewidth.
Many NP-complete problems can be solved efficiently if the treewidth is bound.
We are implementing a minimum degree based heuristic and a minimum fill-in heuristic.
They are described in section 4.3 in the paper linked below.
We can submit a pull-request probably within the next 2 weeks.

[https://dspace.library.uu.nl/bitstream/handle/1874/22171/2005-018.pdf](https://dspace.library.uu.nl/bitstream/handle/1874/22171/2005-018.pdf)

dschult:
Fixed in #2951

PR #2951, opened 2018-04-24
Treewidth and tree decomposition heuristics (#2935)

KaneTW:
See #2935 for details.

 We implemented two heuristics for an upper bound on the treewidth: a minimum degree based heuristic and a minimum fill-in heuristic.
They are described in section 4.3 in this paper:
https://dspace.library.uu.nl/bitstream/handle/1874/22171/2005-018.pdf

The implementation is finished and we would like to get your feedback.
We will complete the documentation in the next few days.

Some points we are unsure about:
- Internally we used a custom graph structure (dict-of-sets) because this turned out to be much faster on our benchmark. We can change this to the default nx.Graph.
- To enable different heuristics, we pass the heuristic as a class. E.g. the min-degree has internal state (keeps a heap), thus it couldn't simply be a function. We could change the heuristics so they don't need state, but the min-degree-heuristic would be slower. 
  - Currently the heuristic classes are iterators, but that might make it unclear that internal state can change. For clarity we could use explicit functions.

# Performance results

<table>
  <tr>
    <td>heuristic</td>
    <td>node search</td>
    <td>graph structure</td>
    <td>graph size</td>
    <td>runtime</td>
  </tr>
  <tr>
    <td>min-degree</td>
    <td>heap</td>
    <td>custom</td>
    <td><= 10000</td>
    <td>113.7s</td>
  </tr>
  <tr>
    <td>min-degree</td>
    <td>linear</td>
    <td>custom</td>
    <td><= 10000</td>
    <td>162.3s</td>
  </tr>
  <tr>
    <td>min-degree</td>
    <td>heap</td>
    <td>nx.Graph</td>
    <td><= 10000</td>
    <td>305.9s</td>
  </tr>
  <tr>
    <td>min-degree</td>
    <td>linear</td>
    <td>nx.Graph</td>
    <td><= 10000</td>
    <td>401.9s</td>
  </tr>
  <tr>
    <td>min-fill-in</td>
    <td>-</td>
    <td>custom</td>
    <td><= 1000</td>
    <td>24.3s</td>
  </tr>
  <tr>
    <td>min-fill-in</td>
    <td>-</td>
    <td>nx.Graph</td>
    <td><= 1000</td>
    <td>399.2s</td>
  </tr>
</table>


dschult:
Thanks for submitting this.  It's not close to my area so please excuse and explain my mistakes in interpretation.

My comments are mostly on coding and API.  I'm not sure the iterators should be built as classes. Typically iterators yield results based on an unchanging data structure. In this case, the data structure is changing throughout the iteration process. 

So I suggest an API where a function is given to the algorithm with calling signature ```heuristic(graph)``` where graph is a dict-of-sets adjacency. The function should return a node or ```None``` to signify that the algorithm stop.  I'm not close enough to this code to know if my suggestion makes sense in this case. Can you make a judgement for whether this makes sense?

If data should be stored from one call to the next (like for MinDegreeHeuristic), you can create a class to store that data and have the function be ```lambda graph: MinDegreeHeuristic.best_node(graph)```.  But for other heuristics that don't store data from one call to another a function will do.

What do you think?

networkx/algorithms/approximation/treewidth.py
```diff
@@ -0,0 +1,272 @@
...
+    """
+    
+    # make dict-of-sets structure
+    graph = {}
+    for u in G:
+        graph[u] = set()
+        for v in G[u]:
+            if u != v:  # ignore self-loop
+                graph[u].add(v)
+
```
dschult: Consider ```graph = {set(G[n]) for n in G}```  or if ignoreing self-loop is key, then ```{set(G[n]) - set([n]) for n in G}```.  Either is arguably clear and almost certainly faster.

networkx/algorithms/approximation/treewidth.py
```diff
@@ -0,0 +1,272 @@
...
+                for j in range(i + 1, len(nbrs)):
+                    if nbrs[j] not in self._graph[nbrs[i]]:
+                        num_fill_in += 1
+                        # break inner loop if this can't be min-fill-in node anymore
+                        if num_fill_in >= min_fill_in:
+                            break
+
+                if num_fill_in >= min_fill_in: # break outer loop
+                    break
+
```
dschult: This loop structure doesn't seem to need the individual loops. Maybe a set operation can get the number of nodes you want without the loop overhead.  I'm not sure of the exact logic -- does this or something similar work?

    for (deg, node) in degree_list:
        nbrs = graph[node]
        for nbr in nbrs:
            num_fill_in += len(nbrs - graph[nbr])
            if num_fill_in >= min_fill_in:
                break

networkx/algorithms/approximation/treewidth.py
```diff
@@ -0,0 +1,272 @@
...
+    heuristic_iterator = heuristic_class(graph)
+
+    for elim_node in heuristic_iterator:
+        # Connect all neighbours with each other
+        nbrs = graph[elim_node]
+        for u in nbrs:
+            for v in nbrs:
+                if u != v and v not in graph[u]:
+                    graph[u].add(v)
+
```
dschult: Consider ```for u, v in itertools.combinations(nbrs, 2):``` and maybe rely on self-loops already being eliminated (if they really are and if you document that trait -- or check them here, but don't bother to eliminate them earlier).

Johannes-S:
Thank you for the feedback, we changed everything you mentioned.

networkx/algorithms/approximation/treewidth.py
```diff
@@ -0,0 +1,239 @@
...
+            if v not in graph[u]:
+                graph[u].add(v)
+
+        # push node and its current neighbors on stack
+        node_stack.append((elim_node, nbrs))
+
+        # remove node from graph
+        for u in graph:
+            if elim_node in graph[u]:
+                graph[u].remove(elim_node)
```
dschult: Could be:  ```for u in graph[elim_node]:``` to avoid looping over all the nodes not connected to elim_node. Then I guess you don't have to check if elim_node is in graph[u].

    for u in graph[elim_node]:
        graph[u].remove(elim_node)

dschult:
Final to-do-list:

- Check line-lengths (<80 chars) and blank lines containing white-space and other PEP8 stuff. Check PEP8 style (perhaps using the pep8 utility easily installed with ```pip install pep8```).

- Doc_string for two public facing functions.
    Make a one-line description, then a blank line, then the rest of the text describing the function.  For example with ```treewidth_min_degree``` End the lin after the first sentence and put a blank line before the next sentence which will appear in the paragraph following the one-line description.

- Doc_string for ```min_fill_in_heuristic``` (not as important because not public-facing, but ...)
    Start with a synopsis sentence similar to the MinDegreeHeuristic class.   

These are all minor things. I think it is ready to merge otherwise.

Johannes-S:
Thanks for your comments, we have fixed all problems you mentioned. Now the pep8 checker does not return any error.

dschult:
travis has been giving errors for the past couple of days while building the documentation. I'm going to merge even with that error.

Thanks very much for this!

PR #2961, opened 2018-05-03
Allow bipartite configuration model for empty graph

wuhaochen:
Fix a bug when run bipartite configuration model with empty degree sequence.


dschult:
Thanks!

PR #2963, opened 2018-05-03
Add test case for not complete random bipartite graph.

wuhaochen:
Add test to cover not complete random bipartite graph.


dschult:
Thanks!

PR #2117, opened 2016-05-03
 Branchings.py - Removed graphs and branchings structures in favor of an unique for unrolling the circuits

Loveuse:
Removed the structures that contain the graphs and branchings of all levels and inserted an unique structure, called unroll, that stores the edges involved in a circuits for all levels, in order to rebuild the graph at every level in the second phase of the algorithm.

PR #2965, opened 2018-05-06
Preserve edge attributes when using branching algorithms

leamingrad:
This pull request contains a patch to allow the `Edmonds` class to preserve the edge attributes of the source graph when working out an optimum.

The function is disabled by default, so all of the branching algorithms gain a new `preserve_attrs` keyword argument to enable it. Tests are added for both preserving and discarding attributes.

Feedback is welcome - I am happy to make changes. I am not sure exactly how this change will interract with the change in #2117 - looking at the diff there, this change is orthogonal, but I would guess it might cause merge conflicts.

### DIFF
--- a/networkx/algorithms/tree/branchings.py
+++ b/networkx/algorithms/tree/branchings.py
@@ -272,7 +272,8 @@ class Edmonds(object):
         # sure that our node names do not conflict with the real node names.
         self.template = random_string(seed=seed) + '_{0}'
 
-    def _init(self, attr, default, kind, style):
+
+    def _init(self, attr, default, kind, style, preserve_attrs):
         if kind not in KINDS:
             raise nx.NetworkXException("Unknown value for `kind`.")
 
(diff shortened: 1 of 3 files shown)


networkx/algorithms/tree/branchings.py
```diff
@@ -600,37 +616,50 @@ def is_root(G, u, edgekeys):
...
+                        kind='max', style='branching')
     return B
 
 
-def minimum_branching(G, attr='weight', default=1):
+def minimum_branching(G, attr='weight', default=1, preserve_attrs=False):
     ed = Edmonds(G)
-    B = ed.find_optimum(attr, default, kind='min', style='branching')
+    B = ed.find_optimum(attr, default, preserve_attrs,
+                        kind='min', style='branching')
```
chebee7i: Existing users who wrote code like this:
```
ed.find_optimum(attr, default, kind, style)
```
will be broken by the placement of `preserve_attrs` before `kind` and `style` in the function signature of `find_optimum`. Since the API is not using keyword-only arguments, I'd recommend placing `preserve_attrs` last in the argument list in the definition of `find_optimum`.

networkx/algorithms/tree/tests/test_branchings.py
```diff
@@ -334,3 +334,46 @@ def test_edmonds1_minbranch():
...
+    assert_equal(B[0][1]['otherattr2'], 3)
+
+    G = nx.MultiGraph()
+    G.add_edges_from(edgelist * 2)  # Make sure we have duplicate edge paths
+
+    ed = branchings.Edmonds(G)
+    B = ed.find_optimum('weight', preserve_attrs=True)
+
+    assert_equal(B[0][1][0]['otherattr'], 1)
+    assert_equal(B[0][1][0]['otherattr2'], 3)
```
chebee7i: Make this two tests?  Then, a failure for one but success for the other is more informative in the test output.

networkx/algorithms/tree/branchings.py
leamingrad: Thats a good point - I had thought the Edmonds object was generally internal-only, so happy to make the change

networkx/algorithms/tree/tests/test_branchings.py
leamingrad: Sounds sensibe

leamingrad:
I think all the suggestions have been acted on now - please let me know if there is anything else.

PR #2972, opened 2018-05-12
Add O(m+n) Adjacency Lists Sorting Algorithm

hungz23:
In this pull request, I add an O(m+n) Algorithm for sorting all adjacency lists of a graph G.
It can be useful for some Greedy Algorithm in Graph Theory.

### DIFF
--- a/examples/algorithms/adjacency_sort.py
+++ b/examples/algorithms/adjacency_sort.py
@@ -1,0 +1,38 @@
+"""
+Sorting the adjacency lists of a graph.
+Complexity: O(m+n)
+
+Input: The unsorted adjacency lists of a graph G = (V, E)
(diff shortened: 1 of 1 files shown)


hungz23:
In this pull request, I add an O(m+n) Algorithm for sorting all adjacency lists of a graph G.
It can be useful for some Greedy Algorithm in Graph Theory.

MridulS:
Thanks for this @hungz23 sorry for the delay in getting this reviewed!

NetworkX has made a bunch of changes in the last 5 years and this function can be now rewritten with just something like
``` python
{i: list(G.adj[i]) for i in order}
```

So if anyone requires this in an implementation of any algorithm they should be able to use this one liner.

PR #2976, opened 2018-05-16
Add bipartite layout

wuhaochen:
Add a simple layout for bipartite graph.

I also noticed that all the functions in layout.py used numpy but some checked if numpy is installed while others didn't. Shall we make them more consistent? Personally I don't think checking and raise a new error is necessary. If we remove all the checks we can bring up the code coverage a bit as well.

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -869,7 +869,6 @@ def rescale_layout(pos, scale=1):
         scaled positions. Each row is a position.
 
     """
-    import numpy as np
     # Find max length over all dimensions
     lim = 0  # max coordinate for all axes
     for i in range(pos.shape[1]):
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -29,7 +29,8 @@ from __future__ import division
 import networkx as nx
 from networkx.utils import random_state
 
-__all__ = ['circular_layout',
+__all__ = ['bipartite_layout',
+           'circular_layout',
            'kamada_kawai_layout',
            'random_layout',
            'rescale_layout',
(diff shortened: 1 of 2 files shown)


dschult:
Yes - I agree that we should remove the tests for import numpy/scipy.
We can just import and any error message will be pretty clear.
Let's do that.

dschult:
In the past, people have asked that the layout functions all supply arguments for the center of the layout and the scale of the layout. This could be helpful, for example, when wanting a placement of multiple networks in the same figure. 

This API asks for a width and height. Is there a way to make that work with a center and scale?  Maybe we would need: center, scale and aspect-ratio. There might be a better way...  Ideas?

networkx/drawing/layout.py
```diff
@@ -866,10 +869,12 @@ def rescale_layout(pos, scale=1):
         scaled positions. Each row is a position.
 
     """
+    import numpy as np
     # Find max length over all dimensions
     lim = 0  # max coordinate for all axes
     for i in range(pos.shape[1]):
-        pos[:, i] -= pos[:, i].mean()
+        center = (max(pos[:, i]) + min(pos[:, i])) / 2
```
wuhaochen: This probably should go to another PR. The original code offset all coordinates by the mean of all the coordinate, but if the distribution is not balanced, the center won't be the mean. I don't see why the mean is used here. Is there a particular reason to use the mean?

The change doesn't break any test either, should we add tests to cover the case?
(The previous version will break my test for bipartite layout but we might want tests dedicate to test the correctness for this function as well?)

dschult: Hmmm...   Changing the definition of center (center of range vs center of mass) would change existing code. So, yes -- this should be in a separate PR. I don't think we've had this discussion before.

Using the mean as center (center of mass) makes "typical" points in the middle at the cost of creating some white space on one side. There's also a notion that one point should not affect the layout significantly.
Switching to center meaning midpoint (center of range) uses the entire range (no white space on any side) at the cost of sometimes having most points on one side.

There may be other reasons/arguments.  We should check how other layout engines handle it.  

And yes, this should be tested to at least encode what we do.

Can we keep it as the mean being the center for this PR and I'll open an issue to reconsider "center".

PR #2982, opened 2018-05-22
Add is_perfect_matching function

arcondello:
Add a method to check whether a maximum matching is perfect.


arcondello:
No, I think it's good. Thanks!

Issue #2977, opened 2018-05-16
change in pandas breaks doctest in python 3.6


wuhaochen:
It looks like that the recent change in pandas: https://github.com/pandas-dev/pandas/pull/19884 is breaking one of the doctest in python 3.6.

I'm not sure what would be the best solution here since py2 and py3 behave differently and the difference is not crucial at all. One way to fix this would be use an `OrderedDict` but it feels like an overkill.

```
======================================================================
FAIL: to_pandas_edgelist (networkx.convert_matrix)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/python/3.6.3/lib/python3.6/doctest.py", line 2199, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for networkx.convert_matrix.to_pandas_edgelist
  File "/home/travis/venv/lib/python3.6/site-packages/networkx/convert_matrix.py", line 192, in to_pandas_edgelist
----------------------------------------------------------------------
File "/home/travis/venv/lib/python3.6/site-packages/networkx/convert_matrix.py", line 221, in networkx.convert_matrix.to_pandas_edgelist
Failed example:
    df
Expected:
       cost source target  weight
    0     1      A      B       7
    1     9      C      E      10
Got:
      source target  weight  cost
    0      A      B       7     1
    1      C      E      10     9
```

PR #2983, opened 2018-05-22
Make to_pandas_edgelist behave consistent across py2 and py3

wuhaochen:
Address #2977. Using OrderedDict here might be an overkill though.

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -232,8 +232,9 @@ def to_pandas_edgelist(G, source='source', target='target', nodelist=None,
     source_nodes = [s for s, t, d in edgelist]
     target_nodes = [t for s, t, d in edgelist]
     all_keys = set().union(*(d.keys() for s, t, d in edgelist))
-    edge_attr = {k: [d.get(k, float("nan")) for s, t, d in edgelist] for k in all_keys}
     from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


dschult:
hee hee -- looks like we were both working on this at the same time.  (see #2984)

What do you think?  My fix specifies the order of the columns in the doctest.
Your change uses an OrderedDict so that the default column order is the same as expected.
I believe the DataFrame can also be created with a list instead of a dict to constrain the order.

wuhaochen:
Ha, what a coincidence. I feel like using OrderedDict is reasonable here since the original change in pandas is because of dict is ordered by default in py3. The performance should actually be better with OrderedDict since it makes pandas avoid sorting the dict keys.

My overkill concerning is purely aesthetic since we don't really care about the order here. In such sense I guess I would prefer mine to #2984.

wuhaochen:
Whee, it seems the order still need some more tweak.

dschult:
I think the new error indicates that OrderedDict will be harder to maintain long term.
And I also appreciate your sense of aesthetics about having source and target printed first in the examples.

I'll put the order as you created it (source, target, cost, weight), but not add dependence on OrderedDict in this part of the code.  I suspect that in N years (as N -> infinity) we won't have OrderedDict anymore.

wuhaochen:
Yes, we should all use python 3 in the end. I'm ok with whatever works here.

dschult:
see #2984

Issue #3000, opened 2018-06-06
Add '_' to legal .gml key names


monchin:
I found that when I'd like to write a .gml file, the key must be '[A-Za-z][A-Za-z0-9]*$', but I think it is usual that '\_' appears in a legal variable name, so maybe it is good to add '\_' in it.

monchin:
I tried to create a new pull request but I failed. 
[https://github.com/networkx/networkx/pull/3001](url)
I'm not quite sure the reason it failed to pass the travis-ci check. It would be appreciated if anyone could help me.

PR #3001, opened 2018-06-06
Add '_' to legal key variable names when writing .gml files(#3000)

monchin:


dschult:
Don't worry about the failure on Travis-ci.  It you look at the log of the error it is not related to your code. It looks like one of the tests that uses random graph generation is no longer working sometimes.

The GML standard http://www.fim.uni-passau.de/index.php?id=17297&L=1
explicitly specifies the keys to allow  '[A-Za-z][A-Za-z0-9]*$'
Are you advocating a change to the standard?

Issue #1393, opened 2015-03-06
`create_using` should be a class, not an instance of a class


jfinkels:
The `create_using` idiom appears in several places throughout the code. As a keyword argument to, for example, the `from_numpy_matrix` function, it specifies the type of graph that will be returned. The function expects this keyword argument to be an _instance_ of a graph, and it proceeds to clear the graph and add nodes and edges to it.

Why should `create_using` be an instance of a graph class if it is just going to be cleared? Instead, `create_using` should be a graph _class_ which should then be instantiated by the function when creating the graph that it will eventually return. Not only does this make more sense with respect to way the keyword argument is being used, this way, the responsibility of instantiating the graph that will be eventually be returned moves from the caller to the callee: this helps to prevent the caller from fiddling with the object when it is being created.

chebee7i:
I've desired this on occasion as well. Maybe we can meet partway and query if `create_using` is a callable, and if so, call it to create the instance to be populated?

jfinkels:
Do you want to meet partway for backwards compatibility? With a major version increment coming up, this is a good opportunity to make non-backwards compatible changes. Is there a place in the code or tests where using an instance instead of a class is really required?

MridulS:
I did a quick search with grep. I didn't find any place where instance is really required.

ysitu:
If we are to do this, let's do it right, not with a mix of what we like and we don't like.

chebee7i:
But suppose I have a graph already with some nodes and edges that was provided to me, and then I want to use some graph generator to add more nodes (I dont't want it cleared). It seems like callables and instances are equally valid use cases. If we don't support an instance, then the user would have to write a silly lambda function to return the instance. Both approaches seem fine to me, but checking if `create_using` is callable seems pretty simple for us to do. In fact, there might even be a few places where we already do such a check, not sure though.

jfinkels:
That situation could be solved by taking the original graph, generating the new graph, and using one of the union or compose operators.

chebee7i:
I might be thinking of the `weight` parameter wrt to existing code checking if it is callable.

chebee7i:
@jfinkels those operators assume you want disjoint combinations, which may not be what is desired. I try to think of the generators as functions that add edges to the graph with a particular property, etc. In that sense, they should always be able to work with existing graphs---the final graph may not have the properties of the graph generator, but that is a decision the user made. Now whether existing graphs are supported directly or through a lambda function or similar, I could see it either way.

[...]

PR #3004, opened 2018-06-08
Enable create_using to be a constructor

dschult:
Related to #1393

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -24,7 +24,12 @@ Improvements
 
 API Changes
 -----------
+empty_graph has taken over the functionality from
+nx.convert._prep_create_using which was removed.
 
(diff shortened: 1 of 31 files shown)


dschult:
This change makes it possible to use ```create_using=nx.DiGraph``` instead of requiring an instance as in ```create_using=nx.DiGraph()```.  In fact, using a constructor instead of an instance is now the method that the doc_strings promote. Both work however. 

The processing of ```create_using``` has been centralized in ```nx.empty_graph``` and the sometimes-used function ```_prep_create_using``` has been removed.

For developers: your function taking parameter ```create_using``` should call ```nx.empty_graph(0, create_using, default=nx.MultiGraph)``` or similar.  The default argument is used ```if create_using is None```.  That way your function can specify the default constructor without needing to check the input and without creating an instance.

PR #3019, opened 2018-06-18
GED: support self-loops, multigraphs and digraphs

aparamon:


networkx/algorithms/similarity.py
```diff
@@ -641,15 +643,24 @@ def match_edges(u, v, pending_g, pending_h, Ce, matched_uv=[]):
...
+            # NOTE: inf remembered from Ce construction
             for k, i in zip(range(m), g_ind):
-                g = pending_g[i]
+                g = pending_g[i][:2]
                 for l, j in zip(range(n), h_ind):
-                    h = pending_h[j]
-                    if not any(g in ((p, u), (u, p)) and h in ((q, v), (v, q))
-                               or g == (u, u) and h == (v, v)
+                    h = pending_h[j][:2]
+                    if nx.is_directed(G2) or nx.is_directed(G2):
```
dschult: Can you look at this if statement again -- are both supposed to be G2?

aparamon: Uh-oh -- thanks for spotting this!
Please expect the fix commit later today.

dschult:
Nice -- and I like all the new tests. 
Does it work with MultiDiGraph?  Should it test that?

aparamon:
@dshult support for multi-graphs and digraphs is orthogonal, so it's expected that for MultiDiGraphs it "just works," -- but I can add an additional test if you consider it desirable.

dschult:
Glad to hear they are orthogonal -- that makes it easier.
I think that'd be good to add a simple test if that's easy.

dschult:
It looks like the test actually uses a MultiGraph instead of MultiDiGraph.
Any other features you want to add to this? It looks ready after that change

PR #3040, opened 2018-06-28
LR planarity checking

KaneTW:
We implemented a planarity check based on the paper "The Left-Right Planarity Test"[1]

The code is able to certify the correctness of the result by giving a combinatorial embedding as a planarity certificate and Kuratowski subgraph as a counterexample. Our unit tests also verify that the certificates are correct.

We've encountered the following issues and would like your feedback regarding these:
- Recursion depth will be exceeded on larger graphs (10000 nodes).
  - Fix would be to use a DFS visitor, but it needs to be more complex than the one already present in networkx
  - We have worked on a fix, but it would make the code much harder to read.
  - Workaround is the standard recursion depth adjustment.
- Runtime complexity of the planarity check is superlinear. Reasons:
  - Sorting of a neighbor list by using the default sort method: A custom sort method could improve the complexity. Effectively, this is O(|V|) in the average case due to the planarity condition |E| <= 3*|V|-6. Worst case is O(|V| log |V|) for a graph with dominating vertices.
  - Insertion in a neighbor list: Requires a better data structure with constant time insert. 
- Is the chosen path `networkx/algorithms/planarity.py` fine or should we add a new category and use e.g. `networkx/algorithms/planarity/checking.py`.

If you have any issues you'd like us to improve, please tell us.

--
[1]: http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.217.9208

### DIFF
--- a/doc/reference/algorithms/index.rst
+++ b/doc/reference/algorithms/index.rst
@@ -50,6 +50,7 @@ Algorithms
    mis
    node_classification
    operators
+   planarity
    reciprocity
    rich_club
    shortest_paths
--- a/doc/reference/algorithms/planarity.rst
+++ b/doc/reference/algorithms/planarity.rst
@@ -1,0 +1,9 @@
+*********
+Planarity
+*********
+
+.. automodule:: networkx.algorithms.planarity
+.. autosummary::
+   :toctree: generated/
+
+   check_planarity
--- a/doc/reference/algorithms/planarity.rst
+++ b/doc/reference/algorithms/planarity.rst
@@ -7,3 +7,5 @@ Planarity
    :toctree: generated/
 
    check_planarity
+.. autoclass:: PlanarEmbedding
+   :members:
--- a/networkx/algorithms/__init__.py
+++ b/networkx/algorithms/__init__.py
@@ -31,6 +31,7 @@ from networkx.algorithms.matching import *
 from networkx.algorithms.minors import *
 from networkx.algorithms.mis import *
 from networkx.algorithms.operators import *
+from networkx.algorithms.planarity import *
 from networkx.algorithms.reciprocity import *
 from networkx.algorithms.richclub import *
 from networkx.algorithms.shortest_paths import *
--- a/networkx/algorithms/planarity.py
+++ b/networkx/algorithms/planarity.py
@@ -1,0 +1,419 @@
+from collections import defaultdict
+import networkx as nx
+
+__all__ = ["check_planarity"]
+
+
+def check_planarity(G, counterexample=False):
+    """Checks if a graph is planar and returns a counterexample or an embedding
+
+     A graph is planar iff it can be drawn in a plane without
+     any edge intersections.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+    counterexample : bool
+        A Kuratowski subgraph (to proof non planarity) is only returned if set
+        to true
+
+    Returns
+    -------
+    is_planar : bool
+        Is true if the graph is planar
+
+    certificate :
+        If the graph is planar this is a planar embedding (dict).
+        If the graph is not planar and counterexample is true,
+        this is a Kuratowski subgraph.
+
+    Notes
+    -----
+    A (combinatorial) embedding consists of cyclic orderings of the incident
+    edges at each vertex, given such an embedding there are multiple approaches
+    discussed in literature to drawing the graph (subject to various
+    constraints, e.g. integer coordinates), see e.g. [2].
+
+    The planarity check algorithm and extraction of the combinatorial embedding
+    is based on the Left-Right Planarity Test [1].
+
+    A counterexample is only generated if the corresponding parameter is set,
+    because the complexity of the counterexample generation is higher.
+
+    References
+    ----------
+    .. [1] Ulrik Brandes:
+        The Left-Right Planarity Test
+        2009
+        http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.217.9208
+    .. [2] Takao Nishizeki, Md Saidur Rahman:
+        Planar graph drawing
+        Lecture Notes Series on Computing: Volume 12
+        2004
+    """
+
+    planarity_state = LRPlanarity(G)
+    embedding = planarity_state.lr_planarity()
+    if embedding is None:
+        # graph is not planar
+        if counterexample:
+            return False, get_counterexample(G)
+        else:
+            return False, None
+    else:
+        # graph is planar
+        return True, embedding
+
+
+def get_counterexample(G):
+    """Obtains a Kuratowski subgraph
+
+    Raises nx.NetworkXException if G is planar.
+
+    The function removes edges such that the graph is still not planar.
+    At some point the removal of any edge would make the graph planar.
+    This subgraph must be a Kuratowski subgraph.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+
+    Returns
+    -------
+    subgraph : NetworkX graph
+        A Kuratowski subgraph that proves that G is not planar.
+
+    """
+    # copy graph
+    G = nx.Graph(G)
+
+    if check_planarity(G)[0]:
+        raise nx.NetworkXException("G is planar - no counter example.")
+
+    # find Kuratowski subgraph
+    subgraph = nx.Graph()
+    for u in G:
+        nbrs = list(G[u])
+        for v in nbrs:
+            G.remove_edge(u, v)
+            if check_planarity(G)[0]:
+                G.add_edge(u, v)
+                subgraph.add_edge(u, v)
+
+    return subgraph
+
+
+class Interval(object):
+    """Represents a set of return edges
+
+    All return edges in an interval induce a same constraint on the contained
+    edges, which means that all edges must either have a left orientation or
+    all edges must have a right orientation.
+    """
+    def __init__(self, low=None, high=None):
+        self.low = low
+        self.high = high
+
+    def empty(self):
+        """Check if the interval is empty"""
+        return self.low is None and self.high is None
+
+    def copy(self):
+        """Return a copy of this interval"""
+        return Interval(self.low, self.high)
+
+    def conflicting(self, b, planarity_state):
+        """Return True if interval I conflicts with edge b"""
+        return (not self.empty() and
+                planarity_state.lowpt[self.high] > planarity_state.lowpt[b])
+
+
+class ConflictPair(object):
+    """Represents a different constraint between two intervals
+
+    The edges in the left interval must have a different orientation than
+    the one in the right interval.
+    """
+    def __init__(self, left=Interval(), right=Interval()):
+        self.left = left
+        self.right = right
+
+    def swap(self):
+        """Swap left and right intervals"""
+        temp = self.left
+        self.left = self.right
+        self.right = temp
+
+    def lowest(self, planarity_state):
+        """Return the lowest lowpoint of a conflict pair"""
+        if self.left.empty():
+            return planarity_state.lowpt[self.right.low]
+        if self.right.empty():
+            return planarity_state.lowpt[self.left.low]
+        return min(planarity_state.lowpt[self.left.low],
+                   planarity_state.lowpt[self.right.low])
+
+
+def top_of_stack(l):
+    """Returns the element on top of the stack."""
+    if not l:
+        return None
+    return l[-1]
+
+
+class LRPlanarity(object):
+    """A class to maintain the state during planarity check"""
+    def __init__(self, G):
+        # copy G without adding self-loops
+        self.G = nx.Graph()
+        self.G.add_nodes_from(G.nodes)
+        for e in G.edges:
+            if e[0] != e[1]:
+                self.G.add_edge(e[0], e[1])
+
+        self.roots = []
+
+        # distance from tree root
+        self.height = defaultdict(lambda: None)
+
+        self.lowpt = {}  # height of lowest return point of an edge
+        self.lowpt2 = {}  # height of second lowest return point
+        self.nesting_depth = {}  # for nesting order
+
+        # None -> missing edge
+        self.parent_edge = defaultdict(lambda: None)
+
+        # oriented DFS graph
+        self.DG = nx.DiGraph()
+        self.DG.add_nodes_from(G.nodes)
+
+        self.ordered_adjs = {}
+
+        self.ref = defaultdict(lambda: None)
+        self.side = defaultdict(lambda: 1)
+
+        # stack of conflict pairs
+        self.S = []
+        self.stack_bottom = {}
+        self.lowpt_edge = {}
+
+        self.left_ref = {}
+        self.right_ref = {}
+
+        self.embedding = {}
+
+    def lr_planarity(self):
+        """Execute the LR planarity test
+
+        Returns
+        -------
+        embedding: dict
+            If the graph is planar an embedding is returned. Otherwise None.
+        """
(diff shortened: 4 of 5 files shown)


dschult:
Hi!  Thanks for this....

I have been too focused on the upcoming release to look at this much yet. I'll try to answer your questions and then dig into it more soon.

Recursion:
many of our algorithms start as a recursive method -- but recursion uses a lot of RAM (hence the recursion depth limit). So its better for many reasons to implement a non-recursive method. If you like I always suggest leaving the recursive code there (with recursive in the function name) as it can be used for testing/comparing and for pedagogy.

You can certainly construct your own depth-first-search.  It's not a long piece of code. And there are many possible variants depending on when you want something returned. So we actually have a number of places where variants of dfs show up in the code.  Many many years ago Aric made a version that was very flexible and could allow the user to construct whichever dfs variant they wanted. But it was removed (I think because Aric decided it was harder for the user and for us to maintain than taking the dfs code and changing it to do what you want -- perhaps other reasons that I don't recall too).

Neighbor sorted_list with constant time insertion requires a different data structure. That's fine if there's an easy way to include it and maintain it. There are a few places in the code where we use dict-of-set for the neighbors -- separate from a Graph object. What exactly do you need? Would heapq or deque (from collections) be sufficient? There was a recent post #3057 about using a Fibonacci heap. Perhaps they have some ideas.

If you decide not to create such a data structure then its still worth having this code in networkx.

File structure:  Start by including all planarity functions in ```algorithms/planarity.py```.  We can refactor if needed.

I guess my main question for you would be:  What are the big questions in planarity?  This seems like a good function to have.  Are there basic algorithms used a lot for planarity that we are missing? Low hanging fruit ideas are appreciated and we can mark them as good for people looking for something to contribute (like "Good First Issue").  I don't have much experience with them so appreciate any hints.

Thanks again!

Johannes-S:
Hi, thanks for your feedback.
Here is a selection of useful algorithms related to planarity:
- Triangulation (we are currently working on it)
- Planar drawing (we are currently working on it)
- Planar generation (we are currently working on it)
- Dual graph (easy)
- 5-coloring algorithm


A datastructure for constant time insertion after a specific element (in the middle) would be a linked list. We cannot use pythons deque, because it does not allow the insertion in the middle.
So we are now adding a datastructure that allows a constant insertion time and will add the non recursive version.

dschult:
Nice -- I'll put the label "Good First PR" on this issue to see if anyone decides to take on Dual graph.

There is a StackOverflow answer [about linked lists in python](https://stackoverflow.com/questions/2154946/python-linked-list-o1-insert-remove) that suggests if you usually add in the same place, or change that place occasionally, you can get efficient O(1) insertion by using two lists. One "before" the pointer of where to add and one "after". Then to move the pointer, you remove from one list and add to the other. So long as it uses python its acceptable here.

Thanks and looking forward to more. :)

Johannes-S:
Thanks for the StackOverflow link. However this would not really help here, because we do not have to iterate over the whole list. What we need is a constant time insert at a specific position in the list (to insert an outgoing edge directly after another given outgoing edge).

We have now implemented our own data structure which would help here.
We have implemented two versions: One implements the nx.Graph interface and the other one does not. I have added links to preliminary versions of each so you can have a look at it:
1. Subclass of nx.Graph : https://gist.github.com/Johannes-S/01d9243f86dbda0eb30ca5bb8729fe7c
2. No nx.Graph subclass: https://gist.github.com/Johannes-S/dacb6e55f0db4745cc7fd8e0e3c5c000

Which version would you prefer?

As we have mentioned, we are currently also working on the triangulation and a planar drawing algorithm, which would then also make use of these data structures.

dschult:
I've only got questions (not suggestions) at this point. Still getting up to speed on the data structure and the algorithm.

It took me a while to figure out that cw and ccw referred to clockwise and counterclockwise respectively. That concept is a little confusing to me. How can you tell which direction you are going if you don't have an embedding to work from?  Is the direction arbitrary for the first edge and once chose determined after that?  That would align with planar graphs' symmetry of flipping your view to "from below" so cw becomes ccw etc.  Maybe in long run a few sentences to discuss cw and ccw could be helpful.

I'm not sure what the difference between half edges and directed edges is. Perhaps more importantly, I don't understand half edges yet.  Are you constructing half edges as you go around a face and then the other half edge comes from the other face and it must have opposite polarity (cw or ccw) to be valid.?.?  I hope that is close to correct...otherwise ignore the rest of this paragraph. :)    Could you use a directed graph in place of this structure -- where directed edges represent half-edges. You cannot have directed edges created where they already exist. And at some point (going around a different face) the other direction of the same edge will show up. So to check for a planar embedding you go around the faces and then check that each edge has the opposite oriented edge.  Is this correct?

Also, why do you need to know the "next node" of the neighbor node in the dict-of-dict structure? You say

> They are dicts that map a node v to a dict that maps each neighbor of v to the next neighbor of v in clockwise direction (or counter clockwise direction).

Couldn't you find the "next neighbor" by looking up the neighbor in the outer dict? Is a face determined by two edges? (three nodes?) I think there is something basic I'm missing here.

Thanks :)

Johannes-S:
First I have listed a few general remarks, which I think might help to understand the algorithm:
- The LR-planarity algorithm uses a directed graph `self.DG`, which is a spanning tree of the original graph. (For every edge in the original graph we have only one direction in the directed graph)
- All edges of the original graph that are not part of the directed graph are so called back edges. Every back edge starts at a node in the directed graph and ends at a position in the directed graph (of lower height)
- One main idea of the LR-planarity test is to partition the back edges into categories whether they must go LEFT around the lower part of the spanning tree, or RIGHT around it. The algorithm does this by considering different cases that force two back edges to be in opposite categories (one LEFT means the other cannot be in LEFT) or that force them to be in the same category (either both LEFT of both RIGHT). The main difficulty of the algorithm is so efficiently compute a solution to these constraints.

>  [Are you constructing half edges as you go around a face and then the other half edge comes from the other face and] **it must have opposite polarity (cw or ccw) to be valid.?**

Half edges do not have a cw or ccw polarity. We just save the order in which the outgoing half-edges appear in the embedding. So for example a node v might have three neighbors w1, w2 and w3. Now a normal graph class would not know in which order these neighbors would appear in a planar embedding. So these nodes could appear in clockwise order (around v) as follows: [w1, w2, w3] and then the counter clockwise order is [w3, w2, w1]. So the half-edges do not contain a clockwise / counter clockwise parameter, but it just allows us to say how we want to view the neighborhood.

The methods `add_half_edge_cw` and `add_half_edge_ccw` are just named this way because the half edge that is added, can either be added clockwise to the `reference_neighbor`, or counterclockwise. With the example from above: `add_half_edge_cw(start_node=v, end_node=x, reference_neighbor=w2)` would lead to the following clockwise neighbor ordering of v [w1, w2, x, w3] (and of course the counter clockwise ordering [w3, x, w2, w1]). 
If we would instead call `add_half_edge_ccw(start_node=v, end_node=x, reference_neighbor=w2)` then we get the following clockwise neighbor ordering of v: [w1, x, w2, w3] (and the counter clockwise ordering [w3, w2, x, w1]).

> Couldn't you find the "next neighbor" by looking up the neighbor in the outer dict?

We are not only interested in any next neighbor but to exactly the immediate next neighbor in counter clockwise or clockwise direction.

Note that we require the ordering at both sides of the edge. Therefore we have to include both directions. A half-edge data structure is only valid if both edges are present, so I think it would not make sense to consider it as a directed graph (because if  (v, w) is present (w, v) must also be present). However because of how the algorithm works it is easier to first add a half edge in one direction and later the half edge into the other direction, otherwise it would be nicer to enforce that both half edges have to be added at the same time. Therefore we have added the method check_structure, which checks that every half-edge has its opposite half-edge.

> Are you constructing half edges as you go around a face [...]

The LR-planarity algorithm does not really consider faces, so we don't construct half edges as we go around a face. 
However the half-edge structure allows us to compute the number of faces later to verify that an embedding is actually correct. This is done in the check_intersection method, which compares it to Euler's formula (N - E + F = 2) to proof that the returned embedding is actually planar.

> Finally, you often mention "constant time"... I always like to tweak the system by asking "constant as what variable changes?"

The parameter of the time complexity we give is the number of nodes or the number of edges of the input graph. Note that the maximum number of edges is proportional to the number of nodes for planar graphs (E &leq; 3*V-6 if V>2). 

> How can you tell which direction you are going if you don't have an embedding to work from? Is the direction arbitrary for the first edge and once chose determined after that? 

Not sure if I understand your question correctly. But of course there might be multiple cases where the resulting embedding is not unique for a specific input graph. E.g. a start graph could have an arbitrary order on the neighbors for the central node. I think the non determinism in the first depth first search (choosing some neighbor of a node) is determining which of the possible embeddings is returned. I would have to look at the code and the paper again to make sure that this is correct.

dschult:
Maybe you can tell me what you mean by a "half-edge".  That might help much of the rest.
I'll get there...   :)

As for which data structure to use: I think the second gist you list is probably a better data-structure for this problem as it doesn't carry any of the "baggage" of subclassing graphs. But I leave that decision up to you since you know the algorithm and code the best.

Thanks!

Johannes-S:
## Regarding Half-Edges
The term half-edge is not an invention from us, it is in the literature also referred to as [doubly connected edge list (DCEL)](https://en.wikipedia.org/wiki/Doubly_connected_edge_list).
We use the term half-edge, because even though the graph structure is undirected (an edge in one direction means that the graph also contains the other direction) the two directions of an edge are still different. For example a normal edge always has two faces as its neighbors (maybe one face is the outer face). Because half-edges always come in pairs, we can say that one half-edge belongs to one face and the other belongs to the other face. The following explanation goes into more detail on what this means.

I have added an image below to explain what half-edges are and what we use it for in more detail.


I have labeled the nodes (1-7) and the faces (A-G). 

#### What would the half-edge data structure contain in this case?
The half-edge data structure saves to order of neighbors (or outgoing half-edges) for a node in clock-wise or counter-clockwise order.
```python
{
  1: [2, 4, 3],
  2: [5, 4, 1],
  3: [1, 4, 6],
  4: [1, 2, 5, 7, 6, 3],
  5: [7, 4, 2],
  6: [3, 4, 7],
  7: [6, 4, 5]
}
```
The data above shows how the embedding would be represented in the current (unoptimized) implementation. For each node it displays its neighbors in clockwise order. Note that the first node in the list is arbitrary, so the neighbor list of node 4 could also be `[5, 7, 6, 3, 1, 2]`.

**Example: How would we obtain the clockwise half-edge of (4, 5)?**

1. Obtain the neighbor list of node 4. Result: `nbrs=[1, 2, 5, 7, 6, 3]`
2. Find the index of node 5. Result: `idx=2`
3. The neighbor of the desired half-edge is at the index `idx+1`. Result: `nbrs[idx+1]=7`
4. The clockwise half-edge of (4, 5) is (4, 7)

Of course the lookup in the list in step 2 is not very efficient, which is the main difference in our more optimized data structure.

### What can the half-edge data structure be used for?
In my drawing you can see that each half-edge belongs to exactly one face. For example the half-edge (4, 1) belongs to face B and the half-edge (1, 4) belongs to face A.
The half-edge data structure allows us to determine all edges that belong to the same face.
For example if we start with the half-edge (4, 1) and want to obtain all other edges that belong to the same face (B) we do the following:

1. Query the data-structure for the counter-clockwise neighbor of (1, 4) --> We obtain (1, 2)
2. Query the data-structure for the counter-clockwise neighbor of (2, 1) --> We obtain (2, 4)
3. Query the data-structure for the counter-clockwise neighbor of (4, 2) --> We obtain (4, 1), which is the same half-edge we started with

So we have found all half-edges (and therefore nodes) that belong to face B. Note that we flip the half-edge for our queries, because the data structure saves the order of **outgoing** half-edges for a node.

We can repeat this process for all half-edges that we have not encountered yet to get all faces of the graph. We can therefore obtain the number of faces of a planar graph, which cannot be determined that easily if you just have a normal graph structure.

dschult:
That's great -- it *is* what I was thinking in terms of directed edges.  I like the way it can make finding the faces easy too.

I think the NetworkX DiGraph might be a reasonable structure to store this "Dcel".
Indeed, the code below creates such a structure (without the height/lowpt/etc filled in).

    # set up graph
    G = nx.path_graph(9)
    nx.add_path([4,7,9])
    
    DG = nx.DiGraph(G)       # creates two directed (half) edges in place of each undirected edge
    nx.set_edge_attributes(DG, None, 'cw')  # set up edge attr to point to cw/ccw reference
    nx.set_edge_attributes(DG, None, 'ccw')  # initialize reference to None
    (4, DG[4][5]['cw'])    # shows the next half-edge clockwise of (4, 5)
    (4, DG[4][5]['ccw'])    # similar for next counter-clockwise of (4, 5)
    # make half-edge ((4, 7) just clockwise of (4, 5)
    if DG[4][7]['cw'] is None:
        DG[4][7]['cw'] = DG[4][5]['cw']
        DG[4][5]['cw'] = 7
    else:
        print("cannot add half-edge. It is already present.")
    
    # You could store the node attributes on the DiGraph too.
    DG.node[0]['height'] = 0
 
I'm not saying it will be better to store all that on a DiGraph -- might not even be possible given all the manipulation you need to do while maintaining the order of the neighbors. But it is possible that it could work.

dschult:
NetworkX allows any hashable node other than ```None```. There are a number of places in the code where we use ```None``` as an indicator of no input so we had to disallow ``None``` as a node.  :)

Johannes-S:
Hello,
We have added a commit to this pull request improving the following:
- Removed recursion (we kept the easier recursive versions and added the suffix: `[...]_recursive`)
- Added a PlanarEmbedding class (subclass of nx.DiGraph)
- Improved documentation
- Lower RAM consumption (by freeing variables that are no longer needed)
 
We have rebased the commits with the master branch so the tests run successfully.

We ran some benchmarks on larger graphs. We did use the planar graphs from https://users.dcc.uchile.cl/~jfuentess/datasets/graphs.php. The maximum memory consumption of the LR-Planarity algorithm is less than 3.5 times the memory consumption of the input graph.

| Edges [Million] | Max RAM [GB] | `nx.Graph` RAM [GB] | `nx.read_edgelist` time [hours] | `check_planarity` time [hours] | `check_structure` time [hours] |
| --------------- | ------------ | ------------------- | ------------------------------- | ------------------------------ | ------------------------------ |
|  3 |   8 |  3 | 0.0 | 0.1 | 0.0 |
| 15 |  39 | 12 | 0.1 | 1.3 | 0.3 |
| 30 |  77 | 24 | 0.3 | 2.9 | 0.6 |
| 45 | 125 | 37 | 0.5 | 4.8 | 1.0 |
| 60 | 155 | 48 | 0.6 | 6.5 | 1.3 |
| 75 | 196 | 61 | 0.8 | 8.4 | 1.7 |

Notes regarding collums in the benchmark:
- `nx.Graph` RAM [GB]: The ram consumption just for importing the graph into NetworkX
- `nx.read_edgelist` time [hours]: The time it takes to import the graph into NetworkX
- `check_planarity` time [hours]: The time the LR-Planarity algorithm takes
- `check_structure` time [hours]: The time it takes to verify that the resulting embedding is correct (Just to make sure everything is correct)


A build of the documentation is uploaded here: https://teamnotjava.github.io/networkx-doc/reference/algorithms/planarity.html

dschult:
This is a good way to handle it.  I'm not sure how I missed those tests.  Looking online instead of in my editor I suspect. Anyway. Is it ready to merge?  I think it looks good to go.

Johannes-S:
Yes, it is ready to merge.
Thanks for your feedback!

PR #3043, opened 2018-07-02
 Fix up edge_dfs and add edge_bfs

### LINKED ISSUE #2792: Add an edge_bfs traversal to match edge_dfs
`networkx.algorithms.traversal.bfs_edges` does not correctly return all the edges of a graph if a node has more than one in-edge. My expectation would be (and critical in my application) is that in a traversal, _all_ edges in the graph are returned at some point in the iteration.

Here's an illustration

```
import networkx as nx
from networkx.algorithms.traversal import bfs_edges
from random import shuffle

G = nx.DiGraph()
e_in = [(1, 2), (1, 3), (2, 4), (2, 5), (3, 6), (3, 7), (4, 8), (6, 8), (5, 9), (7, 9), (8,10), (9,10)]
# shuffle(e_in) # test that bfs_edges generates correct order
G.add_edges_from(e_in)
e_out = [e for e in bfs_edges(G, 1)]
print('Received: ', e_in)
print('Returned: ', e_out)
print('Set difference: ', set(e_in) ^ set(e_out))
# Set difference:  {(6, 8), (9, 10), (7, 9)}

# visualisation
# import graphviz as gv
# gv.Source(nx.drawing.nx_pydot.to_pydot(G).to_string())
```

The issue is in `generic_bfs_edges` in https://networkx.github.io/documentation/latest/_modules/networkx/algorithms/traversal/breadth_first_search.html, where a node gets excluded from forming more edge pairs if it gets visited once - hence the issue if a node has more than one in-edge. However, it's not a solution to completely exclude this visitor-mechanism, because then the respective child-nodes get added to the queue multiple times.

```
from collections import deque, Counter
def generic_bfs_edges_no_visitors(G, source, neighbors=None):
    visited = {source}
    queue = deque([(source, neighbors(source))])
    while queue:
        parent, children = queue[0]
        try:
            child = next(children)
            if child not in visited:
                yield parent, child
                # visited.add(child)
                queue.append((child, neighbors(child)))
        except StopIteration:
            queue.popleft()

e_out = [e for e in generic_bfs_edges_no_visitors(G, 1, G.successors)]
print('Received: ', e_in)
print('Returned: ', e_out)
print('Set difference: ', set(e_in) ^ set(e_out))
# Set difference:  set()
print('Multiset difference: ', Counter(e_out) - Counter(e_in))
# Multiset difference:  Counter({(8, 10): 1, (9, 10): 1})
```

Here's my proposed fix. I don't know enough about the rest of the library to see if this would break anything, but getting graph traversal right sounds pretty fundamental to me.

```
def generic_bfs_edges_proposed_fix(G, source, neighbors=None):
    visited = {source}
    queue = deque([(source, neighbors(source))])
    while queue:
        parent, children = queue[0]
        try:
            child = next(children)
            yield parent, child # moved this line out of the if-condition!
            if child not in visited:
                visited.add(child)
                queue.append((child, neighbors(child)))
        except StopIteration:
            queue.popleft()

e_out = [e for e in generic_bfs_edges_proposed_fix(G, 1, G.successors)]
print('Received: ', e_in)
print('Returned: ', e_out)
print('Set difference: ', set(e_in) ^ set(e_out))
# Set difference:  set()
print('Multiset difference: ', Counter(e_out) - Counter(e_in))
# Multiset difference:  Counter()
```

### LINKED ISSUE #2840: dfs_edges does not correctly traverse directed multi graph?
Somewhat related to #2792, I expected ```dfs_edges``` to return all edges in a connected component. 
However, this doesn't seem to be the case:

```python
import networkx as nx
g = nx.MultiDiGraph()
g.add_node(1)
g.add_node(2)
g.add_node(3)
g.add_edge(1, 2, label = 'A')
g.add_edge(1, 2, label = 'B') # Another edge between 1, 2
g.add_edge(1, 3, label = 'C')
print(g.edges()) # [(1, 2), (1, 2), (1, 3)] -- Three edges overall, two between nodes 1, 2 
print(list(nx.dfs_edges(g))) # [(1, 2), (1, 3)] -- Only two edges. Also, unclear which of the edges between 1, 2 is chosen.
```
Is there anything I'm missing here? Should I use some other function?
If not, I'll probably implement something for my application, and I'll be happy to PR if there's interest.

Thanks!

### LINKED ISSUE #3037: Return values of `edge_dfs` could use cleanup
From the documentation of `edge_dfs` in networkx 2.2: 

> If orientation is ‘reverse’ or ‘ignore’, then edge takes the form (u, v, key, direction) where direction is a string, ‘forward’ or ‘reverse’,  that indicates if the edge was traversed in the forward (tail to head) or reverse (head to tail) direction, respectively.

I often find myself wanting to use `edge_dfs` with the `original` and with the `reverse` direction (dynamically set) in the same loop, but in the former case I get a tuple of two or three values, and in the latter case I get a tuple of three or four  values to work with, depending on whether I am working with a multigraph. I think it would be cleaner if `edge_dfs` always returned the direction value if the `direction` parameter to `edge_dfs` is not `None`. 

Here is a wrapper that I think would do what I am talking about (untested code):

```python
def my_edge_dfs(G, source=None, orientation=None):
    add_orientation = False
    if orientation is None:
        orientation = 'original'
    elif orientation == 'original':
        add_orientation = True
    for edge in nx.edge_dfs(G, source=source, orientation=orientation):
        if add_orientation:
            edge = edge + (orientation,)
        yield edge
```

What do people think?

dschult:
- add docs to explain that dfs_edges is not edge_dfs
Fixes #2840 
- clean up edge_dfs return values to unify length of yielded tuple
Fixes #3037  
- add edge_bfs to match edge_dfs
Fixes #2792

### DIFF
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -329,9 +329,11 @@ def recursive_simple_cycles(G):
     return result
 
 
-def find_cycle(G, source=None, orientation='original'):
-    """
-    Returns the edges of a cycle found via a directed, depth-first traversal.
+def find_cycle(G, source=None, orientation=None):
+    """Returns a cycle found via depth-first traversal.
+
+    The cycle is a list of edges indicating the cyclic path.
+    Orientation of directed edges is controlled by `orientation`.
 
     Parameters
     ----------
@@ -343,30 +345,29 @@ def find_cycle(G, source=None, orientation='original'):
         is chosen arbitrarily and repeatedly until all edges from each node in
         the graph are searched.
 
-    orientation : 'original' | 'reverse' | 'ignore'
+    orientation : None | 'original' | 'reverse' | 'ignore' (default: None)
         For directed graphs and directed multigraphs, edge traversals need not
-        respect the original orientation of the edges. When set to 'reverse',
-        then every edge will be traversed in the reverse direction. When set to
-        'ignore', then each directed edge is treated as a single undirected
-        edge that can be traversed in either direction. For undirected graphs
-        and undirected multigraphs, this parameter is meaningless and is not
-        consulted by the algorithm.
+        respect the original orientation of the edges.
+        When set to 'reverse' every edge is traversed in the reverse direction.
+        When set to 'ignore', every edge is treated as undirected.
+        When set to 'original', every edge is treated as directed.
+        In all three cases, the yielded edge tuples add a last entry to
+        indicate the direction in which that edge was traversed.
+        If orientation is None, the yielded edge has no direction indicated.
+        The direction is respected, but not reported.
 
     Returns
     -------
     edges : directed edges
-        A list of directed edges indicating the path taken for the loop. If
-        no cycle is found, then an exception is raised. For graphs, an
-        edge is of the form `(u, v)` where `u` and `v` are the tail and head
-        of the edge as determined by the traversal. For multigraphs, an edge is
-        of the form `(u, v, key)`, where `key` is the key of the edge. When the
-        graph is directed, then `u` and `v` are always in the order of the
-        actual directed edge. If orientation is 'ignore', then an edge takes
-        the form `(u, v, key, direction)` where direction indicates if the edge
-        was followed in the forward (tail to head) or reverse (head to tail)
-        direction. When the direction is forward, the value of `direction`
-        is 'forward'. When the direction is reverse, the value of `direction`
-        is 'reverse'.
+        A list of directed edges indicating the path taken for the loop.
+        If no cycle is found, then an exception is raised.
+        For graphs, an edge is of the form `(u, v)` where `u` and `v`
+        are the tail and head of the edge as determined by the traversal.
+        For multigraphs, an edge is of the form `(u, v, key)`, where `key` is
+        the key of the edge. When the graph is directed, then `u` and `v`
+        are always in the order of the actual directed edge.
+        If orientation is not None then the edge tuple is extended to include
+        the direction of traversal ('forward' or 'reverse') on that edge.
 
     Raises
     ------
@@ -394,7 +395,17 @@ def find_cycle(G, source=None, orientation='original'):
     [(0, 1, 'forward'), (1, 2, 'forward'), (0, 2, 'reverse')]
 
     """
-    out_edge, key, tailhead = helper_funcs(G, orientation)
+    if not G.is_directed() or orientation in (None, 'original'):
+        def tailhead(edge):
+            return edge[:2]
+    elif orientation == 'reverse':
+        def tailhead(edge):
+            return edge[1], edge[0]
+    elif orientation == 'ignore':
+        def tailhead(edge):
+            if edge[-1] == 'reverse':
+                return edge[1], edge[0]
+            return edge[:2]
 
     explored = set()
     cycle = []
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -20,7 +20,6 @@ from itertools import tee
 
 import networkx as nx
 from networkx.utils import not_implemented_for, pairwise
-from networkx.algorithms.traversal.edgedfs import helper_funcs
 
 __all__ = [
     'cycle_basis', 'simple_cycles',
--- a/networkx/algorithms/tests/test_cycles.py
+++ b/networkx/algorithms/tests/test_cycles.py
@@ -221,6 +221,12 @@ class TestFindCycle(object):
         x_ = [(0, 1, FORWARD), (1, 0, FORWARD)]
         assert_equal(x, x_)
 
+    def test_digraph_reverse(self):
+        G = nx.DiGraph(self.edges)
+        x = list(find_cycle(G, self.nodes, orientation='reverse'))
+        x_ = [(1, 0, REVERSE), (0, 1, REVERSE)]
+        assert_equal(x, x_)
+
     def test_multidigraph_ignore(self):
         G = nx.MultiDiGraph(self.edges)
         x = list(find_cycle(G, self.nodes, orientation='ignore'))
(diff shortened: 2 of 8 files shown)


h-vetinari:
Hey, I was just thinking about the old issue about this - what a fortuitous coincidence! :)

Very happy to see that you picked this up - even more so because I wouldn't have had an idea of all the additional functionality (MultiGraphs etc) that you needed.

Unfortunately, this still does not solve my original problem -- that edges come breadth-first, guaranteed. I used your definition of `edge_bfs` to test, and found the following counter-example (haven't had time to pare this down to the minimal counter-example possible, unfortunately...):
```
G = nx.DiGraph()
e_in = [(1, 2), (2, 4), (2, 5), (1, 6), (6, 7), (4, 9), (5, 9), (7, 8), (8, 9), (8, 10), (9, 10)]
G.add_edges_from(e_in)
e_out = [e for e in edge_bfs(G, 1)]
print(e_out)
# [(1, 2), (1, 6), (2, 4), (2, 5), (6, 7), (4, 9), (5, 9), (7, 8), (9, 10), (8, 9), (8, 10)]
### (9, 10) __must__ come after (8,9) ###

# visualisation
import graphviz as gv
gv.Source(nx.drawing.nx_pydot.to_pydot(G).to_string())
```



The order does not change with `orientation='original'`:
```
[e for e in edge_bfs(G, 1, orientation='original')]
# [(1, 2, 'forward'), (1, 6, 'forward'), (2, 4, 'forward'), (2, 5, 'forward'), (6, 7, 'forward'), (4, 9, 'forward'), (5, 9, 'forward'), (7, 8, 'forward'), (9, 10, 'forward'), (8, 9, 'forward'), (8, 10, 'forward')]
```

I may not have the complete overview over graph-theory, and maybe your implementation works exactly as intended in my case above. But if so, would it be possible to add a keyword-option `orientation='enforce'` or something like that?

To clarify what I need: In my application, it is crucial that no edge is added to the queue whose originating node has not received all in-edges (as counted by the appearance of the corresponding edge in the `edge_bfs`-iteration).

dschult:
Ack!  I didn't see you comment before I merged -- But I'll just open a new one.

Thanks for testing this out! 

Let me make sure I understand your need: For the example you show above is the problem that edge (9, 10) comes before (8, 9)?  Any other edges off?

dschult:
If I understand correctly, you want a ```topological_sort```.
Typically people using a topological sort want a list of nodes. 
But I guess i can imagine wanting the edges. Are you sure you want the edges? Maybe there is some way to get them from the ordered nodes.

dschult:
I think the following gives a list of the edges for a topological sort. I would like a reference for this but I can't find any.

    nodelist = list(nx.topological_sort(G))
    while nodelist:
        n = nodelist.pop(0)
        for nbr in nodelist:
            if nbr in G[n]:
                print((n,nbr))

h-vetinari:
> Let me make sure I understand your need: For the example you show above is the problem that edge (9, 10) comes before (8, 9)? Any other edges off?

In the example above, there was nothing else off -- it doesn't matter what the exact order is between different side branches, but the crucial point is the following: Imagine the nodes being workers who take their incoming edges as inputs and produce the outgoing edges, and that the work cannot be executed before all inputs have been received (in the case above, node 9 would need to be executed to produce `(9,10)`, but cannot because it has not received the input from 8 yet).

> If I understand correctly, you want a topological_sort.
Typically people using a topological sort want a list of nodes.
But I guess i can imagine wanting the edges. Are you sure you want the edges? Maybe there is some way to get them from the ordered nodes.

I didn't know `topological_sort`, but looking at https://networkx.github.io/documentation/latest/reference/algorithms/generated/networkx.algorithms.dag.topological_sort.html
> > A topological sort is a nonunique permutation of the nodes such that an edge from u to v implies that u appears before v in the topological sort order.

this looks to be exactly would solve the specification above (regarding the nodes, at least).

> I think the following gives a list of the edges for a topological sort. I would like a reference for this but I can't find any.

I played around with it a little(!) bit, and it looks good so far. Thanks a lot! Do you think this could make it into 2.2 still?

dschult:
Actually, upon further reading, it looks like you should make the "tasks" into the nodes and the people be the edges. This is the standard way to handle your problem. If the directed graph comes with people as the nodes, then you can shift to the "line graph" which takes the edges as nodes of a new graph. The nodes are connected if the edges in the original graph share a node (that is, together make a path of length 2).  Anyway, try:

    print(list(nx.topological_sort(nx.line_graph(G))))
    [(1, 6), (6, 7), (7, 8), (8, 10), (8, 9), (1, 2), (2, 5), (5, 9), (2, 4), (4, 9), (9, 10)]

I don't think I will put this into networkx. But maybe a comment in the docs for top_sort...

h-vetinari:
I'm not sure I understand you correctly -- you mean "people" = "workers" (=nodes for me) and "tasks" = "inputs" (edges for me)?

If so, inverting the conceptual setup like this is not really possible, and would make application/interpretation/visualisation that much harder.

But just calling `line_graph` as an intermediate step in the calculation is perfectly fine -- I don't need to have a purpose-built function. I'd just like to have a way that works out of the box, and your `list(nx.topological_sort(nx.line_graph(G)))` does so beautifully.

I didn't know either of those functions, and would never have been able to come up with that reasoning, but I can (sort of) understand it now, and appreciate its beauty. Thanks!

dschult:
Looking at the top couple of hits for google on "topological sort scheduling" makes me think this is the "usual" way to view scheduling problems. 

You wanted a list of edges where each edge represented a input, and each node represented a person who could supply that input (given appropriate inputs). To view it the other way, think of the inputs as the nodes. You are looking for an order of the inputs -- just as before, only now those are nodes instead of edges. The edges in the scheduling problem are the dependencies of the inputs. Which must be done before the others. You could also associate a person with each dependency (though I suspect the same person would associate with many dependencies.

In any case, I'm glad the "top_sort of line_graph" approach works.  There is no reason you have to morph your problem just to solve it. :)  Best Wishes ---

Issue #2352, opened 2017-01-18
Depth-limited Breadth First Search


Abdulhakeem135:
It would be very good to have a new parameter to the breadth-first search function like "distance" which is  None by default – the maximum distance from the start nodes to traverse. 

bfs_tree(G, source,**distance=None**, reverse=False)

I want to start breadth first search from start node for a specific distance.

jfinkels:
I believe this is the same as #2258, but for breadth-first search instead of depth-first search.

Abdulhakeem135:
Yes, it is exactly the same. I also found this #2116. But it might be better to rename the variable as breadth_limit, since we have here a breadth-first search technique. I am not sure if this has been done or not yet. This is my first contribution to open-source and still do not know what is going one here.

mohdkashif93:
If nobody else is working on this issue, can I work on it ?

dschult:
Yes, that would be helpful! Why don't you structure it as a module within ```algorithms/traversal``` with naming similar to ```edgedfs.py``` along with a test module ```tests/test_edgebfs.py``` Look around at other code for example, the github networkx wiki for odds and ends and ask questions here.

Thanks very much!

dschult:
Sorry... I got confused in my previous answer about depth_limit for ```bfs_*``` functions. Please see my edits to that message.  Thanks

gmyr:
I would also be interested to help out on this issue, but Mohammed was quicker. :)
Are there any other issues on this project a new-comer to open source development could tackle?

dschult:
Take a look at the issues with the tag: "Good first issue", or "Needs PR".
Some are quite involved -- others more straightforward.

Some very interesting ones that will get you reading the literature as well as coding are #1609 and #1457

dschult:
@gmyr would you like to take a look at this one (depth limit on breadth-first-search) #2352
we've got a PR but it has been almost 3 weeks since any activity and I think the code would be quick to get right.  Let me know -- if you don't, I will get to it :)

mohdkashif93:
Sorry for the delay. I am still working on it and will get this done by this weekend positively. Sorry for the inconvenience caused.

gmyr:
I will take a shot at it if it's still open after the weekend. :)

PR #3045, opened 2018-07-03
Depth limit for Breadth First Search

jaideep-seth:
#2352
@dschult this is my first attempt at NetworkX, is the PR alright?

PS:  networkx/algorithms/traversal/tests/breadth_first_search.py is a duplicate of networkx/algorithms/traversal/breadth_first_search.py and i have used it as i was unable to access the file from the parent directory. It can be deleted . 

-Thank you!


dschult:
Thanks for this --  Some comments:

- remove all your imports (```from breadth_first_search``` ) and just use ```nx.bfs_edges``` or ```nx.bfs*```
  You can remove the new file from tests that way too.  Go through the nx interface instead of directly 
  to the files.
- leave blank line after your name in contribute.rst file. Also remove the ```:orphan:``` line at beginning.
- your line ```step = list(neighbors(source))``` uses the neighbors of ```source```.  That doesn't work for depths greater than 2 -- or am i missing something. see [this stackoverflow question](https://stackoverflow.com/questions/10258305/how-to-implement-a-breadth-first-search-to-a-certain-depth/16923440)
- The examples/tests in the docstring are fairly exhaustive. Might be better to include one example and put the rest of the more test-like examples into the test files.

dschult:
See #3077

Issue #2720, opened 2017-10-24
Return all permutations of topological_sort


suranap:
Currently topological_sort returns only one ordering. Would there be interest in a version that returns all possible correct sorts? Then the standard `topological_sort` function would just return the first one.

I have a need for this currently. If there's interest I can make a PR when I've written it.

hagberg:
Yes, this would be an interesting addition to the package.  Please update with a PR if you like.

aweltsch:
If there is noone working on it currently, I would like to work on this.

dschult:
Sure!  Thanks!
-- do you have a good algorithm to try to implement for it? This is potentially memory intensive as well as potentially time intensive.

aweltsch:
I'm currently implementing https://doi.org/10.1016/0020-0190(74)90001-5
My plan is to implement both the recursive as well as the iterative version and then give the option to choose the algorithm (default will be iterative)
I already implemented the recursive version (as it is significantly easier to do). Obviously, the recursive procedure comes with the usual problems related to recursion depth (linear in the number of nodes). Also I don't know any easy way to use the yield statement with recursive function calls, so there will be huge memory requirements (as there are up to`n!` on a graph with `n` nodes) 

I did not yet examine the dynamic programming/iterative procedure, but I guess there we can use the yield keyword, so we don't have to hold all topological sorts in memory.

If there is no interest in the recursive version at all, I'll just leave it and will only include the iterative version. I just chose to implement the recursive one, because it's so easy to do, and then we have at least one working version, until I find time to implement the iterative one.

dschult:
If the recursive version is short and easy to read let's keep it in (not exposed in ```__all__```) and use it in the tests. I think it's worthwhile to have multiple versions for both testing and pedagogy.

Thanks!

aweltsch:
Ok, I decided to replace the recursive algorithm with the iterative one, as they are very similar. I kept the recursive one in a separete commit, to record it in the project history. I hope this is ok.

Another question: I'm a very big fan of property based testing on randomized instances. Does it make sense to add something like this to the tests?
I'm thinking about creating a random dag by selecting random pairs of the form `(i, j)` with `i < j`.
For graphs up to 10 nodes, we could then compare the algorithm output with a naive filtering of permutations (take permutations of nodes that are topological sorts of the graph (easy to check))
[...]

PR #3049, opened 2018-07-05
Enumerate all topological sorts of a digraph

aweltsch:
Issue #2720 
Implementation of an enumeration algorithm for topological sorts. Based on https://doi.org/10.1016/0020-0190(74)90001-5

### DIFF
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -288,6 +288,7 @@ def lexicographical_topological_sort(G, key=None):
 @not_implemented_for('undirected')
 def all_topological_sorts(G):
     """Returns a list of all topologicals sortings of the directed graph G.
+
     Implements the recursive algorithm given in [1].
 
     Parameters
@@ -446,6 +447,7 @@ def is_aperiodic(G):
     else:
         return g == 1 and nx.is_aperiodic(G.subgraph(set(G) - set(levels)))
 
+
 @not_implemented_for('undirected')
 def transitive_closure(G):
     """ Returns transitive closure of a directed graph
@@ -479,6 +481,7 @@ def transitive_closure(G):
         TC.add_edges_from((v, u) for u in nx.dfs_preorder_nodes(G, source=v)
                           if v != u)
     return TC
+
 
 @not_implemented_for('undirected')
 def transitive_reduction(G):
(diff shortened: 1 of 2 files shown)


aweltsch:
I felt like I checked everything, but I just saw that I remoevd 2 lines from dag.py
Should I add one of the ugly "Fix flake8 formatting"-commits, or is there another way now that I have already published these changes in my repo?

dschult:
You should just change them back and make another commit with those changes.  It is true that this makes more total commits -- and there are ways to "squash commits" together if you want to search for that -- but it undoes what you already did as if nothing happened. And when github merges it squashes all your commits by default anyway -- so the long term history is just one commit for the entire PR.

Other items of note:
- Check the PEP8 style of your changes (perhaps using the convenient program pep8 available as 
  ```pip install pep8```) 
- Line 342: ```q = current_sort[-1]``` and the next line . ---  Isn't this just ```q=current_sort.pop()```?
- Line 380:   isn't that just an ```append```?
- Add space after the first line of the docstring. The first line gets used as a short description and the rest should appear as a paragraph, but separated from the first line with a blank line
- Probably good to add an example to the doc_string.

:)

aweltsch:
I readded the lines and improved the docstring alittle.
concerning lines 342 and 380 (now 357 and 395): This is neither a `pop` nor `append`, as both `pop` and `append` change `current_sort` in-place. We do not want this, because we yield current sort.
Actually it might make more sense to return immutable tuples instead of a list, this would make it possible to apply use `pop`/`append`. Either way, it makes sens to include a comment to explain this issue. The way it's done now, does not seem to be obvious.

dschult:
You could ```yield list(current_sort)``` to make/yield a copy of the list. 
That clarifies the separation of the list in the algorithm from the lists that have been released.
tuple would work too of course.

Does this algorithm work for MultiGraph? We should either have a not_implemented_for, or include tests to ensure it does what we expect.

aweltsch:
Ok, I changed the code to yield a copy of the list.
I adjusted the code to work for multigraphs, by iterating over all out_edges of a node to decrease the degrees correctly when "removing" the node from the graph.

dschult:
This looks good to me.  Anything else before I merge?

aweltsch:
I have nothing else, I'd say it's finished.
Thanks for reviewing :+1:

dschult:
Thanks for submitting!!! :)

PR #3056, opened 2018-07-10
Update dependencies and test 3.7 on xenial

jarrodmillman:
Xenial is not officially supported yet.

See https://github.com/travis-ci/travis-ci/issues/9815

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,19 +11,11 @@ environment:
 
   matrix:
     - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.13"
+      PYTHON_VERSION: "2.7.15"
       PYTHON_ARCH: "32"
 
     - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.13"
-      PYTHON_ARCH: "64"
-
-    - PYTHON: "C:\\Python34"
-      PYTHON_VERSION: "3.4.4"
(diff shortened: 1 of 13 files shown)


jarrodmillman:
Pygraphviz doesn't work on Python 3.7 yet:  pygraphviz/pygraphviz#164

dschult:
Removing the ```default_opener``` command allows the tests to finish.
All ```default_opener``` does on linux is run ```xdg-open```, but that could be opening the file using a variety of programs depending on the xenial setup.  I don't see any reason that ```adg-open``` would hang and cause the tests to not complete.  I'll need some help tracking down how the configuration of xenial will allow ```xdg-open``` to finish and return control.

The osx test on 3.7 is still not working, but it looks like it just doesn't run the tests.  The log says  ```script.sh``` is started, but there is no output after that and the test is marked as failed.

jarrodmillman:
I thought I already had a fix for xdg-open and pngs in
```
tools/travis/linux_install.sh
```
The relevant part is
```
  # needed for view_graphviz and default_opener
  DIR=~/.local/share/applications
  mkdir -p $DIR
  FILE=$DIR/png.desktop
  cat <<EOF >$FILE
[Desktop Entry]
Name=png
MimeType=image/png;
Exec=/usr/bin/file
Type=Application
Terminal=true
NoDisplay=true
EOF

  xdg-mime default png.desktop image/png
```

I'll take a look and see what changed.

PR #3058, opened 2018-07-10
Unify random state treatment across package

dschult:
The goal is to allow a single RNG created with  ```numpy.random.RandomState()``` to produce all pseudorandom outcomes if desired. Default behavior is still to use the global RNGs supplied by ```numpy``` and Python's ```random``` package.  The easiest way to make those outcomes deterministic throughout the NetworkX package is to set the state of each of those two generators, and forget about setting anything else (use the default ```seed``` argument of None).  To make every random number created by a single RNG, create that numpy RNG and pass it into each ```seed``` argument for every function you call that has one. You cannot use a ```random``` style RNG for all functions.

Based on discussion in #1764 

This PR involves the following:

 - finding and update all code that uses random numbers but doesn't yet allow seed to be set.
 - creating decorators and utility functions to allow easy use by developers of random numbers. Contributors can use either the ```random``` package or ```numpy.random```.  Decorators take care of processing the input ```seed``` to turn it into an appropriate RNG.
 - update all code using random numbers to use these tools so they work together.

This PR does not update all code to use the ```numpy.random``` interface by default. Indeed, most code uses the ```random``` interface. That conversion is more than I'm up for before this release.


doc/reference/randomness.rst
```diff
@@ -0,0 +1,86 @@
...
+Many users will be satisfied with this level of control.
+
+For people who want even more control, we include an optional argument
+to functions that use an RNG.  This argument is called `seed`, but
+determines more than the seed of the RNG. It tells the function which
+RNG package to use, and whether to use a global or local RNG.
+
+.. nbplot::
+
+    >>> seed=None    # use the global default RNG of either package
```
jfinkels: If I'm understanding this correctly, these are examples of how I might set the `seed` keyword argument in an invocation of a NetworkX function. It might be clearer to show these examples as used in an actual function call. For example,

```python
>>> random_graph(n, m, p, seed=None)
>>> random_graph(n, m, p, seed=42)
# etc.
```

dschult: Good -- thanks!  I agree...

dschult:
I'm done tweaking this PR and it is ready for people to look at.
It fixes a lot of places where we used random numbers but didn't allow setting the RNG.
It makes setting the seed affect a local RNG instead of the global RNG.
It allows a single numpy RNG to be used for all functions if desired.
The name for the control argument for RNGs is ```seed``` rather than ```random_state```.

At this point it does not rewrite code using Python's ```random``` to use ```numpy```.

Feedback requested. :)

PR #3066, opened 2018-07-14
Bugfix for issue 3065

### LINKED ISSUE #3065: kamada_kawai_layout fails with dim=1
`nx.kamada_kawai_layout(G, dim=1)` Fails when called without initializations through `pos` argument.
This is cased by the default initialization which calls [circular_layout](https://github.com/networkx/networkx/blob/27dd93dda901171631fba92dc02a521cfc9e057e/networkx/drawing/layout.py#L667).

But `circular_layout()` ignores parameter `dim=1` and overwrites it with 2 without any notification. 
`circular_layout` should throw an exception if it cannot handle `dim=1`.

```
import networkx as nx

def create_bipartite_graph(n, m):
    assert n <= m
    dist_matrix = np.ones((n+m, n+m), dtype=float) * 255
    matches = n + np.random.permutation(m)[:n]
   
    dist_matrix[np.arange(n), matches] = 75
    dist_matrix[matches, np.arange(n)] = 75
    
    G = nx.Graph()
    
    edges = (dist_matrix <= 75).nonzero()
    dists = dist_matrix[edges]
    edges = (edges[0], edges[1], 
             [{'weight': x} for x in dists])
    edges = zip(*edges)
    G.add_edges_from(edges)
    return G

G = create_bipartite_graph(10, 10)
pos = nx.kamada_kawai_layout(G, dim=1) # THIS FAILS
pos = nx.kamada_kawai_layout(G, dim=1, pos=np.random.rand(20)) # THIS WORKS
```

gfetterman:
Fix #3065 

`circular_layout()` now raises an exception when `dim < 2`.

Also, `kamada_kawai_layout()` now uses a uniformly-spaced linear layout when `dim < 2` and `pos` hasn't been specified, rather than passing a bad parameter value to `circular_layout()`.

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -130,6 +130,7 @@ def circular_layout(G, scale=1, center=None, dim=2):
         Dimension of layout.
         If dim>2, the remaining dimensions are set to zero
         in the returned positions.
+        If dim<2, a ValueError is raised.
 
     Returns
     -------
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -674,7 +674,7 @@ def kamada_kawai_layout(G, dist=None,
     if pos is None and dim >= 2:
         pos = circular_layout(G, dim=dim)
     elif pos is None:
-        pos = {n: pt for n,pt in zip(G, np.linspace(0, 1, len(G)))}
+        pos = {n: pt for n, pt in zip(G, np.linspace(0, 1, len(G)))}
     pos_arr = np.array([pos[n] for n in G])
 
     pos = _kamada_kawai_solve(dist_mtx, pos_arr, dim)
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -629,7 +629,7 @@ def kamada_kawai_layout(G, dist=None,
     pos : dict or None  optional (default=None)
         Initial positions for nodes as a dictionary with node as keys
         and values as a coordinate list or tuple.  If None, then use
-        circular_layout().
+        circular_layout() for dim > 2 and a linear layout for dim == 1.
 
     weight : string or None   optional (default='weight')
         The edge attribute that holds the numerical value used for
@@ -671,8 +671,10 @@ def kamada_kawai_layout(G, dist=None,
                 continue
             dist_mtx[row][col] = rdist[nc]
 
-    if pos is None:
+    if pos is None and dim >= 2:
         pos = circular_layout(G, dim=dim)
+    elif pos is None:
+        pos = {n: pt for n,pt in zip(G, np.linspace(0, 1, len(G)))}
     pos_arr = np.array([pos[n] for n in G])
 
     pos = _kamada_kawai_solve(dist_mtx, pos_arr, dim)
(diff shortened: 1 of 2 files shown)


gfetterman:
(Not sure if a WIP tag is necessary for something this small, but it's my first PR here and I did want to run a couple of things by a more experienced contributor.)

(1) While making the change, I noticed that `shell_layout()` also makes the same assumption that `dim=2`. Is that worth making more explicit in code with an exception, or is its documentation clear enough?

(2) I followed @dschult 's suggestion in the comment on #3065 for the fix to apply, but two other approaches seem equally valid: projecting a circle onto a single dimension (which would closer approximate the spirit of a "one-dimensional circle"), or producing a uniformly-spaced linear layout (which would better match the function's behavior in 2-D). What do you think?

(3) If `circular_layout()` is still going to raise an exception for `dim=1`, is the linear layout I added worth pulling out into a new function, or is it too trivial or unlikely to ever be used anywhere else?

(4) I didn't see any tests for `kamada_kawai_layout()` (just for its cost function), so didn't add anything to test this new functionality. Is this alright?

networkx/drawing/layout.py
```diff
@@ -621,7 +629,7 @@ def kamada_kawai_layout(G, dist=None,
     pos : dict or None  optional (default=None)
         Initial positions for nodes as a dictionary with node as keys
         and values as a coordinate list or tuple.  If None, then use
-        circular_layout().
+        circular_layout() for dim > 2 and a linear layout for dim == 1.
```
dschult: ```>= 2```   instead of ```> 2```

networkx/drawing/layout.py
```diff
@@ -663,8 +671,10 @@ def kamada_kawai_layout(G, dist=None,
                 continue
             dist_mtx[row][col] = rdist[nc]
 
-    if pos is None:
+    if pos is None and dim >= 2:
         pos = circular_layout(G, dim=dim)
+    elif pos is None:
+        pos = {n: pt for n,pt in zip(G, np.linspace(0, 1, len(G)))}
```
dschult: Check for PEP8 code style (space after comma and other things) perhaps using the pep8 command line tool. install with ```pip install pep8```, or read the Python PEP8 document itself.

Also, you can avoid evaluating ```pos is None``` twice by nesting your if statements. Not a big deal either way.

dschult:
Thanks for this!

Thoughts:

- we should add a quick test for ```dim==2``` in ```shell_layout```. And we should add a sentence in the circular layout description of ```dim``` that ```dim<2``` causes an error.
- trying to infer the rules for a dim=1 circular layout may cause surprises. Let's just raise an exception.
- uniformly spaced dim=1 case is easy enough -- let's avoid making a new function for it.
- add a test for the dim=1 case for ```kamada_kawai_layout```.  There are "smoke tests" for kamada_kawai_layout in the test methods that check all functions. The added test can similarly be a "smoke test". Just run it with dim=1 and don't check the output.

gfetterman:
Thanks for the feedback and answers. I believe I've made the requested changes.

dschult:
OK... this looks good to me. I'll look into the codecov stuff -- but can't see why those lines aren't being hit by the tests.  Anything else? Tale the WIP off the title when you are ready for me to merge.

gfetterman:
Don't think so! Thanks for your help.

PR #3068, opened 2018-07-16
faster, type-preserving from_pandas_edgelist

### LINKED ISSUE #3013: nx.from_pandas_edgelist() iterates over DataFrame.values which changes dtypes
I arrived here through:

```
>>> df = pd.DataFrame([[1, 2, '0.1']], columns=['source', 'target', 'other'])
>>> nx.from_pandas_edgelist(df).nodes()
NodeView((1, 2))

>>> df = pd.DataFrame([[1, 2, 0.1]], columns=['source', 'target', 'other'])
>>> nx.from_pandas_edgelist(df).nodes()
NodeView((1.0, 2.0))
```

I initially thought this would be an issue with [`from_pandas_edgelist`](https://github.com/networkx/networkx/blob/9ac4941c1cb25bf47c85d2bb19f7557f5c807694/networkx/convert_matrix.py#L241) but that code just hands off directly to add_edge:

https://github.com/networkx/networkx/blob/9ac4941c1cb25bf47c85d2bb19f7557f5c807694/networkx/convert_matrix.py#L346-L350

which means that [Graph().add_edge](https://github.com/networkx/networkx/blob/e97abea688a471a8422a676ea7c93d079bfa0f1f/networkx/classes/graph.py#L812-L874) must be responsible.

This is in networkx version 2.1

zfrenchee:
Closes #3013

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -323,16 +323,15 @@ def from_pandas_edgelist(df, source='source', target='target', edge_attr=None,
     if edge_attr is True:
         cols = [c for c in df.columns if c is not source and c is not target]
     elif isinstance(edge_attr, (list, tuple)):
-        cols = [c for c in df.columns
-                if c in edge_attr and c is not source and c is not target]
-    elif isinstance(edge_attr, int) and (0 <= edge_attr < len(df.columns)):
-        cols = [df.columns[edge_attr]]
-    elif isinstance(edge_attr, str):
-        cols = [c for c in df.columns if c in [edge_attr]]
+        cols = edge_attr
     else:
-        raise NetworkXError("Invalid edge_attr argument")
+        cols = [edge_attr]
 
-    eattrs = zip(*[df[col] for col in cols])
+    try:
+        eattrs = zip(*[df[col] for col in cols])
+    except (KeyError, TypeError) as e:
+        msg = "Invalid edge_attr argument: %s" % edge_attr
+        raise nx.NetworkXError(msg)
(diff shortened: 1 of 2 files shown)


zfrenchee:
@dschult please review both commits independently: 
- The first commit makes a simple improvement for the use case I wanted. 
- The second commit is a more dramatic makeover of the whole method, which I think is much more elegant. 
If you don't like the way the new method looks, please still consider the first commit alone. I'd be happy to rollback the second and merge just the first.

dschult:
I like the second version better -- easier to read and hopefully understand.
I haven't checked details to make sure there aren't any strange differences.

How hard would it be to allow ```int``` column specifiers?

-----------------
The testing errors are due to Python 2 choking on:

    for s, t, *attrs in zip(df[source], df[target], *[df[col] for col in cols]):

which could be replaced (if I understand correctly) with 

    for edgeinfo in zip(df[source], df[target], *[df[col] for col in cols]):
        s, t = edgeinfo[:2]
        attr = edgeinfo[2:]

zfrenchee:
> How hard would it be to allow int column specifiers?

Easy, would just need to add another case in the if block. 

> which could be replaced (if I understand correctly) with

That would definitely preserve functionality. Unfortunately though, taking slices here will definitely slow things down I think.

dschult:
Another option?:     

    for s, t, attrs in zip(df[source], df[target], zip([df[col] for col in cols])):

zfrenchee:
@dschult 

> Another option?:

I like that idea! That didn't exactly work, but what I have now does seem to, and hopefully it's python2 compliant.

PR #3070, opened 2018-07-16
Degree centrality bug fix.

daniel-leicht:
1) Fixed a division by zero bug in the degree_centrality functions (division by zero was raised when len(G)==1), if there is one node in the graph it's centrality is set to 1.

2) Removed redundant dict initialization line.


dschult:
Thanks for this!  

Could you add tests for these cases in the tests/test_degree_centrality.py file?

PR #3072, opened 2018-07-17
Raise ValueError when empty list is passed to functions in all.py, fixes #2973

### LINKED ISSUE #2973: Behavior of `nx.union_all` when `graphs=[]`
I'm running into a case where I'm passing in an empty list to `nx.union_all`, and it returns a None type. 

While this is not necessarily the wrong thing to do, it is not documented.

Intuitively, I would expect the result of union with no inputs to be an empty graph, but the issue here is that you don't know what type the graph should be. Therefore I think the best behavior would be to raise a ValueError indicating that the input cannot be empty. This would make it more clear where the code is failing. 

Current behavior:

```python

>>> nx.union_all([nx.path_graph([1, 2])])
<networkx.classes.graph.Graph at 0x7f6fb15d1ac8>

>>> nx.union_all([nx.path_graph([1, 2]), nx.path_graph([3, 4])])
<networkx.classes.graph.Graph at 0x7f6fb1477ac8>

>>> print(nx.union_all([]))
None

```

Proposed Behavior:

```python
>>> print(nx.union_all([]))
ValueError: Cannot union_all an empty list
```

gmyr:
I raise a ValueError when any of the functions in all.py (i.e., union_all, disjoint_union_all, compose_all, and intersection_all) receives an empty list as an argument.

Fixes #2973.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -119,6 +119,7 @@ is partially historical, and now, mostly arbitrary.
 - Mads Jensen, Github: `atombrella <https://github.com/atombrella>`_
 - Edward L. Platt, `elplatt <https://elplatt.com>`_
 - James Owen, Github: `leamingrad <https://github.com/leamingrad>`_
+- Robert Gmyr, Github: `gmyr <https://github.com/gmyr>`_
 
 Support
 -------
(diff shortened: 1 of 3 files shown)


Erotemic:
Thanks for tackling this PR. This LGTM from a functional perspective.

dschult:
Thanks!

Issue #2975, opened 2018-05-16
all_simple_paths() took a lot of time to execute


hffqyd:
Hello, thanks for your awesome networkx. 

It took a lot of time while I was using networkx to find all paths in a graph. The code I used was like below:

```python
import networkx as nx
import matplotlib.pyplot as plt
% matplotlib inline

graph = nx.Graph()

nodes_list  = [1, 2, 3, 4, 5, 6]
edges_list  = [(1, 2), (1, 3), (1, 4), (2, 5), (3, 5), (3, 6), (4, 6)]

for n in nodes_list: 
    graph.add_node(n)
for i, e in enumerate(edges_list): 
    graph.add_edge(*e)

nx.draw_networkx(graph)

path_iter = nx.all_simple_paths(graph, 4, 5)

pathes = []
for p in path_iter:
    pathes.append(p)
print (pathes)
```

It took me 0.5 second to find all paths when the graph containing 53 nodes and 103 edges on my computer.  But when the nodes and edges increased to more than 1500 and 2800, respectively, it had been running for more than 20 hours already. 

Is there any way to decrease the search time to a tolerable level? 

Thank you very much.

ericmjl:
My guess here is that `all_simple_paths`'s complexity increases super-linearly with the number of nodes and edges in the graph. Inherent here is a fundamental problem with the implemented algorithm, and perhaps even the problem definition itself -- finding all simple paths might not necessarily have an efficient solution. 

For your actual problem, do you need to find the number of simple paths between two nodes? If we were to reframe the problem as "the number of paths between two nodes of length `k`", if I am calculating this correctly, raising the adjacency matrix to the power `k` gives you the number of paths that exist between two nodes of length `k`, and this is an extremely efficient operation to perform.

hffqyd:
Thank you very much! I used the `cutoff` argument of `all_simple_paths` to decrease the calculating time to a second.

winni2k:
Hi,
A big thanks for networkx from me as well!  

For my problem I really do want every path. I did a line_profile of all_simple_paths, and it looks like line 270 below is an obvious place to do some optimizing inside networkx (line 263 is probably my fault, as I'm using my own graph implementation).

```
Function: _all_simple_paths_multigraph at line 255

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   255                                           @profile
   256                                           def _all_simple_paths_multigraph(G, source, target, cutoff=None):
[...]

PR #3074, opened 2018-07-19
In all_simple_paths(). Replace list with OrderedDict for speedup

winni2k:
See discussion of this at https://github.com/networkx/networkx/issues/2975#issuecomment-406354042


dschult:
Looks like Codecov wants a test of the multigraph cases. (which wasn't there before but now that we touched that code, it is flagging it.  I'll add that soon...

dschult:
Thanks for doing this!!

winni2k:
np

winni2k:
This fix should remove a major roadblock in my work...

Issue #2869, opened 2018-02-06
Eulerization of a Graph


msramalho:
# Eulerization of a Graph
What I propose is that the **Euler** functionality be expanded to handle the transformation of a non-euler graph into one, with as little additions and removals as possible. This would be useful because many programs need such features and it would be a disruptive feature. 

[File to update](https://github.com/networkx/networkx/blob/master/networkx/algorithms/euler.py)

I would like to do it, but I don't know wether I'll have enough time.

[Instructions](http://www.math.unl.edu/~bharbourne1/M203Spr07/EulerizationHandout.pdf)

dschult:
Maybe this would be best to implement as its own function rather than an extension of the existing euler functionality. This might be a reasonable project for someone looking to contribute.

mtrenfield:
I am willing to work on this.  This would be my first PR for NetworkX and only my second on Github, so I may need a bit of guidance.

Three questions:

1. Should the graph be Eulerized in place by default?
2. Does it make sense to Eulerize weighted graphs?  I can think of a few reasonable ways one might re-assign weights to new edges, but nothing unique.  Should the function throw an exception or raise a warning if the edges are weighted?
3. Where should I look to get a sense of what tests I should write for the function and where should the tests go?

dschult:
I think you should put the code in ```networkx/algorithms/euler.py``` with the tests in ```networkx/algorithms/tests/test_euler.py```.  Take a look at what is in there to see how it is set up. Also read the NetworkX github Wiki to general style ideas. 

Read the links from the original poster.  Also search for "eulerization of a graph" or similar and read what those pages say about how to do this.

It looks like you will need to create a MultiGraph to be able to handle duplicate edges. I think you should ignore edge weights altogether unless you can find a reference that uses them. The tests should be unittests -- that is, tests of specific features and/or aspects of the code. The idea is that they should flag us if we change something that breaks your code. -- one way to do this is to check that the answers are right in some important specific cases. :)  vague enough?

Put your name with the authors at the top and if you want in the ```CONTRIBUTORS.rst``` file.
If you have questions about overall structure ask here -- or submit your unfinished PR and ask your questions in the PR.  In this case, people often put ```[WIP]``` at the beginning of the PR title.

Thanks!

mtrenfield:
That's very helpful.  I have started working on this.

PR #3076, opened 2018-07-20
[WIP] Added function to eulerize simple graphs to algorithms.euler

mtrenfield:
As discussed in [issue 2869](https://github.com/networkx/networkx/issues/2869), I have implemented a function to transform a graph into an Eulerian graph along the lines mentioned in [these lecture notes](http://www.math.unl.edu/%7Ebharbourne1/M203Spr07/EulerizationHandout.pdf) and using the method described [here](https://cs.stackexchange.com/questions/9126/is-there-a-non-brute-force-algorithm-for-eulerization-of-graphs) to find an efficient set of paths.

I have also written test cases for Eulerization of Erdos-Renyi graphs and some complete graphs.

I am labeling this as a WIP because this would be my first commit to NetworkX and expect that there may be things I haven't thought of or things I have done wrong.  I have also not implemented this for directed graphs.  This is because I use `nx.max_weight_matching` to find an Eulerization that doesn't require duplicated too many edges.  I am sure directed graphs can be Eulerized, but I can't see how to do this efficiently.

### DIFF
--- a/networkx/algorithms/euler.py
+++ b/networkx/algorithms/euler.py
@@ -10,14 +10,16 @@
 # Authors:
 #   Nima Mohammadi <nima.irt@gmail.com>
 #   Aric Hagberg <hagberg@lanl.gov>
+#   Mike Trenfield <william.trenfield@utsouthwestern.edu>
 """
 Eulerian circuits and graphs.
 """
+from itertools import combinations
+
 import networkx as nx
+from ..utils import arbitrary_element, not_implemented_for
 
-from ..utils import arbitrary_element
-
-__all__ = ['is_eulerian', 'eulerian_circuit']
+__all__ = ['is_eulerian', 'eulerian_circuit', 'eulerize']
 
 
 def is_eulerian(G):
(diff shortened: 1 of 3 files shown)


networkx/algorithms/euler.py
```diff
@@ -181,3 +183,78 @@ def eulerian_circuit(G, source=None, keys=False):
...
+            if n != m:
+                Gp.add_edge(m,n,weight=1/len(P), path=P)
+
+    # find the minimum weight matching of edges in the weighted graph
+    best_matching = nx.Graph(list(nx.max_weight_matching(Gp)))
+
+    # duplicate each edge along each path in the set of paths in Gp
+    for m,n in best_matching.edges():
+        path = Gp[m][n]["path"]
+        G.add_edges_from([(path[i], path[i+1]) for i,_ in enumerate(path[:-1])])
```
dschult: There is a ```networkx.utils.pairwise``` function for this:

   G.add_edges_from(nx.utils.pairwise(path))

networkx/algorithms/tests/test_euler.py
```diff
@@ -112,3 +112,15 @@ def test_multigraph_with_keys(self):
     @raises(nx.NetworkXError)
     def test_not_eulerian(self):
         f = list(eulerian_circuit(nx.complete_graph(4)))
+
+class TestEulerize(TestCase):
+
```
dschult: You should probably add the corner cases like ```G=nx.Graph()``` and ```G=nx.empty_graph(3)```.
Also, test that it raises an error for a directed graph.

dschult:
This look good!  
I made some comments connected to individual lines of code.

You should also bring the code style up to the PEP8 standards. Its easy as ```pip install pep8``` and then ```pep8 euler.py```.  Most of the trouble is typically spacing issues and 79 char limit on length of line.  You can also read the [PEP8 style guidelines in their official form](https://www.python.org/dev/peps/pep-0008/?). 

Also, the single-use function ```_simple_eulerize_simplegraph``` can be eliminated for simplicity by removing lines 230-233

The Codecov check (click on "Diff" heading of the details to see how Codecov thinks your tests are covering the code) is basically asking you to test the two NetworkXError cases.  It also asks you to test a case where ```n==m``` (which I think can't happen since you used ```combinations``` to create the data.

You should probably add simple tests that input a MultiGraph with multiedges to check that it is handled the way you want.  Just a simple one like a triangle with an extra edge would do it I think -- but maybe there's more there to test.

mtrenfield:
How do you think the case of the null graph be handled?  It could be considered as trivially Eulerian and I could return nx.MultiGraph(), but I could also raise an exception.  I don't have a strong preference for one over the other, but I see that nx.is_connected(nx.Graph()) raises a NetworkXPointlessConcept exceptions, so perhaps I should, too, for the sake of consistency.  Thoughts?

lexelby:
Amazing!  This is going to be super useful for [Ink/Stitch](https://github.com/inkstitch/inkstitch).  Thanks for your work!

Do you folks have any idea when this might make it into a stable release?

PR #3077, opened 2018-07-22
Depth-limited Breadth First Search

mohdkashif93:
Addresses [#2352](https://github.com/networkx/networkx/issues/2352). Added code for depth_limit for BFS along with the tests, references, and examples.

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -57,14 +57,18 @@ def generic_bfs_edges(G, source, neighbors=None, depth_limit=None):
     >>> G = nx.path_graph(3)
     >>> print(list(nx.bfs_edges(G,0)))
     [(0, 1), (1, 2)]
+    >>> list(nx.bfs_edges(G, source=0, depth_limit=2))
+    [(0, 1), (1, 2)]
 
     Notes
     -----
     This implementation is from `PADS`_, which was in the public domain
-    when it was first accessed in July, 2004.
+    when it was first accessed in July, 2004.  The modifications
+    to allow depth limits based on the Wikipedia article
+    "`Depth-limited search`_".
 
     .. _PADS: http://www.ics.uci.edu/~eppstein/PADS/BFS.py
-
+    .. _Depth-limited search: https://en.wikipedia.org/wiki/Depth-limited_search
     """
     visited = {source}
     if depth_limit is None:
@@ -112,6 +116,8 @@ def bfs_edges(G, source, reverse=False, depth_limit=None):
         >>> G = nx.path_graph(3)
         >>> list(nx.bfs_edges(G, 0))
         [(0, 1), (1, 2)]
+        >>> list(nx.bfs_edges(G, source=0, depth_limit=1))
+        [(0, 1)]
 
     To get the nodes in a breadth-first search order::
 
@@ -124,8 +130,11 @@ def bfs_edges(G, source, reverse=False, depth_limit=None):
 
     Notes
     -----
-    Based on http://www.ics.uci.edu/~eppstein/PADS/BFS.py
-    by D. Eppstein, July 2004.
+    Based on http://www.ics.uci.edu/~eppstein/PADS/BFS.py.
+    by D. Eppstein, July 2004.The modifications
+    to allow depth limits based on the Wikipedia article
+    "`Depth-limited search`_".
+    .. _Depth-limited search: https://en.wikipedia.org/wiki/Depth-limited_search
     """
     if reverse and G.is_directed():
         successors = G.predecessors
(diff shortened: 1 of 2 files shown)


networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -152,19 +173,28 @@ def bfs_tree(G, source, reverse=False):
     >>> G = nx.path_graph(3)
     >>> print(list(nx.bfs_tree(G,1).edges()))
     [(1, 0), (1, 2)]
+    >>> H = nx.Graph()
+    >>> nx.add_path(G, [0, 1, 2, 3, 4, 5, 6])
+    >>> nx.add_path(G, [2, 7, 8, 9, 10])
+    >>> dict(nx.bfs_tree(H, source=3, depth_limit=3)
```
dschult: This tests that the depth_limit is 3, but the answer only shows a depth_limit of 1. (1 step away from the source.)

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -54,30 +57,37 @@ def generic_bfs_edges(G, source, neighbors=None):
     >>> G = nx.path_graph(3)
     >>> print(list(nx.bfs_edges(G,0)))
     [(0, 1), (1, 2)]
+    >>> list(nx.bfs_edges(G, source=0, depth_limit=2))
```
dschult: This tests depth_limit of 2 but the full answer has depth limit of 2. So maybe its better to show depth_limit of 1 in this case.

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -190,17 +223,26 @@ def bfs_predecessors(G, source):
     >>> H.add_edges_from([(0, 1), (0, 2), (1, 3), (1, 4), (2, 5), (2, 6)])
     >>> dict(nx.bfs_predecessors(H, 0))
     {1: 0, 2: 0, 3: 1, 4: 1, 5: 2, 6: 2}
+    >>> G = nx.Graph()
+    >>> nx.add_path(G, [0, 1, 2, 3, 4, 5, 6])
```
dschult: Use ```G = nx.path_graph(7)``` instead of adding the first path (to save one line in the example). not a big deal.

mohdkashif93:
Thank you for your comments :-) I will update the code according to pep8 standard and modify the test cases as well. I am sorry I took so long to complete this.

dschult:
Instead of making the test case so simple, try not using the ```dict``` function.  Something like:

    >>> print(sorted(nx.bfs_predecessors(M, source=1, depth_limit=3)))
    [(0, 1), (2, 1), (3, 2), (4, 3), (7, 2), (8, 7)]

Eventually, all dicts will be ordered (python 3.6+), but the older versions have arbitrary ordering so for consistency in printing we often need to sort the results.

mohdkashif93:
Sorry about the multiple commits. It is my first time working with docstrings. Thank you for your comments :) I really appreciate that :D

mohdkashif93:
Please let me know if any other changes are to made :)

dschult:
Okay!  good news -- it passes all the tests. :)

I only see one more issue to overcome. The documentation is having trouble finding the reference for ```depth-limited searching```.  Looking at the [sphinx documentation for citations](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#citations) it looks like the space is not allowed in the name:  ```case-insensitive single words consisting of alphanumerics plus internal hyphens, underscores, and periods; no whitespace```.

Can you rename it with ```-``` or ```_``` replacing all white space?  The place to check for this is the first version in our Travis-CI tests. That one is run with Python 2.7 and has the special attribute that it compiles the documentation. If the active branch is networkx's master branch it also updates the latest documentation pages. But that doesn't happen until the PR is merged with master.  So for now we can't see the documentation, but we can see the error log as it is compiled.

In any case, can you change the citation name to remove white space?
Also, I didn't see an error for this, but most examples of sphinx I have seen use a space after the double-periods.  It looks like you got rid of this space both for the new citation and the existing one. Could you either put those spaces back or explain why they need to be changed? If they need to be changed then we have a lot of places where it should be changed.

You can check the documentation from your own machine using installed python 2.7 and sphinx. After installing packages listed in ```requirements/docs.txt``` go to the doc directory and use ```make html``` and to check the examples, use ```make doctest```.   :)

mohdkashif93:
I removed the space because the size of the line was more than 80 characters. I didn't knew that those spaces were there according to sphinx specification. I assumed it had to do with spacing and formatting. I changed the spacing based on your comments. I think there are no errors in the error log. Please let me know if I am missing something. Again, I really appreciate your comments for the same :) Thanks !

PR #3080, opened 2018-07-22
Allow individual extras selection

johnsom:
Currently networkx only has a definition for "all" dependency extras.
This patch adds support to specify the individual extras required.

Our use case is for pydot where we use nx_pydot but none of the other
"extras".


dschult:
Can't you just pip those individual packages?
NetworkX will work with those packages if they are installed.

johnsom:
Well, they are dependencies of this package if you use those features. Installing "all" extras fails for me with gdal issues (we don't need it). The issue I have is our larger project has standardized on pydot3 and pydotplus and does not allow our projects to use pydot directly because these other packages provide that capability. Since networkx explicitly requires pydot 1.2.3  (https://github.com/networkx/networkx/blob/master/networkx/drawing/nx_pydot.py#L368) this puts me in a tough situation if I can't have networkx install the correct dependency itself.

dschult:
Right -- I understand the trouble with requiring pydot 1.2.3 when pydot3 and pydotplus are your standards (that should be changed however, as the developers of those two packages are now working to make pydot the single package for python3 use of pydot).  We went through similar issues -- first using pydot, then pydotplus and now back to pydot.

How does this change in sphinx help you?  You still can't install with pydotplus, right?

johnsom:
My CI troubles go away if I can install networkx with "networkx[pydot]" in my requirements.
networkx nx_pydot.py code requires pydot, and a specific version.

johnsom:
Also, I'm not changing sphinx here. I just allowing the user to install the "extra" dependencies individually as opposed to having to install all of the with the existing "all" option.
Examples:
Current code: networkx[all]
This patch, in addition,  allows:
networkx[pydot] or networkx[pydot, pygraphviz], etc.

dschult:
Did you want a 'nose' version in there too?

johnsom:
Ah, missed one. Let me get that.

dschult:
Thanks!

Issue #1909, opened 2016-01-05
Iterative Deepening Search


sdsingh:
While looking through networkx code, I didn't find functionality for iterative deepening search. 

Is this functionality available? If not, would there be any interest in adding this to networkx? It is a really simple algorithm with reasonably heavy usage in practice.

dschult:
It is not available currently and could be added as a module in the algorithms/traversal area.
Thanks!

mohdkashif93:
If nobody else is working on this issue, can I work on it?

dschult:
Go for it!

mohdkashif93:
Hi @dschult , when I am trying to create a pull request it is taking changes from the depth limit in BFS feature as well (This was a previous pull request that was merged with the master branch). Is there something else that I need to figure out ? I am not able to understand what exactly am I doing wrong here.

mohdkashif93:
I guess I messed up things really bad. I was trying different things to clean the repo but ended up making more mess than before. I will fix it and get back with the updates. I have reverted the commits on my master branch. Thank you for your help :) I really appreciate that!

mohdkashif93:
Thank you so much for your help! It took some time but I finally got it to work! I will update the code today with examples, tests and documentation. Please let me know if you want any changes in the implementation :)

dschult:
@sdsingh we have a question about implementing IDS for NetworkX.

What should these functions return? IDS is similar to BFS so we could use it to return the things BFS returns, but then people could just use BFS. Can you give an example of where IDS is being used and what that example would want?   Thanks!

dschult:
I think we're not getting any feedback here, so based on reading online, it seems that IDS is most effective in providing early results which can then be used to pursue other approaches. So an iterator yielding results is needed.   I'm guessing that some people will want the edges and some will want predecessors (nodes along with which edges discovered them). Let's start with yielding edges like the ```dfs_edges``` does.  You'll need to keep track of the edges already yielded when you iteratively deepen the search so you don't yield them again.

sdsingh:
@dschult sorry, missed this earlier. 

That sounds right to me. As opposed to BFS, it would be nice to use IDS for memory efficiency. Since IDS actually visits nodes in a DFS order, it uses only O(d) memory, as opposed to O(b^d) memory, while still achieving an equivalent result to BFS on a certain depth. This is useful in games w/ high branching factor, which can quickly get results in a certain time constraint by restricting to low depth.

dschult:
[...]

PR #3081, opened 2018-07-23
[WIP]Iterative Deepening Search

mohdkashif93:
Addresses [#1909](https://github.com/networkx/networkx/issues/1909). I have added the initial logic for Iterative Deepening Search (also known as Iterative Deepening DFS). I have following question about the implementation

1. Here I am returning True or False depending on whether the target node is reachable from the source node or not. Is the correct way or do you want me to return the path from source to target?

TODO:
1. Add examples in docstrings
2. Add tests


mohdkashif93:
Sorry I opened this pull request incorrectly (It has pulled in my previous changes as well). I will make a new one.

PR #3084, opened 2018-07-24
[WIP] Iterative Deepening Search

mohdkashif93:
Addresses [#1909](https://github.com/networkx/networkx/issues/1909). I have added the initial logic for Iterative Deepening Search (also known as Iterative Deepening DFS). I have following question about the implementation

1. Here I am returning True or False depending on whether the target node is reachable from the source node or not. Is the correct way or do you want me to return the path from source to target?

TODO:
1. Add examples in docstrings
2. Add tests

### DIFF
--- a/networkx/algorithms/traversal/depth_first_search.py
+++ b/networkx/algorithms/traversal/depth_first_search.py
@@ -415,3 +415,39 @@ def dfs_labeled_edges(G, source=None, depth_limit=None):
                 if stack:
                     yield stack[-1][0], parent, 'reverse'
         yield start, start, 'reverse'
+
+        
+def iterative_deepening_dfs(G, source, target, depth_limit=None):
+    """Iterate over Graph using Iterative deepening depth-first-search (IDDFS) to
+     find if a path exists from source to target or not. It combines the
+     space optimization of DFS with the time optimality of BFS.
+
+       Parameters
+       ----------
+       G : NetworkX graph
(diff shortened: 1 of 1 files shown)


dschult:
Returning True or False isn't very useful here (as I guess you figured out). BFS would work better for discovering if two nodes are connected. So we have to figure out what service we are providing...

Since IDS is essentially like breadth-first-search, we should probably create a similar module to breadth_first_search.py with functions that yield edges, predecessors, successors, etc all using the same basic algorithm.  I may be missing the point here, so think about it yourself too. What do people want to use IDS for? Are there general uses that we should provide?

mohdkashif93:
From what I understand about the algorithm, it is more useful in a scenario when we have say large graphs, in which traditional BFS/DFS would run out of memory(it makes the assumption that the branching factor is finite). So we can make use of IDS for large graphs, like if a path exists between two nodes, path between the said nodes. I am not sure if predecessors/successors and tree methods will be useful here since one will be using this for large graphs, the output from these 3 methods might make the program run out of memory? (And in case if it does not run out memory, then traditional BFS would work as well, I mean that's the whole point of IDS -- to increase the depth gradually till we reach our goal). So I am thinking of changing the return type as the following: List of edges between source and target if a path exists, None otherwise. 

Please let me know your thoughts on this. :) 

Apologies for the delay in response.

PR #3089, opened 2018-07-25
Add percolation centrality

r4um:
Percolation centrality for a given node, at a given time, as the proportion of ‘percolated paths’ that go through that node. A ‘percolated path’ is a shortest path between a pair of nodes, where the source node is percolated (e.g., infected).

Ref. http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0053095

### DIFF
--- a/doc/reference/algorithms/centrality.rst
+++ b/doc/reference/algorithms/centrality.rst
@@ -97,3 +97,10 @@ Reaching
 
    local_reaching_centrality
    global_reaching_centrality
+
+Percolation
+-----------
+.. autosummary::
+   :toctree: generated/
+
+   percolation_centrality
--- a/networkx/algorithms/centrality/__init__.py
+++ b/networkx/algorithms/centrality/__init__.py
@@ -12,3 +12,4 @@ from .harmonic import *
 from .katz import *
 from .load import *
 from .reaching import *
+from .percolation import *
--- a/networkx/algorithms/centrality/percolation.py
+++ b/networkx/algorithms/centrality/percolation.py
@@ -1,0 +1,131 @@
+# coding=utf8
+#    Copyright (C) 2018 by
+#    Pranay Kanwar <pranay.kanwar@gmail.com>
+#    All rights reserved.
+#    BSD license.
+#
+"""Percolation centrality measures."""
+from __future__ import division
+
+__author__ = """\n""".join(['Pranay Kanwar <pranay.kanwar@gmail.com>'])
+
+import networkx as nx
+
+from networkx.algorithms.centrality.betweenness import\
+    _single_source_dijkstra_path_basic as dijkstra
+from networkx.algorithms.centrality.betweenness import\
+    _single_source_shortest_path_basic as shortest_path
+
+__all__ = ['percolation_centrality']
+
+
+def percolation_centrality(G, attribute='percolation',
+                           states=None, weight=None):
+    r"""Compute the percolation centrality for nodes.
+
+    Percolation centrality of a node $v$, at a given time, is defined
+    as the proportion of ‘percolated paths’ that go through that node.
+
+    This measure quantifies relative impact of nodes based on their
+    topological connectivity, as well as their percolation states.
+
+    Percolation states of nodes are used to depict network percolation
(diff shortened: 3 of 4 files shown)


dschult:
Thanks for this!  The code and the idea to include it looks good.

My quick read suggests this is similar to betweenness centrality only you don't consider all paths, only paths starting from percolated nodes.  Is that correct?  Could you add a paragraph describing the relation between betweenness and percolation centrality?

Along those lines, but from a code perspective: would it make sense to include this in the betweenness module?

You mention "Time" in the docstring.  Can you describe what you mean by that? Is it part of the definition, or part of a setting in which this measure is often used?

r4um:
>My quick read suggests this is similar to betweenness centrality only you don't consider all paths, only paths starting from percolated nodes. Is that correct? Could you add a paragraph describing the relation between betweenness and percolation centrality?

If all nodes are at the same percolation level, percolation centrality is equal to betweenness 
centrality (there is a test for the same included in this PR)

>Along those lines, but from a code perspective: would it make sense to include this in the betweenness module?

I did think and consider it, avoided doing that for primarily the following reasons
1. While they seem similar, percolation centrality has an over the 'analysis time' tone to it, in future may
    add time based arguments (percolation states over time) / analysis.
2. There will be other enhancements, like considering both source and target node percolation states.
3. It needed to hook directly into the pair dependency calculations.

If we still feel it belongs in the `betweenness` module for now, I can add an optional 
`percolation_states` parameter to the `betweenness` centrality function and move/use the `_accumulate_percolation` function. Does that sound good ?

>You mention "Time" in the docstring. Can you describe what you mean by that? Is it part of the definition, or part of a setting in which this measure is often used?

The percolation centrality analysis makes more sense across different percolation states of nodes
across different times in an analysis. Depending on the percolation states of nodes 
at different times, nodes have different percolation centralities.

dschult:
This makes sense...

I think we should leave this in a separate module from betweenness. But please add a paragraph (or more) to the docstring about the connection betweenness and percolation centrality. Also, make the time aspect explicit:  this is a measure of a shapshot in time of a typically evolving process as percolation occurs.  This is your chance to get other people to understand and use this measure. So make the description accessible and complete.

I'll go through the code in more detail at some point soon. If you haven't already done so, can you check that the code matches Python's PEP8 code style which you can read about from [the official PEP8 specification](https://www.python.org/dev/peps/pep-0008/?), or use a convenient tool to check it for you ```pip install pep8``` and then ```pep8 filename```

Thanks!

r4um:
>I think we should leave this in a separate module from betweenness. But please add a paragraph (or more) to the docstring about the connection betweenness and percolation centrality. Also, make the time aspect explicit: this is a measure of a shapshot in time of a typically evolving process as percolation occurs. This is your chance to get other people to understand and use this measure. So make the description accessible and complete.

Thanks, updated the docstring.

>I'll go through the code in more detail at some point soon. If you haven't already done so, can you check that the code matches Python's PEP8 code style which you can read about from the official PEP8 specification, or use a convenient tool to check it for you pip install pep8 and then pep8 filename

Yes flake8 seems happy as it can be
```
git:percolation_centrality=⚡ flake8 networkx/algorithms/centrality/percolation.py
git:percolation_centrality=⚡ flake8 networkx/algorithms/centrality/tests/test_percolation_centrality.py
networkx/algorithms/centrality/tests/test_percolation_centrality.py:3:1: F403 'from nose.tools import *' used; unable to detect undefined names
networkx/algorithms/centrality/tests/test_percolation_centrality.py:42:13: F405 'assert_almost_equal' may be undefined, or defined from star imports: nose.tools
networkx/algorithms/centrality/tests/test_percolation_centrality.py:50:13: F405 'assert_almost_equal' may be undefined, or defined from star imports: nose.tools
networkx/algorithms/centrality/tests/test_percolation_centrality.py:77:13: F405 'assert_almost_equal' may be undefined, or defined from star imports: nose.tools
networkx/algorithms/centrality/tests/test_percolation_centrality.py:82:13: F405 'assert_almost_equal' may be undefined, or defined from star imports: nose.tools
```
Perhaps should be a part of the CI tests too ?

dschult:
Thanks!

PR #3091, opened 2018-07-26
Update docs with all functions left out of rst files

dschult:
Also add script to find these functions

### DIFF
--- a/networkx/algorithms/smallworld.py
+++ b/networkx/algorithms/smallworld.py
@@ -15,7 +15,7 @@ Both coefficients compare the average clustering coefficient and shortest path
 length of a given graph against the same quantities for an equivalent random
 or lattice graph.
 
-For more information, see the Wikipedia article on small-world network [1]_.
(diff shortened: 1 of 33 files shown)


networkx/algorithms/smallworld.py
```diff
@@ -148,20 +148,21 @@ def lattice_reference(G, niter=1, D=None, connectivity=True, seed=None):
...
     -------
     G : graph
-       The latticized graph.
+        The latticized graph.
 
     Notes
     -----
-    The implementation is adapted from the algorithm by Sporns et al. which is
-    inspired from the original work from Maslov and Sneppen (2002) [2]_.
+    The implementation is adapted from the algorithm by Sporns et al. [1]_
```
jfinkels: These sentences have lost their periods.

dschult: Good catch!  It looks like any period after an underscore was removed.
I'll fix it!  :) Thanks!

Issue #2346, opened 2016-12-30
Add bellman-ford as alternative algorithm in generic shortest path solving method


dzhuang:
As documented [here](https://networkx.readthedocs.io/en/latest/reference/algorithms.shortest_paths.html)(**UPDATE**, the url was moved to [here](https://networkx.github.io/documentation/latest/reference/algorithms/shortest_paths.html#module-networkx.algorithms.shortest_paths.generic)), some generic method for solving undirected and directed shortest path problem, including ``shortest_path``, ``all_shortest_paths``, ``shortest_path_length`` and ``average_shortest_path_length``.

Those method are actually using dijkstra algorithm according to the source code. I think it reasonable if we are using those method for problems (e.g., social networks) with non-negative edge weights. However, since the naming of those method are not method-specific, they should be able to deal with situations when there exists negative weighted edges in a graph. Bellman-ford is also able to solve undirected graphs if there exist no negative weighted edges, and for undirected graphs with negative weighted edges, it will raise errors since that graph will be views as a graph having negative loops if converting to a directed graph. Therefore, I think we should integrate bellman-ford method into those methods, at least allow it as an alternative method when users are calling those functions.

AvijitGhosh82:
This was a problem I genuinely faced in my research project, as I had a task of finding the path with the highest product of weights. Taking negative log and then applying shortest path was giving me incorrect answers. 

Will be great if this is solved.

dschult:
Fixed in #3098

dzhuang:
@dschult Thank you for the fix!

scutsmash:
Teacher zhuang,I do not know what you want to say.If you can express it more clearly,I think it would be better.

scutgeorge:
@scutsmash new bee

PR #3098, opened 2018-07-30
Add option to use Bellman-Ford in generic shortest-path functions

gfetterman:
Addresses #2346 

This is one possible way to do this. I reworked some of the machinery to reduce conditional nesting, but there's still redundant code.

Some verbosity is inevitable - we're trying to value-dispatch on multiple parameters, which isn't usually elegant.

Another approach would be a more dramatic refactor in favor of a more functional style of passing functions around and closing them over values as needed. I've briefly sketched this out for myself, and I can share it in a gist if there's interest. My concern is that it would add complexity (at least for the casual reader) for somewhat questionable gain - how likely is it that the generic shortest-path functions will need to be extended similarly soon?

Finally: this module raises `NetworkXNoPath`, `NetworkXPointlessConcept`, and `NetworkXError` exceptions - should I factor them out, per #1705 ? It's not entirely clear to me from that issue when such refactors should occur, since they're potentially interface-breaking.

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -57,6 +57,10 @@ Functions for Clauset-Newman-Moore modularity-max community detection.
 
 Functions for small world analysis, directed clustering and perfect matchings.
 
+The shortest_path generic and convenience functions now have a `method`
+parameter to choose between dijkstra and bellmon-ford in the weighted case.
+Default is dijkstra (which was the only option before).
+
 API Changes
 -----------
 empty_graph has taken over the functionality from
--- a/networkx/algorithms/shortest_paths/generic.py
+++ b/networkx/algorithms/shortest_paths/generic.py
@@ -154,8 +154,6 @@ def shortest_path(G, source=None, target=None, weight=None, method='dijkstra'):
                 for target in paths:
                     paths[target] = list(reversed(paths[target]))
     else:
-        if source not in G:
-            raise nx.NodeNotFound("Source {} not in G".format(source))
         if target is None:
             # Find paths to all nodes accessible from the source.
             if method == 'unweighted':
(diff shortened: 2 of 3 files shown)


dschult:
If you take out the check for ```source in G``` then I think an input ```v``` that ```is not in G``` gives a ```KeyError``` instead of ```NodeNotFound```.  Is that what we want? I guess we should be testing all these errors...:}

gfetterman:
Good point. I should have checked that more exhaustively:

* unweighted functions (path and length) raise `nx.NodeNotFound` errors, but
* there are no checks in the weighted functions (neither Dijkstra nor Bellman-Ford, neither path nor length), so they raise a `KeyError`, as you point out. (There are checks in `goldberg_radzik()` and `bidirectional_dijkstra()`, though.)

Seems like the minimal answer here is to put that check back into `shortest_path_length()`, and also add it in the same place to `shortest_path()`, which did not have it before. I've now done this.

Would a standardization of the shortest-path function input checking be worthwhile at some point? It could break interfaces ...

dschult:
This looks ready to merge.  Anything else planned for this PR?

gfetterman:
Just a couple of things:

1. Now that #3103 is merged, I could re-remove the absent-source guards in `shortest_path()` and `shortest_path_length()`, since `nx.NodeNotFound` will get raised by the innermost functions.
2. I'm still not hitting the coverage delta target - I'd like to fix that.
3. There is the potential refactor of the `NetworkXNoPath`, `NetworkXPointlessConcept`, and `NetworkXError` throws I mentioned above.

I can throw the first two in later today; for the third I'll wait on a response.

dschult:
I added a short description in the release notes for the change in API and I (hopefully) improved the coverage delta target by changing ```elif``` to ```else``` in some places. 

So, you can remove the absent-source guards...

When you talk about a potential refactor of the exceptions. What do you have in mind?
The #1705 isn't very specific in what to do -- other than remove NetworkX from the exception names. And we shouldn't do that until v3.0 due to potential backward compatibility issues.

gfetterman:
Re exceptions: got it - I'll leave them be.

I've removed the absent-source guards, and so I think this PR is ready.

(Apologies for the rebase, by the way - was trying to avoid a merge commit.)

PR #3111, opened 2018-08-04
In 'from_pandas_adjacency', use column names rather than column order.

### LINKED ISSUE #3105: When using 'from_pandas_adjacency', edges are added focusing on column names rather than the order
When using the function `from_pandas_adjacency` (networkx 2.1) to create a graph, edges are added focusing on the ***order*** of columns and rows in the pandas DataFrame.
Personally, I think that it is more useful by adding edges focusing on the ***names*** of columns and rows (indices) rather than the ***order***, because

1. Personally, when column names and row names are given in the pandas DataFrame, I think it is natural to interpret them as node names.
2. In the current behavior, it is necessary for the code calling `from_pandas_adjacency` to ensure that the columns and rows of the pandas DataFrame are in the same order. For example, by writing the following code every time
　　`df = df[df.index]  # Sort columns by row (index) order`
Without this code, there is a possibility that an unintended graph structure may be created.

The examples are as follows.

### Current behavior:
In the following case 1 and case 2, the pandas DataFrames have the same structure except column order. (The columns order of case 1 is ["A", "B", "C"]],  and the columns order of case 2 is ["A "," C "," B "])

dataframe

However, the graph structure created using `from_pandas_adjacency` (networkx 2.1) differs between case 1 and case 2.

graph

**Case 1:**  This case is no problem.

```
# A->B, B->C
data = {
        "A": {"A": 0, "B" : 0, "C": 0},
        "B": {"A": 1, "B" : 0, "C": 0},
        "C": {"A": 0, "B" : 1, "C": 0}
        }
case1_df = pd.DataFrame(data)

G  = nx.from_pandas_adjacency(case1_df, create_using=nx.DiGraph())
pos = nx.spring_layout(G)
nx.draw_networkx_nodes(G, pos=pos, node_color="#ffa1a1")
nx.draw_networkx_labels(G, pos=pos)
nx.draw_networkx_edges(G, pos=pos)
plt.title('case 1')
plt.show()
```
case1_graph

**Case 2:**  
I think that behavior in this case is a little confusing.
Focusing on column names and row names, it is expected that the same graph structure as case 1 will be created. However, in fact, a different graph structure is created. It is because `from_pandas_adjacency` is currently based on the ***order***, not the ***names***.

```
# A->B, B->C (Same as case 1)
data = {
        "A": {"A": 0, "B" : 0, "C": 0},
        "B": {"A": 1, "B" : 0, "C": 0},
        "C": {"A": 0, "B" : 1, "C": 0}
        }
case2_df = pd.DataFrame(data)

# !!! Change the column order from ["A", "B", "C"] to ["A", "C", "B"] !!!
case2_df = case2_df[["A", "C", "B"]]  

G  = nx.from_pandas_adjacency(case2_df, create_using=nx.DiGraph())
pos = nx.spring_layout(G)
nx.draw_networkx_nodes(G, pos=pos, node_color="#ffa1a1")
nx.draw_networkx_labels(G, pos=pos)
nx.draw_networkx_edges(G, pos=pos)
plt.title('case 2')
plt.show()

```
case2_graph

### Expected result: 
I want the result of case 2 to be the same as case 1. The reasons are mentioned at the beginning of this issue. 
If the code is changed, I'm thinking of sending a PR like the following.

**Before:**
https://github.com/networkx/networkx/blob/ab3a6cdece25e359f28526fdc5f9f8ae443503f3/networkx/convert_matrix.py#L180-L190

**After:**
My desire is to avoid the situation that always needs to be careful of columns order when using `from_pandas_adjacency`. I think that there are two approaches.

**[Approach 1]**
In the original code, the columns are sorted after creating the graph. Change this to sort before creating the graph.

```
def from_pandas_adjacency(df, create_using=None):
    # …

    try:
        df = df[df.index]
    except:
        msg = "%s not in columns"
        missing = list(set(df.index).difference(set(df.columns)))
        raise nx.NetworkXError("Columns must match Indices.", msg % missing)

    A = df.values
    G = from_numpy_matrix(A, create_using=create_using)

    nx.relabel.relabel_nodes(G, dict(enumerate(df.columns)), copy=False)
    return G
```

**[Approach 2]**
If the columns and rows are in the different order, throw an exception. Then, before using `from_pandas_adjacency`, I will be able to remember that the columns and rows are in the same order. And, it is possible to avoid trouble that an unintended graph structure is created.

komo-fr:
Fixes #3105.

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -177,14 +177,15 @@ def from_pandas_adjacency(df, create_using=None):
 
     """
 
-    A = df.values
-    G = from_numpy_matrix(A, create_using=create_using)
     try:
         df = df[df.index]
     except:
         msg = "%s not in columns"
         missing = list(set(df.index).difference(set(df.columns)))
         raise nx.NetworkXError("Columns must match Indices.", msg % missing)
+
+    A = df.values
+    G = from_numpy_matrix(A, create_using=create_using)
 
     nx.relabel.relabel_nodes(G, dict(enumerate(df.columns)), copy=False)
     return G
--- a/networkx/tests/test_convert_pandas.py
+++ b/networkx/tests/test_convert_pandas.py
@@ -163,3 +163,14 @@ class TestConvertPandas(object):
         df = nx.to_pandas_adjacency(Gtrue, dtype=int)
         G = nx.from_pandas_adjacency(df)
         assert_graphs_equal(Gtrue, G)
+
+    def test_from_adjacency_named(self):
+        # example from issue #3105
+        data = {"A": {"A": 0, "B": 0, "C": 0},
+                "B": {"A": 1, "B": 0, "C": 0},
+                "C": {"A": 0, "B": 1, "C": 0}}
+        dftrue = pd.DataFrame(data)
+        df = dftrue[["A", "C", "B"]]
+        G = nx.from_pandas_adjacency(df, create_using=nx.DiGraph())
+        df = nx.to_pandas_adjacency(G, dtype=int)
+        pd.testing.assert_frame_equal(df, dftrue)


dschult:
Looks good so far.  Can you add a test to make sure we know if we break using names later?

komo-fr:
@dschult
Thank you for reviewing!
I added test code.

dschult:
Thanks!  This is very helpful...

PR #3114, opened 2018-08-06
Clarify documentation for predecessors() and successors(). Fixes #3113.

### LINKED ISSUE #3113: G.predecessors(node) documentation could benefit from additional explanation
I'm on a _tree_ network similar to this toy,

    [('total_fruits', 'apples'),
     ('total_fruits_and_vegetables', 'total_fruits'),
     ('total_fruits_and_vegetables', 'total_vegetables'),
     ('total_vegetables', 'carrot')]

To find the root node I'm using this solution on StackOverflow: [NetworkX find root_node for a particular node in a directed graph](https://stackoverflow.com/a/47191030/343215). This solution works, but I think the documentation could be clearer. 

The documentation for predecessor [networkx.DiGraph.predecessors](https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.DiGraph.predecessors.html) says: 

> DiGraph.predecessors(n)[source]
> Return an iterator over predecessor nodes of n.

For my tree I expected G.predecessors(node) output to return _all_ of the predecessors to the root,

    [n for n in G.predecessors('apples')]
    >>> ['apples', ''total_fruits', 'total_fruits_and_vegetables']

But what I get:

    [n for n in G.predecessors('apples')]
    >>>['total_fruits']

Finally, I realized _nodes_ refers to the predecessors one level or edge from the node parameter, because the network may not be a tree. 

In this edge case I'm asking if maybe the documentation could benefit further description.

leotrs:
…Graph.successors.

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -769,6 +769,25 @@ class DiGraph(Graph):
     def successors(self, n):
         """Return an iterator over successor nodes of n.
 
+        A successor of n is a node m such that there exists a directed
+        edge from n to m.
+
+        Parameters
+        ----------
+        n : node
+           A node in the graph
+
+        Raises
+        -------
+        NetworkXError
+           If n is not in the graph.
+
+        See Also
+        --------
+        predecessors
+
+        Notes
+        -----
         neighbors() and successors() are the same.
         """
         try:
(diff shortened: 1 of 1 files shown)


leotrs:
See #3114

PR #3117, opened 2018-08-10
Adding second order centrality algorithm to centrality/

erwanlemerrer:
* Adds second_order_centrality module to networkx.algorithms.centrality package

Author implementation of "Second order centrality: Distributed assessment of nodes criticity in complex networks", Elsevier Computer Communications 34(5):619-628, 2011. Cited 96 times.

### DIFF
--- a/doc/reference/algorithms/centrality.rst
+++ b/doc/reference/algorithms/centrality.rst
@@ -107,7 +107,7 @@ Percolation
    percolation_centrality
 
 Second Order Centrality
---------
+-----------------------
 .. autosummary::
    :toctree: generated/
 
--- a/doc/reference/algorithms/centrality.rst
+++ b/doc/reference/algorithms/centrality.rst
@@ -105,3 +105,10 @@ Percolation
    :toctree: generated/
 
    percolation_centrality
+
+Second Order Centrality
+--------
+.. autosummary::
+   :toctree: generated/
+
+   second_order_centrality
--- a/networkx/algorithms/centrality/second_order_centrality.py
+++ b/networkx/algorithms/centrality/second_order_centrality.py
@@ -132,5 +132,6 @@ def setup_module(module):
     from nose import SkipTest
     try:
         import numpy
+        import scipy
     except:
         raise SkipTest("NumPy not available")
--- a/networkx/algorithms/centrality/second_order_centrality.py
+++ b/networkx/algorithms/centrality/second_order_centrality.py
@@ -124,5 +124,13 @@ def second_order_centrality(G):
                                   (np.identity(n) + _Qj(P, i)) *
                                   np.reshape(M[:, i], [n, 1]))[0]
 
-    return {k: v for k, v in enumerate(
-        np.sqrt(H - np.square(M))[0, :])}  # eq 4
+    return dict(zip(G.nodes, np.sqrt(H - np.square(M))[0, :]))  # eq 4
+
+
+# fixture for nose tests
+def setup_module(module):
+    from nose import SkipTest
+    try:
+        import numpy
+    except:
+        raise SkipTest("NumPy not available")
--- a/networkx/algorithms/centrality/second_order_centrality.py
+++ b/networkx/algorithms/centrality/second_order_centrality.py
@@ -33,7 +33,11 @@
 
 __all__ = ['second_order_centrality']
 
-@nx.not_implemented_for('directed')
+import networkx as nx
+import copy
+from networkx.utils import not_implemented_for
+
+@not_implemented_for('directed')
 def second_order_centrality(G):
     """Compute the second order centrality for nodes of G.
 
@@ -81,9 +85,6 @@ def second_order_centrality(G):
        "Second order centrality: Distributed assessment of nodes criticity in
        complex networks", Elsevier Computer Communications 34(5):619-628, 2011.
     """
-
-    import networkx as nx
-    import copy
 
     try:
         import numpy as np
--- a/networkx/algorithms/centrality/second_order_centrality.py
+++ b/networkx/algorithms/centrality/second_order_centrality.py
@@ -65,6 +65,21 @@ def second_order_centrality(G):
     >>> print(sorted(soc.items(), key=lambda x:x[1])[0][0]) # pick first id
     0
 
+    >>> G = nx.empty_graph()
+    >>> second_order_centrality(G)
+    networkx.exception.NetworkXException: Empty graph.
+
+    >>> G = nx.Graph()
+    >>> G.add_node(0)
+    >>> G.add_node(1)
+    >>> second_order_centrality(G)
+    networkx.exception.NetworkXException: Non connected graph.
+
+    Raises
+    ------
+    NetworkXException
+        If the graph ``G`` is empty or non connected.
+
     See Also
     --------
     betweenness_centrality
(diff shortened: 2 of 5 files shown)


dschult:
Should add tests for calling it with an empty graph and calling it with a disconnected graph.
(this will make the coverage of the tests pass the codecov standards).

Also we need to add it to the ```__init__.py``` and ```/doc/reference``` pages.

I will add those tests and others soon if you don't get to it, so no worries...   Thanks!

erwanlemerrer:
Dan, license is "BSD-3-Clause-Clear".
Will add coverage tests tomorrow.

dschult:
Yeah -- that'll teach me not to look before questioning....  its the same as license we use.  Sorry for the confusion...

Thanks very much in advance for tests... :)

erwanlemerrer:
Hey Dan, no worries! 
Added doc and __init__ call, and coverage by asserRaises(). Still got one a bit unclear fail in travis, solely for one build job...

doc/reference/algorithms/centrality.rst
```diff
@@ -105,3 +105,10 @@ Percolation
    :toctree: generated/
 
    percolation_centrality
+
+Second Order Centrality
+--------
```
dschult: This line needs to be the same length as the "section title" line above it.  So in this case quite  bit more minus signs.

networkx/algorithms/centrality/second_order_centrality.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Returns
+    -------
+    nodes : dictionary
+       Dictionary keyed by node with second order centrality as the value.
+
+    Examples
+    --------
+    >>> G = nx.star_graph(10)
+    >>> soc = second_order_centrality(G)
```
dschult: This call should use ```nx.second_order_centrality``` instead of ```second_order_centrality```.  It works fine as is for nosetests, but once the docstring is process by sphinx to buid the html pages, the module context is lost so we have to refer to it through the ```nx.``` syntax.  This also tests the namespace to make sure ```second_order_centrality``` is at the top level.

dschult:
It looks like you are trying to use ```unittest``` for the tests...
You should remove that in favor of using ```nose``` (though ```nose``` is not being activately maintained so we will likely have to move to ```pytest```).  But for this PR use ```nose```.

The tests should be put in a separate file ```tests/test_second_order.py``` or similar. Each folder of code has a ```tests``` subfolder.  Look at the other test modules in ```tests``` for examples. You need to test the special cases as you are doing, but it would also be good to test the main algorithm with some known cases. The docstring does one such example -- perhaps there are others?

If it's too much trouble to get that all working, I can add the tests to make codecov happy.

erwanlemerrer:
Thanks for this clarification, indeed that explains my current trouble! I propose to sort this out and create the test file, but if it fails on next trial (with also new examples), I would be grateful if you can have your expert look on it. Thanks again.

dschult:
That looks better -- you are passing the tests when numpy is installed. 
We need to skip these tests when numpy isn't present. Take a look at other modules (e.g. ```current_flow_betweenness.py```) to see how that is done.  You've got the SkipTest code in the test class but not in the module itself. I think it is an easy fix.

I haven't looked at the actual tests yet

erwanlemerrer:
Dan, it seems that everything is fine now, except codecov that misses 3% coverage (88 instead of 91%). It looks to me that this is because the code is very compact: if I were to arbitrarily add some dummy instructions, coverage would be fine (also seems that other centrality algorithms do not cover those except either). I am correct ? If so, what is best to do now ?

dschult:
Don't worry about the 3% on the coverage. It is low because of the checks for when numpy is not there. Those checks have to be there but coverage complains. We can override and merge even with that failure. 

Why did the Florentine test fail? Is there an example where the second order centrality is known?

erwanlemerrer:
Yes indeed, now that the core code is fine, I will focus on adding some more advanced tests and examples to the test_second_order_centrality.py file.

networkx/algorithms/centrality/second_order_centrality.py
```diff
@@ -106,6 +109,20 @@ def second_order_centrality(G):
         raise nx.NetworkXException("Empty graph.")
     if not nx.is_connected(G):
         raise nx.NetworkXException("Non connected graph.")
+    if [True for (i, j, d) in G.edges(data=True)
```
dschult: The ```any``` function does this without creatingg a list -- not a big deal, but probably easier to read too.
 
    if any(d.get('weight', 0) < 0 for u, v, d in G.edges(data=True)):
        raise(...)

networkx/algorithms/centrality/second_order_centrality.py
```diff
@@ -106,6 +109,20 @@ def second_order_centrality(G):
...
+            if 'weight' in d and d['weight'] < 0]:
+        raise nx.NetworkXException("Graph has negative edge weights.")
+
+    def unbias_graph(G):
+        G = nx.DiGraph(G)
+        d_max = np.max([G.in_degree(i,
```
dschult: Nodes don't have to be integers...and ```in_degree```returns a dict of all nodes if no node is specified.
So the code is probably faster and more readable something like this:

    in_deg = G.in_degree(weight='weight')
    d_max = np.max(in_deg.values())
    for n, deg in d_max.items():
        if deg < d_max:
            G.add_edge(n, n, weight=d_max-deg)

networkx/algorithms/centrality/second_order_centrality.py
```diff
@@ -106,6 +109,20 @@ def second_order_centrality(G):
         raise nx.NetworkXException("Empty graph.")
     if not nx.is_connected(G):
         raise nx.NetworkXException("Non connected graph.")
+    if [True for (i, j, d) in G.edges(data=True)
+            if 'weight' in d and d['weight'] < 0]:
+        raise nx.NetworkXException("Graph has negative edge weights.")
+
+    def unbias_graph(G):
```
dschult: Why create a function for ```unbias_graph``` when you only use it once? You can just include the code directly.

erwanlemerrer:
Thanks for those comments! Function is removed; I had it here, because this is a common operation for random walks on graph, so I will probably later contribute to a proper function "Metropolis-Hastings" unbiasing, in due directory.

dschult:
Looks good -- anything else to change before I merge? I'll look through one more time before I do.

erwanlemerrer:
Targeting next releases, I will probably add the biased version of the algorithm and some more variants, but I suggest to stay for now with this compact, loyal to the paper, and probably more stable version of the algorithm for the first merge ;)

dschult:
Sorry not to have raised this earlier:
Could you change the name of the new module so that it is not the same as the function? -- avoiding potential namespace collisions.  

Given the other names in the centrality folder, maybe change the module to ```second_order.py``` and similarly to tests file. The ```__init__.py``` entry will also need to be changed, but the documentation file (```centrality.rst```) should be fine since it is pulling from the centrality module and not the second_order file.

Thanks!

dschult:
One more thing -- is ```deepcopy``` really needed in ```_Qj```?
Unless it is a matrix of containers, (which I think doesn't make sense here) could you use ```P.copy()```?

erwanlemerrer:
Nice catch for copy! For renaming,  I made the suggested changes, except in test/ as all files in there follow the format with ending centrality.py in names, so that looked consistent as in.

erwanlemerrer:
Dan, renaming of second_order_centrality.py to second_order.py did not remove the first file. So now there are two.. my fault, can you remove it manually?

PR #3121, opened 2018-08-15
Fix scale_free_graph restriction

inexxt:
`scale_free_graph` generator has a restriction, that `alpha + beta + gamma` has to be equal to `1`.
However, while working with random graph parameters, it is often hard to satisfy this requirement, because   of the float representation.

Eg, this simple code doesn't work:
```python
from networkx import scale_free_graph
import numpy as np

np.random.seed(1)
t = np.random.random(3)
t = t / t.sum()
G = scale_free_graph(N, alpha=t[0], beta=t[1], gamma=t[2])
```
neither does that one
```python
from networkx import scale_free_graph
import numpy as np

np.random.seed(1)
t = np.random.random(3)
t = t / t.sum()
G = scale_free_graph(N, alpha=t[0], beta=t[1], gamma=(1 - t[0] - t[1]))
```
or that one
```python
from networkx import scale_free_graph
import numpy as np

t = np.random.dirichlet(np.ones(3, dtype=np.float32), size=1)[0]
G = scale_free_graph(10, alpha=t[0], beta=t[1], gamma=t[2])
```
So, that's why I think just checking if `math.isclose(alpha + beta + gamma, 1.0)` would be better.

### DIFF
--- a/networkx/generators/directed.py
+++ b/networkx/generators/directed.py
@@ -17,7 +17,6 @@ scale-free graphs.
 from __future__ import division
 
 from collections import Counter
-import math
 
 import networkx as nx
 from networkx.generators.classic import empty_graph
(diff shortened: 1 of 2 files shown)


dschult:
This looks good!  Thanks.  But ```math.isclose``` doesnt work for python 2.7 so can we make it

    if abs(alpha + beta + gamma - 1.0) < 1e-9:

inexxt:
@dschult Fixed

dschult:
Great -- It looks like the tests are testing this well. 
I'll try to add to that today before merging.  (unless you want to try that)
Thanks!

dschult:
Thanks!

PR #3127, opened 2018-08-19
Equitable coloring of graphs.

### LINKED ISSUE #3042: ENH: Addition of the Equitable coloring algorithm.
I recently implemented an [equitable coloring](https://en.wikipedia.org/wiki/Equitable_coloring) algorithm for degree bounded graphs from a paper by [Kierstead et. al.](https://link.springer.com/article/10.1007%2Fs00493-010-2483-5) here: [musically-ut/equitable-coloring](https://github.com/musically-ut/equitable-coloring).

From Wikipedia:

>    In graph theory [..] an equitable coloring is an assignment of colors to the vertices of an undirected graph, in such a way that
>
>   * No two adjacent vertices have the same color, and
>   * The numbers of vertices in any two color classes differ by at most one.

----

I wanted to get an idea of whether there is any interest in including `equitable_color` from my package as a potential coloring strategy under `networkx.algorithms.coloring`, next to the `greedy_color` strategies already there.

If so, I could work on a PR which will port the code and the test cases to `networkx`.

musically-ut:
WIP: Initial pull-request for adding equitable-coloring to the `networkx.algorithms.coloring`.

Closes #3042.


dschult:
Looks good!  :) . Thanks --

I haven't worked through all the code, but I notice that you've got tests and integrated the autodoc rst files.  You still need doc_strings as you know.  I would also encourage you to pay special attention to the naming of functions. ```procedure_P``` is not very descriptive to a generic user. But if it is a standard name used in many (whatever many means) sources, then that's OK. If it is only referred to that way in the one paper, then there is perhaps a better name that describes what it is doing. Similar for variable names -- one letter names are often discouraged as being non-descriptive, but we use them for example for graphs in lots of our code because the literature does that. 

Your placement of the code and general style looks good -- again, I haven't gone through it in detail.

musically-ut:
Thanks for the comments!

I've tried to stick as closely as possible to the nomenclature used in the [reference paper][1] and the name `procedure_P` is quite literally the name of Section 3.3. :)

The procedure is not known by any other name outside the paper and what it does is ... not trivial to explain, let alone express as a function name. I empathize with the authors there.

I've used similar reasoning for naming the variables `F`, `C`, `H`, `N`, `L`, `L_prime`, `A_cal`, `A_cal_prime`, etc. as it made following the algorithm along with the paper much easier for me. I _think_ that it will do the same for any maintainers who may have to deal with it in case I am unavailable.

----

Also, I've only added the doc_string to `equitable_color` function because it the only function which _should_ be exposed.

I took hints from the `greedy_coloring_with_interchange.py` file, where the internal classes/methods are not that well documented. However, `equitable_coloring` is exposing a lot of functions at the module level. Should I refactor the code such that fewer functions are exposed (move some functions to the test module) or refine the `doc_string` for some functions (i.e. `is_equitable` and `is_coloring`)?

Functions like `make_C_from_F`, `move_witness`, etc. are strictly internal (but are used by the test-suite). I can prefix them with an `_` to mark them as such, if that is an acceptable convention?

----

I know that this is a large body of non-trivial code and I made the PR expecting it will need significant refactoring/stylistic changes before it can be merged in. So do not hold back. :)

Thanks again!


[1]: https://sci-hub.tw/https://doi.org/10.1007/s00493-010-2483-5

networkx/algorithms/coloring/tests/test_coloring.py
```diff
@@ -97,6 +102,202 @@ def test_seed_argument(self):
...
+
+        params = make_params_from_graph(G=G, F=F)
+
+        nx.algorithms.coloring.equitable_coloring.procedure_P(
+            V_minus=0, V_plus=num_colors - 1, **params
+        )
+        check_state(**params)
+
+    def test_hardest_prob(self):
+        # Tests for two levels of recursion.
```
musically-ut: The graphs here were found using `Z3` which cause the algorithm to hit all different cases. The code to generate them is [here](https://github.com/musically-ut/equitable-coloring/blob/master/tests/generate_graph.py) and they are visualized [here](https://github.com/musically-ut/equitable-coloring/tree/master/tests/test-graphs).

networkx/algorithms/coloring/equitable_coloring.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+def make_H_from_C_N(C, N):
+    return {(c1, c2): sum(1 for node in C[c1] if N[(node, c2)] == 0)
+            for c1 in C.keys() for c2 in C.keys()}
+
+
+def change_color(u, X, Y, N, H, F, C, L):
+    """Change the color of 'u' from X to Y and update N, H, F, C."""
+    assert F[u] == X and X != Y
```
musically-ut: I can drop the assertions from the final version.

musically-ut:
Just checking in, is there anything further I can do to help with the code assimilation?

Thanks!

PR #3131, opened 2018-08-22
Handle negatively weighted edges in bellman-ford, fixes #3130

### LINKED ISSUE #3130: bellman_ford_path seems to work wrong
On this example

```python
sat=0.5
sf = 1
mon = 2

G=nx.MultiDiGraph()
G.add_nodes_from(['start','bread','money:3','eat bread0','finish','buy bread','bread:2','eat bread1'])
G.add_edge('start','bread', weight = 0)
G.add_edge('start','money:3', weight = 0)
G.add_edge('bread','eat bread0', weight = 1/sf-1/sat)
G.add_edge('eat bread0', 'finish', weight = 0)
G.add_edge('money:3','buy bread', weight = 2/mon-1/sf)
G.add_edge('buy bread', 'bread:2', weight = 0)
G.add_edge('bread:2','eat bread1', weight = 1/2-1/sat)
G.add_edge('eat bread1','finish', weight = 0)
G.add_edge('eat bread1', 'bread', weight = 0)

path = nx.bellman_ford_path(G,'start','finish', weight='weight')
w    = nx.bellman_ford_path_length(G,'start','finish', weight='weight')

print(path, w)
```
The output is 
`['start', 'bread', 'eat bread0', 'finish'] -1.0`
But `print(nx.johnson(G, weight="weight")["start"]["finish"])` output is 
`['start', 'money:3', 'buy bread', 'bread:2', 'eat bread1', 'bread', 'eat bread0', 'finish']`
And if one will draw the graph, it`s easy to see, the shortest path weight is -2.5
I am using the last stable version: 2.1

wuhaochen:
When negatively weighted edges are present, skipping nodes with longer distance than destination could miss better path. See example added in test case.

This should fix #3130.

### DIFF
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -1290,10 +1290,6 @@ def _bellman_ford(G, source, weight, pred=None, paths=None, dist=None,
                     if dist_v > cutoff:
                         continue
 
-                if target is not None:
-                    if dist_v > dist.get(target, inf):
-                        continue
-
                 if dist_v < dist.get(v, inf):
                     if v not in in_q:
                         q.append(v)
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -1216,7 +1216,8 @@ def _bellman_ford(G, source, weight, pred=None, paths=None, dist=None,
     G : NetworkX graph
 
     source: list
-        List of source nodes
+        List of source nodes. The shortest path from any of the source
+        nodes will be found if multiple sources are provided.
 
     weight : function
         The weight of an edge is the value returned by the function. The
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -1101,7 +1101,7 @@ def all_pairs_dijkstra_path(G, cutoff=None, weight='weight'):
 
 
 def bellman_ford_predecessor_and_distance(G, source, target=None,
-                                          cutoff=None, weight='weight'):
+                                          weight='weight'):
     """Compute shortest path lengths and predecessors on shortest paths
     in weighted graphs.
 
@@ -1200,12 +1200,12 @@ def bellman_ford_predecessor_and_distance(G, source, target=None,
     weight = _weight_function(G, weight)
 
     dist = _bellman_ford(G, [source], weight, pred=pred, dist=dist,
-                         cutoff=cutoff, target=target)
+                         target=target)
     return (pred, dist)
 
 
 def _bellman_ford(G, source, weight, pred=None, paths=None, dist=None,
-                  cutoff=None, target=None):
+                  target=None):
     """Relaxation loop for Bellman–Ford algorithm
 
     Parameters
@@ -1233,9 +1233,6 @@ def _bellman_ford(G, source, weight, pred=None, paths=None, dist=None,
         dict to store distance from source to the keyed node
         If None, returned dist dict contents default to 0 for every node in the
         source list
-
-    cutoff: integer or float, optional
-        Depth to stop the search. Only paths of length <= cutoff are returned
 
     target: node label, optional
         Ending node for path. Path lengths to other destinations may (and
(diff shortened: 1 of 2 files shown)


dschult:
What description of the Bellman-Ford algorithm is this based on when it was rewritten recently? We need to make sure it is really bellman-ford and not a close substitute.

wuhaochen:
I think it is a correct implementation overall, the code piece I deleted is an attempted optimization that works on positively weighted edges.

The CI failed on doctest, where extra computation is exposed, I'll fix them now.

wuhaochen:
It seem that we have a design decision to make here.

The doctest failed on function [`bellman_ford_predecessor_and_distance`](https://github.com/networkx/networkx/blob/b26b4e928ba0a1fce4513960b59b1c5748678f58/networkx/algorithms/shortest_paths/weighted.py#L1103).

It has this warning in its return section:
       Warning: If target is specified, the dicts are incomplete as they
       only contain information for the nodes along a path to target.

In this commit, the dicts won't be incomplete and this caused difference in the later example.
Should we remove the warning and change the example or force upon the previous behavior?

```
Failed example:
    sorted(pred.items())
Expected:
    [(0, []), (1, [0])]
Got:
    [(0, []), (1, [0]), (2, [1]), (3, [2]), (4, [3])]
```

dschult:
Does the cutoff argument make sense for Bellman-Ford since we allow negative weights? Should that be removed too?

wuhaochen:
Good point. Should we remove the argument at all or add extra check in the end? I'm concerning that removing the argument may have backward compatibility issue.

dschult:
I think we should remove it completely. It will fail noisily so no surprises, and checking at the end doesn't save any time or resources, so is a pointless concept.

I'm also worried that this isn't the Bellman-Ford algorithm.  It seems to have been changed in a major way from the initial code a number of years ago and I want to check if there is a variant of the algorithm that this is following. If not, we should go back to a method with a provably correct solution in finite time -- or a negative cycle.

wuhaochen:
The cutoff parameter was introduced to match dijkstra, but it doesn't apply to graph with negative weights.

Besides, taking a list of nodes as source also doesn't make sense since the algorithm cannot handle it properly, I would suggest to revert it to only handle a single node.

networkx/algorithms/shortest_paths/weighted.py
```diff
@@ -1202,13 +1200,16 @@ def bellman_ford_predecessor_and_distance(G, source, target=None,
...
     dist = _bellman_ford(G, [source], weight, pred=pred, dist=dist,
-                         cutoff=cutoff, target=target)
+                         target=target)
     return (pred, dist)
 
 
 def _bellman_ford(G, source, weight, pred=None, paths=None, dist=None,
-                  cutoff=None, target=None):
-    """Relaxation loop for Bellman–Ford algorithm
+                  target=None):
```
wuhaochen: I'm not sure if this is the best way to cite the algorithm.

networkx/algorithms/shortest_paths/weighted.py
```diff
@@ -1202,13 +1200,16 @@ def bellman_ford_predecessor_and_distance(G, source, target=None,
...
+    """Relaxation loop for Bellman–Ford algorithm.
+
+    This is an implementation of the SPFA variant.
+    See https://en.wikipedia.org/wiki/Shortest_Path_Faster_Algorithm
```
wuhaochen: I'm not sure if this is the best way to cite the algorithm though.

dschult:
Yes -- Please keep it with comments describing the variant of Bellman-Ford.

And your other suggestions for handling single node and multiple nodes in all these algorithms make sense too.  Good work!!  :)

networkx/algorithms/shortest_paths/weighted.py
dschult: I agree, though there seems to be ambiguity over how to cite it.  It is claimed [on a WCIPEG webpage](https://wcipeg.com/wiki/Shortest_Path_Faster_Algorithm) that Yen's improvements on Bellman-Ford included this algorithm. But wikipedia cites a paper by Duan Fanding. All this can be found from links at the bottom of the wikipedia page, so this citation should be good for our purposes.

dschult:
This looks good.  As far as I can tell you've covered all the suggestions made above. Are you ready to merge this?  And Thanks again.

wuhaochen:
I added one more comments and it should be ready to go now.

PR #3136, opened 2018-08-30
Handle noncomparable nodes in isomorphism matcher

wbernoudy:
See description of issue in #3134.

### DIFF
--- a/networkx/algorithms/isomorphism/isomorphvf2.py
+++ b/networkx/algorithms/isomorphism/isomorphvf2.py
@@ -155,6 +155,18 @@ import networkx as nx
 __all__ = ['GraphMatcher',
            'DiGraphMatcher']
 
+def set_min(s):
+    """
+    Returns the minimum of the set is possible. Otherwise, returns one
+    of the items of the set.
+    """
+
+    try:
+        return min(s)
+    except TypeError:
+        # some items in the set could not be compared
+        return next(iter(s))
+
 
 class GraphMatcher(object):
     """Implementation of VF2 algorithm for matching undirected graphs.
@@ -226,7 +238,7 @@ class GraphMatcher(object):
         # P(s) = T1_inout x {min T2_inout}
         if T1_inout and T2_inout:
             for node in T1_inout:
-                yield node, min(T2_inout)
+                yield node, set_min(T2_inout)
 
         else:
             # If T1_inout and T2_inout were both empty....
@@ -234,7 +246,7 @@ class GraphMatcher(object):
             # if not (T1_inout or T2_inout):       # as suggested by  [2], incorrect
             if 1:                                  # as inferred from [1], correct
                 # First we determine the candidate node for G2
-                other_node = min(G2_nodes - set(self.core_2))
+                other_node = set_min(G2_nodes - set(self.core_2))
                 for node in self.G1:
                     if node not in self.core_1:
                         yield node, other_node
@@ -525,7 +537,7 @@ class DiGraphMatcher(GraphMatcher):
         # If T1_out and T2_out are both nonempty.
         # P(s) = T1_out x {min T2_out}
         if T1_out and T2_out:
-            node_2 = min(T2_out)
+            node_2 = set_min(T2_out)
             for node_1 in T1_out:
                 yield node_1, node_2
 
@@ -540,7 +552,7 @@ class DiGraphMatcher(GraphMatcher):
             # If T1_in and T2_in are both nonempty.
             # P(s) = T1_out x {min T2_out}
             if T1_in and T2_in:
-                node_2 = min(T2_in)
+                node_2 = set_min(T2_in)
                 for node_1 in T1_in:
                     yield node_1, node_2
 
@@ -549,7 +561,7 @@ class DiGraphMatcher(GraphMatcher):
 
             # elif not (T1_in or T2_in):   # as suggested by  [2], incorrect
             else:                          # as inferred from [1], correct
-                node_2 = min(G2_nodes - set(self.core_2))
+                node_2 = set_min(G2_nodes - set(self.core_2))
                 for node_1 in G1_nodes:
                     if node_1 not in self.core_1:
                         yield node_1, node_2
--- a/networkx/algorithms/isomorphism/isomorphvf2.py
+++ b/networkx/algorithms/isomorphism/isomorphvf2.py
@@ -120,19 +120,7 @@ Notes
 -----
 
 The implementation handles both directed and undirected graphs as well
-as multigraphs. However, it does require that nodes in the graph are
-orderable (in addition to the general NetworkX requirement that nodes
-are hashable). If the nodes in your graph are not orderable, you can
-convert them to an orderable type (`int`, for example) by using the
-:func:`networkx.relabel` function. You can store the dictionary of
-old-to-new node labels to retrieve the original node labels after
-running the isomorphism algorithm::
-
-    >>> G = nx.Graph()
-    >>> node1, node2 = object(), object()
-    >>> G.add_nodes_from([node1, node2])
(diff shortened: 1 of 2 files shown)


wbernoudy:
Digging a bit into the VF2 code, it looks like the only time it a single node is pulled out from a set is in `candidate_pairs_iter`, which as the name suggests, returns a generator which yields all possible matchings of one available node in G2 to all available nodes in G1. `candidate_pairs_iter` is used only in `match` which recursively attempts to find viable matchings between G1 and G2. If a candidate pair of matched nodes is potentially viable, then the state of the mapping between G2 and G1 is updated, which includes adding the G2 node to `core_2` ([line 920](https://github.com/networkx/networkx/blob/25470bb145895b356113168fd2a51c6848f4f713/networkx/algorithms/isomorphism/isomorphvf2.py#L920)). This means on the next call of `candidate_pairs_iter`, that G2 node will not belong to set of "available" nodes that we call `set_min` on.

Basically, as far as I can tell, we do not need `set_min` to always return the same order of nodes. If you see a mistake in my analysis of the code or would still think it would be better to be sure, then I am happy to modify `set_min` to have a consistent ordering like you suggested.

dschult:
I'll dig into the code more this weekend.  Question: if we don't have to keep the order the same, then we shouldn't be using any ```min``` function at all.  Just ```nx.arbitrary_element``` or ```next(iter(```.  Would this work?

chebee7i:
Late to the party. :)

Just as an FYI, I attempted to document the requirements that this algorithm imposes on graphs in the module docstring: 

https://github.com/networkx/networkx/blob/master/networkx/algorithms/isomorphism/isomorphvf2.py#L124.  

In addition to nodes being hashable (already a NetworkX requirement), it also requires a total order on the nodes. There's also an example of how one can do a pre-computation step that will put a total order on the nodes.

Though it's been a while since I've looked at this code and algorithm, my recollection was that pulling out an arbitrary item has fairly substantial performance implications due to it re-visiting configurations that it has already ruled out.

chebee7i:
I no longer have access to paywall'd PDFs, but fortunately, I was able to find citation [1] from the docstring:

https://github.com/pfllo/VF2/blob/master/doc/A%20(Sub)Graph%20Isomorphism%20Algorithm%20for%20Matching%20Large%20Graphs.pdf

Right before Section 2.1:
>Note that the algorithm explores the search graph in the SSR according to a depth-first search strategy. Using this simple formulation, a state can be reached through different paths. In order to avoid that, during the matching process, the algorithm generates useless and already generated states, a special procedure for generating a successor node is used. An arbitrary, total order relation (denoted by $\prec$) is defined on the nodes of $G_2$ which belong to the set $P(s)$. Since the node insertion order in the partial solution $M(s)$ does not influence the resulting state, the algorithm ignores any pair $(n_i, m_j)$ in $P(s)$ if this set already contains a node $m_k \prec m_j$. This simple strategy allows the algorithm to generate each state only once.

dschult:
Thanks @chebee7i  !! 
The doctring is pretty clear about needing nodes with a total order.  It even talks about relabeling to obtain ordered nodes.  Sorry about not reading that! :{

In addition to the quote shown above from [1], the article [2] also clearly requires a total ordering of the nodes. 

So... should we rely on the docstring to tell people how to relabel? Or should we build in an ordering?
We could create one using the order of ```G.nodes``` and adapt ```min(someset)``` to use that ordering.  The cost in time would be 1) creating the ordering dict, and 2) requiring ```min``` to use a comparison function rather than finding min of the nodes themselves. The benefit would be easy handling of nodes that aren't ordered.  Am I missing any costs or benefits here?

chebee7i:
Yeah that seems about right. My feeling at the time, which was a while ago and things may have changed since then, is that most people use standard data types for their nodes like int, string, homogenous tuples, etc. For these common use cases, it is slower if we automatically try to build our own ordering since those already have a total order.


So my pref was that if you were doing something special where you had custom nodes, then you should define an ordering on them (super easy with `functions.total_ordering`), or relabel them before using the isomorphism class. If we go this route, the requirement should probably be spelled out more explicitly at the top-level functions, however. 

Now, since it is an arbitrary ordering that is required, we can always make it true without affecting the result. So I see the appeal of hiding this requirement from the user. But then you have to worry about inverse mapping the results back to the user nodes. There are a number of ways to interact with this class and each outpoint would need to be handled. For example, some have asked for access to the internal mapping as the search happens. This would require retranslating those nodes at each iteration, etc. I suspect it would get ugly and hard to maintain quickly...in addition to the big perf hit. I could see an argument that only the top-level `is_isomorphic` should auto-construct an ordering though, so long as people are comfortable with the slowdown when your nodes are already ordered.

dschult:
I think we can hide the total orderable requirement from the user without changing the mapping/nodes, etc. We change the ```min``` function so it can find a min from a total ordering.

    ```min(T2_inout)``` becomes ```min(T2_inout, key=total_order.__getitem__)```

The nodes don't change. So there is no issue with making the internal mapping available. That maps nodes, and doesn't care about ordering. The only place where ordering seems to matter is when generating candidate pairs. As you mention above, it is important to avoid considering equivalent pairs multiple times. The total order does that.

Performance: I tried to check performance. Performance in Python often seems to be counter-intuitive. Specifying the ```key``` argument in ```min``` did indeed seem to slow it down about 1% for graphs with 100 to 1000 nodes if the nodes were added in order. BUT if I shuffled the node numbers so they weren't ordered integers, the ```key``` argument method was about 1% faster. Presumably finding the min is slightly faster if the container is already ordered low to high. The vf2 code iterates over nodes from the graph and my ```total_ordering``` dict was based on the iteration order of the graph. So I conjecture that it is slightly faster to do ```min``` with a ```key``` that matches the iteration order of the nodes in the container, then to do ```min``` without a ```key``` on a container that doesn't have its iteration matching the desired order.

None of the changes I tested made a difference in speed of more than 2%. And no version was consistently ahead of the others.

Even stranger, I went ahead and implemented a version that didn't use ```min``` to sort the nodes. It just pulls the first node out of the container based on iteration order. I hoped that would show an example where the order mattered. It hasn't... and I have run it many times on many different size graphs.  Perhaps iteration order is sufficiently reliable for selecting the same element that it doesn't break the algorithm. (I was checking that the list of isomorphisms is equal using vf2 with ```min``` and using vf2 with ```next(iter```. The list of isomorphisms always was equal.)  

Back to timing -- I did not expect that the arbitrary element method was not faster(!) than computing the ```min```.  Starts making me wonder what takes time in Python.  Perhaps calling ```next``` and ```iter``` requires lookup time that cancels iterating over 400 nodes?  

I also was surprised to find that changing 

    T1_inout = [node for node in G1_nodes if node in self.inout_1 if node not in self.core_1]

to 

    T1_inout = [node for node in self.inout_1 if node not in self.core_1]

Actually slowed down the code about 2%.  It seems to me that the second should be faster.  Optimizing Python code is hard to do...

Bottom line: We should probably use the iteration order of ```G.nodes``` as the total ordering for nodes in the graph. It doesn't require orderable node objects. It doesn't consistently take more time -- seems to depend on whether the graph iteration ordering matches the node object ordering.  Time difference is about 1% either way.  It does cost memory with one dict keyed by node to integer.

chebee7i:
I like it! And yeah, that optimization you suggested does seem like it should perform better. Odd. 

Re perf: I vaguely remember all the function calls being a big deal in general, as well as the stack dives from recursion. I just noticed that someone put out a paper on VF2++ (https://www.sciencedirect.com/science/article/pii/S0166218X18300829). Looks very nice. Would be interesting to offload the isomorph ism checks to that LEMON library.

chebee7i:
Opps. Typing on my phone...that's what I get.

dschult:
@wbernoudy does this variation of the changes do what you need it to for your use-case? 
If so, maybe we can merge it before the upcoming release.

PR #3141, opened 2018-09-04
Mathematics and rendering fixes to algorithms.cluster documentation

huonw:
- The documentation for `clustering` had `c_u = ... \sum_{uv}` but used `w` within the sum, so I think it was probably meant to be `c_u = ... \sum_{vw}`.

- The whole file used `` `...` `` instead ``:math:`...` `` for inline math, which meant it rendered as (semi-)raw LaTeX, 

Comparison showing both of these:

image

Rendered documentation at  https://huonw.github.io/networkx/reference/algorithms/clustering.html. Another comparison that's particularly heavy on inline math:

- old: https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.square_clustering.html#networkx.algorithms.cluster.square_clustering
- new: https://huonw.github.io/networkx/reference/algorithms/generated/networkx.algorithms.cluster.square_clustering.html#networkx.algorithms.cluster.square_clustering

### DIFF
--- a/networkx/algorithms/cluster.py
+++ b/networkx/algorithms/cluster.py
@@ -283,7 +283,7 @@ def clustering(G, nodes=None, weight=None):
     .. math::
 
        c_u = \frac{1}{deg(u)(deg(u)-1))}
-             \sum_{uv} (\hat{w}_{uv} \hat{w}_{uw} \hat{w}_{vw})^{1/3}.
+             \sum_{vw} (\hat{w}_{uv} \hat{w}_{uw} \hat{w}_{vw})^{1/3}.
 
     The edge weights `\hat{w}_{uv}` are normalized by the maximum weight in the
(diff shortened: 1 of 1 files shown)


dschult:
This looks great -- thank you very much!
Can you make sure the new source satisfies PEP8 standards?
That most often means checking that the linewidth is 79 chars or less.
But there are other potential issues. You can use ```pip install pep8``` or ```flake8``` or maybe others.

huonw:
Fixed. It seems the only problems in the changes I made was line length.

(Incidentally, is there a reason that `flake8` doesn't run as part of CI? I would've noticed and fixed the problems myself if I'd seen that the tests fail. 🙂 )

dschult:
No reason I know of that we don't check flake8 in CI except we haven't included that.

Issue #3129, opened 2018-08-21
nx.rescale_layout input problem. 


frhyme:
I am using `networkx.drawing.layout.rescale_layout`.

However, in this function, input is fixed to `np.array`.
In fact, other layouts (`nx.spring_layout`, `nx.spectral_layout`, etc.) return a dictionary (key: node label, value: pos tuple (x, y)).

So, for `networkx.drawing.layout.rescale_layout`, I think it would be useful to change the input to dictionary (key: node label, value: pos tuple (x, y)).

To change to the code:

I think this case is more versatile and I would like to ask you what you think.

```python
def rescale_layout(pos, scale=1):
    pos_v = np.array([[v[0], v[1]] for v in pos.values()])
    lim = 0  # max coordinate for all axes
    for i in range(pos_v.shape[1]):
        pos_v[:, i] -= pos_v[:, i].mean()
        lim = max(abs(pos_v[:, i]).max(), lim)
    # rescale to (-scale, scale) in all directions, preserves aspect
    if lim > 0:
        for i in range(pos_v.shape[1]):
            pos_v[:, i] *= scale / lim
    return {k: tuple(v) for k, v in zip(pos.keys(), pos_v)}
```

dschult:
We wrote ```rescale_layout``` with those inputs to reduce the number of times we have to convert back and forth between arrays and dicts.  How about implementing your suggestion as a new function rather than replacing the current one:

    def rescale_layout_dict(pos, scale=1)
        pos_array = np.array([[px, py] for px, py in pos.values()])
        new_pos_array = rescale_layout(pos_array, scale=scale)
        return {v: tuple(p) for v, p in zip(pos, new_pos_array)}

PR #3142, opened 2018-09-04
add new function rescale_layout_dict

frhyme:
the issue was uploaded in [#3129](https://github.com/networkx/networkx/issues/3129)

in function `networkx.drawing.layout.rescale_layout`, input is fixed to `np.array`.
In fact, other layouts (`nx.spring_layout`, `nx.spectral_layout`, etc.) return a dictionary (key: node label, value: pos tuple (x, y)).

So, I think it would be useful to add new function(the input is dictionary(key: node label, value: pos tuple (x, y)).

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -894,6 +894,24 @@ def rescale_layout(pos, scale=1):
             pos[:, i] *= scale / lim
     return pos
 
+def rescale_layout_dict(pos, scale=1):
+    """Return a dictionary of scaled positions keyed by node 
+
+    Parameters
+    ----------
+    pos : A dictionary of positions keyed by node
+
+    scale : number (default: 1)
+        The size of the resulting extent in all directions.
+
+    Returns
+    -------
+    pos : A dictionary of positions keyed by node
+
+    """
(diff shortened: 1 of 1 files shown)


dschult:
Thanks for this!

Just a couple comments -- mostly about getting the function integrated with the whole networkx API.

- You should include the new function in the ```__all__``` variable defined at the top (to have it imported to the networkx namespace.
- You should add "See Also" sections to both rescale function's doc_string to refer to each other
- You should update the documentation page ```doc/reference/drawing.rst``` by adding the name of the new function to the list of layout functions
- You should add a few simple tests to ```networkx/drawing/tests/test_layout.py``` to make sure this function runs and gives reasonable answers.
- The first line defining ```pos_v``` unpacks the position tuple only to repack it as a tuple again. Could do:  ```pos_v = np.array(pos.values())``` or just remove the line and in the next line replace ```pos_v``` with ```np.array(pos.values())```.

frhyme:
Thank you for your comment.
I will reflect all the comments you have mentioned. 
and then I will create a new pull request afterwards.
Thank you.

PR #3145, opened 2018-09-06
add a new function "rescale_layout_dict" again. 

frhyme:
There are some comment to integer it. and I fixed it all. 

1. You should include the new function in the __all__ variable defined at the top (to have it imported to the networkx namespace. ==> **complete**
2. You should add "See Also" sections to both rescale function's doc_string to refer to each other ==> **complete**
3. You should update the documentation page doc/reference/drawing.rst by adding the name of the new function to the list of layout functions ==> **complete**
4. You should add a few simple tests to networkx/drawing/tests/test_layout.py to make sure this function runs and gives reasonable answers. ==> **complete, make simple test for checking empty graph** 
5. The first line defining pos_v unpacks the position tuple only to repack it as a tuple again. Could do: pos_v = np.array(pos.values()) or just remove the line and in the next line replace pos_v with np.array(pos.values()). ==> **little changed it. However, `pos_v = np.array(pos.values())` will raise error "AttributeError: 'dict_values' object has no attribute 'shape'" in `rescale_layout`, so I should changed it to `pos_v = np.array(list(pos.values()))` with converting list**

Thanks for your help. 
If there were more issues or feedback to reflect it, Please let me know.


frhyme:
Maybe There are some problems there.

PR #3146, opened 2018-09-06
add new function "rescale_layout_dict" again

frhyme:
There are some comment to integer it. and I fixed it all.

- You should include the new function in the all variable defined at the top (to have it imported to the networkx namespace. ==> **complete**

- You should add "See Also" sections to both rescale function's doc_string to refer to each other ==> **complete**

- You should update the documentation page doc/reference/drawing.rst by adding the name of the new function to the list of layout functions ==> **complete**

- You should add a few simple tests to networkx/drawing/tests/test_layout.py to make sure this function runs and gives reasonable answers. ==> **complete, make simple test for checking empty graph**

- The first line defining pos_v unpacks the position tuple only to repack it as a tuple again. Could do: `pos_v = np.array(pos.values())` or just remove the line and in the next line replace pos_v with `np.array(pos.values())`. ==> **little changed it. However, `pos_v = np.array(pos.values())` will raise error "`AttributeError: 'dict_values' object has no attribute 'shape'`" in rescale_layout, so I should changed it to` pos_v = np.array(list(pos.values()))` with converting list**

Thanks for your help.
If there were more issues or feedback to reflect it, Please let me know.


networkx/drawing/layout.py
```diff
@@ -894,6 +899,32 @@ def rescale_layout(pos, scale=1):
...
+    See Also
+    --------
+    rescale_layout
+    """
+    import numpy as np
+    
+    if len(pos)==0:
+        ## the empty graph 
+        return {}
+    pos_v = np.array(list(pos.values()))
```
dschult: The AttributeError you mention will occur if you use ```pos.values()``` without the ```np.array``` call.  There should be no AttributeError with ```np.array(pos.values())``` because the object is a numpy array which DOES have a ```shape``` attribute.  So you can take out the ```list``` here.

Also, you should not have to create new pull requests for each change. Simply commit the changes to the same branch you made the original changes to. The changes accumulate, but are "squashed" into a single commit when we merge the pull request into the repository. You can "push" changes to the same branch and they automatically get added to the pull request.  This is better than creating many pull requests because the entire history and comments, etc are stored in a single webpage for that pull request.

The test you created doesn't have any nodes in it. So, the test doesn't actually use any code inside a loop over nodes. A test with one node would do more, but it seems like you really need at least 2 nodes to test the rescale feature. Perhaps create a position dict with 2 nodes and integer values positions. Then set scale=2 and see if the positions double.

dschult:
Fixed by #4154

PR #3165, opened 2018-09-20
fix prefix_tree doc string

kdbhiggins:
Very minor docs typo. `T = nx.prefix_tree(paths)` returns a tuple and `T.predecessors(NIL)` will throw. The other examples all have `T, root = nx.prefix_tree(paths)`.


dschult:
Thanks -- but it looks like more is needed. The docstring says that a DiGraph is returned. But clearly two objects are returned... a DiGraph and a root object.

kdbhiggins:
Thanks for the quick response. I've cleaned up the docstring a bit more.

dschult:
Thanks very much!

PR #3169, opened 2018-09-27
fix zero division

j40903272:


dschult:
Hmmmm.... doesn't ```G.degree(w)``` have to be at least 2? 
```w``` has neighbors ```u``` and ```v```.  

Can you explain a little more in what situation this arises?

j40903272:
It happens when a node has an edge with itself (a weird thing in my dataset).
 `nx.common_neighbors` gives all neighbors of the node, including those with only one edge, which leads  `G.degree(w)` be 1.

dschult:
It looks like the fix will make the index VERY large for any self-loop edges.
That is not how the index was intended. I prefer to put a warning in the docstring not to apply this to self-loops. Perhaps it is worth checking for that in the code, but I prefer it in the docs.
Thanks for bringing up this issue!

PR #3171, opened 2018-09-28
Import ABCs from collections.abc

### LINKED ISSUE #3170: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated
Running code that depends on networkx with the upcoming python 3.8 yields the following warnings:

```
/coarse/jon/Envs/pygit/lib/python3.8/site-packages/networkx/classes/graph.py:23: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping

/coarse/jon/Envs/pygit/lib/python3.8/site-packages/networkx/classes/reportviews.py:95: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping, Set, Iterable
```

The abstract syntax classes are in the `collections.abc` module since python 3.3. Python 2.7 still defines them directly in the `collections` module.

jbarnoud:
Importing Abstract Base Classes directly from the collections module is
deprecated in python 3.8 in favour of the collections.abc module.
Importing from collections is the only option on python 2.7, though.

This PR tries to import ABCs from collections.abc, and fallback to
collections in case of failure.

Fixes #3170

### DIFF
--- a/networkx/algorithms/approximation/kcomponents.py
+++ b/networkx/algorithms/approximation/kcomponents.py
@@ -6,12 +6,7 @@
 #    BSD license.
 import itertools
 from collections import defaultdict
-# Importing Abstract Base Classes directly from collections is deprecated
-# in pythn 3.8. There is no other option in python 2.7, though.
-try:
-    from collections.abc import Mapping
-except ImportError:
-    from collections import Mapping
+from collections.abc import Mapping
 
 import networkx as nx
 from networkx.exception import NetworkXError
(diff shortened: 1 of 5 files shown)


jbarnoud:
@MarkCBell I forgot one, hence the remaining warning. Running `python -W once -m nose networkx` does not issue any ABC related warning anymore. At least none in networkx, there are some related to nose.

dschult:
With the release of NetworkX 2.2, we no longer plan to support Python2.7.
That doesn't necessarily mean we start removing support for 2.7, but it could.
We should figure out what we're doing with 2.7...

jbarnoud:
I rebased against master and removed python 2.7 support since this version of python is not supported anymore.

Issue #3184, opened 2018-10-05
read_gml() differs from documentation


clayadavis:
The documentation for [networkx.readwrite.gml.read_gml](https://networkx.github.io/documentation/stable/reference/readwrite/generated/networkx.readwrite.gml.read_gml.html#networkx.readwrite.gml.read_gml) states the following:

> * label (string, optional) – If not None, the parsed nodes will be renamed according to node attributes indicated by label. Default value: ‘label’.

Expected behavior: `nx.read_gml('karate.gml', label=None)` will read the file and not rename the nodes.

Current behavior: `nx.read_gml('karate.gml', label=None)` reads the file and attempts to rename the nodes according to their 'label' attribute, which may not exist. In fact, any string passed to the `label=` kwarg is currently ignored because the string `'label'` is used instead of the `label` kwarg: https://github.com/networkx/networkx/blob/be7c67553d2d27db4c66ce20942e6e443c39bef5/networkx/readwrite/gml.py#L424

PR incoming.

dschult:
fixed in #3185 3185

PR #3185, opened 2018-10-05
Fix reading labels from gml files (#3184)

clayadavis:
This PR fixes issue #3184 in reading GML files, bringing the behavior in line with the documentation.


dschult:
This looks good.  Can you put in a test or two to make sure we keep this (now correct) behavior?
I haven't looked at the tests carefully in a long time, so if its not easy just say so.

clayadavis:
@dschult I added a test. This is my first NX contribution, so don't hesitate to nitpick if something isn't up to par.

dschult:
Excellent -- Thanks very much!

PR #3190, opened 2018-10-12
Fix deprecated graph.selfloop_edges() method

mikebabst:
The deprecated graph.selfloop_edges() method call to nx.selfloop_edges() was overriding parameters.  This fix passes the parameters through.


dschult:
Thanks!

mikebabst:
You're welcome!  I found this while upgrading from the older 1.x API.

Issue #3191, opened 2018-10-16
pagerank appear to be non deterministic


LoreDema:
If I run pagerank several time I got different  results each time.

PR #3192, opened 2018-10-16
Making pagerank deterministic version available (#3191)

LoreDema:


dschult:
This fix relies on nodes being comparable (sortable) which they aren't in general.

Also, I believe pagerank is deterministic in python 3.6+ where dict's iterate in a deterministic way. but maybe I don't understand the meaning of "deterministic" completely.  Is it enough that running it twice in the same session gives the same answer, on the same machine, or does it need to be the same on all versions of python on all machines?  what is the goal here?   (and Thanks for the idea!)

LoreDema:
Would be ideal for me  to get the same results on all machines, given a version of Python, which is BTW for me 3.5.2. Probably more in general would be awesome to get also same results with different versions of Python, that's way I have introduced the sorting, didn't know that nodes are not comparable in general.

dschult:
I'm going to close this for lack of handling any hashable node.

PR #3193, opened 2018-10-17
Speed up simple_cycles

GxvgiuU:
Really appreciate your work on this wonderful lib. These are codes written by genius!

And, I found below example,
```
import networkx as nx
g = nx.DiGraph()
for i in range(10000):
    g.add_edge(i, i + 1)
g.add_path([10005, 10007, 10005])
g.add_path([10014, 10016, 10014])
list(nx.simple_cycles(g))
```
runs surprisingly slow. 
After I look into this, I would like to make 2 changes.

In Johnson's paper,
>  Our definitions exclude graphs with loops (edges of the form (v, v)) and multiple edges between the same vertices. 

So I would like to remove self-cycle edges out. This reduces huge amount of time. If you really want to do this (to be back compatible, although not consistent with its raw paper def), it would still be faster to check it in advance within O(n) time (I would add it if needed).

Another one is the component subgraph can be induced by `scc` nodes, which will also reduce some meaningless operation related with stack/blocked compared to subgraph `subG`.

Hope to make this lib even better! Would appreciate hearing your thoughts.

### DIFF
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -181,9 +181,20 @@ def simple_cycles(G):
     # Also we save the actual graph so we can mutate it. We only take the
     # edges because we do not want to copy edge and node attributes here.
     subG = type(G)(G.edges())
-    sccs = list(nx.strongly_connected_components(subG))
+    sccs = [scc for scc in nx.strongly_connected_components(subG)
+            if len(scc) > 1]
+
+    # Johnson's algorithm exclude self cycle edges like (v, v)
+    # To be backward compatible, we record those cycles in advance
+    # and then remove from subG
+    for v in subG:
+        if subG.has_edge(v, v):
+            yield [v]
+            subG.remove_edge(v, v)
+
     while sccs:
         scc = sccs.pop()
+        sccG = subG.subgraph(scc)
         # order of scc determines ordering of nodes
         startnode = scc.pop()
         # Processing node runs "circuit" routine from recursive version
@@ -192,7 +203,7 @@ def simple_cycles(G):
         closed = set()   # nodes involved in a cycle
         blocked.add(startnode)
         B = defaultdict(set)  # graph portions that yield no elementary circuit
-        stack = [(startnode, list(subG[startnode]))]  # subG gives comp nbrs
+        stack = [(startnode, list(sccG[startnode]))]  # sccG gives comp nbrs
         while stack:
             thisnode, nbrs = stack[-1]
             if nbrs:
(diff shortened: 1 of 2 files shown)


dschult:
I'm confused.  
1) Your example has no self-loops. And I don't see how handling them at the start saves time.
2) Your idea to switch to ```sccG``` instead of ```subG``` is only used for looking up neighbors as in ```sscG[startnode]```.  But looking up neighbors is the same speed for sccG as it is for subG. The speedup occurs when we compute the strongly connected components, but that already uses ```H = subG.subgraph(scc)``` so we've saved that time already.

Are you seeing a speedup? Where is it coming from?

GxvgiuU:
> 1. Your example has no self-loops. And I don't see how handling them at the start saves time.

The first thing came to me is that I do not consider self-loops. The speed-up happens in L185 which reduce huge amount of strongly-connected-components as single nodes. If cut that down, there would be less pop and other operations in `while sccs` loop. 

I found the time difference using the example above. Actually, I think run this alg on a path like really long linked list could lead to the diff.

But there are self-cycles in test cases and doc tests. So I add that at the beginning.


> 2\. is only used for looking up neighbors as in `sscG[startnode]`

At L217 when to expand stack, this also reduces the number of neighbors needed to explore.

dschult:
Ahh...  got it.  By throwing out length 1 scc you reduce the number of sccs a huge amount when there are directed paths.   But that made some self-loops not appear in the tests, so you added code to yield them.  So it's the handling of singleton components that speeds up the code.

And I am mistaken about the ```sccG[nextnode]``` as well. I was getting confused between connected_component (which is for Graph) and strongly_connected_component (which is for DiGraph).   Thanks for helping me straight it out.

In my poking around, I notice that one of the tests:  ```test_simple_cycles``` on line 62 checks for all entries in cc to be in ca.  But it doesn't test that all entries in ca are in cc.  Could you add a line after the loop with something like: ```assert_equal(len(cc), len(ca))```.  Similarly for line 93 and maybe other places.

Thanks for the speed-ups!!

GxvgiuU:
I have amended the tests. But didn't find where to change around L93.

GxvgiuU:
nop. pls merge it.

GxvgiuU:
nop. pls merge it.

GxvgiuU:
nop. pls merge it.

GxvgiuU:
nop. pls merge it.

GxvgiuU:
nop, let's merge it

GxvgiuU:
nop, let's merge it.

PR #3194, opened 2018-10-18
Added dual Barabasi-Albert model

niemasd:
I added an extension of the Barabasi-Albert model that, when a new node is added, creates `m1` new edges with probability `p` and creates `m2` new edges with probability `1-p`. The graph is initialized with `p*m1 + (1-p)*m2` nodes. In other words, when `p = 1`, it reduces to Barabasi-Albert with `m = m1`, when `p = 0`, it reduces to Barabasi-Albert with `m = m2`, and when `0 < p < 1`, it's a mixture of the two.


dschult:
Thanks!   
- you reference the BA paper. Does that have the dual in it? Any reference for the dual?
- could you add some tests in e.g. ```tests/test_random_graphs.py```?

dschult:
:)
I don't think you need to put a manuscript out there in order to cite it here.  But when you do publish it, it'd be great to come back here and add it to the documentation. :)

niemasd:
The short write-up is now on [arXiv](https://arxiv.org/abs/1810.10538), and I've updated the reference in my generator function, so the pull request is now 100% good-to-go!

dschult:
It looks like a second function ```multi_...``` slipped into the last commit.  It needs docs and tests. Or you can pull it out and add it in a later PR.

niemasd:
Woops, sorry about that! I was playing around with some other ideas and forgot to remove it before I updated the dual-BA citation. It's removed in the most recent commit. Thanks for checking!

dschult:
Thanks very much!!

niemasd:
Thanks! I see that the code has been incorporated (https://networkx.github.io/documentation/latest/_modules/networkx/generators/random_graphs.html), but it doesn't appear in the docs (https://networkx.github.io/documentation/latest/reference/generators.html). I can't seem to figure out what I should have edited to make it appear. Any guidance would be much appreciated!

dschult:
Oops.. Sorry about that -- I didn't check the docs. 
Open a new PR I guess.

The way to add this function to the docs is to edit the file:  doc/reference/generators.rst
Find the section for the module where your function is and add it to the list of functions.

The Sphinx doc making program automatically generates the rst files from the docstrings -- but you have to list which functions to include.  To run this on your local machine you'll need to ```pip install sphinx``` or conda or whatever package system you use. Then in the doc/ directory use ```make html```. The result is put in doc/build/html

If that is any trouble at all let me know and I'll make the change.

niemasd:
Woops, sorry about that! I'm a bit new to Python package development! I've added the reference to `doc/reference/generators.rst` in https://github.com/networkx/networkx/pull/3221

Issue #3204, opened 2018-10-22
Node attr dict values container not customize-able


chebee7i:
For subclasses of the primary graph classes, it doesn't look like we allow the developer to specify a factory for the values of `G.nodes`. The code does:
```
self._node[node_for_adding] = attr
```
instead of calling a user-provided factory to create the object and then doing `obj.update(attr)`. Similarly for the code in `add_edge` etc. This seems like an unintended omission given that we allow developers to specify a `edge_attr_dict_factory`.

Similarly, it doesn't look like we allow the developer to control `self.graph`.

dschult:
We even have a node factory variable set up to do this: ```node_dict_factory```. We just don't use it. And I'm pretty sure we did used to use it.  It is certainly intended to be used.  I consider this a bug -- certainly doesn't agree with the docs that describe it.

We haven't got a factory for ```self.graph``` and maybe we should. It's only ever created once (there is only one such object) and it can be overwritten easily as ```G.graph=mydict()```. But for consistency I guess we can make a factory for it part of the class.

chebee7i:
We do use `node_dict_factory`, but it is only used in `__init__` to create `self._node`. It seems to be documented correctly, perhaps. It's just not as useful right now, preventing an analogous `ThinNodeGraph` subclass.  When I looked at it, it seemed like using a new node attr factory would also simplify some of the code in `add_node`, `add_nodes_from`, `add_edge`, etc.

dschult:
Ahh... got it...    so something named like:  ```node_attr_dict_factory```
would love to simplify some of that code in ```add_node```, ```add_edge```, etc too.
:)

PR #3205, opened 2018-10-23
implemented node_attr_dict_factory and graph_attr_dict_factory.

stsouko:
this is a first commit for implementing this issue #3204.
if this approach will be satisfying, I will make changes to all types of graphs in this PR.


networkx/classes/digraph.py
```diff
@@ -457,28 +463,15 @@ def add_nodes_from(self, nodes_for_adding, **attr):
...
-            # CPython throws TypeError on n not in self._succ,
-            # while pre-2.7.5 ironpython throws on self._succ[n]
             try:
-                if n not in self._succ:
-                    self._succ[n] = self.adjlist_inner_dict_factory()
-                    self._pred[n] = self.adjlist_inner_dict_factory()
-                    self._node[n] = attr.copy()
-                else:
-                    self._node[n].update(attr)
+                hash(n)
```
dschult: You are only testing for whether the node is hashable, but this code is supposed to check if the node is already in the DiGraph.

stsouko: old code 
```
try:
    if n not in self._succ:
        self._succ[n]
```
this code tried to add|update node.

but if exception raised 
```
nn, ndict = n
```
this code try to unpack (node, data) tuple and after  
```
if nn not in self._succ:
```
this code is duplicate of previous try block.

hash(n) is more simple check IMO.

dschult:
I'll try to start a more complete comment here in the main conversation.
There are many things I like about the proposed changes. But it changes 
much more than the PR was intended to do. The main focus was to allow 
a custom node attribute dict factory similar to that for ```edge_attr_dict_factory```.
Also a graph attr dict factory seems reasonable. 
Some of the "extra" changes break/change functionality, some are simply
style changes without obvious benefit. I will focus mainly on the things I'd
like to see remain the same rather than on the large number of good changes.
That is unfortunate in terms of the tone here, but I think it most efficient.
Don't take it the wrong way...

The doc changes look good.

Leave the assignments (that seem to do nothing) which move the class variables that hold factories to instance variables (L293 in digraph and similar in graph).

Leave the node checking with ```self._succ``` and ```self._adj``` rather than shift to ```self._node```

Leave the code that adds the node in ```add_nodes_from``` rather than pushing it to ```add_node``` If ```add_nodes_from``` becomes a loop with calls to ```add_node```, we might as well remove it. Users can do that.

Leave the code in ```add_edge``` that adds a new edge_attr_dict... that was optimized -- not sure it still is of course, but those changes require more testing and should be in a different PR.

Leave the code to add edges in ```add_edges_from``` rather than off-loading to add_edge.

Leave the code in ```copy``` unchanged.

Thanks very much for this!

stsouko:
I found that multigraph and multidigraph in add_edges_from use add_edge method.
I implemented add_edges_from without add_edge usage.
I can move all changes to separate PR If need.

stsouko:
I have prepared performance tests.

existing code slower.

```
def old_n():
    u = 1
    v = 2
    _adj = {1: {}, 2: {}}
    datadict = _adj[u].get(v, {})
    datadict.update({})
    _adj[u][v] = datadict
    _adj[v][u] = datadict


def new_n():
    u = 1
    v = 2
    _adj = {1: {}, 2: {}}
    if v not in _adj[u]:
        attr_dict = _adj[u][v] = _adj[v][u] = {}
        attr_dict.update({})
    else:
        _adj[u][v].update({})


def old_e():
    u = 1
    v = 2
    _adj = {1: {2: {}}, 2: {1: {}}}
    datadict = _adj[u].get(v, {})
    datadict.update({})
    _adj[u][v] = datadict
    _adj[v][u] = datadict


def new_e():
    u = 1
    v = 2
    _adj = {1: {2: {}}, 2: {1: {}}}
    if v not in _adj[u]:
        attr_dict = _adj[u][v] = _adj[v][u] = {}
        attr_dict.update({})
    else:
        _adj[u][v].update({})


from timeit import timeit


print('old code. new edge', timeit('old_n()', 'from __main__ import old_n', number=100000))
print('old code. existing edge', timeit('old_e()', 'from __main__ import old_e', number=100000))
print('new code. new edge', timeit('new_n()', 'from __main__ import new_n', number=100000))
print('new code. existing edge', timeit('new_e()', 'from __main__ import new_e', number=100000))
```

old code. new edge 0.13050702500186162
old code. existing edge 0.13641421899956185
new code. new edge 0.11319191099755699
new code. existing edge 0.11032685999816749

dschult:
Thanks... This makes it easier for me... I restarted the hung test.

PR #3208, opened 2018-10-24
Fix `simple_paths` targets parameter

GxvgiuU:
Hey, I've recently use `simple_paths` function. I found https://github.com/networkx/networkx/pull/3074 which replaces `list` to `OrderedDict` speeds up it a lot. Thanks to that. But after that, I found https://github.com/networkx/networkx/pull/3138/ a little misleading.

In the doc, there is no warning about it when we pass in parameters as targets, those should be **out-degree 0**. Otherwise it may not return all_simple_paths as this function name indicates.

Actually, I noticed the added doc itself is likely a misuse.
```
        >>> G = nx.complete_graph(4)
        >>> for path in nx.all_simple_paths(G, source=0, target=[3, 2]):
        ...     print(path)
        [0, 1, 2]
        [0, 1, 3]
        [0, 2]
        [0, 3]
```
in this complete_graph, nodes [2, 3] are not out-degree 0. So the result paths miss some simple path like [0, 1, 2, 3], [0, 2, 1, 3] ...

There also exists same mistakes in the test.  `path_graph` used in the test not creating using `DiGraph` do not have out-degree 0 nodes.

there is also a bug for the cutoff part in if branch (`# len(visited) == cutoff`)
```
G = nx.DiGraph()
G.add_edges_from([(0,1), (1,2), (1,3)])
list(nx.all_simple_paths(G, source=0, target=[2, 3], cutoff=2))
```
in this example, pervious code would only get 1 path (which should be 2 paths of course).

However, I think in one pass of DFS, we could enumerate `all_simple_paths` to many targets, **including but not limit to** leaf nodes (as out-degree 0). pls check the code below. 

If that is OK, we do not need to restrict targets to be leaf nodes when we pass in more than one targets parameters.

### DIFF
--- a/networkx/algorithms/tests/test_simple_paths.py
+++ b/networkx/algorithms/tests/test_simple_paths.py
@@ -95,10 +95,31 @@ def test_all_simple_paths_with_two_targets_emits_two_paths():
     assert_equal(set(tuple(p) for p in paths), {(0, 1, 2, 3), (0, 1, 2, 4)})
 
 
-def test_all_simple_paths_with_two_targets_in_line_emits_one_path():
+def test_digraph_all_simple_paths_with_two_targets_emits_two_paths():
+    G = nx.path_graph(4, create_using=nx.DiGraph())
+    G.add_edge(2, 4)
+    paths = nx.all_simple_paths(G, 0, [3, 4])
+    assert_equal(set(tuple(p) for p in paths), {(0, 1, 2, 3), (0, 1, 2, 4)})
+
+
+def test_all_simple_paths_with_two_targets_cutoff():
+    G = nx.path_graph(4)
+    G.add_edge(2, 4)
+    paths = nx.all_simple_paths(G, 0, [3, 4], cutoff=3)
+    assert_equal(set(tuple(p) for p in paths), {(0, 1, 2, 3), (0, 1, 2, 4)})
+
+
+def test_digraph_all_simple_paths_with_two_targets_cutoff():
+    G = nx.path_graph(4, create_using=nx.DiGraph())
+    G.add_edge(2, 4)
+    paths = nx.all_simple_paths(G, 0, [3, 4], cutoff=3)
+    assert_equal(set(tuple(p) for p in paths), {(0, 1, 2, 3), (0, 1, 2, 4)})
+
+
+def test_all_simple_paths_with_two_targets_in_line_emits_two_paths():
     G = nx.path_graph(4)
     paths = nx.all_simple_paths(G, 0, [2, 3])
-    assert_equal(set(tuple(p) for p in paths), {(0, 1, 2)})
+    assert_equal(set(tuple(p) for p in paths), {(0, 1, 2), (0, 1, 2, 3)})
 
 
 def test_all_simple_paths_ignores_cycle():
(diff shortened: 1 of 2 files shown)


dschult:
@winni2k can you comment on this?

networkx/algorithms/tests/test_simple_paths.py
```diff
@@ -129,18 +146,34 @@ def test_all_simple_paths_cutoff():
     assert_equal(set(tuple(p) for p in paths), {(0, 1), (0, 2, 1), (0, 3, 1)})
 
 
+def test_all_simple_paths_non_trivial():
```
winni2k: I cannot easily tell from looking at this test whether the test was written correctly. Can you summarize the expected result in the function name?

winni2k:
I should probably respond to the original post as well:

>In the doc, there is no warning about it when we pass in parameters as targets, those should be out-degree 0. Otherwise it may not return all_simple_paths as this function name indicates.

That is correct.  The way the code is currently written, `all_simple_paths` returns all simple paths ending when the first target is found.

>There also exists same mistakes in the test. path_graph used in the test not creating using DiGraph do not have out-degree 0 nodes.

I am in favor of testing with more inputs. However, the algorithm does not care about out-degree, so I'm not sure about this statement. Maybe more interestingly: We are dealing with a single-edge and a multigraph implementation.  If we are going to test with multiple graph inputs, then multigraphs should be included as well.

> in this example, pervious code would only get 1 path (which should be 2 paths of course).

Good catch!

>However, I think in one pass of DFS, we could enumerate all_simple_paths to many targets, including but not limit to leaf nodes (as out-degree 0). pls check the code below.

Do you mean that your suggestions would not necessarily enumerate all simple paths to all targets?

winni2k:
>I just draw some nodes and edges casually and make it just a little more complex (so one may not see through it at once).

Ok, let's try this: Did you come up with this test in a test-driven manner? That is, does this test catch something that no other test catches? That might be one way to name/describe the purpose of the test.

After taking a closer look at the test, it appears to me that this test is testing the `cutoff` parameter. I would suggest replacing the "non-trivial" test with one along the lines of "test_all_simple_paths_with_two_targets_cutoff" with a cutoff of 2. Alternatively, maybe just remove the test if it's testing something that another test will catch. Lastly, you could leave it in, but be prepared for the next contributor to delete it if they do not understand its point...

winni2k:
@dschult: barring any more changes from @GxvgiuU, this looks good to me.

GxvgiuU:
Hi, I just rename it to `test_all_simple_paths_on_non_trivial_graph`. The `test_all_simple_paths_with_two_targets_cutoff` has already been used earlier. 

And, since there is a `nx.trivial_graph` function and a test named `test_trivial_path`, I think it is OK to express on its opposite situation.

dschult:
Did you all figure out whether ```all_simple_paths``` should stop when it reaches the first target or continue through targets to other targets to get all paths?

winni2k:
We are returning all simple paths between the source and every target. If all targets have degree 1, then the behavior of this function is not changed. If some targets have degree > 1, then this function should now return the same paths that would be retrieved if every target was fed into `all_simple_paths` one at a time. 

@GxvgiuU, is my summary correct?

winni2k:
I agree with @GxvgiuU that this behavior is likely more intuitive to most users and that we should adopt it.

PR #3214, opened 2018-10-27
edges and nodes addition fixes

stsouko:
Graphs add_edge, add_edges_from, add_node, add_nodes_from methods speedup.
Graphs init cleaning.

this is second PR after #3205.

also I want to add all fixes regarding this looking through the whole project.


stsouko:
in this https://github.com/networkx/networkx/pull/3205#issuecomment-433143535 performance tests presented.

PR #3215, opened 2018-10-29
converter node_attr_dict override fixed.

stsouko:


dschult:
Can you explain what you're thinking here?
I saw this in #3214 too and tried to figure it out... something about possibly adding nodes to ```G._node``` without adding nodes to the adjacency structures. Did I get it?

stsouko:
for subclassed Graph with custom node_attr_dict instantiation with incoming_graph_data which is common Graph object, existing node attrs will be common dicts.
not node_attr_dict custom objects.

this PR fix this error.

dschult:
Thanks!!

PR #3223, opened 2018-11-06
Make approximation subpackage available in networkx namespace.

sandoracs:
Hi!

I guess, the approximation subpackage should be imported too. What is your opinion about it?

### DIFF
--- a/networkx/algorithms/__init__.py
+++ b/networkx/algorithms/__init__.py
@@ -1,3 +1,4 @@
+from networkx.algorithms.approximation import *
 from networkx.algorithms.assortativity import *
 from networkx.algorithms.boundary import *
 from networkx.algorithms.bridges import *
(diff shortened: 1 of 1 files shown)


sandoracs:
I guess, there is an issue with the algorithms/connectivity tests.

dschult:
That's almost certainly a name collision between the approximation namespace and the connectivity namespace. Don't have time to look now. But adding the approx namespace to the whole broke a lot of links... so I suspect a namespace collision.

sandoracs:
Yes, it sounds logical. Let me take a look.

sandoracs:
I've found the following in the related init file.

> """Approximations of graph properties and Heuristic functions for optimization
problems.
    .. warning:: The approximation submodule is not imported in the top-level
        ``networkx``.
    These functions can be imported with
    ``from networkx.algorithms import approximation``.
"""

Is it a design decision or can we solve it by avoiding namespace collisions?

hagberg:
I remember it being intentional to not automatically import the approximation package.   
In addition to the (fixable) name conflicts I argued that this might easily confuse users.  If you don't know the context of the approximation validity (which can be a little complicated) you can easily get bad answers or maybe just plan wrong answers.

sandoracs:
I see. Thank you for the explanation.

Issue #3213, opened 2018-10-27
Memory savings for edges without attributes


dschult:
Try the suggested delayed dict-of-dict strategy suggested in [this stackoverflow post](https://stackoverflow.com/questions/43894902/delay-creating-python-dict-until-set-update/50654122#50654122) for edge attributes. It should result in ThinGraph storage requirements until an attribute is looked up or set, then same memory usage as current system.

mohdkashif93:
I have read the answer, it seems really promising. Is it okay if I take a look into it ?

PR #3229, opened 2018-11-07
add_nodes_from fixed for custom hashable node_attr_dict.

stsouko:
node_attr_dict may be hashable dict-like object and this example give unexpected result:

```
class hashable_dict(dict):
    # bad class, but enough to check add_nodes_from method
    def __hash__(self):
        return hash(tuple(self.items()))

class CustomGraph(nx.Graph):
    node_attr_dict_factory = hashable_dict

G = CustomGraph()
G.add_node(1, w=1)

H = nx.Graph()
H.add_nodes_from(G.nodes(data=True))
H.nodes[1]  # raise KeyError
H.nodes  # NodeView(((1, {'w': 1}),))
```

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -10,6 +10,7 @@
 #            Pieter Swart <swart@lanl.gov>
 """Base class for directed graphs."""
 from copy import deepcopy
+from collections import Mapping
 
 import networkx as nx
 from networkx.classes.graph import Graph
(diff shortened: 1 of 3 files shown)


dschult:
I think protecting people against using hashable-dicts as attribute dicts is better done by documenting that they "don't do that!"....  :)

We're adding a fair amount of argument checking of node types for everyone using ```add_nodes_from``` in order to protect the few people who 1) want to use hashable attribute dicts, 2) want to use ```add_nodes_from```, and 3) want to specify the hashable attribute dicts as part of the ```add_nodes_from``` input.   

I haven't checked the time difference for normal users, but I prefer that in the description for ```node_attr_dict_factory``` we warn people that if they make their dicts hashable, they shouldn't use ```add_nodes_from``` with those dicts.  Indeed, maybe it would be better to tell them that if they make their dicts hashable, they should subclass to redefine ```add_nodes_from```.

I'd like to hear other's opinions on the ```node_attr_dict_factory``` addition and whether we should protect users from themselves. Perhaps I'm being too rigid...

stsouko:
dict-like objects can be a slotted storage (`__slots__`).
this is a way for memory saving issued in #3213.
this objects may be hashable.

only one limitation for this type of storage need: subclassing of collections.abc.Mapping for correct processing.

stsouko:
ok. I prepared tests. this PR reduces speed up to 5 times.

Issue #3153, opened 2018-09-11
Generalized Modularity for Community Detection


xiaoylu:
Is there any way we can integrate the generalized modularity into the current implementation of 
> networkx.algorithms.community.quality
> networkx.algorithms.community.modularity_max

The definition of the generalized modularity is just a simple extension of modularity with one extra \gamma parameter. And it can easily improve the resolution limit of modularity-based community detection.

See Newman's paper in 2016.

[Equivalence between modularity optimization and maximum likelihood methods for community detection](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.94.052315)

elplatt:
Probably easy for the quality module. In the modularity_max module, the change in modularity is calculated at each step, rather than the modularity itself. The formulas for the modularity updates would need to be determined for both directed and undirected networks.

xiaoylu:
Update multiple places in the modularity_max module to change the modularity definition.
Appreciate that if someone can review #3260

willpeppo:
hi I'd like to start contributing to networkx and it looks like this is still open.

somyabhargav:
Hello! I would like to contribute to networkx and wanted to know if there is something still unresolved here that I can help with. Thanks.

dschult:
Closed by #3260

PR #3238, opened 2018-11-13
added resolution parameter to modularity

gveltri:
Added resolution parameter, gamma, to modularity in algorithms.community.quality.modularity.

See [Statistical Mechanics of Community Detection](https://arxiv.org/abs/cond-mat/0603718)

#3153

### DIFF
--- a/networkx/algorithms/community/quality.py
+++ b/networkx/algorithms/community/quality.py
@@ -247,19 +247,20 @@ def coverage(G, partition):
     return intra_edges / total_edges
 
 
-def modularity(G, communities, weight='weight'):
+def modularity(G, communities, weight='weight', resolution=1):
     r"""Returns the modularity of the given partition of the graph.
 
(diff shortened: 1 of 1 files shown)


dschult:
Thanks -- 
Would it be possible to add a simple test that makes sure we don't mess this up later?
Even just computing one very simple case with resolution=2 and checking values would be helpful.

dschult:
This addresses one major part of #3153. It implements the generalized case for the ```modularity`` function. 

Does it make sense to generalize other functions in this module? 
What about in the ```modularity_max.py``` module? 
If you'd prefer those changes to be in another PR thats fine.

gveltri:
It looks like Xiaoylu has added it along with the modularity maximization change. I could add a test to this if we want to split out the modularity resolution change from xiaoylu's pull request #3153.

dschult:
Let's work on getting #3260 merged

PR #3243, opened 2018-11-13
Place pydot version requirement in setup.py rather than enforcing it at runtime.

benjaminp:


dschult:
pydot is not required to run NetworkX. It is an optional package, and one of the ways to interact with GraphViz.  But we don't want to check it on installation since users don't have to have it and sometimes don't...

benjaminp:
This doesn't make `pydot` required. It means that if you do `pip install 'networkx[pydot]'`, pip will make sure the right pydot version is installed.

dschult:
Ahh.... sorry... i misunderstood...   Thanks...
Can we leave the runtime check in also for people who aren't installing that way?

benjaminp:
Part of the purpose of the PR is to remove the usage of `pkg_resources`. networkx doesn't declare a dependency on setuptools, so using `pkg_resources` is technically illegal. Furthermore, `pkg_resources` is an extremely heavy-weight import. It doesn't a lot of work at import-time. Depending on `pkg_resources` just to compare some version numbers is painful.

I suppose I could reimplement the version check some other way, but I wonder how necessary it is. networkx has many optional dependencies. Why does only `pydot` receive this special version-checking treatment?

dschult:
The history with pydot is that it sat for a long time unable to work with Python3. A few different updates brought it into Python3 compatibility. We moved support to one of those versions of pydot for a while. Eventually @johnyf took over from @erocarrera and the others worked with him to bring it into line with modern packages. We added a hard (checked) dependency when we changed back to support for the main package.  

The history does not adequately answer your question however: why does pydot get special attention to version number when the other packages don't.

I don't have a good answer for that. It seems reasonable to take it out.

dschult:
Does anyone have objections to removing the version checking code inside NetworkX for pydot?
We don't check versions of dependency packages other than for pydot.

benjaminp:
What's the rationale for not expressing the version requirement in `setup.py`?

jarrodmillman:
The current official minimum version is in a requirements file.  We use that for testing, so we need to keep it there.  I would like to have just one official location for minimum versions, but currently there is more than one location.  And I wasn't aware of the one in ``nx_pydot.py``, which is why I didn't update it when I last updated the minimum required version of pydot.  (Thanks for catching it.)

Eventually, we should either modify ``setup.py`` to get the minimum versions from the requirements file or create a script that edits ``setup.py`` based on the requirements files (then we would run the script as part of the release process).  If you want to create another PR to do that, I would be happy to review it.

benjaminp:
I've now updated the PR to only delete the version check in `nx_pydot.py`.

I see what you mean about duplicate version constraint declarations. One way to unify the requirement declarations would be to put everything in `setup.py` and use `setup.py nose` for testing.

chebee7i:
IIRC, one of the preferred variants of pydot wasn't even an officially released package at the time, due to a long-standing unsolved status on the direction of the official pydot package and its successor(s). This is why setup.py was not an option. I suspect the PRs or mailing list can shed light on this if you want to know more about the history. Though my memory could be incorrect in this instance, usually there are semi-rational reasons why developers make particular decisions.

Issue #3237, opened 2018-11-12
write_graphml doesn't like custom object 


gabri94:
I'm using the attributes of my graph to store custom object that i need for my software. 
I'd like to be able to export it in graphml without removing this object, but i understand that graphml cannot compose the xml for a custom object. 
However I expected to be able, by defining the __str__ or __repr__ method, to allow the graphml method to use its string representation in the graphml. 

By looking at the code I don't think this has been implemente, but would it be possible?

dschult:
How hard would it be to 
1) copy the graph
2) iterate over the edges and replace each custom object with its string representation
3) write the result as graphml
That might be the most straightforward way.

gabri94:
Yes, this was my idea. That would also work for the lists
Are u going to implement it soon or I can give it a try?

dschult:
Please go ahead and write some code! :)  If you have questions, ask... and its fine to submit a PR that is incomplete for feedback.  It is easy to update and add commits to the PR.  There might even be a way to attach the PR to this issue, but the easy way is to make a PR that simply says "Fixes #3237" somewhere in the description.    Thanks!!

gabri94:
can you give it a look? i used str because it was already there, and i just had to change 1 line of code.

PR #3248, opened 2018-11-17
custom object composition in graphml

gabri94:
Hi, i realized that is not possible to write graph with custom object as attribute.
I was replacing them by their str representation just before saving, but I think this could be a cleaner and more general solution for the problem.
I already had open an issue #3237.

I'm not sure whether should be use str or repr. As a first try i used str, because it was much cleaner and linear than redefining make_str()
But I'm hear for your feedback.

### DIFF
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -462,9 +462,12 @@ class GraphMLWriter(GraphML):
         type in the keys table.
         """
         if element_type not in self.xml_type:
-            msg = 'GraphML writer does not support %s as data values.'
-            raise nx.NetworkXError(msg % element_type)
-        keyid = self.get_key(name, self.xml_type[element_type], scope, default)
+            # If element_type is unknown we fall back to a string representation
+            keyid = self.get_key(name, "string", scope, default)
+            # msg = 'GraphML writer does not support %s as data values.'
+            # raise nx.NetworkXError(msg % element_type)
+        else:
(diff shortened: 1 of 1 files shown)


dschult:
It looks like you have change the ```keyid``` to be "string" for unknown data types, but don't you have to change the code that generates the string?  It's been a long time since I looked at this code -- and I am traveling so am short on time to investigate further right now.

Please also include a test in ```tests/test_graphml.py``` that writes a graph that has a custom object as a data attribute. Then read it back in and check that the result gives what you expect.  The doc_string for the function will also need to be updated.

Thanks!

dschult:
I actually don't know which would cause fewer surprises -- str or repr?
What would be best in your use-case?

gabri94:
In my use case it would be the same since I made my custom object myself. Idk in a more general way what should be the difference

dschult:
There are [some discussions ](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr) about the difference between repr and str.  My understanding is that typically: ```str``` is a readable description of the object. ```repr``` is some text that unambiguously describes the object -- best case: repr  shows python code that could regenerate the object.

In many cases they are the same. One that isn't is: 
str("3")==str(3)=='3'  but 
repr("3")=="'3'" has an extra set of quotes, so it is the code that would create "3".

In the end it probably doesn't matter much, but if the goal is to be able to recreate the object perhaps ```repr``` would be better. If the goal is to make it readable, the ```str``` is better.

PR #3255, opened 2018-11-25
Remove ticks and tick labels from draw_networkx(), close #3254.

### LINKED ISSUE #3254: Remove axis ticks and tick labels from default network visualization
The axis ticks are districting and in most cases unnecessary.

elplatt:
* Removes left and bottom ticks and tick labels for draw_networkx() and all related draw_networkx_*() functions.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -412,6 +412,13 @@ def draw_networkx_nodes(G, pos,
                                  linewidths=linewidths,
                                  edgecolors=edgecolors,
                                  label=label)
+    plt.tick_params(
+        axis='both',
+        which='both',
+        bottom=False,
+        left=False,
+        labelbottom=False,
+        labelleft=False)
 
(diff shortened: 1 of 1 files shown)


dschult:
I don't like the idea of hard coding these options into our code. The user can do this (along with many other matplotlib options) after the plot is created. If we do it in our code, then we have more to maintain and we haven't added functionality per se.  We could provide parameters for all the different ticks options, but that sounds like it could lead to requests for options about many plot features.  My leaning is toward a piece-meal approach where each function does a minimal thing and the user puts the functions together to create what they want.  Does this make sense or is there more to the story?

elplatt:
The motivation is to make the defaults friendly to beginners and "just
work". The ticks can easily be added back, but are going to be unwanted in
most cases and possibly confusing/distracting. Being able to make a usable
visualisation with a single line of code really appeals to
students/newcomers.

dschult:
Right...  That makes sense to me... I'll take a closer look and probably merge soon...

PR #3261, opened 2018-11-30
Fix typo in docstring

todoaskit:
Fix typo in docstring


ericmjl:
@dschult also looks small enough that I'll just merge it.

Issue #3266, opened 2018-12-05
Add Floyd-Warshall for average_shortest_path_length


wuhaochen:
I just find that the function `average_shortest_path_length` by default is using the Dijkstra's algorithm to computer all pairs of shortest path, which is inefficient for dense graph. It doesn't support using Floyd-Warshall algorithm. I'm no sure if we should make this as the default as well.

I can send a PR if you think this change is reasonable.

dschult:
Right now the ```average_shortest_path_length``` function allows methods dijkstra, bellman-ford or unweighted. It would make sense to add the option of using Floyd-Warshall. 

The default method for most shortest path computations is dijkstra.  It should stay that way -- that is what "most people" will want. If you have a dense network then you should change from the default.

Thanks!

wuhaochen:
Sounds good. I just noticed one more thing there.
Currently `average_shortest_path_length` only require a directed network to be weakly connected and count the length of unreachable pairs as 0. I don't think this is a correct behavior. For example, in a network with two nodes a and b and only an edge a, b, the function will return 0.5.

I think the right thing to do here is to require it to be strongly connected. Should I open a separate PR for this?

dschult:
Yup -- two PRs are appropriate for these two items.

I think the definition of average shortest path length should depend on a literature search if possible. The application to directed graphs ought to do what is standard in the literature -- or provide both possibilities if both are popular in the literature.

(I'm actually not sure why an average of 0.5 for the graph a->b is bad.... isn't that what you would want?)

wuhaochen:
One of the issue is that average for the graph a<->b is 1. Adding edges reduced connectivity should be counter-intuitive.

dschult:
Ahh..  you are pointing out that the other path (from b to a) has infinite shortest path length so the average should be (1+\infty)/2 or something like that... that's why it shouldn't be allowed.  Am I understanding?

it's hard to find examples where average shortest path is used... even for undirected graphs. Are there any papers/books with this defined for directed graphs?

dschult:
In the case of strongly connected graphs, the current code gives the same result. So restricting it to only work for that type of directed graph means we are removing something without adding anything except possibly protecting a user from misunderstanding what is happening. I would prefer to do this with the docs rather than with code.  The same could be said for disconnected undirected graphs. But there we have an obvious alternative -- take average of each component. 

What do we suggest to users who want the average for a weakly connected graph?

wuhaochen:
I think this is a good point.

[...]

PR #3268, opened 2018-12-05
Require strongly connected to compute average shortest path length.

wuhaochen:
As mentioned in #3266


MridulS:
This was fixed in https://github.com/networkx/networkx/pull/6003

PR #3276, opened 2018-12-21
adjust convergence condition in power method for pagerank

BlueWhaleZhou:
The err is computed using l1 norm, which can be equally represented by (N * \delta x[i]) ( \delta x[i] is the difference of the ranking value of the node x[i]).

However, if we have err < N * tol, it makes the condition close to \delta x[i] < tol. The problem here is that if we have a very large graph (more that 10^7 nodes), the initialized value for x[i] is already less than tol (i.e., 1e-6), so the change will be very possibly less than tol, therefore the loop will stop in just few iteration without convergence.

Thank you!


dschult:
Will this make it hard to coverage to tolerance for some cases (?small graphs?)

dschult:
I think this case can be easily covered by reducing ```tol``` when you have a problem (say due to a large graph).

Issue #3230, opened 2018-11-07
Deprecation warnings in import, DiGraph.reverse


wkschwartz:
```python
(venv) $ python -Wa
Python 3.7.0 (default, Jun 29 2018, 20:13:13) 
[Clang 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import networkx as nx
.../venv/lib/python3.7/site-packages/networkx/classes/graph.py:23: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping
.../venv/lib/python3.7/site-packages/networkx/classes/reportviews.py:95: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping, Set, Iterable
.../venv/lib/python3.7/site-packages/networkx/classes/reportviews.py:95: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping, Set, Iterable
.../Python.framework/Versions/3.7/lib/python3.7/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
.../Python.framework/Versions/3.7/lib/python3.7/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
.../Python.framework/Versions/3.7/lib/python3.7/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
>>> nx.__version__
'2.2'
>>> g = nx.DiGraph()
>>> nx.reverse(g)
.../venv/lib/python3.7/site-packages/networkx/classes/digraph.py:1180: DeprecationWarning: G.fresh_copy is deprecated. Use G.__class__ instead
  H = self.fresh_copy()
<networkx.classes.digraph.DiGraph object at 0x112f52c88>
```

These deprecated lines of code are still in master as of the time I'm writing this:

https://github.com/networkx/networkx/blob/636afa0777b1c11922beb4b1cc91c2f217421066/networkx/classes/graph.py#L23

[...]

PR #3278, opened 2018-12-22
replace deprecated fresh_copy with __class__

AntoinePrv:
Related to #3230

All the mentioned depredations were already fixed. I did find two more (if they're relevant)


dschult:
The test failure is not due to this change. I think the numpy version on Travis changed so it prints differently. We'll fix that in another PR.

Could you also make sure all the imports of ```collections``` are correctly changed?  (You probably did already, but I just wanted to be sure...  #3171 did a bunch of that already.)

PR #3279, opened 2018-12-23
fix frozen function

eyalb1:
frozen should have kwargs otherwise when calling a frozen graph method with a keyword you get
a TypeError : frozen() got an unexpected keyword argument ..


dschult:
Thanks!

PR #3280, opened 2018-12-24
Handle singleton input in add_star/cycle

### LINKED ISSUE #3277: Add node when singleton input to add_star/cycle
This is basically a continuation of issue #2726. (The add_star/path/cycle functions need to add a node when the input is a singleton.)

Need to expand #2759 (for add_path) to add_star and add_cycle.

hridaydutta123:
Fixes #3277

### DIFF
--- a/networkx/classes/function.py
+++ b/networkx/classes/function.py
@@ -137,7 +137,7 @@ def is_directed(G):
     return G.is_directed()
 
 
-def frozen(*args, **kwargs):
+def frozen(*args):
     """Dummy method for raising errors when trying to modify frozen graphs"""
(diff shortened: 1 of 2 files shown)


dschult:
The ```frozen``` changes are new today -- so you'll need to rebase to the latest version of master: I think it should just be

1) update a master branch; 
2) change to your PR branch; 
3) ```git rebase master``` .   # but things may come up here... I usually use ```git rebase -i master```

Also, it looks like your changes for ```add_star``` aren't passing tests because it is adding a path instead of a star.

dschult:
See #3285

PR #3282, opened 2018-12-28
Lessen cyclomatic complexity in directed_laplacian_matrix

cthoyt:
This commit just flips the polarity of the first so this can be written more succinctly with an elif


dschult:
I'd rather not get into rearranging code for succinctness. Readability is a better measure of code-quality (and sometimes agrees with succinctness). In this case, I'd rather not make the change -- if only to avoid a precedent.  Thanks for the suggestion!

cthoyt:
@dschult fair enough. Thanks for taking the time to look into this.

PR #3284, opened 2018-12-31
Fix several DeprecationWarning: invalid escape sequence

BoboTiG:
Hello,

This is a little patch to fix warnings such as:
```python
networkx/networkx/algorithms/cluster.py:537: DeprecationWarning: invalid escape sequence \m
```

In one place, I had to shorten the output of a function to fix
```python
ValueError: line 31 of the docstring for networkx.algorithms.cluster.generalized_degree has inconsistent leading whitespace: '3: Counter({3: 4}), 4: Counter({3: 4})}'
```
Let me know if you are not OK with that :)

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>

### DIFF
--- a/networkx/algorithms/approximation/clique.py
+++ b/networkx/algorithms/approximation/clique.py
@@ -68,7 +68,7 @@ def max_clique(G):
 
 
 def clique_removal(G):
-    """ Repeatedly remove cliques from the graph.
+    r""" Repeatedly remove cliques from the graph.
 
     Results in a $O(|V|/(\log |V|)^2)$ approximation of maximum clique
     and independent set. Returns the largest independent set found, along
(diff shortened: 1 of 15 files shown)


Moradnejad:
Can you please explain what r stands for and what does it do? @BoboTiG

dschult:
@Moradnejad see [this stackoverflow question](https://stackoverflow.com/questions/2081640/what-exactly-do-u-and-r-string-flags-do-and-what-are-raw-string-literals) about ```r" "```

@BoboTiG   Thanks for this!  The shortening of output means the user can't see what the full output should be. I'm probably OK with that, but wonder what options there might be for a workaround that does show the full output.

BoboTiG:
I did not find something good because of the trailing backslash. A solution would be to let the entire output on the same line, but IIR it will break some docstring lint. Will try and see if we can dsable thel int of that docstring.

BoboTiG:
@dschult seems OK now :)

PR #3288, opened 2019-01-03
Add two generators for Harary graph

ws4u:
This pull request implements two generators for the Harary graph, which achieves the maximum node connectivity among all graphs of the same number of nodes and number of edges. The way of constructing Harary graph is described by the famous mathematician Frank Harary in his 1962 work "The Maximum Connectivity of a Graph." 

The two generators use different parameters to generate Harary graph. For more details about them, please refer to the comments for the code. Please review and feedback. Thank you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,4 +47,3 @@ networkx.egg-info/
 
 # PyCharm project file
 .idea
-.vscode/settings.json
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ networkx.egg-info/
 
 # PyCharm project file
 .idea
+.vscode/settings.json
--- a/networkx/generators/harary_graph.py
+++ b/networkx/generators/harary_graph.py
@@ -8,14 +8,18 @@
 #
 """Generators for Harary graphs
 
-This module gives two generators for the Harary graph, which has 
-the maximum node connectivity among all graphs with given number of nodes
-and given number of edges. [1]_.
+This module gives two generators for the Harary graph, which was
+introduced by the famous mathematician Frank Harary in his 1962 work [1]_.
+The first generator gives the Harary graph that maximizes the node
+connectivity with given number of nodes and given number of edges.
+The second generator gives the Harary graph that minimizes
+the number of edges in the graph with given node connectivity and
+number of nodes.
 
 References
 ----------
-.. [1] F. T. Boesch, A. Satyanarayana, and C. L. Suffel,
-"A Survey of Some Network Reliability Analysis and Synthesis Results," Networks, pp. 99-107, 2009.
+.. [1] Harary, F. "The Maximum Connectivity of a Graph."
+       Proc. Nat. Acad. Sci. USA 48, 1142-1146, 1962.
 
 """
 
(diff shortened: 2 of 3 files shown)


ws4u:
The error encountered by Travis CI is not from this PR. Please help fix this error. Thanks.

dschult:
This looks pretty good -- I have only two suggestions:

- can you make sure it passes a pep8 style checker (pycodestyle which used to be pep8; or flake8 or maybe others).  I think linelength may be the only change -- but worth looking.

- Can you change the description from 2 functions that do the same thing to 2 functions which generate Harary graphs, but one maximizes connectivity holding number of nodes and edge fixed while the other minimizes edges holding connectivity and number of nodes fixed.  This might lead to other doc changes, but I'm not sure.

Thanks!

ws4u:
Thanks @dschult for both of your comments. I've fully addressed them. Please have a further check.

dschult:
OK... now some more (hopefully this will be it):

  - remove the .vscode line from .gitignore 

  - Add blank lines after each parameter entry

So that:

    n: integer
       The number of nodes the generated graph is to contain
    m: integer
       The number of edges the generated graph is to contain
   
Becomes:

    n: integer
       The number of nodes the generated graph is to contain
   
    m: integer
       The number of edges the generated graph is to contain
    

  - Make sure each function's docstring starts with a single line (<80 chars)
  and then a blank line.  You may have to move the detailed description of the 
  function to the third line -- and replace it with a first line that is more compact.
  
For example:

    Returns the Harary graph $H_{n,m}$ that maximizes the node
    connectivity with $n$ nodes and $m$ edges.
    This maximum node connectivity is known to be $[2m/n]$. [1]_

Becomes:

    Returns the Harary graph with fixed number of nodes and edges.

    The Harary graph $H_{n,m}$ is the graph that maximizes node connectivity
    while keeping fixed the numbers of $n$ nodes and $m$ edges.

    This maximum node connectivity is known to be $[2m/n]$. [1]_

ws4u:
Thanks @dschult . I'll revise based on the three comments above. I have a question on how to express "the floor of $2m/n$" and "the ceiling of $kn/2$". What are the best latex codes for them in the docstrings?

dschult:
I know of two latex codes for fractions like that and then the way you have it, which is actually pretty good since it doesn't make the line any taller:

    ${2m \over n}$ and $\frac{2m}{n}$ and $2m/n$

ws4u:
You misunderstood me. Since $2m/n$ and $kn/2$ may not be integers, I want to get the floor of $2m/n$ and the ceiling of $kn/2$ respectively. Are there easy latex codes to express them in docstring? Or I just say "the floor of $2m/n$" and "the ceiling of $kn/2$" without using math symbols? Thanks.

dschult:
I did misunderstand... sorry about that.
In LaTeX there are packages that provide special symbols, e.g. 
$\left\lceil 2m/n \right\rceil$ and $\left\lfloor 2m/n \right\rfloor$
But that makes the docstring hard to read in text/code form. 

I would suggest using the more 'code-like' form:
ceil($2m/n$) and floor($2m/n). 
That is what we have used elsewhere e.g. ```edge_augmentation```.

ws4u:
Thanks @dschult for your suggestions. I've used them in the docstrings. Any further feedback is welcome.

dschult:
Thanks very much for this!

ws4u:
@dschult I see these two generators for Harary graph have been included in the NetworkX 2.3 release, but they cannot be found in the Reference. Do I need to do something to fix this? Thanks.

dschult:
Yes.... I apologize for not checking this. Each module gets automatically included in the docs by Sphinx  if you create a corresponding file/entry in the /doc/reference/ directory.  In this case you should add to /doc/reference/generators.rst

The next question is which section of generators.rst to put the new functions. You'll need to add a new section that will cover your new module. Follow the syntax for other modules in that file. You can test locally by building the docs locally.  

Thanks for noticing and asking... :)

ws4u:
Thanks @dschult . I have created the PR https://github.com/networkx/networkx/pull/3407 for this.

Issue #2404, opened 2017-03-27
Directed Laplacian


antoniopugliese:
Question as in [this pull request](https://github.com/networkx/networkx/pull/741#issuecomment-289179061)  ( #741 ).

The `directed_laplacian_matrix` function is actually a `normalized_directed_laplacian_matrix` function.
This function uses the normalized formula
L = I - (\Phi^{1/2} P \Phi^{-1/2} + \Phi^{-1/2} P^T \Phi^{1/2} ) / 2

For a correct `directed_laplacian_matrix` the combinatorial formula should be used
L = \Phi - (\Phi P + P^T \Phi) / 2
This new function can be named `directed_combinatorial_laplacian_matrix` as suggested by @aweinstein.

The two formulas are obtained from the same paper:
Fan Chung (2005). Laplacians and the Cheeger inequality for directed graphs. Annals of Combinatorics, 9(1), 2005 ([available here](http://www.math.ucsd.edu/~fan/wp/dichee.pdf)).

antoniopugliese:
Possible implementation of the function on [my profile](https://github.com/antoniopugliese/networkx/blob/master/networkx/linalg/laplacianmatrix.py).

I haven't tested it at all. Please do not implement as is before testing it.

PR #3297, opened 2019-01-11
Directed combinatorial laplacian as in #2404

j-petit:
Hi,

I gave the integration of the directed combinatorial laplacian a try as suggested in #2404 . This includes some tests. As the calculation oft the transition probability matrix is the same as in the normalized directed laplacian, I refactored it into a separate function. Open to any feedback!


dschult:
Thanks!!

PR #3307, opened 2019-01-22
WIP: add connectionstyle parameter

guidoeco:
This is a cosmetic change to the `draw_networkx_edges()` to pass the `connectionstyle` parameter to `matplotlib.patches.FancyArrowPatch` to create curved arc of rounding radius `rad`. For example, connectionstyle='arc3,rad=0.2'.

The default is `None` which is ignored by `matplotlib.patches.FancyArrowPatch`

See :py:class: `matplotlib.patches.ConnectionStyle` and :py:class: `matplotlib.patches.FancyArrowPatch` for more info.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -435,6 +435,7 @@ def draw_networkx_edges(G, pos,
                         node_size=300,
                         nodelist=None,
                         node_shape="o",
+                        connectionstyle=None,
                         **kwds):
     """Draw the edges of the graph G.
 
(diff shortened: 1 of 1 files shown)


dschult:
Does this allow two directed edges in opposite directions between a pair of nodes to be seen as two edges? What other use cases do you see for this? Do the styles have to be the same for all edges, or is there a way to make them different even though the arrows are in the same patch? (I guess you can always draw some edges with one style and then draw others with a different style using a second function call...)

guidoeco:
Apologies in the delay for replying. 
I think the answer is yes to both your questions. The use-case here is to pass the `arc3,rad=0.1` style parameter to the `matplotlib.patches.FancyArrowPatch` call. This draws edges as arcs on a per-edge basis as you can see in the example below. This is bi-directional edge pairs are drawn with arcs where single edges aren't. 
w014rb

dschult:
Does the picture you show use a single call to ```draw_networkx_edges```? Or is it many calls?

That is, could ```connectionstyle``` (which should be ```connection_style```) be a list of strings?

guidoeco:
The arrows in the examples are added with two calls as follows:
`nx.draw_networkx_edges(MDG, p, edgelist=[('N047', 'N046'), ('N047', 'N045')], width=1, node_size=nodesize, arrowsize=24, arrowstyle="simple")`
and
`nx.draw_networkx_edges(MDG, p, edgelist=[('N047', 'N048'), ('N048', 'N047')], width=1, node_size=nodesize, arrowsize=24, arrowstyle="simple", connection_style="arc3,rad=0.1")`

The suggested implementation is based on passing on an additional optional style `connectionstyle` parameter in the `matplotlib.patches.FancyArrowPatch` API

guidoeco:
Cool. Thanks

PR #3312, opened 2019-02-01
Implement ISMAGS subgraph isomorphism algorithm

### LINKED ISSUE #3216: ISMAGS subgraph isomorphism algorithm
Hi all,

First off, thanks for the library; I make heavy use of networkx, and it always seems to have the tools I need already implemented.

Recently however, this was not the case ( :'( ). I work with molecules which to a layman can be described as highly symmetric graphs, and I needed to find subgraph isomorphisms. When using the available VF2 algorithm, this caused a combinatorial explosion of equivalent isomorphisms wrecking performance.

As a result I had to implement the ISMAGS (Index-based Subgraph Matching Algorithm with General Symmetries) [1]. This algorithm find all subgraph isomorphisms that are not symmetrically equivalent. I'm still in the progress of polishing and cleaning, but do you think this would make a nice addition to networkx? If so, I'd be more than happy to open a PR.

[1] https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0097896

pckroon:
Hi all!

As discussed some time ago in #3216 I made a Python implementation of the ISMAGS algorithm [1]. I'm not connected to the original developers/authors of the paper in any way. The algorithm takes symmetries in the subgraph into account and only produces symmetry unique isomorphisms. I also needed to find all largest common induced subgraphs exploiting symmetry in the same way, so that's added as well.
I realize it's a lot of code to review, so take your time, and please comment on things that need to change :). This code has originally been made for (and merged in) https://github.com/marrink-lab/vermouth-martinize/ but none of us really knows graph theory so I may have walked into some pitfalls...

Fixes #3216  
[1] https://doi.org/10.1371/journal.pone.0097896


dschult:
I'm optimistically putting this under release 2.3. Please don't be disappointed if it ends up being in 2.4. But I would like it to be merged and its just a matter of time and how fast the release happens etc.

Thanks!

pckroon:
Great :)
No worries either way, I know it's a lot of code to go through, so no rush.
As an aside, the doctests aren't working (`NameError: name 'nx' is not defined`). Any advice on how to solve that?
I also have a small change to make the algorithm a little faster, I'll push that together with the fix for the doctests.

dschult:
doctest errors:   Looks like you'll need to add ```import networkx as nx``` perhaps after ```import itertools```. That allows the doctest to pick up ```nx```. There maybe a fancier way, but that works. 

Then you need to add ```nx.``` before ```isomorphism``` so it can find that name, and ```ISMAGS``` before ```_get_permutations_by_length``` so it can find that name.

The formatting of output is a little more tricky. You are supposed to have the test output formatted exactly as the output of the doctest.  You've got your results nicely with line-breaks, but the doctest output is all on one line.  And, if you put it all on one line in your expected answer, then you aren't following PEP8 guidelines for readability.  Also, the order of the list and the order of the elements in the dict that is inside the list may differ from run to run and so are hard to pass the doctest system.

One option is to have the test create a variable to hold the answer -- then check that the answer is what its supposed to be without outputting the answer.  If order is giving trouble, replace ```list``` with ```set```.  For example:

    >>> largest_subgraph = set(ismags2.largest_common_subgraph())
    >>> answer = {{1: 0, 0: 1, 2: 2}, {1: 0, 0: 1, 3: 2}, {2: 0, 0: 1, 1: 2},
    ... {2: 0, 0: 1, 3: 2},{3: 0, 0: 1, 1: 2}, {3: 0, 0: 1, 2: 2}}
    >>> largest_subgraph == answer
    True

Your mileage may vary. :}

pckroon:
Cheers, I'll have a look.

pckroon:
That should do it. To be honest I'm rather annoyed that the order of dictionary items (!) matters for doctest, but what can you do.

PR #3314, opened 2019-02-05
WIP: Implement Incremental Topological Sort (Issue 1457)

aweltsch:
I started to add an algorithm that can be used to incrementally update the topological sorting of a directed graph. I'm not 100% content with the API design. It would be great to get some feedback.

I am particularly concerned with the placement of the corresponding data structure (`class IncrementalTopologicalSort`). Is the `dag.py` file the right place?

Also I wondering if I should derive the class from the `nx.DiGraph` or something like that.

Any additional feedback is greatly appreciated!
Thank you!

### DIFF
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -415,6 +415,128 @@ def all_topological_sorts(G):
             break
 
 
+class IncrementalTopologicalSort():
+    """
+    FIXME
+    """
+    def __init__(self):
+        """Initialize a datastructure that enables incremental topological sorting.
+        """
+        self.graph = nx.DiGraph() # FIXME enable MultiDiGraph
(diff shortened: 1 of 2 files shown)


dschult:
Can you provide a short description of the goal of this feature? What are you trying to achieve relative to the existing algorithms for topological sort (that is -- why create an incremental system)?
The next question is whether you are implementing the published algorithm's data structures as well as program flow. I haven't read it, but your question about whether it should subclass ```nx.DiGraph``` might be answered by whether the paper uses a graph-like data structure.

Thanks!

MridulS:
Thanks for your work @aweltsch!

This hasn't been touched in a while. I will go ahead and close this for now. Thanks again for your contributions!

PR #3320, opened 2019-02-13
[WIP] Steiner multigraph

suricactus:
I am proposing a change in `networkx.algorithms.approximation.steinertree` to make it work with multigraphs. This affects Dijkstra algorithms results too. The change is not complete yet, but I would be happy to get some feedback from the core maintainers on the current progress as I think the main task is done, what's left are some fixes here and there.

- [ ] update documentation of the updated functions
- [ ] update tests of the updated functions
- [x] add more tests for Steiner algorithm
- [x] fix for unweighted multigraph

However, there is a breaking change, as Dijkstra algorithm funtions no longer returns 2 items in a tuple, but 3 (the new one for the tuples, if any, see for example `nx.single_source_dijkstra`). I am planning to use it for my project, so I need to finish it for sure.

### DIFF
--- a/networkx/algorithms/approximation/steinertree.py
+++ b/networkx/algorithms/approximation/steinertree.py
@@ -50,7 +50,7 @@ def metric_closure(G, weight='weight'):
 
 
 @not_implemented_for('directed')
-def steiner_tree(G, terminal_nodes, weight='weight'):
+def steiner_tree(G, terminal_nodes, weight='weight', metric_closure=None):
     """ Return an approximation to the minimum Steiner tree of a graph.
 
     Parameters
@@ -60,6 +60,12 @@ def steiner_tree(G, terminal_nodes, weight='weight'):
     terminal_nodes : list
          A list of terminal nodes for which minimum steiner tree is
          to be found.
+
+    weight :
+         The weight to be used for calculating the shortest path
+
+    metric_closure:
+         A precalculated metric_closure of the graph
 
     Returns
     -------
@@ -81,7 +87,7 @@ def steiner_tree(G, terminal_nodes, weight='weight'):
     """
     # M is the subgraph of the metric closure induced by the terminal nodes of
     # G.
-    M = metric_closure(G, weight=weight)
+    M = metric_closure if metric_closure else metric_closure(G, weight=weight)
     # Use the 'distance' attribute of each edge provided by the metric closure
     # graph.
     H = M.subgraph(terminal_nodes)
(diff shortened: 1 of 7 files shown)


dschult:
We have had proposals before to allow the shortest path algorithms work with multigraphs. My impression is that it is better to use the graph algorithms and either morph the multigraph to a graph before calling the routine, OR use the "weight" functionality to create a function that looks at all edges between the two nodes and selects the shortest weight. The former requires altering the whole graph once before starting the search.  The latter requires recomputing which multiedge is best each time the two nodes are examined.

I'm not sure of the details of your methodology, but can you give a big picture description of why you need to do this and how your method might improve over these two alternatives?  Also, it would help if there is a reference to a published algorithm for this method of handling multigraphs.

suricactus:
I saw there was a previous proposal for this change, but without any PR (at least I didn't find that). From my perspective, it means it's a useful feature if it's requested several times.

In my case, I have a bunch of geographical points in the space (nodes), connected with one or more lines (edges). Each of the lines may have a different weight. 

From you two workarounds, I have some arguments against. If I turn the multigraph to graph with least cost edges, then I need to copy the graph several times if I want to use it with several different properties for weights. On the second suggestion, if I correctly understand you, it needs a lot of redundant calculations each time I get the shortest path to see which edge was used.

Basically, what am I doing here is using the exact same logic that is already in place for graphs, but also returning the keys of the used edges, not only the path.

dschult:
I was envisioning one copy of the multigraph to a graph with computation of the "best" edge between each pair of connected nodes. That "best" edge weight could be computed and stored in different attributes -- one for each way you'd like to find the best from the multiedges.  So, you don't need to copy the graph many times. You just need to create an edge attribute for each method you'd like to compute the "best" multiedge.  

The second method (using a weight function to scan through the multiedges and find the best one "on-the-fly") has redundant computations compared to changing to a Graph.  But it doesn't have redundant calculations compared to handling the multigraph. With a MultiGraph you must examine each multiedge between two nodes. With a weight function approach you can also examine each multiedge between the two nodes. Those methods are comparable in terms of computations. The redundancy is when comparing to converting to a Graph -- where finding the "best" edge is done once (for each way you want it computed) for each pair of nodes.

You identify the real problem directly in your final paragraph.  We are treating edges differently for MultiGraph and for Graph.  The distinction between 2-tuples and 3-tuples makes it hard to construct algorithms that work for either. There is a lot of checking in the code. This is a fairly deep concern with many implications for the code. We've thought about ways to handle edges more generally, but it almost always impacts performance for the simple Graph case.

I am fine with making a version of steiner that works with multigraphs so long as it doesn't impact the performance for highly used functions like Djikstra. I encourage you to find ways to avoid searching every multiedge in a djikstra manner and instead process each node-pair that has an edge. They amount to the same work -- but the latter allows you to use the djikstra built for the simple case.

dschult:
I'm going to close this PR based on a decision not to encourage using shortest path algorithms on multigraphs, but instead convert them to simple graphs with appropriate edge weights.

PR #3323, opened 2019-02-21
Fixed error message for single_target_shortest_path_length

scchow:
Discovered a small copy-paste error in the error message for `single_target_shortest_path_length`.

Currently, if the node passed into `single_target_shortest_path_length` is not defined, it will throw an error that `source` is not defined. Replaced `source` with `target` since that is the variable defined for this function.


dschult:
Thank you!

PR #3335, opened 2019-03-06
Adjust function "is_list_of_ints". (fix #3322)

### LINKED ISSUE #3322: is_graphical treats numpy int as not integers
There is a minor issue with using `is_graphical` function with numpy array.
When passing a numpy array with integers as the argument, `is_graphical` automatically returns `False` because numpy `int` type is not an instance of python `int` type.

Also there is no warning regarding the data type of input so I was very confused for a while when I had a clearly graphical degree sequence passed to the `is_graphical` and was returned `False`.

I would suggest considering numpy int as a valid integer type, otherwise I have to recreate a list of python integers from the array of numpy integers.

Also I think it is reasonable to raise an exception or print a warning when the data type of the input degree sequence is not a valid integer, because this case is fundamentally different from not being able to pass HH or EG tests. 

```
>>> import networkx as nx
>>> import numpy as np
>>> a = np.array([2,2,2])
>>> b = [2,2,2]
>>> nx.is_graphical(a)
False
>>> nx.is_graphical(b)
True
```

KmolYuan:
Following script allows numpy array or any iterable object with integers to pass the test:

```python
>> from numbers import Integral
>> import numpy as np
>> isinstance(np.int32(0), Integral)
True
>> from typing import Iterable, Iterator, ItemsView
>> isinstance(np.array([0, 1, 2]), Iterable)
True
>> isinstance(np.array([0, 1, 2]), (Iterator, ItemsView))
False
```

Allowed types:

+ List[int], Tuple[int, ...]
+ range
+ Dict[int, Any]
+ ndarray[dtype=int]

### DIFF
--- a/networkx/utils/misc.py
+++ b/networkx/utils/misc.py
@@ -18,13 +18,13 @@ True
 #    Pieter Swart <swart@lanl.gov>
 #    All rights reserved.
 #    BSD license.
-from collections import defaultdict
-from collections import deque
+from collections import defaultdict, deque
+from typing import Iterable
+from numbers import Integral
 import warnings
 import sys
 import uuid
 from itertools import tee, chain
-import networkx as nx
 
 # itertools.accumulate is only available on Python 3.2 or later.
 #
@@ -93,10 +93,10 @@ def flatten(obj, result=None):
 
 def is_list_of_ints(intlist):
     """ Return True if list is a list of ints. """
-    if not isinstance(intlist, list):
+    if not isinstance(intlist, Iterable):
         return False
     for i in intlist:
-        if not isinstance(i, int):
+        if not isinstance(i, Integral):
             return False
     return True
 
--- a/networkx/utils/misc.py
+++ b/networkx/utils/misc.py
@@ -19,7 +19,7 @@ True
 #    All rights reserved.
 #    BSD license.
 from collections import defaultdict, deque
-from typing import Iterable, Iterator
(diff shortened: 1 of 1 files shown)


dschult:
Should this be renamed as ```is_container_of_ints```?  That's a real question... What should it be called if it accepts more than lists?

Also, the function is used in only a couple of places in the code. In at least one, a return value of False triggers an attempt to convert from numpy.int64 to int. With this change perhaps that code can be simplified.

KmolYuan:
How about `Collection[Integral]`?

```python
isinstance(np.array([0, 1, 2]), Collection)  # True
isinstance([0, 1, 2], Collection)  # True
isinstance((0, 1, 2), Collection)  # True
isinstance({0, 1, 2}, Collection)  # True
isinstance(range(0, 1, 2), Collection)  # True

def a():
    yield 0
isinstance(a(), Collection)  # False

isinstance([0, 1, 2], Collection[Integral])  # TypeError: Subscripted generics cannot be used with class and instance checks
```

KmolYuan:
It's hard to check the instance for `numpy.array` without import NumPy.

In traditional way:

```python
hasattr(np.array([0, 1, 2]), '__len__')
hasattr(np.array([0, 1, 2]), '__getitem__')
hasattr(np.array([0, 1, 2]), '__iter__')
```

dschult:
Closed via #3617 
Thanks for your help!

PR #3351, opened 2019-03-19
Create moral graph

julien-klaus:
I have written a function for calculating the moral graph of a given directed graph. 
The moralization is often used for probabilistic learning and specifically when working with graphical models.


julien-klaus:
Hey, thank you for the response. I have changed the location of the files and added moral to the documentation in algorithms/index.rst and algorithms/moral.rst. Hope this is all, pushed it.

dschult:
The core of the function can be replaced by many fewer statements that might be faster:

    HH=G.to_undirected()
    for preds in G.pred.values():
        HH.add_edges_from(itertools.combinations(preds, r=2))

Note that this handles the case of len(preds)<=1.  It also avoids creating a list of preds for each node.
Coding style varies quite a bit by coder, so see if you like that suggestion (and maybe time it).

Also -- can you change the function name from ```get_moral_graph``` to ```moral_graph```?
Again, if you think ```get_``` is better, let's hear your reasons. But our names typically don't have ```get_``` when we are computing/creating that object.

You can consider putting your name in ```/CONTRIBUTORS.rst``` . You don't have to.

julien-klaus:
Hi, I have tested and timed the other approach and it is much faster, so I changed it. 
Thank you for the suggestion.

julien-klaus:
I have no idea what this error is and in the report, I did not see the error coming from my files.

dschult:
The error is not related to your code.... its a version difference between testing environments that has cropped up recently.  This looks good to go.

PR #3355, opened 2019-03-25
Update pyyaml and fix yaml.load(input) deprecation warning

jarrodmillman:
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation


jarrodmillman:
@dschult I would like to require the latest pyyaml 5.1 due to the fact the `load` function could be easily exploited.  It was released recently, but it is easy to install and isn't required.  Does this seem reasonable?

https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

dschult:
Yes, it seems reasonable to increase the version of pyyaml in the extras file.
:)

Issue #1120, opened 2014-04-23
Better shortest paths for Multigraphs...


chebee7i:
Most (all?) of the shortest path algorithms do not give sufficient information for multigraphs.  In those situations, knowing the source and target is not enough---one must also know the key of the edge.  These algorithms touch _everything_ though. So I'm not sure it's wise to change them. Perhaps though, we can offer more general implementations and then have the existing implementations use them to create the outputs?

Also, its often that case that one is interested in the edges of the shortest path. In situations like these, it might even be simpler to get output of the form:  [(u,v,key,weight),...].  Without the weight, you'd have to query the graph again.

Obviously, we can't cater to every possible output format, but I think the issue with multigraphs is serious enough that we should do something. For the various outputs, perhaps we can provide some helper functions for typical cases.

Thoughts on this?

dschult:
This Issue discusses shortest paths for Multigraphs. The question is whether the shortest_path routines should return something like [(u, v, key, weight), ...] for each edge in the path. 

My impression is that shortest_path for most multigraph problems would use the edge with minimum weight between two nodes. Can we write a wrapper for the output that moves along the found shortest path and adds keys and weights after-the-fact? For that matter, wouldn't it be possible to shrink the multigraph to a graph by only keeping minimal edges between each pair of connected nodes?  Maybe I don't understand the use-case for multigraphs and shortest path.

chebee7i:
A pre-processing step that keeps only minimal edges could work: parse the existing graph, construct a new graph with the same edge keys, and then run shortest path on that reduced graph. An after-the-fact solution could work too.

Aesthetically though, both of those solutions seem a bit awkward. The shortest path algorithms are well-defined on multigraphs and the lack of a key seems more like an oversight.

Suppose after the wrapper was created, I then find that I'm interested in *all* shortest paths on a multigraph. With the pre- and post-processing solutions, we now have to write additional code that will make sure to visit all edges between A and B that have the same weight (or we have to extend the wrapper to cover this new use case). This begs the question: in what other ways might we have to extend the wrapper for other algorithms that *should* be returning a complete specification of the edge in the first place.

[...]

PR #3358, opened 2019-03-27
Index edges for multi graph simple paths

MartinPJorge:
Modifies `_all_simple_paths_multigraph()` to return the traversed edges rather than the nodes.
The returned edges contain their associated keys, so user can distinguish across edges connecting same pair of nodes.

If we have following `MultiGraph` with these nodes and edges (with their keys):
```txt
   (1,2,'k0')
    ________
  /          \                     
1 ___________ 2  ___________ 3
   (1,2,'k1')    (2,3,'k0')
```

The modified `_all_simple_paths_multigraph()` now returns:
```python
import networkx as nx

mg = nx.MultiGraph()
mg.add_edge(1,2,key='k0')
mg.add_edge(1,2,key='k1')
mg.add_edge(2,3,key='k0')
paths = nx.all_simple_paths(mg, source=1, target=3)
print(list(paths))

>>> [[(1,2,'k0'),(2,3,'k0')], [(1,2,'k1'),(2,3,'k0')]]
```
rather than
```python
>>> [[1,2,3],[1,2,3]]
```


dschult:
Hmmm...  This PR points out that we use node-paths. That is, we store and report paths as lists of nodes rather than lists of edges. There is a bijection for simple graphs between node-paths and edge-paths. But for MultiGraphs, node-paths do not uniquely specify edge-paths.

I think mixing the two output structures will lead to a convoluted API.  Sometimes a function will return a list of nodes and other times a list of edges. That's going to be hard for users to debug and for us to maintain.

Does anyone have any comments on paths specified as lists for edges?

How hard would it be to make a new function that returns simple paths as edge-paths, that is as lists of edges instead of lists of nodes. For non-multigraphs it could use the current functions and then convert to edge-paths. For multigraphs it would return what you've put into this PR.
What do you think?

MartinPJorge:
> How hard would it be to make a new function that returns simple paths as edge-paths, that is as lists of edges instead of lists of nodes. For non-multigraphs it could use the current functions and then convert to edge-paths. For multigraphs it would return what you've put into this PR.

That might be the best solution. I think it should not take much effort, since for non multi graphs one can use `nx.utils.pairwise` to get a list of edges from a given list of nodes, e.g.,

```python
G = nx.complete_graph(4)
paths = nx.all_simple_paths(G, source=0, target=3)
for path in map(nx.utils.pairwise, paths):
  print(list(path))

[(0, 1), (1, 2), (2, 3)]
[(0, 1), (1, 3)]
[(0, 2), (2, 1), (1, 3)]
[(0, 2), (2, 3)]
[(0, 3)]
```

I can create a function called `all_simple_paths_edges()` that mimics this behaviour for non-multi graphs, and uses the function of the pull request for multi graphs. Something like this:
```python
# Non multi graph
G = nx.complete_graph(4)
paths_edges = nx.all_simple_paths_edges(G, source=0, target=3)
for path_edges in paths_edges:
  print(list(path_edges))

[(0, 1), (1, 2), (2, 3)]
[(0, 1), (1, 3)]
[(0, 2), (2, 1), (1, 3)]
[(0, 2), (2, 3)]
[(0, 3)]

# multi graph
mg = nx.MultiGraph()
mg.add_edge(1,2,key='k0')
mg.add_edge(1,2,key='k1')
mg.add_edge(2,3,key='k0')
paths_edges = nx.all_simple_paths_edges(mg, source=1, target=3)
print(list(paths_edges))

[[(1,2,'k0'),(2,3,'k0')], [(1,2,'k1'),(2,3,'k0')]]
```

Do you agree?

dschult:
Yes -- that looks good to me. Let's think about name (you get to choose): 
paths_edges, path_edges, edge_paths    [with the all_simple in front, but I think the version without all_simple will become a name for the data-structure of specifying a path by making a list of edges.
What should we call a list of edges that specify a path.

:)

MartinPJorge:
Hi,

Last commit includes `all_simple_edge_paths` as discussed.
In the [other pull request](https://github.com/networkx/networkx/pull/3393/commits/6a3766782facc89c86f6c3d84339c1f90516bd8c) I coded the `all_trails` and `all_trails_as_edge_paths`.

Both pull requests, this and the mentioned above, just have Travis error in Python 3.5 in the docstring. The problem is related to the order in which the docstring code examples present the output, vs. the expected order. Honestly I don't think it's an error, because even python 3.5 obtains a correct output.

Do you think it's ready to accept the pull requests https://github.com/networkx/networkx/pull/3393 https://github.com/networkx/networkx/pull/3358?

dschult:
Same comment as on the other PR.
Perhaps a ```sorted()``` function will help the test results.
We need to make the tests pass independent of the order of 
dict reporting (and thus order of Graph reporting).

MartinPJorge:
Done!

Including `sorted()` did the job.
This PR and #3393 pass the checks and have no conflicts.

Now it's ready to be safely merged.

dschult:
Great -- Thanks for this.  Some things to change:

- First line in doc_string should be on one line (<80 chars). Maybe: 

     """Generate lists of edges for all simple paths in G from source to target

- In the first example in the doc_string, use integers instead of strings that look like integers. '1' -> 1
- same example, use ```g.add_egdes_from([(1, 2), (2, 4), (1, 3), (3, 4)])```
- you don't need ```import networkx as nx``` in the examples. It is automagically included.
- remove blank lines in examples created by ```>>>```
- remove spurious lines added to ```test_simple_paths.py```

- add some tests?

jtrakk:
This PR may be helpful in constructing a solution to shortest paths in `MultiDiGraphs`. https://github.com/networkx/networkx/issues/1120

PR #3370, opened 2019-03-30
Enables altering alpha of specific nodes and edges

haiyangdeperci:
Matplotlib bugs related to alpha channels have been resolved, thus I see no point in keeping code pertaining to that. It is currently quite difficult to affect the alpha of particular nodes and edges. Currently, when passing rgba or 8-digit hex values for nodes and edges, alpha is forced to be uniform (1, if not stated otherwise). This pull requests changes this annoying behaviour and lets the user set it up himself or herself.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -288,7 +288,7 @@ def draw_networkx_nodes(G, pos,
                         node_size=300,
                         node_color='#1f78b4',
                         node_shape='o',
-                        alpha=1.0,
+                        alpha=None,
                         cmap=None,
                         vmin=None,
                         vmax=None,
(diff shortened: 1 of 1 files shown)


dschult:
Hmm.. sorry, I'm not sure what the impact of this change is. (It's been a while since I looked at this code.)  What is the current behavior that this changes? What is the new behavior?

It looks like this change makes it so people can't specify a single number and have it apply to the whole collection.

haiyangdeperci:
Currently, you cannot alter the alpha of a specific node or edge. You have to apply it to all of them. If you specify alpha with rgba list / array that you provide to the function via node / edge color parameter, it'll be ignored. The alpha will remain uniform for all the nodes / edges. This change enables the user to set a variable alpha per node / edge.

joseppinilla:
This also applies to nodes. I think this should change and I can make it part of the change in [PR #3395](https://github.com/networkx/networkx/pull/3395).

It would allow three different modes:
Global alpha
List of alphas
*RGBA values in `edge_color `(or `node_color`)

*Currently overridden by the default alpha=1.0, which is unexpected behaviour.

@haiyangdeperci can you take a look at [PR #3395](https://github.com/networkx/networkx/pull/3395)?

dschult:
@haiyangdeperci do the changes that @joseppinilla made for alpha values satisfy your needs?

haiyangdeperci:
@dschult Yes, I believe so. Thanks for implementing that.

PR #3377, opened 2019-04-03
AT-free graph recognition

haakonhr:
Dear all,

This is my first contribution to networkx and to any open-source project so bear with me. I have created a module called `asteroidal.py`which contains a method for recognition of AT-free graphs. I plan to implement more functionality for asteroidal recognition and some other graph recognitions which I have developed, but I try to keep my first pull request small.

### What does the PR contain
`asteroidal.py` implements the method `is_at_free` which takes a graph and checks whether it contains any asteroidal triples or not, i.e. if it is AT-free or not. AT-free graphs is a graph class for which several classical NP-complete problems are solvable polynomial time, independent set and coloring (even list coloring) amongst them.

### Documentation
I have tried to set up sphinx locally to check what my docstrings would look like, but have not been able yet. I'm particularly curious about the math codes.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -126,6 +126,7 @@ is partially historical, and now, mostly arbitrary.
 - Brian Kiefer, Github: `bkief <https://github.com/bkief>`_
 - Julien Klaus
 - Weisheng Si, Github: `ws4u <https://github.com/ws4u>`_
+- Haakon H. Rød, Gitlab: `haakonhr <https://gitlab.com/haakonhr>`_, `<https://haakonhr.gitlab.io>`_ 
 
 Support
 -------
--- a/doc/reference/algorithms/asteroidal.rst
+++ b/doc/reference/algorithms/asteroidal.rst
@@ -6,4 +6,5 @@ Asteroidal
 .. autosummary::
    :toctree: generated/
 
-   is_at_free 
+   is_at_free
+   find_asteroidal_triple
--- a/doc/reference/algorithms/asteroidal.rst
+++ b/doc/reference/algorithms/asteroidal.rst
@@ -1,0 +1,9 @@
+**********
+Asteroidal
+**********
+
+.. automodule:: networkx.algorithms.asteroidal
+.. autosummary::
+   :toctree: generated/
+
+   is_at_free 
--- a/networkx/algorithms/__init__.py
+++ b/networkx/algorithms/__init__.py
@@ -1,4 +1,5 @@
 from networkx.algorithms.assortativity import *
+from networkx.algorithms.asteroidal import *
 from networkx.algorithms.boundary import *
 from networkx.algorithms.bridges import *
 from networkx.algorithms.chains import *
--- a/networkx/algorithms/asteroidal.py
+++ b/networkx/algorithms/asteroidal.py
@@ -125,10 +125,7 @@ def is_at_free(G):
     >>> nx.is_at_free(G)
     False
     """
-    asteroidal_triple = find_asteroidal_triple(G)
-    if asteroidal_triple:
-        return False
-    return True
+    return find_asteroidal_triple(G) is None
 
 
 @not_implemented_for("directed")
--- a/networkx/algorithms/asteroidal.py
+++ b/networkx/algorithms/asteroidal.py
@@ -14,9 +14,9 @@ An asteroidal triple in a graph G is a set of three non-adjacent vertices
 u, v and w such that there exist a path between any two of them avoiding the
 neighborhood of the third. More formally, v_j, v_k belongs to the same
 connected component of G - N[v_i]. A graph which contains no asteroidal
-triples is called an AT-free graph. AT-free graphs is a graph class
-for which many NP-complete problems are solvable in polynomial time.
-Amongst them, independent set and coloring.
+triples is called an AT-free graph. The class of AT-free graphs
+is a graph class for which many NP-complete problems are solvable
+in polynomial time. Amongst them, independent set and coloring.
 """
 import networkx as nx
 from networkx.utils import not_implemented_for
@@ -27,13 +27,14 @@ __all__ = ["is_at_free"]
 @not_implemented_for("directed")
 @not_implemented_for("multigraph")
 def find_asteroidal_triple(G):
+    r"""Find an asteroidal triple in the given graph.
 
     An asteroidal triple is a triple of non-adjacent vertices such that
     there exists a path between any two of them which avoids the closed
     neighborhood of the third. It checks all independent triples of vertices
     and whether they are an asteroidal triple or not. This is done with the
     help of a data structure called a component structure.
-    The component structure encodes information about which vertices belongs to
+    A component structure encodes information about which vertices belongs to
     the same connected component when the closed neighborhood of a given vertex
     is removed from the graph. The algorithm used to check is the trivial
     one, outlined in [1]_, which has a runtime of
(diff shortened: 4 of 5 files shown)


dschult:
Hi!  Thanks for this -- it looks promising. :)

Here are some comments:

- you can put your name in CONTRIBUTORS.rst in the base directory if you want (don't have to)
- When using ```not_implemented_for``` the docstring says:

    Multiple types are joined logically with "and".
    For "or" use multiple @not_implemented_for() lines.

So, you have restricted MultiDiGraph, but not DiGraph or MultiGraph. I think you'll need two lines instead.

- I usually test locally using ```nosetests --with-doctest *optional_filename*```, but that only tests one environment, so I rely on the CI testing to check the other environments. Realistically I usually only test the files where I've changed stuff, so I guess I rely on CI to test everything even for my environment.
- The ```from nose.tools import *``` at the top of your test file pulls in lots of tools like ```assert_true``` and ```assert_false``` which you should use instead of e.g. ```assert_equal( --- , False)```.  There is even an ```assert_raises``` which you can read more about and/or look for examples.  Some people would say you should replace the ```*``` with each specific tool you use, but we're not too picky about that.
- In your tests, you use 3 lines for each test and the lines are spread out because you have many cases. If many more tests get added it may be hard to track. You can reduce the number of lines for each graph to check and complete testing of each graph before moving to the next. For example:

    is_at_free = nx.asteroidal.is_at_free
    assert_false(is_at_free(nx.cycle_graph(6))
    assert_true(is_at_free(nx.path_graph(6))
    ...

- You should add a ```asteroidal.rst``` file so this function shows up in the docs. Look at /doc/reference/algorithms for examples. If you have questions, ask -- it can be tricky at times.
- In the docstring, you state that the function checks if there are AT, but the name is ```is_at_free```. The phrase "check if there are AT" kind of implies True means there are AT. Perhaps reword that first line?
- In the docstring Parameters section, after ```certificate : boolean``` you should add: ```(default: False)``` to show it is optional.
- The docstring indicates the return value is boolean when sometimes it isn't. Perhaps you need to document the certificate case more carefully. 

- line 82: you use an entire if structure to return True or False when you can just ```return first_cond and second_cond and third_cond```.  Some would say you can in-line the conditions so when the first is False python doesn't have to compute the other two. That could make the line very long, so other approaches include nested ```if```s, or strategic line breaks and spacing to retain readability. Probably can just be left as is too.
- line 150: I think you have an extra layer of subgraph not needed. You only need the nodes from the components, so you don't have to create a subgraph to isolate those nodes. The loop can start immediately after G_reduced is created in line 149:

    for cc in nx.connected_components(G_reduced):
            label += 1
            for u in cc:
                row_dict[u] = label

Note that the inner loop can just iterate over ```cc``` instead of ```cc.nodes```.

haakonhr:
Thanks very much for the feedback and for the tips. I'll comment on the bullet points.

* not_implemented is fixed. The and/or distinction passed me by.
* The tests have been refactored into what should be more readable and more condensed. I also changed the grouping to per graph instead of the arrange-act-assert structure.
* asteroidal.rst is added. I used rich_club.rst as a template.
* The docstrings have been updated: the first line in `is_at_free` has been re-formulated to try and avoid any misunderstandings, as suggested. The `certificate` parameter is also documented better, w.r.t how it affects the output. However, I was not able to find any examples of multiple return so I just used the same structure as in the parameters section.
* I also included your other suggestions to make the code more concise and to avoid unecessary checks.

In addition, I included one more test where `certificate` is set to True. However, it currently relies on the label names to be passed. Is this bad practice? The asteroidal triple is not unique, but as long as the algorithm loops through from the beginning (or the end for that matter) then the same asteroidal triple will be detected.

dschult:
Hi ... Sorry for the delay in my response. 

I've been thinking about the ```certificate``` keyword and the 2-tuple return values for ```is_at_free```.  I think it would be better (and match our API elsewhere) if you create 2 functions:

    is_at_free(G)  # return True if there are no AT in G
    find_asteroidal_triple(G)  # return a 3-tuple of nodes, or None if G is AT-free

Basically, ```find_asteroidal_triple()``` would be the function you have now (with simplified return code).  And ```is_at_free()``` can just call find_asteroidal_triple and then check if the output is None. What do you think?

Finally, you define a function ```_is_asteroidal_triple``` which is a long, but single logical expression.
That function only gets used once. It's probably easier to read and faster (python function calls are somewhat expensive) to inline the function (put the 3-line logical statement in the main code and remove the function).  Have you tried this on larger graphs of say 100 or 1000 nodes? Is performance going to be an issue?

haakonhr:
Thanks for the reply!

I agree that the tuple is inelegant and it also did feel a bit forced. Your solution is much cleaner and the naming is more intuitive while it is not necessarily obvious what kind of certificate the `certificate` parameter refers to. I'll change this. 

Thanks for pointing out the overhead of doing function calls. I think the current implementation is more readable, but an inline comment can probably do the same job. It might be interesting to check how large a potential speed-up is. I'll try to look into this soon.

As to performance, I used a similar implementation for checking different real-world graphs (interference graphs where nodes consist of radio broadcasters in the US) and denser, transformed versions of them, with sizes ranging from 50 to 1000 vertices where the smaller ones were within the second and the largest ones would take roughly 15 minutes, if memory serves me right. I did a quick benchmark just now using some `%time` magic in a notebook on random graphs of sizes from 10 to 500 nodes. The largest all took less than a minute. Removing the function call also saves some time (roughly 20-30%) so I'll change this too. However, I do not know how accurate the `%time` is. I will see if I get some time later today or tomorrow to a more thorough (and scientific) benchmark.

dschult:
This looks pretty good.  Thanks very much for your work!

- Do you want to expose other functions besides ```is_at_free```? Perhaps ```find_asteroidal_triple```?
- The if statement in ```is_at_free``` is really testing whether the result is ```None```, so it should use ```is None```. Actually that whole function is probably better written as a one-liner:

    return find_asteroidal_triple(G) is None

This is close to being merged.

haakonhr:
Thanks again for the feedback!

I exposed `find_asteroidal_triple` as well as it is the function containing most of the logic and for the user it might be nice to be able to get the certificate, if it exists.

`is_at_free` have also been re-written to the oneliner you suggested.

dschult:
Looks good.  But you need to add ```find_asteroidal_triple``` to the .rst doc file.
:)

haakonhr:
Sorry for the oversight.

I have added it to the .rst file and also tried to clean up the commit history a bit. I'm still not a git poweruser so I hope I didn't mess up anything along the way.

I also have a couple of other potential contributions, but I thought I would ask your opinion first as to whether you think they make sense in networkx:
- More recognition algorithms for different $H$-free graphs where H are small graphs such as the claw, the chair, the bull and so on (it would need some work in order to generalize it)
- A utility to transform certain boolean formulas that arise in frequency planning problems into a corresponding constraint graph. It might be a tad too specialized for networkx?
- A feature generation utility similar to the package tsfresh (https://tsfresh.readthedocs.io/en/latest/) that takes a graph and computes a bunch of properties of them. I used this to generate features to try and characterize different kinds of networks in frequency assignment problems. It might also make sense as a standalone library. What do you think?

dschult:
I find it hard to know what some of the suggestions might become. But my first impression is that the first bullet is quite in line with what NetworkX provides. So that fits well. The second sounds like it would make a good "example" (in the examples folder) but developing a complete utility program would probably be better supported as a separate package. The third also seems large enough and self-contained enough to be a separate package, but I don't really know what your aiming for. The tsfresh is clearly a package for time series analysis. If you are envisioning a package that could capture lots of computed quantities for networks, then NetworkX tries to do that. :) If you envision a utility that would choose which computations to make and collect the data for many graphs, that feels like a front-end for networkx that might be better as a separate package -- though obviously there would be exchange of ideas and code between the two.

PR #3385, opened 2019-04-09
New Feature - Resistance Distance

bkief:
Added memory efficient algorithm for calculating resistance distance on a undirected graph

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ networkx.egg-info/
 
 # PyCharm project file
 .idea
+
+# VS Code settings
+.vscode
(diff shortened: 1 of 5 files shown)


dschult:
This looks like a good addition. Placement in the package is good. I only have a couple of comments/requests:

- remove the unused import of ```bisect```
- add a test using a MultiGraph to check those 2 lines of code
- add test of square matrix and right size matrix for _laplacian_submatrix (can use ```assert_raises``` or ```@raises``` decorator)
- check PEP8 style using the pycodestyle package or similar style checker. I see / operators without spaces and some lines are longer than 79 chars. There might be other things.

Thanks!

bkief:
@dschult Should be good to go!

PR #3387, opened 2019-04-09
Limit all_simple_paths by weight

harisbal:
This a first effort to allow limiting the results of `nx.all_simple_paths` not only by the length of the path but as well as by the distance covered as specified by the `weight` attribute.
The suggested approach is most likely inefficient but it could work as a basis for further improvement.


dschult:
Hi! Thanks for this and sorry for the delay. The PR seems to have differences related to the release 2.3 made about the same time as your PR.  Could you rebase to the current master or maybe just start fresh and ```git push --force``` onto this branch? You could also start a new PR if that is easier. I think you have added one function and a couple of tests, but it is hard to parse with all the other cruft in the github diff.

harisbal:
At this version multiple combinations of weights with their respective cutoffs can be supplied.

PR #3390, opened 2019-04-11
Add minimum source and target margin to draw_networkx_edges.

rajeee:
When the node sizes are especially small, and they have text labels, the base and the head of the arrows/edges can interfere with the text, thereby making reading the node labels hard. I have added a simple option to specify the minimum margin the edges would maintain from their source and target nodes.


dschult:
@ericmjl   This might be of interest to nxviz also.

PR #3393, opened 2019-04-16
Find all trails in a directed graph

MartinPJorge:
## Trail
This pull request includes a function to obtain all the trails of a given directed graph. A trail is a walk without repeated edges.

In a directed graph like:
```txt
 1 ---->  2 ---->  3
  ^     /  ^     /
  |____/   |____/
```
a possible trail could be the one traversing `[(1, 2), (2, 3), (3, 2)]`.

## Implementation
Function `all_trails()` transforms a directed graph into another graph where each directed edge connects two unique nodes, e.g., `(1,2)` connects nodes `(s1_2, d2_1)`. Hence the graph shown above becomes:

```txt
 s1_2         s2_1 <--  --> s2_3           s3_2
 ^   \      /  ^      \/     ^    \      /  ^
 |    \    /   |      /\     |     \    /   |
 |     \  /    |     /  \    |      \  /    |
 |      \/     |    /    \   |       \/     |
 |      /\     |   /      \  |       /\     |
d1_2 <--  --> d2_1          d2_3  <--  --> d3_2

```
If one executes `all_simple_paths()` over this graph, the retrieved solutions are trails.

### DIFF
--- a/networkx/algorithms/simple_paths.py
+++ b/networkx/algorithms/simple_paths.py
@@ -14,12 +14,14 @@ from networkx.utils import pairwise
 __author__ = """\n""".join(['Sérgio Nery Simões <sergionery@gmail.com>',
                             'Aric Hagberg <aric.hagberg@gmail.com>',
                             'Andrey Paramonov',
-                            'Jordi Torrents <jordi.t21@gmail.com>'])
+                            'Jordi Torrents <jordi.t21@gmail.com>',
+                            'Jorge Martín Pérez <jmartinp@it.uc3m.es>'])
 
 __all__ = [
     'all_simple_paths',
     'is_simple_path',
     'shortest_simple_paths',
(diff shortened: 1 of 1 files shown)


dschult:
You can add a ```sorted()``` to the test to make sure that the order of the returned values is treated the same (and test as the same). We need to design the tests to pass independent of the order in a dict (and thus in a graph). Even though CPython is moving to fixed order dicts, Pypy and others are not.

dschult:
Some suggested changes:

- doc_string for all_trails should start with a single line (not blank lines)
- When defining a trail, don't rely on users knowing what a "walk" is. Define that too.
- doc_string talks about "path"s. Should that be "trail"?
- remove ```import networkx as nx``` from examples. Also remove blank lines (```>>>```)
- same comments on doc_string for second function.

- need some tests..

PR #3395, opened 2019-04-17
Fix edge_color inconsistency with node_color and description.

### LINKED ISSUE #3394: Parameter edge_color is inconsistent with node_color and description
When using draw_networkx_nodes and draw_networkx_edges, the formats supported for the colors (node_color and edge_color) by the two methods are inconsistent.
The description of the node_color and edge_color parameters is also inconsistent with what is supported.

This is OK, although RGBA is not mentioned as a supported format:
```
import networkx as nx
G = nx.complete_graph(5)
nx.draw_networkx_nodes(G, nx.spring_layout(G), node_color=(0.0,0.0,1.0,1.0))
nx.draw_networkx_edges(G, nx.spring_layout(G), edge_color='r')
```

This throws an error:
```
import networkx as nx
G = nx.complete_graph(5)
nx.draw_networkx_nodes(G, nx.spring_layout(G), node_color=(0.0,0.0,1.0,1.0))
nx.draw_networkx_edges(G, nx.spring_layout(G), edge_color=(0.0,0.0,1.0,1.0))
```

[matplotlib.pyplot.scatter](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.scatter.html) says:

> Note that c should not be a single numeric RGB or RGBA sequence because that is indistinguishable from an array of values to be colormapped. If you want to specify the same RGB or RGBA value for all points, use a 2-D array with a single row. Otherwise, value- matching will have precedence in case of a size matching with x and y.

joseppinilla:
Both draw_networkx_nodes and draw_networkx_edges support RGB or RGBA tuples.
Fixes #3394

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -181,11 +181,11 @@ def draw_networkx(G, pos=None, arrows=True, with_labels=True, **kwds):
        Size of nodes.  If an array is specified it must be the
        same length as nodelist.
 
-    node_color : color string, or array of floats, (default='#1f78b4')
-       Node color. Can be a single color format string,
-       or a  sequence of colors with the same length as nodelist.
-       If numeric values are specified they will be mapped to
-       colors using the cmap and vmin,vmax parameters.  See
+    node_color : color or array of colors (default='#1f78b4')
+       Node color. Can be a single color or a sequence of colors with the same
+       length as nodelist. Color can be string, or (rgb) or (rgba) tuple of
+       three floats from 0-1. If numeric values are specified they will be
+       mapped to colors using the cmap and vmin,vmax parameters. See
        matplotlib.scatter for more details.
 
     node_shape :  string, optional (default='o')
@@ -207,11 +207,11 @@ def draw_networkx(G, pos=None, arrows=True, with_labels=True, **kwds):
     width : float, optional (default=1.0)
        Line width of edges
 
-    edge_color : color string, or array of floats (default='r')
-       Edge color. Can be a single color format string,
-       or a sequence of colors with the same length as edgelist.
-       If numeric values are specified they will be mapped to
-       colors using the edge_cmap and edge_vmin,edge_vmax parameters.
+    edge_color : color or array of colors (default='k')
+       Edge color. Can be a single color or a sequence of colors with the same
+       length as edgelist. Color can be string, or (rgb) or (rgba) tuple of
+       three floats from 0-1. If numeric values are specified they will be
+       mapped to colors using the edge_cmap and edge_vmin,edge_vmax parameters.
 
     edge_cmap : Matplotlib colormap, optional (default=None)
        Colormap for mapping intensities of edges
@@ -320,11 +320,11 @@ def draw_networkx_nodes(G, pos,
        Size of nodes (default=300).  If an array is specified it must be the
        same length as nodelist.
 
-    node_color : color string, or array of floats
-       Node color. Can be a single color format string (default='#1f78b4'),
-       or a  sequence of colors with the same length as nodelist.
-       If numeric values are specified they will be mapped to
-       colors using the cmap and vmin,vmax parameters.  See
+    node_color : color or array of colors (default='#1f78b4')
+       Node color. Can be a single color or a sequence of colors with the same
+       length as nodelist. Color can be string, or (rgb) or (rgba) tuple of
+       three floats from 0-1. If numeric values are specified they will be
+       mapped to colors using the cmap and vmin,vmax parameters. See
        matplotlib.scatter for more details.
 
     node_shape :  string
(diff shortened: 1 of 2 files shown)


dschult:
Thank you for making the docs consistent -- and also fixing the documented default value for edge color.  It is good to keep the options up-to-date with what matplotlib has to offer.

Given the ```scatter``` warning on 1d-arrays of length 3 and 4, I think we should stick with requiring ```len``` to be 1 (if not string_like).  Your example code from #3394 would then print a warning. But making ```node_color``` a 2D-array with 3 or 4 elements would work without warnings:

    nx.draw_networkx(G, node_color=(0.0,0.0,1.0,1.0))

should give a warning unless there are 4 nodes in G.

    nx.draw_networkx(G, node_color=[(0.0,0.0,1.0,1.0)])

should be allowed and use the same blue color for all nodes.  What do you think?

joseppinilla:
I think it'd be OK to have the warning, like you mention, and discourage the use of a single tuple. But I think the issue is in edge_color not behaving the same way.
If you try both:
```
nx.draw_networkx(G, edge_color=(0.0,0.0,1.0,1.0))
```
or:
```
nx.draw_networkx(G, edge_color=[(0.0,0.0,1.0,1.0)])
```

They have different errors, instead of having the warning from scatter.

dschult:
Hmmm.. I think we should make it work as flexibly as possible.

    nx.draw(D, edge_color=[(0.0,0.0,1.0,1.0)])
    nx.draw(D, edge_color=(0.0,0.0,1.0,1.0))     # one color except if 4 edges in G. Then 4 nums into cmap
    nx.draw(D, edge_color=[(0.0,0.0,1.0)])
    nx.draw(D, edge_color=(0.0,0.0,1.0))       # one color except if 3 edges in G.
    nx.draw(D, edge_color='green')
    nx.draw(D, edge_color='g')
    nx.draw(D, edge_color=['green'])
    nx.draw(D, edge_color=['g'])

should all work for node_color and for edge_color.  I think we can do this.

joseppinilla:
I think it's best to leave the color parsing to matplotlib as much as possible.
So... I suggest removing the conversion from _edge_color_ to _edge_colors_ (lines 584-607)
And pass edge_color to LineCollection (line 610).

Something else would have to be done for arrows, but... I suggest following that idea of passing _edge_color_ directly.
What do you think?

dschult:
I think that will make node_color and edge_color follow different rules for color assignment -- in an undocumented way (without referring to matplotlib docs). I do like the idea of using the defaults that matplotlib has chosen. It'd be nice if node_color and edge_color encoded colors the same way, but that isn't a design goal per se.

The value we should add to matplotlib here is checking/implementing any Graph Theory constraints before calling matplotlib.  Are there any in this case? Previous developers (often using older versions of matplotlib) have made it possible to assign colors as a single value or as an array of values. Matplotlib does this now. Maybe there isn't any reason to add to the possibilities provided by matplotlib. (well -- except if we want node_color to be similar to edge_color in the way it handles colors?)

joseppinilla:
I had some time to experiment again. These new changes can handle most anything you throw at it, but mostly by passing things directly to matplotlib. The special "graph-theoretic" case is when we pass a list of numbers and I handle that first. Instead of case by case (undirected/directed) as was done before.

There are some things that could be removed, but I like that this way it can really handle almost anything.

```nx.draw(G, edge_color=None) # Defaults to black. Use alpha=0.0 instead to remove edges.
nx.draw(G, edge_color=[(0.0,0.0,1.0,1.0)])
nx.draw(G, edge_color=(0.0,0.0,1.0,1.0))
nx.draw(G, edge_color=[(0.0,0.0,1.0)])
nx.draw(G, edge_color=(0.0,0.0,1.0))
nx.draw(G, edge_color='green')
nx.draw(G, edge_color='g')
nx.draw(G, edge_color=['green'])
nx.draw(G, edge_color=['g'])
nx.draw(G, edge_color=['g','k','r']) # matplotlib can handle these
nx.draw(G, edge_color=['g','k','b','r'])
```
You can change G between directed, undirected, len(G)==3 or 4 and should all work.

I also added the support for iterables of colors that are not exactly the same as the number of edges (line 668). Matplotlib handles these, so I thought it would be better to have that too. But I left a stub with an error message if that's a better idea.

dschult:
This looks good. Can you add test(s) that runs the commands you put here as examples?
That will help make sure we don't regress when making other changes later. Thanks for this.

joseppinilla:
I created the tests only for directed (arrows), since `edge_color` for undirected graphs is passed directly to `matplotlib`, and there is coverage for `cmap` and `edge_color` in other tests.

dschult:
We should include some tests for undirected graphs even if ```edge_color``` is passed directly to ```matplotlib``` just so we know when they have changed their API.

Your added test function goes way beyond just testing color (Thanks!) so maybe just add another function that does something like:

    for G in (self.G, self.G.to_directed()):
        # smoke test of color inputs 
        nx.draw(G, edge_color=[(0.0,0.0,1.0,1.0)])
        nx.draw(G, edge_color=(0.0,0.0,1.0,1.0))
        nx.draw(G, edge_color=[(0.0,0.0,1.0)])
        nx.draw(G, edge_color=(0.0,0.0,1.0))
        nx.draw(G, edge_color='green')
        nx.draw(G, edge_color='g')
        nx.draw(G, edge_color=['green'])
        nx.draw(G, edge_color=['g'])
        nx.draw(G, edge_color=['g','k','r']) # matplotlib can handle these
        nx.draw(G, edge_color=['g','k','b','r'])

joseppinilla:
I'm running into an issue related to this [PR #3370](https://github.com/networkx/networkx/pull/3370)
I needed to make modifications to how alpha is interpreted (make default alpha=None) to make things work well.

dschult:
Changing the default alpha from 1.0 to None makes sense to me if it causes matplotlib to assign the value 1.0 to alpha.  Those changes look good. Do we need tests for alpha now?
Did your commit finish the alpha handling in a way that does what #3370 wants?
I couldn't tell from your comment whether this is done or still needing to fix something.
Thanks!

joseppinilla:
Fixed. And covered by tests in lines 88, 92 and 95 of test_pylab.py

PR #1878, opened 2015-12-18
Implemented eulerian_path().

humberto-ortiz:
This is a refactoring of https://github.com/networkx/networkx/pull/1488

Edwardo RIvera was my undergraduate student, and has since graduated. 
I'd like to see his work merged upstream. I think I have addressed most of the concerns
noted on the pull request.

PR #1488, opened 2015-04-28
Added euler path for Graphs, Digraphs and Multigraphs

edwardo2013:
We added the file find_euler_path.py which has an implementation of the euler path algorithm. It is general: It works for Graphs, Digraphs and Multigraphs. We added some tests after the function. It uses the eulerian_circuit() function, which is implemented in euler.py . We make the comments following the aforementioned file.

Please tell us is this is the way to document the function. If not, provide information/files to add the required documentation.

About the code: the function receives a networkx graph G. If G is eulerian, then just returns the eulerian circuit. If not, then verify degree condition of nodes to see if it has an euler path. If it does not have it, raise an error. Then, at line 145 the code to find the euler path is given (this works because we know that we have the conditions to have an euler path). It returns a generator which has the euler path. 

Please review this code, we would be happy to see it merged in the networkx project. Please let us know if some changes must be made (emails are in the file) or if it is posible to make the code more compact.

Thanks,
Edwardo S. Rivera

PR #3399, opened 2019-04-21
Implemented Euler Path functions

ja-ox:
Added functions to check if a graph has an Euler Path, whether it is semi Eulerian, and to return an Euler Path if one exists.

Also adds initial test cases for new functions.

Supersedes previous similar (closed) pull requests #1878 and #1488. See comment [here](https://github.com/networkx/networkx/pull/1878#issuecomment-485263242).

To do: a function implementing the [BEST theorem](https://en.wikipedia.org/wiki/BEST_theorem) to return _all_ Euler Paths (for a directed graph).

### DIFF
--- a/doc/reference/algorithms/euler.rst
+++ b/doc/reference/algorithms/euler.rst
@@ -12,4 +12,3 @@ Eulerian
    is_semieulerian
    has_eulerian_path
    eulerian_path
-   find_path_start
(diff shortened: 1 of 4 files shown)


dschult:
It seems like a good idea to add these functions.   Thanks!

You can/should keep the ```_find_path_start``` function private, and in the tests when you need to call it use something like:

    find_path_start = nx.algorithms.euler._find_path_start

The doc_strings need work -- Use one-line descriptions, followed by blank line and then paragraph defining the terms you use. Then define the parameters and return values, notes, "see also"s, references and examples.  Look at existing functions for how to make it look. It would be good to have a reference for most of these since they are definition heavy.

You might need to update the relevant *.rst files in /doc/reference to make the docs show the functions.

PR #3400, opened 2019-04-22
Floyd-Warshall Optimization

Pascal-Ortiz:
This is a continuation of my unfinished PR #3400. It implements a low level and very simple optimization of the `floyd_warshall_predecessor_and_distance()` function in order to avoid multiple recomputations of dictionary references inside of the inner loops.  

I abreviated `pred[w]` by `Pw`.  As noticed by Dan Schult, the variable's name is not very suggestive but the shortcut is only for performance and has no semantics and this kind of shortcut is commun, see for instance [this code](https://github.com/networkx/networkx/blob/master/networkx/algorithms/communicability_alg.py#L94) from NetworkX.

I have done some benchmarks against random graphs from 20 up to 4000 nodes and for various densities: performances are always better, and on my machine the gain is from 22% up to 38%, cf. benchmark code at the end of this message.

In the original code, predecessors are stored in a default dictionary `pred`. As a consequence, the shortcut `Pw = pred[w]` has sometimes an unexpected side effect: it creates an empty dictionary that doesn't appear in the return value of the function. Networkx test doesn't detect this error. So I added a test to cover this case.

```python
# Benchmark code
from time import perf_counter
import networkx as nx

def random_edges(n, density, max_weight=100):
    from random import randrange
    M=n*(n-1)//2
    m=int(density*M)
    edges=set()
    wedges=[]
    while len(edges)<m:
        L=(randrange(n),randrange(n))
        if L[0]!=L[1] and L not in edges:
            w=float(randrange(max_weight))
            wedges.append(L+(w,))
            edges.add(L)
    return wedges 


def floyd_warshall_predecessor_and_distance_opt(G, weight='weight'):
    # Optimised version
    dist = defaultdict(lambda: defaultdict(lambda: float('inf')))
    for u in G:
        dist[u][u] = 0
    pred = defaultdict(dict)
    undirected = not G.is_directed()
    for u, v, d in G.edges(data=True):
        e_weight = d.get(weight, 1.0)
        dist[u][v] = min(e_weight, dist[u][v])
        pred[u][v] = u
        if undirected:
            dist[v][u] = min(e_weight, dist[v][u])
            pred[v][u] = v
    for w in G:
        Dw=dist[w]
        Pw=pred[w]
        for u in G:
            Du=dist[u]
            Pu=pred[u]
            for v in G:
                d=Du[w] + Dw[v]
                if Du[v] > d:
                    Du[v] = d
                    Pu[v] = Pw[v]
    return {u:pred[u] for u in G if pred[u]}, dict(dist)


def read(filename):
    lines=open(filename).read().strip().split('\n')
    n=int(lines[0])
    m=int(lines[1])
    lines=lines[2:]
    L=[line.split() for line in lines]
    return n, [(int(a), int(b), float(w)) for (a,b, w) in L]

def wedges2graph(wedges):

    G = nx.DiGraph()
    G.add_weighted_edges_from(wedges)
    return G

def perf(n, density, repeat=1):
    cumul=cumul_opt=0
    check=True
    for _ in range(repeat):
        wedges=random_edges(n, density)
        G=wedges2graph(wedges)
        
        start= perf_counter()
        pred, dist=nx.floyd_warshall_predecessor_and_distance(G)
        cumul+= perf_counter()-start

        start_opt= perf_counter()
        pred_opt, dist_opt=floyd_warshall_predecessor_and_distance_opt(G)
        cumul_opt+= perf_counter()-start_opt  
        
        check=(pred_opt==pred) and (dist_opt==dist)
        if not check:
            raise Exception
    return cumul/repeat, cumul_opt/repeat


nnodes=[400]
density=[0.2]

durations=[]
repeat=1

for n in nnodes:
    for d in density:
        print("n =", n, "d =", d)
        duration, duration_opt=perf(n, d, repeat)
        durations.append((n, d, duration, duration_opt))
        print("%.2f, %.2f" %(duration, duration_opt))
        print("Gain: %.0f %%" %((duration-duration_opt)/duration*100))
        



Output:


n = 400 d = 0.2
9.01, 6.15
Gain: 32 %
```


dschult:
To be considered for adoption, you need to describe what the PR is about -- provide justification that this enhances rather than de-hances (my made up word) -- etc.  Also, your variable names are fairly cryptic abbreviations. Can you ease readability by choosing more meaningful names? (that won't affect whether this gets accepted or not -- just a suggestion really)

dschult:
Great! Thanks for this update...

Github is good for keeping track of who has said what and submitted which changes. 
One step is to add comments to the bottom of the pull request even if you already have comments you could edit. That way github sends an email to people "following" the pull request.  Editing a comment doesn't trigger any notifications of the changes.

Looks like you have figured out how to nicely display code in the comments! Thanks for that performance data.

Since we're going on about names and its good practice for github anyway, could you change ```Du```, ```Pu```, etc to ```dist_u``` and ```pred_u```.  Also, in your test, you name one variable ```path``` when it is predecessors. Please change to ```pred```.

Please add back the comments about defaultdict magick near lines 108 and 115. If they are incorrect please let me know. But it is better to have them than not IMO.

If I understand your test and change in procedure, you are removing ```u``` as a key of the ```pred``` dict if it has no predecessors. The function is supposed to return all-pairs results, so I think it needs to have all nodes as keys - even if their result is empty. So I think I'd like you to remove that change (and adapt the corresponding test). I may not be understanding completely though, so push back if you have other reasons for this change.

Thanks!

Pascal-Ortiz:
> Great! Thanks for this update...
> 

You are welcome!

 
> Since we're going on about names and its good practice for github anyway, could you change `Du`, `Pu`, etc to `dist_u` and `pred_u`. Also, in your test, you name one variable `path` when it is predecessors. Please change to `pred`.


Thanks for the suggestion, i will change the variable names.


> 
> Please add back the comments about defaultdict magick near lines 108 and 115. If they are incorrect please let me know. But it is better to have them than not IMO.
> 

OK. By the way, in the original `floyd_warshall_predecessor_and_distance` function, the return instruction provided is :

```python
    return dict(pred), dict(dist)
```

where `pred` is a defaultdict. But after all, a defauldict is also a dict, so why casting `pred` with the `dict` constructor? If I understand well, this cast was meant to "forget" the defaulting character of `pred`. And indeed, how handy they are, defaultdicts can change his own value in your back: evaluating a simple expression like `pred[k]` may change `pred` contents and this can be surprising. So it was a good idea to cast to dict. 

But on the other hand, cf. the `return` instruction above, `dist` is a defaultdict of defaultdicts, hence `dict(dist)` is now a dict of defaultdicts. As a result, a user of the `floyd_warshall_predecessor_and_distance` function will get back a bunch of defaultdicts and doesn't need them. Using defaultdict here was only useful for the implementator of the function, not for the user.

So a better way to return from `floyd_warshall_predecessor_and_distance` function could be :


```python
non_empty_pred = {u:pred[u] for u in G if pred[u]}
non_defaultdict_dist = {u:dict(dist[u]) for u in G}

return non_empty_pred, non_defaultdict_dist

```






> If I understand your test and change in procedure, you are removing `u` as a key of the `pred` dict if it has no predecessors. 

Because this is the way the initial implementation works and of course don't want to break it. The assignment `Pw = pred[w]` causes a defaultdict side effect and I'm only adjusting it.


> The function is supposed to return all-pairs results

For clarity, suppose graph has `n` nodes. The initial function `floyd_warshall_predecessor_and_distance` **asymmetrically** returns two dicts:

- the first one is a dict of non-empty dict
- the other one is a dict of `n` defaultdicts keyed by the `n` nodes

The second dict returns all-pairs results but the first one does not. As an illustration of the implementation behavior (networkx version 2.2):


```python
import networkx as nx
print("Networkx version:", nx.__version__, end='\n\n')

n=5
wedges=[(0, 1, 2.), (1, 2, 3.), (0, 2, 6.), (3, 4, 1.)] 
G = nx.DiGraph()

G.add_weighted_edges_from(wedges)

pred, dist=nx.floyd_warshall_predecessor_and_distance(G)

# for readability, remove defaultdict reference
non_defaultdict_dist = {u:dict(dist[u]) for u in G}
print("Predecessors:")
print(*pred.items(), sep='\n')
print()
print("Distances:")
print(*non_defaultdict_dist.items(), sep='\n')

```

outputting



```bash
Networkx version: 2.2

Predecessors:
(0, {1: 0, 2: 1})
(1, {2: 1})
(3, {4: 3})

Distances:
(0, {0: 0, 1: 2.0, 2: 5.0, 3: inf, 4: inf})
(1, {1: 0, 2: 3.0, 0: inf, 3: inf, 4: inf})
(2, {2: 0, 0: inf, 1: inf, 3: inf, 4: inf})
(3, {3: 0, 4: 1.0, 0: inf, 1: inf, 2: inf})
(4, {4: 0, 0: inf, 1: inf, 2: inf, 3: inf})

```

dschult:
Very helpful comments...  Thank you!

I believe the thinking was that converting the outer dict was good for removing the defaultdict nature, and doesn't affect performance. Is it also possible that the resulting dict-of-defaultdict would be useful? (not sure).
I haven't tested it, but perhaps converting to dict-of-dict takes too much time.

I did not realize that the two returned dicts differ in how they handle nodes without predecessors.
I can argue that pred should not have nodes if there are no preds, and that dist should have all nodes. But I don't know whether that is true in practice. 

I would prefer not to add lots of post-processing. And I'd prefer to leave dist with all-pairs information. Maybe the best solution to make them consistent is to add empty entries to pred for each node. But again, I'm not sure that consistency between t he two dicts is necessary. Maybe they should be different. Can you think of a rationale for making them the same or for making them different? The rationale should be based on how they would be used by the user.

dschult:
Maybe you should try leaving ```pred_w``` and ```pred_u``` out. They only save time when the ```if``` is true, so it might not save much time compared to the time needed to remove empty dicts later. Check the speed difference, but it is an idea for how to handle the empty dicts.

Pascal-Ortiz:
> Maybe you should try leaving `pred_w` and `pred_u` out. They only save time when the `if` is true, so it might not save much time compared to the time needed to remove empty dicts later. Check the speed difference, but it is an idea for how to handle the empty dicts.


To check the speed difference, I have done some tests:

```
n = 200 d = 0.1
1.05, 0.65, 0.70
Gain: 38 %, Gain var 34 %
----------------------------
n = 200 d = 0.25
1.11, 0.72, 0.71
Gain: 35 %, Gain var 37 %
----------------------------
n = 200 d = 0.4
1.13, 0.72, 0.73
Gain: 36 %, Gain var 36 %
----------------------------
n = 800 d = 0.1
78.31, 55.58, 57.30
Gain: 29 %, Gain var 27 %
----------------------------
n = 800 d = 0.25
84.60, 60.81, 61.72
Gain: 28 %, Gain var 27 %
----------------------------
n = 800 d = 0.4
89.38, 66.25, 67.23
Gain: 26 %, Gain var 25 %
----------------------------
```
As an example, explanation for the last case:

directed graph has 800 vertices with 0.4 as density

Durations:

- NetworkX : 89.38s
- my PR: 66.25 meaning a gain of 26% wrt Networkx time execution
- your above suggestion: 67.23 meaning a gain of 25% wrt Networkx time execution

It results that variable memorization for predecessors doesn't lead to a significant gain. And, as you have noticed, leaving `pred_w` and `pred_u` out prevents post-processing.

On the other hand, similar functions with name ending in `_predecessor_and_distance ` return a full dictionary.

dschult:
Can you remove the .kate-swp file from the PR?
Also, please put back the comment lines about defaultdicts at about line 108 of dense.py
Thanks

PR #3401, opened 2019-04-23
Add bethe hessian matrix

jg-you:
The Bethe Hessian matrix or deformed Laplacian matrix generalizes the graph Laplacian.
 It is great for clustering, and also for counting the number of communities in dense networks, see:

[Saade, Alaa, Florent Krzakala, and Lenka Zdeborová. "Spectral clustering of graphs with the bethe hessian." Advances in Neural Information Processing Systems. 2014.](https://arxiv.org/abs/1406.1880)

and
[Le, Can M., and Elizaveta Levina. "Estimating the number of communities in networks by spectral methods." arXiv preprint arXiv:1507.00827 (2015).](https://arxiv.org/abs/1507.00827).

This PR implements this family of matrices and its spectrum, for undirected networks, with tests and documentation.

### DIFF
--- a/doc/reference/linalg.rst
+++ b/doc/reference/linalg.rst
@@ -25,6 +25,14 @@ Laplacian Matrix
    normalized_laplacian_matrix
    directed_laplacian_matrix
 
+Bethe Hessian Matrix
+--------------------
+.. automodule:: networkx.linalg.bethehessianmatrix
+.. autosummary::
+   :toctree: generated/
+
+   bethe_hessian_matrix
+
 Spectrum
 ---------
 .. automodule:: networkx.linalg.spectrum
@@ -32,6 +40,7 @@ Spectrum
    :toctree: generated/
 
    laplacian_spectrum
+   bethe_hessian_spectrum
    adjacency_spectrum
    modularity_spectrum
 
--- a/networkx/linalg/__init__.py
+++ b/networkx/linalg/__init__.py
@@ -9,3 +9,5 @@ import networkx.linalg.laplacianmatrix
 from networkx.linalg.algebraicconnectivity import *
 from networkx.linalg.modularitymatrix import *
 import networkx.linalg.modularitymatrix
+from networkx.linalg.bethehessianmatrix import *
+import networkx.linalg.bethehessianmatrix
--- a/networkx/linalg/bethehessianmatrix.py
+++ b/networkx/linalg/bethehessianmatrix.py
@@ -1,18 +1,18 @@
-"""Bethe Hessian matrix of graphs.
-"""
+# -*- coding: utf-8 -*-
 #    Copyright (C) 2004-2019 by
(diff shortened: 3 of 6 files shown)


jg-you:
I do not think that the codecov/patch check is correctly flagging this PR. All relevant tests pass.

dschult:
The code coverage check simply means that your new code has 86.2% of it's lines covered by tests. That is supposed to be an incentive for you to write tests that execute the lines of code not executed by the other tests. These are usually special cases, raising exceptions, branches of if statements that rarely happen, etc.  Think about if there are lines of code you should be testing that you aren't. If there are, include tests. If  not, don't worry about it. we can/do merge PRs that don't pass this fairly arbitrary coverage level.

Thanks for your contributions!

jg-you:
Ah thanks for the clarification. I'll add some extra test and comment back.

jg-you:
Alright, now I test everything added in the PR, 

see :

https://codecov.io/gh/networkx/networkx/compare/f210354d2657c01c648896bfc02236b8ef312555...231d720fddd04ba4e5eab53f53bca235744ae5b0/diff

Somehow codecov is detecting a coverage decrease in a part of the code that I did not alter in the PR: isomorphvf2.py and joint_degree_seq.py.

dschult:
Yeah -- I have seen that happen before with other PRs and couldn't figure out what codecov was doing to flag those two files. Didn't have time to track it down yet...

dschult:
You'll need to add/update some rst files to get this new function into the documentation. 

Also, why do you use ```eigvalsh()```?  Is there a better function -- perhaps one that can avoid the call to ```.to_dense()```

jg-you:
RE: eigvalsh. The bethe hessian is symmetric therefore the specialized algorithm for Hessian matrix will be faster than the general one (in fact adjacency_spectrum could use it too in the undirected case, as well as modularity_spectrum in that case).  
Regarding sparsity: since the spectrum functions return full spectra, there are no way around todense(). Only if the functions gave the leading X eigenvalue would we be able to harness the sparse data structure with scipy.sparse.linalg.

RE: Doc. I completely forgot the rst updates, sorry. I'v added them.

A postscript: There are a few inconsistencies in the linalg doc (missing parts for `spectrum` and weird oredering), as well as missed opportunity for using eigvalsh as I noted above. I can fix all that in another PR "Tidy up linalg" once this one is pulled in?

dschult:
One more change:  Can you shift the heading ("authors", imports, comments, etc) to match the [suggested ordering on the wiki page](https://github.com/networkx/networkx/wiki#user-content-top-of-the-module).  Essentially, match the ordering/style in [richclub.py](https://github.com/networkx/networkx/blob/master/networkx/algorithms/richclub.py)

Note that one aspect of that is the copyright will still have the same names (and you can add yours to that list) but the list of authors can/should be you since you wrote this code.

jg-you:
Done. Thank you so much for your code review!

PR #3409, opened 2019-04-29
Binomial trees generator

efraimrodrigues:
Recent research has been developed on binomial trees and networkx doesn't have a generator yet. So here it is my contribution on generating binomial trees.

More information about binomial trees can be found here: https://www.sciencedirect.com/science/article/pii/S097286001730289X


dschult:
Thank you for this. 

- Can you put the docs from the "Notes" section as a paragraph after the doc_string heading. That is, start with the current one-line description, put a blank line, then put a paragraph description of what the definition is. Furthermore, can you be more specific about what "linking" 2 trees means (how are they connected/linked?)

- I think the code can be simplified -- since you know the sizes and all nodes are integers.

Something like the following should be faster:

    def binomial_tree(n):
        G = nx.empty_graph(1)
        N = 1
        for i in range(n):
            edges = [(u + N, v + N)  for (u, v) in G.edges]
            G.add_edges_from(edges)
            G.add_edge(0,N)
            N *= 2
        return G

efraimrodrigues:
I just committed your suggestions.

efraimrodrigues:
Thanks for your suggestions @dschult . I just committed them.

networkx/generators/classic.py
```diff
@@ -195,6 +196,37 @@ def barbell_graph(m1, m2, create_using=None):
         G.add_edge(m1 + m2 - 1, m1 + m2)
     return G
 
+def binomial_tree(n):
+    """Returns the Binomial Tree of order n.
+
+    Parameters
+    ----------
+
```
dschult: remove blank line after Parameters heading.

networkx/generators/classic.py
```diff
@@ -195,6 +196,37 @@ def barbell_graph(m1, m2, create_using=None):
...
+        Order of the binomial tree.
+
+    Returns
+    -------
+    G : NetworkX graph
+        A binomial tree of $2^n$ vertices and $2^n - 1$ edges.
+
+    Notes
+    -----
+    The binomial tree of order 0 consists of a single vertex.
```
dschult: Move this info from the Notes section into text one blank line under the first line of the doc_string.
So there will be no Notes section -- only a paragraph or two near the top.

dschult:
You'll also need to add a line to the /doc/reference/generators.rst file to get the documentation to include this function.  Thanks!

Issue #3388, opened 2019-04-09
Add function for group betweenness centrality


nandahkrishna:
Recently I worked on a program to calculate the betweenness centrality of a group of nodes in a graph, rather than the individual betweenness of each node. I was wondering if I could add a function to do the same, as I believe it's a useful centrality measure. Would this be fine?
<img width="333" alt="GBC" src="https://user-images.githubusercontent.com/33784207/55827819-b4b39c00-5b28-11e9-96ad-ad5326373cf9.png">

dschult:
One image of a formula without definitions or context is not enough information to say if this is useful.
Do you have a reference/article that describes this centrality measure?

nandahkrishna:
Sorry about that, should have linked you to my references. Group centrality measures were discussed by Everett and Borgatti in 1999, as part of their extensive studies on social network analysis. Here's the link to their paper: http://www.casos.cs.cmu.edu/events/summer_institute/2006/reading_list/borgatti/Borgatti_Centrality_of_Groups.pdf

Ulrik Brandes gives us a brief mention of group betweenness in this paper, on page 12: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.72.9610&rep=rep1&type=pdf

The importance of group betweenness centrality is also highlighted in the introduction of these papers: 
https://pdfs.semanticscholar.org/f3d6/dfef4f99896c31ea8dbb521b4a785fd63ca9.pdf
http://www.cs.ucsb.edu/~arlei/pubs/sdm18.pdf

And here's a paper highlighting an application of this measure: https://www.elektrokomponenten.ch/media/files/09e4150bd20e58c913000000.pdf

I do hope these help, do tell me if you require anything else. If you find this measure worthy of inclusion, would it be fine if I try to implement it, being a first time contributor? Thank you for your time.

dschult:
Yes -- these links would be good to include in the docs for the function(s) too.

It might be best to create a new module in networkx/algorithms/centrality/ named something like group_betweenness.py 
If you have questions about how to organize the code or how to use github, ask.  It's often helpful to create a "close to finished" PR (rather than wait for it to be polished) so we can look at the code and ask questions -- and you can ask questions.  Adding new commits and pushing to your github automatically updates the PR in networkx after the PR is created.

nandahkrishna:
[...]

PR #3421, opened 2019-05-04
Adding module for group centrality measures

nandahkrishna:
Added group.py, a new module, to networkx.algorithms.centrality for group centrality computation functions. Added function for group betweenness centrality. Please do review the code and suggest any changes/improvements. Thanks.
(Future work: Adding functions for group closeness and group degree centralities)
Issue: #3388

### DIFF
--- a/networkx/algorithms/centrality/group.py
+++ b/networkx/algorithms/centrality/group.py
@@ -99,13 +99,14 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
     # accumulation
     for pair in combinations(V_C, 2):  # (s, t) pairs of V_C
         try:
-            paths = list(nx.all_shortest_paths(G, source=pair[0],
-                                               target=pair[1], weight=weight))
+            paths = 0
             paths_through_C = 0
-            for path in paths:
+            for path in nx.all_shortest_paths(G, source=pair[0],
+                                              target=pair[1], weight=weight):
                 if set(path) & C:
                     paths_through_C += 1
-            betweenness += paths_through_C / len(paths)
+                paths += 1
+            betweenness += paths_through_C / paths
         except nx.exception.NetworkXNoPath:
             betweenness += 0
     # rescaling
--- a/networkx/algorithms/centrality/group.py
+++ b/networkx/algorithms/centrality/group.py
@@ -8,7 +8,6 @@
 #    BSD license.
 #
 """Group centrality measures."""
-from __future__ import division
 from itertools import combinations
 
 
@@ -27,7 +26,7 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
     .. math::
 
     c_B(C) =\sum_{s,t \in V-C; s<t} \frac{\sigma(s, t|C)}{\sigma(s, t)}
-    
+
     where $V$ is the set of nodes, $\sigma(s, t)$ is the number of
     shortest $(s, t)$-paths, and $\sigma(s, t|C)$ is the number of
     those paths passing through some node in group $C$. Note that
@@ -44,14 +43,14 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
       centrality is to be calculated.
 
     normalized : bool, optional
-      If True the betweenness values are normalized by `2/((|V|-|C|)(|V|-|C|-1))`
-      for graphs, and `1/((|V|-|C|)(|V|-|C|-1))` for directed graphs where `|V|`
-      is the number of nodes in G and `|C|` is the number of nodes in the group C.
-      
+      If True, group betweenness is normalized by `2/((|V|-|C|)(|V|-|C|-1))`
+      for graphs and `1/((|V|-|C|)(|V|-|C|-1))` for directed graphs where `|V|`
+      is the number of nodes in G and `|C|` is the number of nodes in C.
+
     weight : None or string, optional (default=None)
       If None, all edge weights are considered equal.
       Otherwise holds the name of the edge attribute used as weight.
-      
+
     Returns
     -------
     betweenness : float
@@ -68,7 +67,7 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
     betweenness centrality of nodes. Group betweenness is also mentioned in
     his paper [2]_ along with the algorithm. The importance of the measure is
     discussed in [3]_.
-    
+
     The number of nodes in the group must be a maximum of n - 2 where `n`
     is the total number of nodes in the graph.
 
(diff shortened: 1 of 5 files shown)


nandahkrishna:
Not sure why the codecov check is failing. Have I made some mistake?

dschult:
Looks good!   Just a few comments for now...

- we don't need __future__ anymore: we have dropped support for python 2, the future is NOW!
- check your code style using pep8 or pycodestyle (which is the new name for pep8)
- Replace the ```_``` variable name with something meaningful. e.g. ```for path in paths:```
The time to use ```_``` is when you won't use that variable again

e.g.:

    for u,_ in G.edges:
        print(u) # code that doesn't use the second node of an edge

networkx/algorithms/centrality/group.py
```diff
@@ -0,0 +1,132 @@
...
+       Group Centrality Maximization via Network Design.
+       SIAM International Conference on Data Mining, SDM 2018, 126–134.
+       https://sites.cs.ucsb.edu/~arlei/pubs/sdm18.pdf
+       
+    """
+    betweenness = 0  # initialize betweenness to 0
+    V = set(G.nodes()) # set of nodes in G
+    C = set(C) # set of nodes in C (group)
+    V_C = set(V - C) # set of nodes in V but not in C
+    pairs = [_ for _ in combinations(V_C, 2)] # (s, t) pairs of V_C
```
dschult: Don't need to define pairs... just put combinations into the loop below:

    for pair in combinations(V_C, 2):

networkx/algorithms/centrality/group.py
```diff
@@ -0,0 +1,132 @@
...
+    # accumulation
+    for pair in pairs:
+        try:
+            paths = [_ for _ in nx.all_shortest_paths(G, source=pair[0],
```
dschult: Can avoid ```[_ for _ in generator]``` in favor of ```list(generator)```.

networkx/algorithms/centrality/group.py
```diff
@@ -0,0 +1,132 @@
...
+                                        target=pair[1], weight=weight)]
+        except (nx.exception.NetworkXNoPath, nx.exception.NodeNotFound):
+            paths = []
+        paths_through_C = []
```
dschult: ```paths_through_C``` could get very big here (as big as ```paths```). You only need to store how many there are. Perhaps there's a way to do that. 

While you are doing that think about ```paths```. I think you don't need a list of all the paths. If you loop over the generator you can handle each path in turn (and count how many you've handled). That should save a lot of memory for large graphs.

networkx/algorithms/centrality/group.py
```diff
@@ -0,0 +1,132 @@
...
+        for _ in paths:
+            if set(_) & C:
+                paths_through_C.append(_)
+        try:
+            betweenness += len(paths_through_C) / len(paths)
+        except ZeroDivisionError:
+            pass
```
dschult: If you ignore a zero division, what value do you return for betweenness? Make it explicit.

networkx/algorithms/centrality/group.py
```diff
@@ -0,0 +1,132 @@
...
+    # rescaling
+    betweenness = _group_rescale(betweenness, len(G), len(C),
```
dschult: Go ahead and put the rescaling code into the same function. That makes it easier to read.

networkx/algorithms/centrality/tests/test_group_betweenness_centrality.py
```diff
@@ -0,0 +1,24 @@
...
+from nose.tools import *
+import networkx as nx
+
+
+class TestGroupBetweennessCentrality:
+    def test_group_betweenness_centrality_1(self):
+        """Group betweenness centrality in P5 for single node group"""
+        G = nx.path_graph(5)
+        C = [1]
+        b = nx.algorithms.centrality.group.group_betweenness_centrality(G, C, weight=None, normalized=False)
```
dschult: You should only need: ```nx.group_betweenness_centrality```
Using it that way tests that the namespace/import code is working properly.

nandahkrishna:
Alright, thank you so much for the suggestions. I'll make the changes and commit them.

networkx/algorithms/centrality/group.py
nandahkrishna: The reason I'm storing all the shortest paths for each (s, t) pair is because I need them to test whether any element of C is part of any of those paths. 

The total number of (s, t) shortest paths which pass through any element of C are counted. This is divided by the total number of (s, t) shortest paths and added to the accumulator. Once we do this for all (s, t) in V - C, we end accumulation and give the group betweenness value.

I've changed paths_through_C to a simple counter variable as that makes it more efficient, but I'm unable to think of a way to remove paths. Any help would be appreciated!

networkx/algorithms/centrality/group.py
nandahkrishna: Done.

networkx/algorithms/centrality/group.py
nandahkrishna: Replaced with list(generator).

networkx/algorithms/centrality/group.py
nandahkrishna: There is no updation done in that case. Have changed this to show updation as value of betweenness += 0.

networkx/algorithms/centrality/group.py
nandahkrishna: Done. The function has been removed.

networkx/algorithms/centrality/tests/test_group_betweenness_centrality.py
nandahkrishna: Done.

networkx/algorithms/centrality/group.py
dschult: Right now, you make ```paths``` into a list and then loop over the entries in ```paths```. You only process one path at a time. So, you don't need to hold all the other paths in memory while you process the one path.  I'm not sure, but I think you could loop with ```for path in nx.all_shortest_paths(...):``` and have the loop check if the path goes through C, and also count how many paths we process (need that to avoid ```len(paths)``` below). But check t he logic -- I might be missing something.

nandahkrishna: Oh okay, I get it. Think I missed the obvious there, and started thinking about how to remove the generator altogether. Thanks!

dschult:
OK... we're close. 

- Did you check the pep8 style? (```pip install pycodestyle``` then either ```pep8 group.py tests/test_group.py``` or ```pycodestyle group.py tests/test_group.py```)
- Could you change some names in the tests... 

First, could you rename the file ```test_group.py```?  
(easier to find if you are working on the group.py file.)

Second, can you make the test names more informative.  For example, the first could be: 
```def test_group_betweenness_single_node(self):```  
In each case, you can change the number to a short phrase. That helps when Travis/Appveyor report an error and helps with readability generally.

nandahkrishna:
> * Did you check the pep8 style? (`pip install pycodestyle` then either `pep8 group.py tests/test_group.py` or `pycodestyle group.py tests/test_group.py`)

Yes, the code style has been checked with pycodestyle, and no errors were reported. Also used the online resource http://pep8online.com to check again.

> * Could you change some names in the tests...

Sure, I'll do this.
Thanks.

dschult:
Thanks very much for this! 
Also -- now you know the process... you can add more functions with a new PR.
:)

nandahkrishna:
Thank you so much for guiding me through this! It's my first contribution and I'm extremely grateful for all your help!

PR #3425, opened 2019-05-07
Fix centrality betweeness subset

michelb7398:
This proposed changeset fixes a bug in the implementation of the `subset centrality betweeness`.

I believe that the existing implementation fails to take into account the case where a node lies on multiple shortest paths, and only accounts for the number of predecessors of those nodes.
This changeset adds a test case that illustrates the problem, along with a proposed fix. I also took the liberty of fixing what apperas to be a mistake in some tests fixtures names.

### DIFF
--- a/networkx/algorithms/centrality/betweenness_subset.py
+++ b/networkx/algorithms/centrality/betweenness_subset.py
@@ -192,15 +192,16 @@ def betweenness_centrality_source(G, normalized=True, weight=None,
 
 
 def _accumulate_subset(betweenness, S, P, sigma, s, targets):
-    delta = dict.fromkeys(S, 0)
-    target_set = set(targets)
+    delta = dict.fromkeys(S, 0.0)
+    target_set = set(targets) - {s}
     while S:
         w = S.pop()
+        if w in target_set:
+            coeff = (delta[w] + 1.0)/ sigma[w]
+        else:
+            coeff = delta[w] / sigma[w]
         for v in P[w]:
-            if w in target_set:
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for This!!

Please help me work through this as it has been a long time since I looked at betweenness and I didn't write the code in the first place. :)
As far as I can tell, you have replaced division by len(P[w]) with multiplication by sigma[v]/sigma[w].
Your description suggests this handles multiple shortest paths (through the same predecessor?) in a way the current code does not. That all seems reasonable and the example you show also makes sense -- though I don't understand why we scale undirected graphs by 0.5 for this subset measure; but that is independent of this PR. :}

Could you comment on why it is correct to remove ```s``` from ```target_set```?
Your update of the test function doc_strings is great.  
Could you make sure the changes still adhere to pep8 standards (probably have to adjust line length in some cases)? 

Thanks again!

michelb7398:
@dschult i added a commit that adresses the pep8 issue you mentionned. Is it okay for you ?

As for the core modification, i believe it is now in accordance with the paper and equations that that are mentionned in `betweenness_subset.py` on lines 23 and following. I also am not entirely sure where the scaling factor comes from. I believe it is a way to ensure that the centrality betweeness is a specific case of the subset centrality betweeness; but i believe there could perhaps be better ways to achieve that, that would also improve interpretability of centrality values between graphs. If time permits, i could tackle that in a next PR.

As for removing `s` from the target set, it doesn't change the results; it is just because iterating on the current node doesn't is superfluous.

dschult:
Thank You!

PR #3435, opened 2019-05-11
Implemented subgraph monomorphism

diogo-cruz:
It is now possible to search for subgraph monomorphisms by using `subgraph_is_monomorphic()` and `subgraph_monomorphisms_iter()` in the same way that `subgraph_is_isomorphic()` and `subgraph_isomorphisms_iter()` is used, respectively.

### DIFF
--- a/networkx/algorithms/isomorphism/isomorphvf2.py
+++ b/networkx/algorithms/isomorphism/isomorphvf2.py
@@ -317,11 +317,11 @@ class GraphMatcher(object):
             yield self.mapping
         else:
             for G1_node, G2_node in self.candidate_pairs_iter():
-                if self.syntactic_feasibility(G1_node, G2_node, mono=mono):
+                if self.syntactic_feasibility(G1_node, G2_node):
                     if self.semantic_feasibility(G1_node, G2_node):
                         # Recursive call, adding the feasible state.
                         newstate = self.state.__class__(self, G1_node, G2_node)
-                        for mapping in self.match(mono=mono):
+                        for mapping in self.match():
                             yield mapping
 
                         # restore data structures
--- a/networkx/algorithms/isomorphism/isomorphvf2.py
+++ b/networkx/algorithms/isomorphism/isomorphvf2.py
@@ -81,8 +81,7 @@ context, 'subgraph' always means a 'node-induced subgraph'. Edge-induced
 subgraph isomorphisms are not directly supported, but one should be
 able to perform the check by making use of nx.line_graph(). For
 subgraphs which are not induced, the term 'monomorphism' is preferred
-over 'isomorphism'. Currently, it is not possible to check for
-monomorphisms.
+over 'isomorphism'.
 
 Let G=(N,E) be a graph with a set of nodes N and set of edges E.
 
@@ -294,7 +293,7 @@ class GraphMatcher(object):
         for mapping in self.match():
             yield mapping
 
-    def match(self):
+    def match(self, mono=False):
         """Extends the isomorphism mapping.
 
         This function is called recursively to determine if a complete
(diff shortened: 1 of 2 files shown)


diogo-cruz:
travis-ci is throwing an error in the function `random_degree_sequence_graph`, which is in no way related to my commits, so I don't know what is causing the issue.

dschult:
The error for the random test is because someone (probably me) merged the test without making it set a seed for the RNG. So occasionally it doesn't pass the test.  It should be investigated further -- but not as part of your PR. 

The codecov test complaint that is showing up now is because many of the lines of code which ```return False``` aren't being tested. Can you add tests that make sure each way to make it fail is tested?  

Thanks for this!

diogo-cruz:
Thanks for the feedback.
I've added some tests for multigraphs. I don't know enough about the subject to write non-trivial tests, unfortunately.

dschult:
Non-trivial tests are probably beyond the scope for this sort of github interface anyway. I'm not looking for anything beyond very simple tests that just test one thing at a time (unit-tests).  For example, if you look at the Codecov report "diff" for this PR: https://codecov.io/gh/networkx/networkx/pull/3435/diff
It shows each line as green if the line was executed as part of the tests. Red means it wasn't executed and Yellow means it is an "if" statement (or other branch) where only one of the two possibilities was encountered in the tests.

The first line that is red is line 371 and that simply means you don't have any tests which cause this criteria for ```subgraph_is_monomorphic``` to return False. The fix is to add a test using this function where the subgraph is not monomorphic. Scanning through the report it looks like lots of criteria that make the routine return False are not tested.  Coming up with tests should amount to finding simple cases where they aren't monomorphic. (that's not always easy-- but it shouldn't be too tricky for most of those cases.

diogo-cruz:
I think I've added the necessary tests. The if's left untested were already so before my changes.

dschult:
Thanks very much!  Is there anything else for the code itself?

I know the docs are not very complete in this module -- they don't follow the usual style of other NetworkX functions. And I don't think we should try to fix that in this PR. But I hate to add code without some basic additions to the docs.

Can you add a paragraph in the module's head doc_string about monomorphisms, what they are and when they would be of interest? Also, for each function with a new argument ```mono```, could you add a line in the docstring stating what it means and how to use it.

Finally, can you check pep8 compliance using ```pycodestyle``` (installable via pip).
Thanks for all this!

diogo-cruz:
Thanks! I actually rewrote the code, since it made more sense for `mono` to be an option for `self.test` instead of being an argument.

I put a small explanation in the doc_string about monomorphisms, and added an explanation of the `self.test` new option where the previous options appeared.

I checked the pep8 compliance, whatever was not following it before my PR is still not following it. They are mostly if statements that, if fixed, would become harder to read. All else should be fixed now.

dschult:
If someone requests a monomorphism_iter, that changes ```self.test```. Can they later request an isomorphism_iter?

diogo-cruz:
Yes, both `isomorphisms_iter` and `subgraph_isomorphisms_iter` change the `self.test` variable to the appropriate string, when they're called.

dschult:
Ahh... I see now -- I was only looking at the diff... not the whole file.   Good...  But now I have another question:

Might someone want to do a graph-graph monomorphism? Or is that just a graph-subgraph monomorphism and we don't have to worry about that showing up some day.
(I just want to make sure "mono" is sufficient rather than using "mono_subgraph" or something.

diogo-cruz:
Yes, it is the same thing, in fact it might also be referred to as graph monomorphism. I'm not aware of any situation where they are different.

dschult:
OK.. I think this is ready to merge. Anything else to add before I do that?

diogo-cruz:
Thanks! No, go ahead, I think everything is here.

PR #3437, opened 2019-05-11
Added group closeness and group degree centralities

nandahkrishna:
Closeness centrality measure implemented for groups. Improvement of module created in PR #3421.
References:
http://www.analytictech.com/borgatti/group_centrality.htm
http://wwwconference.org/proceedings/www2014/companion/p689.pdf
Please do tell me if there is anything I should correct.

### DIFF
--- a/doc/reference/algorithms/centrality.rst
+++ b/doc/reference/algorithms/centrality.rst
@@ -73,6 +73,7 @@ Group Centrality
    :toctree: generated/
 
    group_betweenness_centrality
+   group_closeness_centrality
 
 Load
 ----
--- a/doc/reference/algorithms/centrality.rst
+++ b/doc/reference/algorithms/centrality.rst
@@ -74,6 +74,9 @@ Group Centrality
 
    group_betweenness_centrality
    group_closeness_centrality
+   group_degree_centrality
+   group_in_degree_centrality
+   group_out_degree_centrality
 
 Load
 ----
--- a/networkx/algorithms/centrality/group.py
+++ b/networkx/algorithms/centrality/group.py
@@ -3,10 +3,10 @@
 #    Aric Hagberg <hagberg@lanl.gov>
 #    Dan Schult <dschult@colgate.edu>
 #    Pieter Swart <swart@lanl.gov>
-#    Nanda H Krishna <nanda.harishankar@gmail.com>
 #    All rights reserved.
 #    BSD license.
 #
+# Authors: Nanda H Krishna <nanda.harishankar@gmail.com>
 """Group centrality measures."""
 from itertools import combinations
 
@@ -14,7 +14,7 @@ from itertools import combinations
 import networkx as nx
 
 
-__all__ = ['group_betweenness_centrality']
+__all__ = ['group_betweenness_centrality', 'group_closeness_centrality']
 
 
 def group_betweenness_centrality(G, C, normalized=True, weight=None):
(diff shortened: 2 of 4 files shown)


nandahkrishna:
Added group degree centrality as well, reference - the first link above.

dschult:
Thanks for this!

A number of lines in the doc_strings need to be indented. In particular, after each parameter and return type, and after the references.  Take a look at the other doc_strings for existing functions.

```set(G)``` is perhaps better than ```set(G.nodes())```. should be faster and perhaps easier to read.
```set(V - S)``` is not needed since V and S are already sets:  ```V - S``` is better. 

Do these work if S contains something that is not a node in G? Do they raise a reasonable exception?

nandahkrishna:
> A number of lines in the doc_strings need to be indented.

Ah yes, I will take a look and indent them accordingly.

I will additionally make the changes you have requested. So far I have not checked if the nodes are present in G, so I will add a suitable exception such as NodeNotFound or NetworkXException.

nandahkrishna:
Not sure why, but it seems like a totally unrelated test is failing in the Travis build, for build 5148.5 alone.

Screenshot of logs:
Screenshot 2019-05-22 at 01 32 12

dschult:
The test failure was due to a test (in an unrelated part of the code) which uses random numbers not setting a seed for te random number generator. We have a couple of those -- I am trying to find them and fix them.  So Thank You for finding one and sending the screen shot.

Almost always, restarting the CI test (an option when looking at the CI-travis report or CI-appveyor list of test results for each environment) without any change to the code will generate new random numbers and the trouble will not appear. You effectively made that happen by creating a change to the code which initiated a re-run of all the tests. 

I will open an issue to fix ```random_degree_sequence_graph``` so it sets a seed on the RNG.

nandahkrishna:
> The test failure was due to a test (in an unrelated part of the code) which uses random numbers not > setting a seed for the random number generator.

Ah I see. Glad I could help. Any more changes required in this PR?

dschult:
Now its going to seem like I'm back tracking on what I said earlier. But I'm actually not. :} 
I am hesitant about so much CPU power checking the input for ```NodeNotFound```. I think that without those explicit checks, each function already raises an appropriate exception. Now that you have the tests in place for NodeNotFound, can you try removing the code that checks for nodes in S that aren't in G? That would retain the speed while still making sure we have good exceptions.

It is possible that one or more exceptions is NetworkXError instead of NodeNotFound. If so, change the test to check for NetworkXError. Otherwise you could try to change the exception in the other parts of the code. But I think that should be done in another PR. That will be a rather big change I suspect.

If you want me to do these last picky things, just let me know.  Thanks,

nandahkrishna:
No problem, I'll try these out.

nandahkrishna:
I just happened to check for `group_betweenness_centrality`, if the node isn't in the graph, no exception is raised. This is because the algorithm is designed to work on `set(G) - set(C)` which doesn't check if nodes in C aren't members of G. So I think we need to retain the check for betweenness.

On the other hand, for `group_closeness_centrality()`, which calls `multi_source_dijkstra_path_length()`, `NodeNotFound` is raised for nodes in the group but not in G.

For `group_degree_centrality()`, `NetworkXError` is raised when a node in the group isn't in G.

So I think it's safe to say that I can remove the `raise NodeNotFound` in the degree and closeness functions. Shall I do so in this PR or create a new one?

dschult:
Sounds good. Leave the code in ```group_betweenness_centrality``` and take it out in the others. And do it in this PR. No advantage to waiting.

nandahkrishna:
I think modifying the `neighbors()` function directly would cause many tests to fail, and would involve changing a lot of code. So the best solution would be to catch the NetworkXError and raise a NodeNotFound instead, to avoid meddling too much with the important classes and tests.

dschult:
Can you revert the exception changes. Just let it be a NetworkXError (and check for that in the tests). I've got it on my list to go through the code and make the exception system consistent. I don't want to add any code now that adds to what has to happen then -- especially if it means adding code to check the inputs.  No one will be confused by NetworkXError with a meaningful message.

nandahkrishna:
Alright, I'll revert the change.

nandahkrishna:
Any other changes required? If not, can this be merged? No hurry though, just asking!

PR #3440, opened 2019-05-13
#3439: use multiprocessing

jazzmuesli:
A very early draft, please criticise and suggest something better.


dschult:
This is a potentially important step toward parallel work using NetworkX. Decisions made here can affect a lot of code long term. So it is good to implement for one function and see what issues it brings up. I can imagine this becoming a large support issue if we implement it for many functions. There are so many ways to set up parallel systems: synch vs asynch, one machine vs remote connections, etc. 

I wonder if we could start by making this an example. Then people don't expect it to "just work" in the way they want it to. They expect to have to switch it from Pool.map to Pool.starmap and they have the code easily accessible to take it and change as needed. (people are reluctant to play with code from the package in a way they aren't for example code.)

What do you think? Am I just acting like an old fuddy duddy? Is this type of usage easy enough now?

PR #3444, opened 2019-05-17
Incremental Closeness Centrality (undirected, unweighted graphs)

mikedeltalima:
Feedback is welcome, especially ideas for testing.
Based on http://sariyuce.com/papers/bigdata13.pdf

Example test output:
incremental time: 0.2489505890000001
regular time: 0.29470588600000003
incremental time: 0.008035898000000152
regular time: 0.19043970500000018
incremental time: 0.006714191999999564
regular time: 0.17528513400000012
incremental time: 0.007777294000000268
regular time: 0.17169830299999944
incremental time: 0.007007126000000419
regular time: 0.17890726899999976


networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -0,0 +1,134 @@
...
+    ----------
+    .. [1] Freeman, L.C., 1979. Centrality in networks: I.
+       Conceptual clarification.  Social Networks 1, 215--239.
+       http://www.soc.ucsb.edu/faculty/friedkin/Syllabi/Soc146/Freeman78.PDF
+       [2]Sariyuce, A.E. ; Kaya, K. ; Saule, E. ; Catalyiirek, U.V. Incremental
+       Algorithms for Closeness Centrality
+       http://sariyuce.com/papers/bigdata13.pdf
+    """
+    if G.is_directed():
+        raise NetworkXNotImplemented("not implemented for directed graphs")
```
nandahkrishna: I think it would be better to use `not_implemented_for` from `network.decorators` here, to remove `DiGraph` support. You could check the other functions on how it's used, something like this:
```
from network.decorators import not_implemented_for
...
@not_implemented_for('directed')
def incremental_closeness_centrality(...):
...
```

nandahkrishna: I think it would be better to use `not_implemented_for` from `network.decorators` here, to remove `DiGraph` support. You could check the other functions on how it's used, something like this:
```
from network.decorators import not_implemented_for
...
@not_implemented_for('directed')
def incremental_closeness_centrality(...):
...
```

networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if prev_cc is not None and len(prev_cc) != len(G.nodes()):
+        raise NetworkXError(
```
nandahkrishna: These lines don't seem to be tested. You could have a test case wherein the exception would be raised - check out other tests with such cases (such as `test_group.py` or `test_katz.py` to see how to handle exceptions in tests.

networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -0,0 +1,134 @@
...
+            sp = path_length(G_prime, n)
+            totsp = sum(sp.values())
+            if totsp > 0.0 and len(G_prime) > 1:
+                closeness_centrality[n] = (len(sp) - 1.0) / totsp
+                # normalize to number of nodes-1 in connected part
+                if normalized:
+                    s = (len(sp) - 1.0) / (len(G_prime) - 1)
+                    closeness_centrality[n] *= s
+            else:
+                closeness_centrality[n] = 0.0
```
nandahkrishna: So in some cases, the closeness value is zero. You could have a test case where such a value is set, as the statement doesn't seem to be covered by tests.

nandahkrishna:
I hope my suggestions were helpful!

networkx/algorithms/centrality/incremental_closeness.py
mikedeltalima: thanks!

networkx/algorithms/centrality/incremental_closeness.py
mikedeltalima: 👍

mikedeltalima:
@nandahkrishna could you take a look and make sure your comments are addressed?

nandahkrishna:
The checks pass, I think it looks good. Maybe @dschult could offer more valuable inputs.

dschult:
Hi -- A couple of comments, mostly about documentation, but the main question is where this stands relative to the existing algorithm.

Is this algorithm guaranteed to give the same results as closeness_centrality?
Is it guaranteed to be faster than closeness_centrality?
Are there cases when you would want to use one as opposed to the other?

Can you also briefly discuss whether this should be a new module or be included in closeness.py?

docs:

- The docstring for your new function should start with a single line description. Then there should be a blank line and then a paragraph or two describing what the function does in more detail. The single line is used by Sphinx when creating links to the function in the docs. You should also stick to <80 chars in a line so try to make the description short.  Then you can explain more later.

- you can add much of your first sentence to the start of the second paragraph to make it become the new "first" paragraph that describes what is going on.

- can you add some description of what incremental refers to and when you want/need to use it?

- The top of the module should follow the format described in the github networkx wiki... let me know if you have questions.

- I'm not totally convinced this should be a separate module. It might make sense to add it to closeness.py.  Either way we need to add something to /doc/reference/algorithms/centrality.rst.  Either a new line naming the function, or a new section describing the module.

mikedeltalima:
Thanks @dschult ! Happy to follow the wiki and guidelines for the module. On the other hand, would you rather it be part of closeness.py ?

mikedeltalima:
Is this algorithm guaranteed to give the same results as closeness_centrality? -- Yes
Is it guaranteed to be faster than closeness_centrality? -- No
Are there cases when you would want to use one as opposed to the other? -- Use it when you have an undirected, unweighted graph that you are updating often and that you want to know the closeness centrality for.

networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -114,7 +114,10 @@ def incremental_closeness_centrality(G,
     nodes = G_prime.nodes()
     closeness_centrality = {}
     for n in nodes:
-        if abs(du[n] - dv[n]) <= 1:
+        n_from_u = du.get(n)
+        n_from_v = dv.get(n)
+        if (n_from_u is not None and n_from_v is not None
```
dschult: Perhaps better to use:

    if n in du and n in dv and abs(du[n] - dv[n]) <= 1:

dschult:
After thinking about it more, and based on the idea that the output is the same as for closeness_centrality, could you put it in the module ```closeness.py```? And the tests in the corresponding test module. You will need to add a line to the centrality.rst doc file with the new function's name. I think this will be more clear and better for the package.
Thanks!

networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Notice that higher values of closeness indicate higher centrality.
+
+    Parameters
+    ----------
+    G : graph
+      A NetworkX graph
+    edge: tuple
+      The modified edge (u, v) in the graph.
+    prev_cc: list
```
dschult: I think prev_cc is a dict.

networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -0,0 +1,140 @@
...
+       http://sariyuce.com/papers/bigdata13.pdf
+    """
+
+    if prev_cc is not None:
+        shared_items = set(prev_cc.keys()) & set(G.nodes())
+        count_shared = len(shared_items)
+        if len(prev_cc) != count_shared or len(G.nodes()) != count_shared:
+            raise NetworkXError(
+                "Previous closeness centrality list does not correspond to given\
+        graph.")
```
dschult: Perhaps:

    if prev_cc is not None and set(prev_cc) != set(G):
            raise NetworkXError('prev_cc and G do not have the same nodes')

networkx/algorithms/centrality/incremental_closeness.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Just aliases G
+    G_prime = G
```
dschult: How is ```G_prime``` helpful? Could you just use G throughout?

networkx/algorithms/centrality/incremental_closeness.py
mikedeltalima: True, thanks. I don't see any issues about using mypy; is that in the cards?

networkx/algorithms/centrality/incremental_closeness.py
mikedeltalima: Yes, I'll remove this

networkx/algorithms/centrality/incremental_closeness.py
mikedeltalima: sounds good

mikedeltalima:
@dschult Just checking you'd prefer this to be a separate function and not part of `closeness_centrality` using e.g. a boolean argument and extra kwargs.

dschult:
Yes, I think it should be a separate function so each is clear. We should add each function to the other's "See Also" section in the docstring.

mikedeltalima:
@dschult this is ready for further review when you get a chance

networkx/algorithms/centrality/closeness.py
```diff
@@ -116,16 +119,156 @@ def closeness_centrality(G, u=None, distance=None, wf_improved=True):
...
+    connected part separately.
+
+    This algorithm is applicable to undirected graphs without edge weights.
+
+    References
+    ----------
+    .. [1] Freeman, L.C., 1979. Centrality in networks: I.
+       Conceptual clarification.  Social Networks 1, 215--239.
+       http://www.soc.ucsb.edu/faculty/friedkin/Syllabi/Soc146/Freeman78.PDF
+       [2]Sariyuce, A.E. ; Kaya, K. ; Saule, E. ; Catalyiirek, U.V. Incremental
```
dschult: I think you need the leading dots for the second reference too.

networkx/algorithms/centrality/closeness.py
```diff
@@ -116,16 +119,156 @@ def closeness_centrality(G, u=None, distance=None, wf_improved=True):
...
+
+@not_implemented_for('directed')
+def incremental_closeness_centrality(G,
+                                     edge,
+                                     prev_cc=None,
+                                     insertion=True,
+                                     wf_improved=True):
+    r"""Incremental closeness centrality for nodes.
+
+    Compute closeness centrality for nodes using level-based work filtering
```
dschult: Can you add some words about what incremental is? You've got "edge" as an input and no description of edges anywhere in the description of the function. Actually, it would be good to re-read all the docstrings with an eye of an inexperienced user who doesn't quite know what this whole incremental closeness thing is anyway. Using a phrase like "level-based work filtering" just replaces one jargon (incremental) with another. Try to get the true essence of the measure/method down to a couple sentences that inexperienced users can understand.

networkx/algorithms/centrality/closeness.py
```diff
@@ -116,16 +119,156 @@ def closeness_centrality(G, u=None, distance=None, wf_improved=True):
...
+      If True, scale by the fraction of nodes reachable. This gives the
+      Wasserman and Faust improved formula. For single component graphs
+      it is the same as the original formula.
+
+    Returns
+    -------
+    nodes : dictionary
+      Dictionary of nodes with closeness centrality as the value.
+
+    See Also
```
dschult: Update the see also section

networkx/algorithms/centrality/closeness.py
```diff
@@ -116,16 +119,156 @@ def closeness_centrality(G, u=None, distance=None, wf_improved=True):
...
+    # Leave the graph as we found it
+    if insertion:
+        G.remove_edge(u, v)
+    else:
+        G.add_edge(u, v)
+
+    return closeness_centrality
+
+
+def _node_closeness_centrality(len_G, sp, wf_improved):
```
dschult: Can you please leave this code inline (not as a separate function). It makes it harder to find what code is where and the inline version is pretty compact and readable in each function. I know it is the same in each function, but I don't think it is worth building a separate function.

networkx/algorithms/centrality/tests/test_closeness_centrality.py
```diff
@@ -1,9 +1,10 @@
 """
-Tests for degree centrality.
+Tests for closeness centrality.
 """
 from nose.tools import *
 import networkx as nx
-
+import timeit
+from random import sample
```
dschult: Tests with random parts are difficult to debug. Since these are unit-tests, it is OK to restrict it to a specific sequence of random values. So, use a seed with any call to a function with a random seed argument (like ```fast_gnp_random_graph```). Also, instead of using random.sample it is sufficient to use ```nx.utils.arbitrary_element(G)``` and then no random number is needed in that section.
The tests should not include timing so please remove that code (and the corresponding print statements).
You could leave it in the code as comments to show people how to time it if that is complicated, but in this case it doesn't seem too complicated to time the two approaches.

Thanks...

dschult:
These changes look good. Anything else to do for this PR?

mikedeltalima:
I still need to write a better description in the docstring

mikedeltalima:
@dschult ready to go. Please let me know what you think of the explanation I added. Revisions welcome.

dschult:
It looks like the documentation system had an error processing the lines that include

    Let G' = (V, E ∪ uv). T

I believe the trouble is the unicode symbol for "union" which probably needs to be replaced by a LaTeX command for union (maybe \cup) and the expression should be put into math mode with something like   :math:`G' = (V, E \cup uv)`
the docstring for closeness_centrality uses math mode but in my quick look at that I don't see any "inline" math; only full paragraphs of math equations. The inline syntax is :math:`<stuff>`

dschult:
Now it looks like the minus sign is a unicode character too. (see bottom of error log for first config in travis-ci.)

dschult:
OK... I'm ready to merge this PR.  Anything else before I do that?

mikedeltalima:
Sounds good! Nothing else on my end.

PR #3445, opened 2019-05-17
Faster transitive closure computation for DAGs

alabarre:
This is an improved algorithm for computing the transitive closure of a DAG using topological_sort (so an exception will be raised if the input graph is not a DAG): traverse vertices following a reverse topological order, and connect each vertex to its descendants at distance 2 as we go. The topological order guarantees that each successor v of a node u is connected to all descendants of u in the transitive closure, and therefore it is enough to connect u to nodes at distance 2 from u rather than attempting to explore all descendants.

The algorithm is probably folklore but I didn't find a reference in my books, feel free to add one if you can.

I also added a function to compute the descendants of a node at a fixed distance. `bfs_successors` worked just as well in my tests but resulted in code that was harder to read. Moreover, the new function can be of general interest and is less memory-hungry since each iteration only stores nodes at a fixed distance, whereas `bfs_successors` will keep previous nodes as well.

A taste of the kind of results one can expect using the time function (I used assert to check that both computed closures have the same vertex and edge sets): 

    $ python3 better_transitive_closure_dag.py 
    Testing transitive closure on a random DAG with 800 nodes and 239532 edges
    Running time for NetworkX's TC:             17.484230041503906
    Running time for TC dag-optimized:           3.222315549850464

I'd gladly add my benchmarks but I'm at loss as to where to include them. I shamelessly copied the tests for `transitive_closure` in the meantime.

### DIFF
--- a/doc/reference/linalg.rst
+++ b/doc/reference/linalg.rst
@@ -33,17 +33,6 @@ Bethe Hessian Matrix
 
    bethe_hessian_matrix
 
-Spectrum
----------
-.. automodule:: networkx.linalg.spectrum
-.. autosummary::
-   :toctree: generated/
-
-   laplacian_spectrum
-   bethe_hessian_spectrum
-   adjacency_spectrum
-   modularity_spectrum
-
 Algebraic Connectivity
 ----------------------
 .. automodule:: networkx.linalg.algebraicconnectivity
@@ -73,3 +62,15 @@ Modularity Matrices
 
    modularity_matrix
    directed_modularity_matrix
+
+Spectrum
+---------
+.. automodule:: networkx.linalg.spectrum
+.. autosummary::
+   :toctree: generated/
+
+   adjacency_spectrum
+   laplacian_spectrum
+   bethe_hessian_spectrum
+   normalized_laplacian_spectrum
+   modularity_spectrum
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -693,7 +693,7 @@ def antichains(G):
     .. [1] Free Lattices, by R. Freese, J. Jezek and J. B. Nation,
        AMS, Vol 42, 1995, p. 226.
     """
-    TC = nx.transitive_closure(G)
+    TC = nx.transitive_closure_dag(G)
     antichains_stacks = [([], list(reversed(list(nx.topological_sort(G)))))]
     while antichains_stacks:
         (antichain, stack) = antichains_stacks.pop()
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -564,7 +564,7 @@ def transitive_closure(G):
 
 
 @not_implemented_for('undirected')
-def transitive_closure_dag(G):
+def transitive_closure_dag(G, topo_order=None):
     """ Returns the transitive closure of a directed acyclic graph. This
     function is faster than the function `transitive_closure`, but will fail
     if the graph has a cycle.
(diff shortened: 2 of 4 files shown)


dschult:
Questions about ```descendants_at_distance```:

- It looks like you don't need a deque. I mean that the order of of the nodes is not important so appendleft could just be append with a list or even add with a set.  Is there any other reason to use deque that I am missing? If not, then we should use list or set.
- As noted in and near your TODO, the could be changes to use ```G[vertex]``` instead of ```G.successors(vertex)``` and it would work for directed or undirected graphs. Then it should be put into breadth_first_search.py since it is quite general and perhaps helpful to other routines.

```transitive_closure``` does feel like a misfit in dag.py.  Can you recommend another place for it?

In ```transitive_closure_dag``` you use a list topo_order so you can index it. But it is perhaps better to use ```reversed(topo_order)``` instead of the indexing. That allows you to not require a list, only an iterable. 

The first line of the doc_string should be split from the rest of that first paragraph.  Sphinx uses the first line, followed by a blank line, as a short summary when it creates links to the function in the reference.  So:  one-line, blank line, paragraphs separated by blank lines.

This looks quite good. Thank you.

alabarre:
> It looks like you don't need a deque. I mean that the order of of the nodes is not important so appendleft could just be append with a list or even add with a set.  Is there any other reason to use deque that I am missing? If not, then we should use list or set.


Indeed, I was too focused on the breadth-first search aspect. I've replaced the deque with a set.

> As noted in and near your TODO, the could be changes to use `G[vertex]` instead of `G.successors(vertex)` and it would work for directed or undirected graphs. Then it should be put into breadth_first_search.py since it is quite general and perhaps helpful to other routines.

Fixed, thanks. I also removed the "Notes" section which thereby became obsolete.

> `transitive_closure` does feel like a misfit in dag.py. Can you recommend another place for it?

I see two options:

1. move it to a placeholder for directed graph algorithms; there doesn't seem to be one yet. 
1. move it to a dedicated file for transitive closure algorithms, along with the one for DAG's, which would then also include algorithms for undirected graphs (although in that case I can't think of anything other than replacing every connected component with the complete subgraph on the corresponding vertex set). I don't know if there exist enough improved algorithms for specialised cases to warrant this however.

> In `transitive_closure_dag` you use a list topo_order so you can index it. But it is perhaps better to use `reversed(topo_order)` instead of the indexing. That allows you to not require a list, only an iterable.

I replaced the slice with a call to `reversed`, but I still need to cast `topo_order` as a `list` when invoking `topological_sort`, since it is a generator and `reversed` demands a sequence as input.

> The first line of the doc_string should be split from the rest of that first paragraph. Sphinx uses the first line, followed by a blank line, as a short summary when it creates links to the function in the reference. So: one-line, blank line, paragraphs separated by blank lines.

I didn't know that, I fixed it.

dschult:
Hmmm.. not sure the reversed vs indexing suggestion helped then... and I learned something. :)

Let's leave transitive closure where it is for now. If we get anything else related to it I'd want to refactor into a separate module. 

Can you do one more check using pycodestyle to check for PEP8 compliance. Is there anything else for this PR?

alabarre:
> Hmmm.. not sure the reversed vs indexing suggestion helped then... and I learned something. :)

Well, there are discussions about which approach is faster ... in any case, it is probably more readable as it now stands, and your argument made sense. Let's leave it as is until someone else has a good reason to change it again.

> Can you do one more check using pycodestyle to check for PEP8 compliance. Is there anything else for this PR?

Done, I fixed a minor issue (line too long for one import). There are two other PEP8 issues in `is_aperiodic` (a poorly named variable `l`, and also `gcd` is apparently deprecated) but since I'm not the author I'd rather leave it to him / her to fix it.

I have nothing more to add or change so we can probably close the discussion.

Issue #3446, opened 2019-05-17
Providing a topological order to DAG-related algorithms


alabarre:
A few functions in `dag.py` rely on `topological_sort`. Chances are people who know they are working on a DAG in an application may need to compute a topological order anyway, and it would make sense to allow them to feed that order to functions that need it instead of having each function recompute such an order for their own needs. 

Would you agree to transform `some_dag_function(G, ...)` as follows?

```python
    def some_dag_function(G, ... , order=None):
        if order is None:
            order = topological_sort(G)
        # rest of the function
```

dschult:
Yes, this looks like a good addition.

Would this be worth making into a decorator like ```not_implemented_for```?
That may be more trouble than its worth. I'm fine with whichever way it's implemented.

Is the name ```order``` descriptive enough that it won't be confused with something else?
It seems like ```topolocial_sort_order=None``` is too long. ```top_sort``` seems short enough to be convenient, but still descriptive enough that it won't be confused with another concept. I'm OK with whatever you choose to make the keyword. I'm just asking for you to think about it.

alabarre:
I think I'll go for the simpler option of appending a default parameter. Decorators are probably more elegant but I don't have a broad view of `networkx` and how this might affect or break existing code, so it's probably safer to start that way and let others improve on it if they wish. I'll try and get this done today, it shouldn't take long.

alabarre:
I've created a pull request but I can't seem to figure out how to tag this issue accordingly.

PR #3447, opened 2019-05-19
Additional topo_order parameter for functions that rely on topological_sort

alabarre:
As previously discussed [here](https://github.com/networkx/networkx/issues/3446), I added an optional `topo_order` parameter to functions that rely on `topological_sort`so that users can provide their own topological order if they have already computed one.

I left the following two functions alone:

1. `has_cycle` in `dag.py`: if you call `has_cycle`, then either you haven't computed a topological order yet; or you already have, and in that case you already know the result that `has_cycle` would return.
1. `_relabel_inplace` from `relabel.py`: it did not seem like a sensible choice to me. I'm having trouble picturing a user wanting to provide a topological order there, but maybe it's just because I've never had a need for the function that calls it.

### DIFF
--- a/networkx/algorithms/components/semiconnected.py
+++ b/networkx/algorithms/components/semiconnected.py
@@ -15,7 +15,7 @@ __all__ = ['is_semiconnected']
 
 
 @not_implemented_for('undirected')
-def is_semiconnected(G):
+def is_semiconnected(G, topo_order=None):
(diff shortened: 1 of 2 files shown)


dschult:
This looks ready to merge. Anything else to add here?

alabarre:
No, I think I'm done. Thanks!

Issue #3432, opened 2019-05-09
Weights of betweenness centrality fail for MultiDiGraph


FBenita:
Following up my original question [here](https://stackoverflow.com/questions/56028683/how-to-add-edge-length-as-a-weight-in-betweeness-centrality-using-osmnx-networkx) I noticed that the weight function of betweenness centrality fails for `MultiDiGraph`. 

Although I found it using OSMnx package, this example illustrates the issue:
```
import networkx as nx

graph = nx.Graph()
graph.add_weighted_edges_from([(1,2,1), (2,3,2), (1,3,1)])
print(nx.betweenness_centrality(graph))
# {1: 0.0, 2: 0.0, 3: 0.0}
print(nx.betweenness_centrality(graph, weight="weight"))
# {1: 0.5, 2: 0.0, 3: 0.0}

multi_di_graph = nx.MultiDiGraph()
multi_di_graph.add_weighted_edges_from([(1,2,1), (2,3,2), (1,3,1)])
print(nx.betweenness_centrality(multi_di_graph))
# {1: 0.0, 2: 0.0, 3: 0.0}
print(nx.betweenness_centrality(multi_di_graph, weight="weight"))
# {1: 0.0, 2: 0.0, 3: 0.0}
```
What I did to solve the issue I converted a `MultiDiGraph` `G` into a undirected simple `Graph`   `G1` by using `G1 = nx.Graph(G)`. After that the weight function of betweenness centrality works.

dschult:
I can verify that this is a bug in NetworkX. We should either be ruling out MultiGraphs or we need to adapt the shortest_path routines to work with them. We have it all working in the shortest_path routines, but due to performance issues (from a long time ago -- who knows if they still exist), the betweenness_centrality code uses it's own custom code for djikstra.  And that code doesn't work for MultiGraph or MultiDiGraph. 

We should either 
1) make the specialized code work for MultiGraphs, 
2) remove the specialized code in favor of the more generic code, or 
3) force the user to convert MultiGraphs to Graphs before running betweenness_centrality.

The potential downsides of 1) and 2) are performance issues.
We should make sure this doesn't slow down the non-multi-case too much.

gboeing:
@FBenita as a follow-up, I'm also pushing some commits today to update the OSMnx stats module to convert MultiDiGraphs to DiGraphs before running betweenness_centrality.

PR #3454, opened 2019-05-25
No multigraphs for betweenness

nandahkrishna:
Aims to fix issue #3432 by adding `@not_implemented_for('multigraph')` in `betweenness_centrality`. If the generalised shortest path functions should be used, or if the custom Dijkstra code should be modified rather than not implementing `betweenness_centrality` for multigraph, let me know. Thanks!

### DIFF
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -14,12 +14,14 @@ from itertools import count
 
 import networkx as nx
 from networkx.utils import py_random_state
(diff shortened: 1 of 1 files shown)


dschult:
A quick test for timing the ```_weight_function``` fix used for dijkstra suggests a more than 10% slowdown for graphs if we accommodate multigraphs. I suggest we make ```betweenness_centrality``` ```not_implemented_for``` multigraphs (as this PR does).
So I'm going to merge this.

PR #3462, opened 2019-06-01
Naive k-truss algorithm implementation.

DrPiranoid:
Added algorithm for calculating maximal k-truss to core.py. Added unit tests to test_core.py.

### DIFF
--- a/doc/reference/algorithms/core.rst
+++ b/doc/reference/algorithms/core.rst
@@ -11,3 +11,4 @@ Cores
    k_shell
    k_crust
    k_corona
+   k_truss
--- a/networkx/algorithms/core.py
+++ b/networkx/algorithms/core.py
@@ -355,11 +355,13 @@ def k_corona(G, k, core_number=None):
     return _core_subgraph(G, func, k, core_number)
 
 
+@not_implemented_for('directed')
+@not_implemented_for('multigraph')
 def k_truss(G, k):
-    """Returns the k-truss of G.
+    """Returns the k-truss of `G`.
 
-    The k-truss is the maximal subgraph of G which contains at least three
-    vertices where every edge is incident to at least k triangles.
+    The k-truss is the maximal subgraph of `G` which contains at least three
+    vertices where every edge is incident to at least `k` triangles.
 
     Parameters
     ----------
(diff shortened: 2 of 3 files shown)


dschult:
Thanks very much for this. We also need an addition to the docs so t his will show up in the reference docs. Go to /doc/reference/algorithms/core.rst and add your function to the list of functions there.

There may be some changes to improve performance, but this code looks pretty good. I suspect if we improve in, say, speed, that we will lose in memory footprint or something. One thing to consider is going through the edges one node at a time so you don't have to lookup the neighbors as often. Probably not a big change in speed, but hard to tell without looking. 

    for u in H:
        nbrs_u = set(H[u])
        for v in nbrs_u:
            if len(nbrs_u & H[v]) < k:
                to_drop.append((u, v))

Also, you currently remove each edge twice -- once in each direction. If you do switch to edges one node at a time you can store the seen nodes in a set and don't consider them later in the other direction.

    seen = set([])   # and then inside the loop, change the if to:
    if v not in seen and len(nbrs_u & H[v]) < k:
    # finally add a line in the loop on u
    seen.add(u)

Another idea -- just before the function definition use the decorators 
@not_implemented_for('directed')
@not_implemented_for('multigraph')

See core_number() for an example. I know many other functions in that module don't use this feature, but that's because they are old and this is a new function, so we might as well use the tools we've got.

Ask if you have questions.

DrPiranoid:
You're welcome! 

I'll try the suggestions you listed for performance and see what line_profiler gives. I thought that the EdgeView only output (u, v) and not (v, u) for simple graphs. So looping over H.edges wouldn't show the same edge twice? Is that not what happens?

I'll definitely use the appropriate decorators!

dschult:
You are correct that H.edges only outputs (u, v) and not (v, u) for the Graph class. The "same edge twice" problem arises if you switch to "one node at a time" processing of the edges.  I was wrong to say your current code does that.  I'll change the comment.

dschult:
I'd like to go ahead and merge this. We can update to tweak performance if needed later in another PR.

Are you OK with that?

DrPiranoid:
I was able to test it out and splitting the set creation of the u and v neighbors led to some significant savings. I pushed it just now.

Feel free to merge away. Thanks for the feedback!

PR #3465, opened 2019-06-05
Wheels are python 3 only

cHYzZQo:
version 2.3 dropped python2 support until py2 is actually deprecated we
shouldn't mark these wheels as universal


harph:
Any update on this?

dschult:
@jarrodmillman is this a simple "yes"? Any other implications?

jarrodmillman:
Seems OK to me.

Issue #3475, opened 2019-06-07
networkx.draw() hides axes on different subplot


przecze:
I try to have a graph visualization on one subplot and a line plot of some values on another subplot. However, after I draw the graph on the first subplot, axes on the other subplot disappear (although I did not allow any shared axes). And I cannot bring them back.
```
import networkx as nx
import matplotlib.pyplot as plt
fig = plt.figure()
ax1 = fig.add_subplot(211)
ax2 = fig.add_subplot(212)
G = nx.graph_atlas(1)
nx.draw(G, ax=ax1)
plt.show()
```
Results in `ax2` having the ticks for both axes hidden, whereas if I run the same code without `nx.draw(...)` line the ticks are visible on both subplots.

How is it possible that `nx.draw(G, ax=ax1)` affects `ax2`?

dschult:
Looks like #3476 fixes this issue.

PR #3476, opened 2019-06-07
Don't remove ticks of other pyplot axes

ImportanceOfBeingErnest:
Fix for issue raised in https://stackoverflow.com/questions/56496486/axes-disappear-after-drawing-networkx-graph-on-another-subplot

`plt.tick_params` operates on the currently active pyplot axes, which is not necessarily the one where the networkx graph resides in. 

This PR makes sure the axes `ax` is operated on. (`ax.tick_params`)


dschult:
Excellent! -- Thanks for tracking this down and putting it into a PR.
I believe this fixes #3475

PR #3477, opened 2019-06-09
Add method to clear edges only

plammens:
Add method `clear_edges` to `Graph` and its subclasses that clears all edges without altering nodes or graph attributes.


dschult:
Could ```G.clear_edges()``` be done with ```H = nx.Graph(G)```?  I guess it doesn't copy the graph attributes so if that matters you would need to add ```H.graph.update(G.graph)```. 

Can you briefly describe a use-case for ```G.clear_edges()```? I'm reluctant to add another method to the base class without due diligence in demonstrating usefulness.
Thanks!

plammens:
I don't see the advantage of making a new graph without the edges with `H = nx.Graph(G)` and returning that. Why make a new Graph object, with all the overhead it entails? Besides, as you said, you would also need to add `H.graph.update(G.graph)`.  And this is an operation that is meant to be performed in-place, exactly like `Graph.clear`. By making a new graph object, operations relying on the object `id` would fail.

plammens:
As for the use cases, here is what I used this functionality for, although I'm sure there are better examples. I was dealing with a "dynamic" geometric graph; i.e., a geometric graph whose "maximum distance" could be changed on demand. The node data (e.g. its geographical position) could change over time too, so every time a request to construct the graph with new maximum distance `d`  was set I would need to clear the edges and then add the edges between nodes that are less than `d`  units of distance apart.

```python
    def construct_graph(self, dist: float):
        """
        Construct geometric graph with a new distance
        :param dist: new distance, in meters
        """
        if dist < 0:
            raise ValueError("distance should be non-negative")
        self.clear_edges() 
        self._construct_graph(dist)  # internal methods to do the actual work...
```

For the time being I used a [workaround](https://github.com/plammens/BicingBot/blob/411a26efa7a4894eeafaf238850f13cb3f62b689/data.py#L110), but that is subpar in comparison to a simply implemented `clear_edges` method (with no need to iterate over each edge).

PR #3484, opened 2019-06-20
Fix dfs_preorder_nodes docstring saying "edges" instead of "nodes"

### LINKED ISSUE #3483: dfs_preorder_nodes docstring says "edges" instead of "nodes"
https://github.com/networkx/networkx/blob/09c3da707d12aec42f3c0b2f91b60007bb436d37/networkx/algorithms/traversal/depth_first_search.py#L284

tbitai:
Fixes #3483


nandahkrishna:
Could you also check the other functions in `depth_first_search.py`? I believe a couple of other functions have the same error in their docstrings.

PR #3485, opened 2019-06-21
Fix for betweenness_subset

### LINKED ISSUE #3481: Apparent bug in `betweenness_centrality_subset`
See this question:

https://stackoverflow.com/questions/56569070/is-this-betweenness-calculation-correct

I believe the issue is that the `_rescale_e` function requires a factor of 0.5 if `G` is not directed, and that this got copied over into the `_rescale` function and wasn't fixed.

The result is that betweenness centrality appears to be off by a factor of 0.5.

nandahkrishna:
Aims to fix #3481 by removing unnecessary rescale factor.


nandahkrishna:
Does the issue affect edge betweenness subset as well?

dschult:
I'm closing this in favor of the documentation changes described in #3841
It can be reopened if needed.

PR #3488, opened 2019-06-23
Added optional center_spring feature to spring_layout for unconnected graphs

joseortiz3:
This addresses https://github.com/networkx/networkx/issues/3325

Previously, spring_layout would not converge for unconnected graphs since each connected piece (component) would fly away to infinity. This optional feature (default = False, i.e. not used) fixes this.

With center_spring=True:
Figure_1

With center_spring=False (current behavior):
Figure_2

This also enables "petri dish" plots (a name I just invented..?) like the following:
montecarlo_orange_vs_historical_corr_blue_with_SIC_labels_and_colors

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -501,6 +501,9 @@ def _fruchterman_reingold(A, k=None, pos=None, fixed=None, iterations=50,
         msg = "fruchterman_reingold() takes an adjacency matrix as input"
         raise nx.NetworkXError(msg)
 
+    # Normalize A so only k controls equilibrium distance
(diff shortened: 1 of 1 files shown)


joseortiz3:
Still needs a new test. I've never made a test before but it seems like I just add a new method to https://github.com/networkx/networkx/blob/master/networkx/drawing/tests/test_layout.py that hits the added code.

joseortiz3:
Whoops, found some numerical stability issues, I'll update later.

joseortiz3:
pending discussion of https://github.com/networkx/networkx/issues/3489

PR #3499, opened 2019-07-04
make bfs_edges exhaustive (fixes #3498)

### LINKED ISSUE #3498: breadth-first search is not exhaustive
I believe it is an error that `bfs_edges` does not yield **all** edges reachable from the source node, but only one edge per reachable target node.

This is true for the [PADS blueprint](http://www.ics.uci.edu/~eppstein/PADS/BFS.py), with the difference that this fact is clearly stated in the docstring there:

> Edges that connect vertices within the same level are not included in the output.

Minimal example:
```Python
import networkx as nx

graph = nx.DiGraph()
graph.add_nodes_from(range(9))
graph.add_edge(0, 1)
graph.add_edge(0, 2)
graph.add_edge(1, 2)
graph.add_edge(2, 3)
graph.add_edge(3, 4)
graph.add_edge(4, 5)
graph.add_edge(5, 6)
graph.add_edge(6, 7)
graph.add_edge(6, 8)
graph.add_edge(7, 8)

edges = nx.bfs_edges(graph, 0)
assert set(edges) == set(graph.edges)
```

bertsky:


bertsky:
Where are those test cases defined?

dschult:
The tests that are failing are in the docstring of bfs_successors.
But I think you should spend time researching the definition of bfs_edges before doing more coding.
It seems many webpages indicate that edges are not traversed in a BFS unless they reach a new node.
Perhaps a differently named function is needed?

bertsky:
> The tests that are failing are in the docstring of bfs_successors.

Oh, I see. Very elegant!

> It seems many webpages indicate that edges are not traversed in a BFS unless they reach a new node.

Many descriptions focus on the problem of _cyclic graphs_, in which case of course never visiting a node twice ensures termination (but so does my solution). However, everything I read so far can also be interpreted in my favour: The traversal itself being BFS, if you want to get the reachable _nodes_ (or a tree), then you would have to ignore edges with the same destination, but if you want to get the _edges_, then there is no reason to do so. 

> Perhaps a differently named function is needed?

Well, you already called the function <i>bfs_<b>edges</b></i> (instead of, say, _bfs_nodes_), so I think this is not a real option. 

But how about my proposal to just add a kwarg, say `exhaustive=False`?

dschult:
Maybe the ```edgebfs``` function provides what you are looking for.
I can't quite tell...
https://networkx.github.io/documentation/latest/reference/algorithms/traversal.html

bertsky:
Closing – not necessary, for #3498 is but a documentation issue.

PR #3503, opened 2019-07-09
Full join operation and cograph generator

efraimrodrigues:
A cograph is a graph containing no path on four vertices. Cographs or $P_4$-free graphs can be obtained from a single vertex by disjoint union and complementation operations.

The generator starts of from a single vertex and performes disjoint union and full join operations on itself at each iteration. The decision on which operation will take place is random.

Full join is the union of G and H in which all edges between G and H are added. The full join operation of two graphs G and H is the same as getting their complement, performing a disjoint union, and finally getting the complement of the resulting graph.

This PR adds the cograph generator and the full join operation.

Reference on cographs: https://www.sciencedirect.com/science/article/pii/0166218X81900135?via%3Dihub

### DIFF
--- a/networkx/algorithms/operators/binary.py
+++ b/networkx/algorithms/operators/binary.py
@@ -360,7 +360,7 @@ def full_join(G, H, rename=(None, None), name=None):
     -----
     It is recommended that G and H be either both directed or both undirected.
 
-    The full join union of a graph G and H is the same as getting
+    The full join operation of graphs G and H is the same as getting
     their complement, performing a disjoint union, and finally getting
     the complement of the resulting graph.
 
--- a/networkx/algorithms/operators/binary.py
+++ b/networkx/algorithms/operators/binary.py
@@ -376,17 +376,26 @@ def full_join(G, H, rename=(None, None)):
     """
     R = union(G, H, rename)
 
-    def add_prefix(node, prefix):
+    def add_prefix(graph, prefix):
         if prefix is None:
-            return node
-        return prefix + repr(node)
+            return graph
+
+        def label(x):
+            if is_string_like(x):
+                name = prefix + x
+            else:
+                name = prefix + repr(x)
+            return name
+        return nx.relabel_nodes(graph, label)
+    G = add_prefix(G, rename[0])
+    H = add_prefix(H, rename[1])
 
     for i in G:
         for j in H:
-            R.add_edge(add_prefix(i, rename[0]), add_prefix(j, rename[1]))
+            R.add_edge(i, j)
     if R.is_directed():
         for i in H:
             for j in G:
-                R.add_edge(add_prefix(i, rename[1]), add_prefix(j, rename[0]))
+                R.add_edge(i, j)
 
     return R
--- a/networkx/algorithms/operators/binary.py
+++ b/networkx/algorithms/operators/binary.py
@@ -360,6 +360,10 @@ def full_join(G, H, rename=(None, None), name=None):
     -----
     It is recommended that G and H be either both directed or both undirected.
 
+    If G is directed, then edges from G to H are added as well as from H to G.
+
+    Note that full_join() does not produce parallel edges for MultiGraphs.
+
     The full join operation of graphs G and H is the same as getting
     their complement, performing a disjoint union, and finally getting
     the complement of the resulting graph.
@@ -371,13 +375,17 @@ def full_join(G, H, rename=(None, None), name=None):
     See Also
     --------
     union
+    disjoint_union
     """
     R = union(G, H, rename, name)
-    nodes = list(R.nodes())
+    nodes = list(R)
     for i in nodes:
-        for j in H.nodes():
-            R.add_edge(i, j)
-
-    R.add_edges_from(H.edges())
+        for j in H:
+            if i != j:
+                if not R.is_directed():
+                    R.add_edge(i, j)
+                else:
+                    R.add_edge(i, j)
+                    R.add_edge(j, i)
 
     return R
--- a/networkx/algorithms/operators/binary.py
+++ b/networkx/algorithms/operators/binary.py
@@ -13,7 +13,7 @@ __author__ = """\n""".join(['Aric Hagberg <aric.hagberg@gmail.com>',
                             'Pieter Swart (swart@lanl.gov)',
                             'Dan Schult(dschult@colgate.edu)'])
 __all__ = ['union', 'compose', 'disjoint_union', 'intersection',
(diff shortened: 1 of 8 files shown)


dschult:
Thank for this cograph stuff! :) . This looks quite good.

Can you make the module header match the description in the NX Github Wiki... The copyright and author recognition and stuff like that.
Make sure the docsrtrings have a single line to start the docstring. Then a blank line and then a paragraph or two describing in more detail what is being done. Your ```full_join``` is great; the ```cograph``` needs work. Also "starts of from" in cograph docs should be "starts off from".

You should replace ```len(list(R.nodes()))``` with ```len(R)``` (and similar with G instead of R).

Shouldn't you test more than the number of nodes? e.g. maybe the case n=2 with two possible answers? I haven't thought about this much.

You'll need to add to the docs to make this show up in the reference.  Look at /doc/reference/ and adjust or add the appropriate ```.rst``` files for these new functions.

Finally, do the usual pycodestyle (PEP8) check to style issues.

efraimrodrigues:
Thanks for the review. 

I exchanged the name of the method `cograph` for `random_cograph` as there may be a nonrandom generator in the future. Can you be more specific on what needs work in `cograph`? There's also a new test which I think might lead to a cograph recognition functionality for networkx.

networkx/generators/cographs.py
```diff
@@ -0,0 +1,66 @@
...
+    """Returns a random cograph with $2 ^ n$ nodes.
+
+    Parameters
+    ----------
+    n : int
+            The order of the cograph.
+    seed : integer, random_state, or None (default)
+            Indicator of random number generation state.
+            See :ref:`Randomness<randomness>`.
+
```
dschult: Add a ```Returns``` block to the docstring
You might also want to use a paragraph before the ```parameters``` section to describe what a cograph is and how the random cograph is created.

networkx/generators/cographs.py
```diff
@@ -0,0 +1,66 @@
...
+    Notes
+    -----
+    This generator starts off from a single vertex and performes disjoint
+    union and full join operations on itself.
+    The decision on which operation will take place is random.
+    """
+    R = nx.empty_graph(1)
+
+    for i in range(n):
```
dschult: Perhaps this code is simpler?

    for i in range(n):
        RR = nx.relabel_nodes(R.copy(), lambda x: x+len(R))
        if seed.randint(0, 1) == 0:    # or you could use if seed.rand() < 0.5
            R = nx.full_join(R, RR)
        else:
            R = nx.disjoint_union(R, RR)
    return R

networkx/generators/cographs.py
efraimrodrigues: `Returns` block added. 

Copied information from the module docstring to the function docstring. Removed `Notes` since the information was inserted in the docstring as a paragraph.

networkx/generators/cographs.py
efraimrodrigues: Thank you for this.

efraimrodrigues:
Simplified the random_cograph function and improved the docstring.

networkx/algorithms/operators/binary.py
```diff
@@ -329,3 +329,55 @@ def compose(G, H):
...
+
+    Graph, edge, and node attributes are propagated from G and H
+    to the union graph.  If a graph attribute is present in both
+    G and H the value from H is used.
+
+    See Also
+    --------
+    union
+    """
+    R = union(G, H, rename, name)
```
dschult: This full_join operation looks like it adds selfloops for nodes of H (but not selfloops for nodes of G). 
Also, I thought the ```union``` operation would add edges from H into R. Do we need to add the edges from H below?  You can use ```list(R)``` instead of ```list(R.nodes())```. Finally, does this assume undirected graphs? Do we need to handle directed graphs (or multigraphs) specially? You can use ```not_implemented_for``` decorators to restrict usage to undirected and to rule out multiedges.

efraimrodrigues: Fixed the possibility of adding selfloops and removed the redundant addition of edges from H into R.

efraimrodrigues:
It makes sense to handle directed graphs. In fact, it now addresses them by adding edges from G to H and from H to G. I added a paragraph in the `Notes` to make this clear. 

I don't think we need to restrict the function from being used on multigraphs, so I added a paragraph in the `Notes` to make it clear that the function doesn't add parallel edges for multigraphs. However if necessary, I can use the `not_implemented_for` to restrict it from being used on multigraphs.

networkx/algorithms/operators/binary.py
dschult: But it adds edges between every two nodes in H...  And why do you need to compute nodes?
Would this work?   

    R = nx.union(G, H, rename, name)
    for i in G:
        for j in H:
            R.add_edge(i, j)
    if R.is_directed():
        for i in H:
            for j in G:
                R.add_edge(j, i)

Actually, don't we have to worry about "rename" and adding a prefix to each node name?
And maybe we should just not include the "name" argument to the union command. People don't use that very often and they can easily do it by hand if they want it. We're moving away from keeping "name" as a special attribute.

efraimrodrigues: Yes. This works.

efraimrodrigues:
@dschult thank you for your suggestions.

Added prefix to the nodes name and removed the `name` argument. 

New tests were added to make sure the `full_join` function doesn't add nodes or more edges than the necessary.

networkx/algorithms/operators/binary.py
```diff
@@ -329,3 +329,64 @@ def compose(G, H):
...
+    G and H the value from H is used.
+
+    See Also
+    --------
+    union
+    disjoint_union
+    """
+    R = union(G, H, rename)
+
+    def add_prefix(node, prefix):
```
dschult: This ```add_prefix``` almost works. But ```repr``` adds quotes around strings. So if the node is string-like we need to avoid using repr. I think you can basically copy the code used for ```union``` on lines 63-75 or so. It creates new graphs G and H which are renamed versions of G and H (it is not changing the inputted graphs).
Then you can use symbols G and H as if they were the original nodes: i.e. ```R.add_edge(i, j)``` works.

efraimrodrigues: Alright!

efraimrodrigues:
Fixed issue with renaming string-like nodes.

dschult:
I think this is ready...  Anything else from your perspective?

efraimrodrigues:
Not really.

Thank you for reviewing this @dschult .

PR #3507, opened 2019-07-14
Fix inverse_line_graph.

hutchison:
Tests included.

### DIFF
--- a/networkx/generators/line.py
+++ b/networkx/generators/line.py
@@ -402,7 +402,7 @@ def _find_partition(G, starting_cell):
             new_cell = [u] + list(G_partition.neighbors(u))
             for u in new_cell:
                 for v in new_cell:
-                    if (u != v) and (v not in G.neighbors(u)):
+                    if (u != v) and (v not in G_partition.neighbors(u)):
                         msg = "G is not a line graph" \
                               "(partition cell not a complete subgraph)"
                         raise nx.NetworkXError(msg)
--- a/networkx/generators/line.py
+++ b/networkx/generators/line.py
@@ -276,6 +276,16 @@ def inverse_line_graph(G):
     -----
     This is an implementation of the Roussopoulos algorithm.
 
+    If G consists of multiple components, then the algorithm doesn't work.
+    You should invert every component seperately:
+
+    >>> from networkx import connected_components
+    >>> root_graphs = []
+    >>> for comp in connected_components(G):
+            root_graphs.append(inverse_line_graph(G.subgraph(comp)))
+
+    Later on you can unite these using ``networkx.union``.
+
     References
     ----------
     * Roussopolous, N, "A max {m, n} algorithm for determining the graph H from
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for this!
Which test showed you that the current function has a bug?
I think the tests for line.py are in tests/test_line.py
Can you move your new tests in there?

hutchison:
> Which test showed you that the current function has a bug?

`inverse_line_graph` had a bug. It produced a wrong root graph for "K_4 minus e".

> Can you move your new tests in there?

Sure, will do.

hutchison:
Should I open up another pull request?

dschult:
No new pull request needed. If you commit to your local and then push to your github copy of networkx it will update the PR on the networkx Github automatically.

Thanks!

hutchison:
Is there anything else that I can/have to do, before this gets merged?

dschult:
Thanks for the gentle bump. I was hoping to get time to check through that code and understand where/why the bug arose in the first place. But I don't think that is going to happen. So...

Can you change the line that you already changed from ```v not in G_partition.neighbors(u)``` to
```v not in G_partition[u]```?  Lookups on a dict are much better than lookups on an iterator that is created from the dict anyway... :} .  That's one reason I wanted to check through that code. It seems non-optimal and this might have other bugs...

Thanks very much!

hutchison:
Done. In the run I replaced all calls to `neighbors()`.

hutchison:
There's also another issue: `inverse_line_graph()` doesn't construct a root graph for isolated nodes or more generally a graph which is composed of components. We should loop over all components and construct a root graph by calling `inverse_line_graph()` for every component. But I have no idea how costly that would be.

dschult:
Perhaps the multiple component version of inverse_line_graph should be included as a Notes section in the docstring telling people (in very general but clear terms) how to do it themselves.

I'm not sure how it would be done exactly... call inverse_line_graph on each component separately?

hutchison:
> Perhaps the multiple component version of inverse_line_graph should be included as a Notes section in the docstring telling people (in very general but clear terms) how to do it themselves.

Okay, I'll try my best.

> I'm not sure how it would be done exactly... call inverse_line_graph on each component separately?

Yes, exactly. In the current implementation `inverse_line_graph` just chooses an arbritrary edge `e` and starts from there. This way only the component containing `e` gets inverted and the other components are being ignored.

Furthermore: currently an isolated node won't be inverted, but that's wrong, because L(K_2) = K_1. In that case `invert_line_graph` should just return a K_2, but how should the nodes be named? `networkx.utils.misc.generate_unique_node()` seems a bit ugly. Is there a better way?

dschult:
Ack!  I hope I didn't mess this up completely.
Did you have more to change/address with this commit?
(maybe the K_1/K_2 case?)

hutchison:
Thanks!

> Did you have more to change/address with this commit?

I was awaiting your thoughts on this. But I'll just write a quick patch to fix this edge case.

dschult:
Great!   Thanks!

PR #3508, opened 2019-07-14
Fix A* returning wrong solution

### LINKED ISSUE #3464: A* (A star) function astar_path returns a wrong answer
Suppose we have a graph below:
astarexample
Find shortest path from N5 to N0.
Numbers in bracket alongside each node are heuristic values estimating distance to n0. (This is a bad example of heuristic function I constructed to use in my classroom.)
The best path from n5 to n0 is:  
n5→n4→n3→n2→n1→n0
What networkx.algorithms.shortest_paths.astar.astar_path returns: 
[n5, n1, n0] 
which is apparently not correct.

matejklemen:
Fixes #3464.

The problem was that each node was allowed to be explored at most once, but sometimes, there are multiple paths through the same node, where the first one is not necessarily the best.  
I fixed it by removing the checks for whether a node was already explored and adding logic to prevent the overriding of starting node's parent. The latter is required because when returning a path, that is done by backtracking until the node's parent is `None`. Without the check, the parent of starting node could be overridden and the backtracking would never end.

The added test contains a smaller version of the graph, mentioned in issue #3464.

### DIFF
--- a/networkx/algorithms/shortest_paths/astar.py
+++ b/networkx/algorithms/shortest_paths/astar.py
@@ -109,14 +109,13 @@ def astar_path(G, source, target, heuristic=None, weight='weight'):
             path.reverse()
             return path
 
-        if curnode in explored:
+        # Do not override the parent of starting node
+        if curnode in explored and explored[curnode] is None:
             continue
 
         explored[curnode] = parent
 
         for neighbor, w in G[curnode].items():
-            if neighbor in explored:
-                continue
             ncost = dist + w.get(weight, 1)
             if neighbor in enqueued:
                 qcost, h = enqueued[neighbor]
--- a/networkx/algorithms/shortest_paths/astar.py
+++ b/networkx/algorithms/shortest_paths/astar.py
@@ -109,9 +109,15 @@ def astar_path(G, source, target, heuristic=None, weight='weight'):
             path.reverse()
             return path
 
-        # Do not override the parent of starting node
-        if curnode in explored and explored[curnode] is None:
-            continue
+        if curnode in explored:
+            # Do not override the parent of starting node
+            if explored[curnode] is None:
+                continue
+
+            # Skip bad paths that were enqueued before finding a better one
+            qcost, h = enqueued[curnode]
+            if qcost < dist:
+                continue
 
         explored[curnode] = parent
 
--- a/networkx/algorithms/shortest_paths/tests/test_astar.py
+++ b/networkx/algorithms/shortest_paths/tests/test_astar.py
@@ -87,6 +87,20 @@ class TestAStar:
         assert_equal(nx.astar_path(XG4, 0, 2), [0, 1, 2])
         assert_equal(nx.astar_path_length(XG4, 0, 2), 4)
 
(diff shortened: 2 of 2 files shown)


matejklemen:
Interesting, the test that is failing @ CI passes locally :thinking: Will investigate.

efraimrodrigues:
> Interesting, the test that is failing @ CI passes locally  Will investigate.

It only happens with Python 3.7.0 upon x32 arch.

dschult:
If the test has random numbers generated it may fail on some architectures and not others just because of the random numbers chosen. To recreate the problem perhaps run it locally with a seed set for the random number generator. Find a seed that creates the error and stick with that seed for the duration of your debugging.    tricky....

matejklemen:
I think I found (and fixed) the cause of that test failing. If a bad path through some node was found before a good path through that same node and if the bad path actually happened to get expanded, the parent node would be incorrectly overridden.
I've added a test for this as well, although I'm pretty sure this was working correctly before this PR.

Another thing I found out was that the randomized test, which checks if A* and Dijkstra produce the same result on the same graph could fail without the solution being wrong. The test does not take into account situations where there are multiple optimal solutions, possibly even of different length.
Maybe changing the test to check the equality between path costs instead could help with this.

dschult:
Good work!  Thanks!

One question about your last comment... how can multiple optimal solutions have different lengths? Wouldn't the shortest one be optimal?

matejklemen:
Here's an example of a case, where 2 optimal paths (of different length) exist. Numbers on the edges are edge weights, bolded blue numbers next to nodes are heuristic values. Consider the case of finding path from `a` to `d`.    
A* finds `a->c->d`, while Dijkstra finds `a->b->c->d`, both paths have a cost of 1.35.
multiple_optimal_paths  
(Sorry about the float weights and heuristics, this was autogenerated and then rounded to 2 decimals)

dschult:
Both of those paths have the same length. So they are both optimal.
In what sense do they have different lengths?

matejklemen:
Sorry, perhaps i used the wrong term, what I meant was that the number of hops on the optimal paths is different. The A* path goes through 3 nodes and the Dijkstra path goes through 4 nodes.

Therefore the `assert_equal(...)` line in the randomized test will fail (lists representing the two paths are not of same length).

dschult:
OK... sounds good.  In cases like that we should allow either path as a correct solution.  Right?

matejklemen:
I think so, yes

matejklemen:
@dschult Just making sure, should I be fixing that test or is that a topic that needs to be further discussed/ is for a separate PR?

dschult:
Could you change the test to check that the answer is one of the two optimal paths?
The paths are short enough that you could define each and then check that the answer is one of the two.  Something like:

    path1 = [ ... ]
    path2 = [ ... ]
    assert_in(nx.astar_path(G, source, target), (path1, path2))

Then commit and push to the same branch and it will update the PR directly.
Thanks for asking... sorry for not being more clear.

matejklemen:
Done - I changed the randomized test into a deterministic one which checks that the returned path is one of the optimal ones.

PR #3514, opened 2019-07-17
Restore checking PyPy3

thedrow:
I don't know why it was removed but PyPy3 now supports Python 3.5 and hopefully the next major PyPy version will support PyPy 3.6.

It's really important for me to know that this library is portable since I'm going to use it quite a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: true
-    - python: pypy3-7.0
+    - python: pypy3.5-7.0
       dist: xenial
       sudo: true
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,9 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: true
+    - python: pypy3-7.0
+      dist: xenial
+      sudo: true
 
 before_install:
   # prepare the system to install prerequisites or dependencies
--- a/networkx/algorithms/bipartite/projection.py
+++ b/networkx/algorithms/bipartite/projection.py
@@ -144,7 +144,7 @@ def weighted_projected_graph(B, nodes, ratio=False):
 
     ratio: Bool (default=False)
         If True, edge weight is the ratio between actual shared neighbors
-        and maximum possible shared neighbors (i.e., the size of the other 
+        and maximum possible shared neighbors (i.e., the size of the other
         node set). If False, edges weight is the number of shared neighbors.
 
     Returns
@@ -255,7 +255,7 @@ def collaboration_weighted_projected_graph(B, nodes):
     >>> G = bipartite.collaboration_weighted_projected_graph(B, [0, 2, 4, 5])
     >>> list(G)
     [0, 2, 4, 5]
-    >>> for edge in G.edges(data=True): print(edge)
+    >>> for edge in sorted(G.edges(data=True)): print(edge)
     ...
     (0, 2, {'weight': 0.5})
     (0, 5, {'weight': 0.5})
@@ -454,7 +454,7 @@ def generic_weighted_projected_graph(B, nodes, weight_function=None):
     >>> # Add some arbitrary weight to the edges
     >>> for i,(u,v) in enumerate(B.edges()):
     ...     B.edges[u, v]['weight'] = i + 1
-    ... 
+    ...
     >>> for edge in B.edges(data=True):
     ...     print(edge)
     ...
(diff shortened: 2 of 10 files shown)


dschult:
It looks like most of the CI errors are due to ordering of the results differing between pypy and what we are "expecting". So, you should update the tests to allow for either ordering.
Thanks!

thedrow:
So either order is fine? It's just the values that matter?
In that case we can convert to sets.

thedrow:
I fixed sorting issues whenever it seems fit.
We're left with some failures which I'm not sure if they are sorting problems or actual bugs.
One failure in `networkx.algorithms.tests.test_planar_drawing.test_invalid_half_edge` appears to be a genuine bug.

Any assistance is welcome.

dschult:
It looks like the embedding used in ```networkx.algorithms.tests.test_planar_drawing.test_invalid_half_edge``` causes an invalid half edge in one ordering of the nodes and a KeyError in another. When I change the order of the dict set in the test as the embedding_data:

    embedding_data = {1: [2, 3], 2: [1, 3], 3: [1, 2], 4: [1, 2, 3]}
    embedding_data = {4: [1, 2, 3], 2: [1, 3], 3: [1, 2], 1: [2, 3]}

I get the KeyError with both pypy3 and regular python.  So it is a bug -- but a bug in the test rather than a bug in the code.  I'm still working on a correction that would not depend on order. Somehow an invalid half edge that does not also lead to a KeyError with a different order.

Any assistance is welcome... :}

@thedrow you can go on and fix up the other errors if you like, or work on this one.

dschult:
OK... found a workaround test:

    embedding_data = {1: [2, 3]}

Then there is no trouble with the order of dict reporting for ```networkx.algorithms.tests.test_planar_drawing.test_invalid_half_edge```

thedrow:
I got the following error when I used your `embedding_data`:

```
======================================================================
FAIL: networkx.algorithms.tests.test_planar_drawing.test_invalid_half_edge
----------------------------------------------------------------------
  File "/home/omer/.local/share/virtualenvs/networkx/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/omer/.local/share/virtualenvs/networkx/site-packages/nose/tools/nontrivial.py", line 67, in newfunc
    raise AssertionError(message)
AssertionError: test_invalid_half_edge() did not raise NetworkXException
    'test_invalid_half_edge() did not raise NetworkXException' = "%s() did not raise %s" % ('test_invalid_half_edge', 'NetworkXException')
>>  raise AssertionError('test_invalid_half_edge() did not raise NetworkXException')
```

thedrow:
Does the order matter in `networkx.algorithms.tests.test_clique.TestCliques.test_cliques_containing_node`?

I got `[[2, 4, 6], [1, 2, 3, 6]] != [[2, 6, 1, 3], [2, 6, 4]]`.
Should the `[1, 2, 3, 6]` list remain in the same order or it doesn't really matter?
If it doesn't, why are you returning a list and not a set?

thedrow:
I fixed another test.
I'm not sure about the rest.

dschult:
Sorry about the false solution... I'm still looking for how to fix invalid_half_edge.
I will look at the others. I think most are "simply" testing for edge tuples. Order matters for tuples, but not for undirected edges, so the representation is sometimes tricky to test.

For cliques we return lists instead of sets probably because that code was written before sets were available in base python.   ;}   I'll take a look to make sure that one is OK too.

thedrow:
Can we replace them with sets?
It's a breaking change so we can bump the major version and forget about it.

Alternatively we should open an issue about it.

dschult:
The problem for test_planar_drawing is that the input used is not a valid embedding. It only has a directed half of the edges connecting node 4. The original order did raise an exception as expected but also introduced self-loops. I think we should change the test to check a valid embedding for an invalid planar graph (i.e. a non-planar graph).

    embedding_data = {1: [2, 3, 4], 2: [1, 3, 4], 3: [1, 2, 4], 4: [1, 2, 3]} 

I'll think about adding more tests for invalid embeddings in issue #3518 
Meanwhile, I'll turn now to looking at the other errors where the order of the 2-tuple representing an edge is causing trouble.

thedrow:
Wonderful.
So we're good to merge this?

dschult:
OK... This seems to pass all tests now. Are there other things needed?

thedrow:
Nope.
Go ahead and merge this.

PR #3515, opened 2019-07-17
Adding non-randomness measures for graphs

erwanlemerrer:
Hello all,

implementation of the non-randomness measures from:
Xiaowei Ying and Xintao Wu, On Randomness Measures for Social Networks,
SIAM International Conference on Data Mining. 2009

The idea of the measures is to quantify how to which extent the inspected graph is carrying communities (first measure) or far from a ER random graph (second one), through a simple float each time.


dschult:
Thanks for this! I've got a number of comments below.

Module/file header: take a look at the networkx github wiki for our standard header/copyright/author recognition. Not all our files have it but we are working toward that.

Check the PEP8 code style using pycodestyle (used to be called pep8). Most changes will be spaces after commas, length of lines, indenting and stuff like that.

In your import of numpy you call your function hits_numpy. :)  You should probably also rearrange that 2 line exception to make the first line define the error string and the second raise the exception:

    msg = "non_randomness requires NumPy: http://scipy.org/"
    raise ImportError(msg)

Code ```if not k:``` should be ```if k is not None:```
Call of ```label_propagation_communities``` can be shortened to ```nx.community.label_propagation_communities```

Define the value for Eqn 4.5 on its own line. Then put that into the return statement.
Can you please come up with two names for these two measures of non_randomness?
Hopefully the paper named them, but even if it didn't, try to find names that will convey the spirit of the measure and/or how it is computed.

:)

erwanlemerrer:
Thanks Dan, hope I fixed all regarding your comments. Pushed again.

dschult:
Thanks -- looks good.

Can you add the NetworkX people to the copyright?

Can you check Pycodestyle again (spaces after commas and between operators)

erwanlemerrer:
yes indeed, pycodestyle was not returning errors since those typos were in the comment section.
Added full copyright, sorry for that!

dschult:
Don't lines 78 and 81 in non_randomness also get flagged by pycodestyle?  They have lots of operations without spaces around them.

erwanlemerrer:
Fixed. No was not flagging weirdly.. (pip 19.1.1 & pycodestyle 2.4.0 under ubuntu)

dschult:
Thanks!

PR #3523, opened 2019-07-24
Prufer with linear complexity

Pascal-Ortiz:
This pull request solves the issue #3521. The new implementation has linear complexity and his based on the paper [Linear-time Algorithms for Encoding Trees as Sequences of Node Labels](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.70.5052&rank=1). 

Time execution has been tested. Some results for a tree with 50,000 vertices:

| Functions | Previous implementation | New implementation |
| :---:         | :---      | :---      |
| `to_prufer_sequence`  |  44.30s    |   0.29s       |
| `from_prufer_sequence`  |   83.65s    |   0.21s       |

PR #3526, opened 2019-07-24
Added safeguards (input graph G) for non-randomness measures 

erwanlemerrer:
Hello,
was not included in previous commits, but original paper:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.215.8394&rep=rep1&type=pdf
lists the requirements for input graph: binary, symmetric, connected, and without self-loops.
Added in this pull request.


dschult:
Looks like you built this pull request on top of another pull request. You should create a new branch for each pull request. In this case it doesn't matter much because most of those commits are already in the networkx repository. So, don't worry about it this time. In the future see some of the comments at https://github.com/networkx/networkx/pull/3523#issuecomment-514621366

This looks good.  Anything else before committing?

erwanlemerrer:
I see thanks for the explanations. Nothing to add (hoping this time it is final!)

PR #3527, opened 2019-07-25
Implement minimum weight full matching of bipartite graphs

### LINKED ISSUE #3520: Minimum weight perfect bipartite matching
I'm wondering if there would be appetite for an algorithm to solve the problem of finding minimum weight perfect bipartite matchings in bipartite graphs: Let *G* = ((*U*, *V*), *E*) be a weighted bipartite graph. We then want to find a perfect matching, i.e. one of cardinality min(|*U*|, |*V*|), so that the sum over all weights in the matching is minimal. This is also known as the [linear assignment problem](https://en.wikipedia.org/wiki/Assignment_problem).

There are algorithms for solving several related problems in NetworkX already, but I didn't seem to be able to find one specialized to this problem. Some related algorithms are the following:

* The linear assignment problem is a special case of a minimum cost flow problem, obtained by adding a source incident to all vertices in *U* and a sink incident to all vertices in *V*, and letting the demand for these two vertices be min(|*U*|, |*V*|) with appropriate signs. This could then be solved using `networkx.algorithms.flow.min_cost_flow`. A specialized algorithm would take advantage of the bipartite structure to improve performance. Moreover, as far as I can tell, `min_cost_flow` assumes that all weights are integral, which we wouldn't have to here.
* On the other hand, `networkx.algorithms.bipartite.maximum_matching` can detect if a feasible solution exists, i.e. a matching with the desired cardinality, but it doesn't take into account the weights.
* `networkx.algorithms.matching.max_weight_matching` takes weights into account but operates on arbitrary graphs and does not require that the matching be perfect in the bipartite case.

A large number of different algorithms exist to solve the problem. Assuming that the inclusion of an algorithm is relevant, the first thing I'm wondering is the following: SciPy already has an efficient implementation of an algorithm to solve the problem in [`scipy.optimize.linear_sum_assignment`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html), and since NetworkX already depends on SciPy, it would be only a few lines of code to simply make use of the SciPy implementation. However, as NetworkX is generally focused on bundling Python implementations of its algorithms, simply making use of the SciPy implementation might be going against the spirit of the library, and it would be preferable to include a Python version of one of the algorithms instead?

fuglede:
This provides an implementation of the minimum weight full matching problem, also known as the linear assignment problem, for complete bipartite graphs by deferring the calculation of the assignment to `scipy.optimize.linear_sum_assignment`.

This closes #3520; note that there, I referred to the matchings as "perfect"; since we also support bipartite graphs in which the two partitions have different cardinalities, "perfect" becomes a bit of a misnomer, so here we use term "full" instead; this follows the terminology used in Karp's *An algorithm to Solve the m x n Assignment Problem in Expected Time O(mn log n)* (doi:10.1002/net.3230100205).

I've marked the PR as "WIP" mainly because this is my first contribution, so I'm sure it has some rough edges. All suggestions for improvements are taken positively. One thing I can see myself:

* As mentioned in the code comments, there's an improvement coming to SciPy 1.4.0 that simplifies the solution here. NetworkX currently requires only 1.1.0; is there a way to ensure that the constraint is removed once 1.4.0 becomes an extra requirement? One could add a check on the SciPy version number but that seems hacky and unrobust.
* I added myself as a contributor. I don't know what sort of precedence exists in terms of the size of contribution, but I'll remove it again if this is too awkward.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -128,6 +128,7 @@ is partially historical, and now, mostly arbitrary.
 - Weisheng Si, Github: `ws4u <https://github.com/ws4u>`_
 - Haakon H. Rød, Gitlab: `haakonhr <https://gitlab.com/haakonhr>`_, `<https://haakonhr.gitlab.io>`_ 
 - Efraim Rodrigues, `GitHub <https://github.com/efraimrodrigues>`_, `LinkedIn <https://linkedin.com/in/efraim-rodrigues/>`_
+- Søren Fuglede Jørgensen, GitHub: `fuglede <https://github.com/fuglede>`_
 
 Support
 -------
--- a/networkx/algorithms/bipartite/matching.py
+++ b/networkx/algorithms/bipartite/matching.py
@@ -1,6 +1,7 @@
 # matching.py - bipartite graph maximum matching algorithms
 #
-# Copyright 2015 Jeffrey Finkelstein <jeffrey.finkelstein@gmail.com>.
+# Copyright 2015 Jeffrey Finkelstein <jeffrey.finkelstein@gmail.com>,
+# Copyright 2019 Søren Fuglede Jørgensen
 #
 # This file is part of NetworkX.
 #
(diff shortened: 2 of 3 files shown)


dschult:
Thanks for this!
It looks pretty good and I don't even see any PEP8 (pycodestyle) things to change though you should check if you haven't. 
I think the Notes section should not start with a blank line. 

The biggest addition I suggest is to include some of the text from your PR description in the docstring -- perhaps under the Notes section, but maybe in the description just before the parameters section. In particular I find the description of "full" vs "perfect" and the citation to be really helpful.

I'm not sure what the best way to handle Scipy 1.1 vs 1.4 is. @jarrodmillman do you have any thoughts?

fuglede:
Thanks for the comments. I checked pep8/flake8, so those should be good.

Regarding the blank line in the Notes section, I did have a look around to see what was common, as it seemed a bit inconsistent with the other sections. It's not quite consistent throughout NetworkX, so I opted to just use the same style as in the rest of `bipartite.matching`. So, let me just go ahead and remove the blank line from the other functions in here as well.

Let me expand on the description with some of the notes from here; having it before the parameters seems logical to me, as that's a good excuse to expand on the problem in general.

PR #3534, opened 2019-08-01
Add spiral layout for graph drawing

michaelfedell:
New method in layout will arrange nodes of a graph in a spiral layout
with variable compactness. Nodes may be positioned with equal angles
(increasing separation further from center), or equidistant (decreasing
angle further from center)

### DIFF
--- a/doc/reference/drawing.rst
+++ b/doc/reference/drawing.rst
@@ -87,4 +87,5 @@ Graph Layout
    shell_layout
    spring_layout
    spectral_layout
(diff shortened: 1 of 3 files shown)


dschult:
Looks like you need to add ```spiral_layout``` to the ```__all__``` variable at the top so the ```__init__``` file can do its import magic and get the function imported to the bigger networkx package.

chrismattmann:
This LGTM! 👍

PR #3535, opened 2019-08-04
Linear prufer coding

### LINKED ISSUE #3521: Prufer schemes: quadratic instead of linear complexity
A tree with set vertices `range(n)` can be encoded by a sequence of `n - 2` integer in `range(n)` known as the tree Prufer sequence and conservely, any such sequence can be decoded into a unique tree. 

The NetworkX module `/networkx/algorithms/tree/coding.py` implements the encoding and decoding schemes by mean of resp. the `to_prufer_sequence` function and `from_prufer_sequence` function. The docstring for each function references a 2009 paper describing linear time algorithms (and not n*log(n) as the docstring claims cf. §2 and §3 in the paper). 

The problem is that performances of the implemented algorithms are unsatisfactory and the time complexity seems to be more quadratic than linear. It results for instance that decoding a Prufer sequence of lenght about 20000 requires 32s.

On the other hand, based on another paper, I implemented Prufer encoding and decoding functions and the previous Prufer sequence need only 0.01s to be decoded: the NetworkX implementation has a time execution problem.

To be more accurate, I have tested the NetworkX decoding function successively on 

- data input of size 2000
- on data input of size 10 x 2000=20000

and time execution raises from 0.32s seconds to about 31.83s: this is a strong indication that the algorithm has quadratic complexity. Here is the test code :

```python
from time import clock
from random import randrange
import networkx as nx

def test_speed(n):
    sequence=[randrange(3) for i in range(n-2)] 
    d=clock()
    G=nx.from_prufer_sequence(sequence)
    print("%d: %.2fs" %(n, clock()-d))    
    
test_speed(2000)
test_speed(20000) 
```

```
2000: 0.32s
20000: 31.83s
```

I suspect that the following part of the source code has not linear complexity, cf. the last line:

```python
# from NetworkX sources
def from_prufer_sequence(sequence):

    # ...    
    # code snipped
    # ...
 
    for v in sequence:
        # ...
        # code snipped
        # ...   
        # the min function executes a loop
        # so we have 2 nested loops of size n (sometimes)
        
        index = u = min(k for k in range(index + 1, n) if degree[k] == 1)
```

The `to_prufer_sequence` function has more or less the same problem.

jj314:
resolves #3521.

I suggest these changes, as the already implemented paper's solution is fine and only a simple fix in networkx' current implementation is required. In contrast, #3523 implements another paper's solution.

### DIFF
--- a/networkx/algorithms/tree/coding.py
+++ b/networkx/algorithms/tree/coding.py
@@ -303,7 +303,7 @@ def to_prufer_sequence(T):
     def parents(u):
         return next(v for v in T[u] if degree[v] > 1)
 
-    index = u = min(k for k in range(n) if degree[k] == 1)
+    index = u = next(k for k in range(n) if degree[k] == 1)
     result = []
     for i in range(n - 2):
         v = parents(u)
@@ -312,7 +312,7 @@ def to_prufer_sequence(T):
         if v < index and degree[v] == 1:
             u = v
         else:
-            index = u = min(k for k in range(index + 1, n) if degree[k] == 1)
+            index = u = next(k for k in range(index + 1, n) if degree[k] == 1)
     return result
 
 
@@ -387,7 +387,7 @@ def from_prufer_sequence(sequence):
     # tree. After the loop, there should be exactly two nodes that are
     # not in this set.
     not_orphaned = set()
-    index = u = min(k for k in range(n) if degree[k] == 1)
+    index = u = next(k for k in range(n) if degree[k] == 1)
     for v in sequence:
         T.add_edge(u, v)
         not_orphaned.add(u)
@@ -395,7 +395,7 @@ def from_prufer_sequence(sequence):
         if v < index and degree[v] == 1:
             u = v
         else:
-            index = u = min(k for k in range(index + 1, n) if degree[k] == 1)
+            index = u = next(k for k in range(index + 1, n) if degree[k] == 1)
     # At this point, there must be exactly two orphaned nodes; join them.
     orphans = set(T) - not_orphaned
     u, v = orphans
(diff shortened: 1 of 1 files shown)


networkx/algorithms/tree/coding.py
```diff
@@ -312,7 +312,7 @@ def parents(u):
         if v < index and degree[v] == 1:
             u = v
         else:
-            index = u = min(k for k in range(index + 1, n) if degree[k] == 1)
```
jj314: The result of this line is always the smallest k satisfying degree[k] == 1, but min doesn't know this and considers others, even after the first satisfying k has been found.

networkx/algorithms/tree/coding.py
```diff
@@ -312,7 +312,7 @@ def parents(u):
...
+            index = u = next(k for k in range(index + 1, n) if degree[k] == 1)
```
jj314: next instead correctly chooses the first satisfying k, avoiding unnecessarily scanning the remainder of the range. Since the index which is shifted forwards, defines the start of the next range, the entire range from 0 to index is only scanned once. Therefore the running time is linear.

networkx/algorithms/tree/coding.py
```diff
@@ -387,15 +387,15 @@ def from_prufer_sequence(sequence):
     # tree. After the loop, there should be exactly two nodes that are
     # not in this set.
     not_orphaned = set()
-    index = u = min(k for k in range(n) if degree[k] == 1)
+    index = u = next(k for k in range(n) if degree[k] == 1)
```
jj314: Same issue as in the encoding part.

networkx/algorithms/tree/coding.py
```diff
@@ -387,15 +387,15 @@ def from_prufer_sequence(sequence):
...
     for v in sequence:
         T.add_edge(u, v)
         not_orphaned.add(u)
         degree[v] -= 1
         if v < index and degree[v] == 1:
             u = v
         else:
-            index = u = min(k for k in range(index + 1, n) if degree[k] == 1)
+            index = u = next(k for k in range(index + 1, n) if degree[k] == 1)
```
jj314: Same problem as in the encoding part.

Issue #3167, opened 2018-09-24
read_graphml fails to open graph generated by write_graphml


abhinavagarwalla:
I get the following error when reading a graph generated by write_graphml, using read_graphml:
**networkx.exception.NetworkXError: Bad GraphML data: no key d10**

On debugging, I found out an issue with write_graphml. 
Below is a snippet from the code that can reproduce the error. If the datatype of _penalty_attr_ attribute passed to the add_edge function is different for different edges, write_graphml creates multiple data keys (in my example, _d10_) in graphml output file. There is no additional key information in the _<key>_ tag which only has description for keys upto _d9_, but _<edge>_ tag contains the extra key _d10_. This results in the key error while parsing graphml file using _read_graphml_.

```
penalty_attr = ''
if random() < 0.5:
    penalty_attr = 1.0  #datatype: double

G.add_edge(from_node,
           to_node,
           penalty = py)
```

Please let me know if I need to rephrase or elaborate on the issue.

dschult:
I can verify that edge attribute types with the same name but different type cause ```write_graphml``` to create two differently named edge attributes -- one to hold edge attributes of each type. Furthermore, the second attribute does not get an entry in the graphml file describing it in a <key> element. That may be why ```read_graphml``` is giving the ```no key d10``` error.

Anyone have suggestions on how to make the writer add the edge attribute to the list of <key> elements? The edge attribute has the same name, but different type.

iamjli:
We ran into this issue as well when setting node attributes using a dictionary generated from a pandas dataframe. 

I'd imagine scenario comes up not infrequently: If there are missing values in the dataframe, this can result in features with mixed types which breaks `nx.write_graphml`.

dschult:
@iamjli you correctly mention that networkx should handle the missing values (or any other case where the same attribute name includes values of different type).

Do you have any suggestions for what the "right" way to handle missing values is?

dschult:
Should we make ```write_graphml``` raise an error if the type of an edge attribute is not consistent?
That would force the user to go back and change the attributes. Kind of a pain, but no surprises that way.  Is that the "right" way to handle missing values?

iamjli:
@dschult That's certainly an acceptable way to handle missing values, but you're right, it's a bit of a pain. 

[...]

PR #3536, opened 2019-08-04
Write mixed-type attributes correctly in write_graphml_lxml

alexhenrie:
If an attribute has different types on different nodes, it needs a separate key for each type. This has always worked fine when calling `write_graphml_xml`, but `write_graphml_lxml` wouldn't write all of the necessary `<key>` elements, which resulted in an invalid GraphML file. (I have added a test case that you can try out to see what I mean.) This patch fixes it so that the two functions produce the same output.

### DIFF
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -644,36 +644,30 @@ class GraphMLWriterLxml(GraphMLWriter):
             element_type = self.xml_type[self.attr_type(k, "graph", v)]
             self.get_key(make_str(k), element_type, "graph", None)
         # Nodes and data
-        attributes = {}
         for node, d in G.nodes(data=True):
             for k, v in d.items():
                 self.attribute_types[(make_str(k), "node")].add(type(v))
-                if k not in attributes:
-                    attributes[k] = v
-        for k, v in attributes.items():
-            T = self.xml_type[self.attr_type(k, "node", v)]
-            self.get_key(make_str(k), T, "node", node_default.get(k))
+        for node, d in G.nodes(data=True):
+            for k, v in d.items():
+                T = self.xml_type[self.attr_type(k, "node", v)]
+                self.get_key(make_str(k), T, "node", node_default.get(k))
         # Edges and data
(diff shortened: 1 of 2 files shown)


alexhenrie:
Fixes #3167

alexhenrie:
@dschult Could you take a look at this?

dschult:
It looks like the new code is creating more children than expected. can you take a simple example and create graphml files with both old and new code to see the difference?

I like the idea of this PR.  Thanks!

alexhenrie:
@dschult Thanks for looking into this. I have not been able to reproduce the problem you described. Could you please post an example program?

dschult:
The example is the one that is failing in Travis-CI tests. You can see which tests are failing by clicking on "Details" next to the Travis-CI tests (preceded by a red X) near the bottom of the Pull Request page on github. The Details show you all the various configurations we automatically test. Any rows that are red have an error. For now, don't bother with the first row which also builds the documentation. Click on, say, the 2nd row and search for the word "fail". The report shows which test failed and t hen a little later gives the error message for the test which you can then find the line number and look up in your code on your computer. You can run the tests yourself using nosetests (install with pip install nose). Then cd to the readwrite directory and run ```nosetests --with-doctest graphml.py tests/test_graphml.py```

The error reported from the test indicates that the expected result gives 3 children while the new version gives 5 children.

alexhenrie:
Got it. My apologies, for some reason I thought that was a pre-existing failure. I just fixed it.

dschult:
nice -- what was the fix? I can't see any difference...
Also, anything else for this PR? It looks ready to me...

alexhenrie:
I needed to move the key creation to after all of the nodes had been searched, instead of creating keys after processing each node. So both the inner and outer loops had to be repeated, and I was originally only repeating the inner loop. There is probably a more efficient way to do this but it would require rewriting quite a bit more of the code in this file and right now the most important thing is just to get the bug fixed.

Anyway, yes, this PR is ready to be merged :-)

alexhenrie:
Thank you!

PR #3538, opened 2019-08-08
Fixes some edge cases for inverse_line_graph().

hutchison:
As previously discussed.

The edge cases are L(K_1) = K_0 and L(K_2) = K_1. Now `inverse_line_graph` correctly inverts K_0 and K_1.


dschult:
Thanks!  
I see what you mean about ```generate_unique_node```.  Maybe there is a better naming convention... Like finding the first 2 integers that aren't already nodes?
Can you explain the deletion of tests/test_line.py?

hutchison:
> Like finding the first 2 integers that aren't already nodes?

I was also thinking about that and it leads to a bunch of „difficulties.“ Let's say one wants to invert a bunch of isolated nodes and does so by inverting them separately on every component. Then one would get a bunch of P₂s with every edge being named `(1, 2)`. Afterwards one wants to unite all these P₂s into the root graph. This will break because of name collisions and one has to rename all nodes by themself.

One possibility would be to create tuples, e.g. `0` → `(0, 0), (0, 1)`. What do you think?

> Can you explain the deletion of tests/test_line.py?

There are already a lot of tests in `generators/tests/test_line.py` which I overlooked at first. In the last commit I merged my tests and the existing tests.

dschult:
I definitely see the problem. If your suggestion about  0 → (0, 0), (0, 1) works let's use it. If it's a pain to implement or there are other issues, then we can keep the uuid method...

hutchison:
It was easy to implement. Let's keep the uuid method for the empty graph, because there's no safe way to infer the name of the node.

PR #3540, opened 2019-08-09
When exporting to GML, write non 32-bit numbers as strings.

tobiasblass:
According to the GML technical report, integers in GML must fit into a
signed 32-bit integer. Values outside this range have to be written as
strings instead.

### DIFF
--- a/networkx/readwrite/gml.py
+++ b/networkx/readwrite/gml.py
@@ -667,6 +667,9 @@ def generate_gml(G, stringizer=None):
                     yield indent + key + ' 1'
                 elif value is False:
                     yield indent + key + ' 0'
+                # GML only supports signed 32-bit integers
+                elif value < -2**31 or value >= 2**31:
+                    yield indent + key + ' "' + str(value) + '"'
                 else:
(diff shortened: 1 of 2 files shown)


dschult:
Looks like a good idea.
I think the < operator won't work if ```value is None```.
Maybe we need a test to check that this works and a fix if it doesn't.

tobiasblass:
> I think the < operator won't work if `value is None`.
> Maybe we need a test to check that this works and a fix if it doesn't.

You're right (I didn't notice this because it only occurs in the `literal_stringizer` case). However, I think there is a different problem: `literal_stringizer` translates None to 'None', which is also invalid GML, isn't it? Shouldn't None be translated to '"None"' (i.e., within quotes)?

tobiasblass:
Looking at the problem in more detail, I noticed that `literal_stringizer` does not have to output valid GML.  It is intended to be used as the `stringizer`-argument for `generate_gml`, and the output of this stringizer is put into quotes anyway.
This means that we don't need a range check in `literal_stringizer`

dschult:
Ahh...  good... thank you!!

PR #3545, opened 2019-08-15
Fixes bug in networkx.algorithms.community.label_propagation.asyn_lpa_communities

filmenczer:
The old convergence test was inconsistent with the reference paper. It would stop only if each node has a single majority label among its neighbors. In a network like below, the bridge node will always have `best_labels` = 2 and the stopping condition will never be met.

ex

This prevented convergence in many cases, including the karate_club_graph. 

Instead, according to the cited reference, a node should be satisfied if it has a majority label (either blue or red in the above example), even if there are multiple tied majority labels (two in the case above).

The algorithm is now consistent with the cited reference and is guaranteed to converge.

### DIFF
--- a/networkx/algorithms/community/label_propagation.py
+++ b/networkx/algorithms/community/label_propagation.py
@@ -91,11 +91,13 @@ def asyn_lpa_communities(G, weight=None, seed=None):
             max_freq = max(label_freq.values())
             best_labels = [label for label, freq in label_freq.items()
                            if freq == max_freq]
-            new_label = seed.choice(best_labels)
-            labels[node] = new_label
-            # Continue until all nodes have a label that is better than other
-            # neighbour labels (only one label has max_freq for each node).
-            cont = cont or len(best_labels) > 1
+            
+            # Continue until all nodes have a majority label
+            if labels[node] in best_labels:
+                cont = cont or False
+            else:
+                labels[node] = seed.choice(best_labels)
+                cont = cont or True
 
     # TODO In Python 3.3 or later, this should be `yield from ...`.
     return iter(groups(labels).values())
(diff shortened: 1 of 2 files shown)


dschult:
I think I understand your suggestion for improvement. I don't understand your code though.
```cont = cont or False``` is the same as ```cont = cont```. Similarly ```cont = cont or True```  is the same as ```cont = True```.  Can you check your logic?

**Comment by filmenczer** [comment]
> I think I understand your suggestion for improvement. I don't understand your code though.
> `cont = cont or False` is the same as `cont = cont`. Similarly `cont = cont or True` is the same as `cont = True`. Can you check your logic?

I had used the same logic as in the previous code but you are absolutely right: `cont = cont or False` is not needed and can be removed; `cont = cont or True` should be replaced by `cont = True`. Thank you @dschult!

**Comment by dschult** [comment]
Good --  could you also add your example as a test (that simply tests that it converges).
The tests are in ```tests/test_label_propagation.py```
Thanks!

**Comment by filmenczer** [comment]
> Good -- could you also add your example as a test (that simply tests that it converges).

Sorry I have zero experience... Do you just mean somethink like this?
```
def test_termination():
    test1 = nx.nx.karate_club_graph()
    test2 = nx.caveman_graph(2, 10)
    test2.add_edges_from([(0, 20), (20, 10)])
    asyn_lpa_communities(test1)
    asyn_lpa_communities(test2)
```
Both of those examples do not terminate but I am not sure it is appropriate to have a test causing an infinite loop with the old code?

dschult:
Yes, that is what I am intending. 

I think testing a loop that is supposed to stop is OK even if previous versions of the code didn't stop. If the new test gets applied to the old code the test will run forever and Travis-CI will time out. I don't see any other way to test that an infinite loop occurred when it shouldn't have.
Thanks!

filmenczer:
Thank you @dschult, I added the convergence tests.

PR #3551, opened 2019-08-20
add Directed Joint Degree Graph generator

tbalint:
Hi,

I would like to contribute the directed_joint_degree_graph function (similar to the undirected joint_degree_graph function). This graph generator is based my peer-reviewed publication with my co-authors: 
B. Tillman, A. Markopoulou, C.T Butts, M.Gjoka, “[Construction of Directed 2K Graphs](https://dl.acm.org/citation.cfm?id=3097983.3098119)”, in Proc. of KDD 2017. [arxiv version](https://arxiv.org/abs/1703.07340).

The main function, directed_joint_degree_graph, generates random graphs with input in/out degree sequences and directed joint degree correlation. I have included tests for checking inputs and the main function. I have followed the guidelines and other graph generators for my code. 

Please let me know if you have any comments or see any issue!

Best,
Balint


dschult:
Thanks for this!
Can you refactor your code to add the functions to the existing ```joint_degree_seq.py``` file, and to put the tests in the ```tests/test_joint_degree_seq.py``` file?

I haven't looked deeply at the code yet. In general, I like to make a new function as you have done, but is the code sufficiently similar to the undirected version to make them a single function with keyword argument ```directed```? (probably not).

tbalint:
I've done the refactoring and also changed the tests to use `assert_true` or `assert_false` instead of simple `assert`. 

Unfortunately, I'd say that a simple keyword argument version to switch between undirected and directed would be too complicated to implement. The basic ideas for graph generation are similar, but there are some crucial differences e.g, to avoid self-loops in directed graphs, to handle the directed edges correctly.

dschult:
That sounds reasonable and the refactoring looks good too. Just a couple more things to work through.

- For Sphinx to get the formatting of the docs right, we need to start with a single-line <80 char description of the function. Follow that with a blank line and then a longer paragraph (or two) that describe in more detail. Sometimes it is a pain to find a short description of the function, but it usually works out better anyway, so sphinx enforces that format.  The short description gets listed along with the function name anywhere Sphinx creates a link for the function.

- PEP8 is a python code style set of guidelines. You can check a module or directory of modules using ```pycodestyle joint_degree_seq.py``` and you can install pycodestyle using ```pip install pycodestyle```.  The usual trouble-spots are 80 character line limits and spacing between operators. Check the test file as well as the module itself.

- We need a pointer in the function in the docs reference. So, find the ```doc/reference/generators.rst``` file and in it the section for joint_degree_sequences and add the two new function names.

Let me know if you have questions.

tbalint:
I've updated the docs, is there a simple way to preview it locally? I'm not familiar with Sphinx.

Let me know if it looks good!

dschult:
Thanks! You can make Sphinx work locally, but it involves more work than a simple pip install.
I usually don't examine the docs locally unless it is a doc-related PR. The first Travis-CI test runs sphinx, so you can check for errors/warnings there. The resulting docs don't actually get deployed unless the branch is the master branch. So we do a little check of that before releases. Not ideal, but it's what we've got. :}

tbalint:
Thanks! If anything comes up, feel free to reach out to me!

PR #3554, opened 2019-08-21
Try to bug Fix #3552

### LINKED ISSUE #3552: draw_networkx_nodes: Error doesn't show my mistakes exactly.
Hi, 

I passed nodelist to draw_networkx_nodes method as it needs.
But if I set nodelist as ndarray in numpy, it shows the following error.

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The nodelist I passed is like as follows.

```
a = np.array([1, 2, 3, 4, 5)]
nx.draw_networkx_nodes(G, pos=pos, nodelist=a)
```
I'm sure G has each node in a, and pos is exactly what it needs.

I managed to solve this problem by transforming ndarray to list like this.

```
a = np.array([1, 2, 3, 4, 5)]
nx.draw_networkx_nodes(G, pos=pos, nodelist=a.tolist())
```

I believe that error confuses us, so we should either allow to pass ndarray or show a more accurate error.

Thank you for your consideration.

Shumpei-Kikuta:
Solve the error as https://github.com/networkx/networkx/issues/3552 , which allows us to use numpy ndarray.
Thank you for your time to check my work.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -389,8 +389,8 @@ def draw_networkx_nodes(G, pos,
     if nodelist is None:
         nodelist = list(G)
 
-    if len(nodelist) == 0 or not nodelist:  # empty nodelist, no drawing
-        return None
+    if len(nodelist) == 0:  # empty nodelist, no drawing
+        return
 
     try:
         xy = np.asarray([pos[v] for v in nodelist])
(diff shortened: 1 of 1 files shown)


dschult:
Now that I look at the code a little closer, there is a cleaner fix: Can you leave off the ```or not nodelist```, at this point in the code the variable nodelist should be a list and the only way it can pass the ```not nodelist``` test is if the length is zero.  So let's just check for zero.

Also, ```return None``` on the very next line should just be ```return``` because that defaults to returning None.

Thanks!  (you can commit the new changes locally to the same branch and then push to your ```origin``` and it should update this PR automagically.)

Shumpei-Kikuta:
I modified my code as you advised.
Thank you for checking.

dschult:
Thanks!

PR #3559, opened 2019-08-28
ENH: sort link output so output is consistent

CJ-Wright:


dschult:
Nodes are not always sortable... so this edge sorting will not work in general.
What are you trying to obtain? If you want the same output each time you run the same code you can use an OrderedGraph, or just Python 3.6+ (which makes dict keys ordered by when then are added).

CJ-Wright:
I totally missed that the edges can be unsortable.

PR #3561, opened 2019-08-29
Updated String Formatting and Error Messages 

manesioz:
Updated error messages, accomplishes the following: 

- Uses proper string formatting conventions consistently (i.e. "{}".format(var) instead of "%s" % var) 
- Uses the actual graph alias instead of the default "G", which makes the error more clear and easier to trace back if you are dealing with multiple graphs

### DIFF
--- a/networkx/algorithms/approximation/connectivity.py
+++ b/networkx/algorithms/approximation/connectivity.py
@@ -182,9 +182,9 @@ def node_connectivity(G, s=None, t=None):
     # Local node connectivity
     if s is not None and t is not None:
         if s not in G:
-            raise nx.NetworkXError('node %s not in graph' % s)
+            raise nx.NetworkXError('node {} not in graph'.format(s))
         if t not in G:
-            raise nx.NetworkXError('node %s not in graph' % t)
+            raise nx.NetworkXError('node {} not in graph'.format(t))
         return local_node_connectivity(G, s, t)
 
     # Global node connectivity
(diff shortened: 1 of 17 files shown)


dschult:
In what sense is ```"{}".format(var)``` more proper than ```"" % var```?
They are both valid Python structures and each has advantages. Can you cite a deprecation of %?

I don't think the term ```graph alias``` describes what you are showing. The output is not very helpful (partly because we haven't made printing graphs very helpful). Maybe it has the ```id``` of the object?

Anyway, you probably should not spend any more time on this without justifying that it is a useful enterprise. Perhaps you can spend the time implementing a graph algorithm instead?

manesioz:
@dschult Thank you for your prompt and detailed response. 

[PEP 3101](https://www.python.org/dev/peps/pep-3101/) states that the new string formatting is 

> intended as a replacement for the existing '%' string formatting operator. 

The main reason for dong so seems to be that 

> The '%' operator is primarily limited by the fact that it is a binary operator, and therefore can take at most two arguments.

Additionally, the `networkx` code base used both string formatting methods inconsistently, so I thought it would be best to keep it consistent and enhance readability. 

You're right about my patch not printing the alias, I apologize. Each instance of the Graph class may have an attribute "name" that we could include in the error message, but since it is not mandatory people may get an unhelpful message if "name" is not defined. Instead, I just changed "G" to "graph" since I think that is more useful ("G" is a common instance name for Graphs in the `networkx` docs but it that may not be clear to the user).

dschult:
I've decided to leave the % string formatting operator because many times we only need a binary operation and it is perhaps more readable. Similarly I am going to leave G instead of graph. While I know it is bad coding practice to use short variable names, this is so embedded in our documentation and code and I haven't run into any downsides in terms of misunderstandings. 
So I'm going to close this PR.

PR #3574, opened 2019-09-09
add internet_as_graph generator

lucabaldesi:
Hi,
I am a researcher at the University of Trento, Italy and I would like to contribute to the NetworkX project adding a random graph generator algorithm I implemented following a state-of-art autonomous system network model.

This algorithm, called "Internet AS graph", follows the approach detailed in [1] and it outputs a NetworkX Graph whose nodes can be "tier 1", "mid-level", "customer" or "content provider" BGP ASs and whose edges model economic relationships between ASs (peering, or service provider).

I implemented the generator mainly for the project I have been working on, "Internet on Fire", part of the "[Fed4FIRE](https://www.fed4fire.eu/)" H2020 experiment project group.
However, since AS and BGP protocol analysis is still an active field, I think it can also be useful to the community.

I prepared the files as described in the [documentation](https://networkx.github.io/documentation/latest/developer/contribute.html) but I am obviously available for any needed changes.

Best,
Luca

[1] A. Elmokashfi, A. Kvalbein and C. Dovrolis, "On the Scalability of BGP: The Role of Topology Growth," in IEEE Journal on Selected Areas in Communications, vol. 28, no. 8, pp. 1250-1261, October 2010.

### DIFF
--- a/doc/reference/generators.rst
+++ b/doc/reference/generators.rst
@@ -219,11 +219,11 @@ Stochastic
 
 AS graph
 --------
-.. automodule:: networkx.generators.internet_as_graph
+.. automodule:: networkx.generators.internet_as_graphs
 .. autosummary::
    :toctree: generated/
 
-   internet_as_graph
+   random_internet_as_graph
 
 
 Intersection
--- a/doc/reference/generators.rst
+++ b/doc/reference/generators.rst
@@ -217,6 +217,15 @@ Stochastic
    stochastic_graph
 
 
+AS graph
+--------
+.. automodule:: networkx.generators.internet_as_graph
+.. autosummary::
+   :toctree: generated/
+
+   internet_as_graph
+
+
 Intersection
 ------------
 .. automodule:: networkx.generators.intersection
--- a/networkx/generators/__init__.py
+++ b/networkx/generators/__init__.py
@@ -12,7 +12,7 @@ from networkx.generators.duplication import *
 from networkx.generators.ego import *
 from networkx.generators.expanders import *
 from networkx.generators.geometric import *
-from networkx.generators.internet_as_graph import *
+from networkx.generators.internet_as_graphs import *
 from networkx.generators.intersection import *
 from networkx.generators.joint_degree_seq import *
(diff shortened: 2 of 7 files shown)


dschult:
Hi! Thanks for contributing this functionality!

It's going to take me a little bit to work through the code, but I have some comments based on my first read-through.

- the function name and the module name are the same which will probably cause trouble down the line. Can you come up with a module name or function name that are different? For example, is this function part of a general group of functions so that in the future someone could put another function into this module? e.g. random_as_graphs.py

- The docstring for each function should start with a single line description (<80 chars) followed by a blank line and the a paragraph explaining for fully what is going on. Sphinx uses the first line in the docs whenever it links to this function. 

- Use pycodestyle to check the PEP8 style guide for Python style. You can install pycodestyle with pip and it runs on the command line. We're not draconian about style, but if there isn't a reason to break the style we might as well stick to it. Usually things it complains about are 80 char lines, spacing or short variable names.

lucabaldesi:
Hi and thanks for the feedback!
I updated file names, docstrings and proof-checked the style against pycodestyle.

dschult:
This looks good!  Thanks!!

I have three items -- the second is probably just pickiness, so you don't have to do it.
1) can you shorten the first line of the docstring for the ```random_internet_as_graph``` function so it is one line? Maybe abbrev IASM and then describe what that acronym is after a blank line.
2) can you move the ```__init__``` method of your class to the top of the methods? (optional)

3) names...   Does it make sense to use ```random_internet_AS_graph``` instead of lowercase ```random_internet_as_graph``` which reads like the word "as" instead of the letters A, S... 
Use whichever you think works best for users... Names are arbitrary, but surprisingly important...

lucabaldesi:
about your items:
1) [DONE], I did not notice it during previous commit....
2) [DONE], I like it better too
3) I absolutely think names are important and I actually picked the lowercase version of the name because I do like the underlying ambiguity :) It can be interpreted both ways, "internet AS graph" and generating a "graph like the internet".

dschult:
Excellent!  Thanks very much for this PR.

Issue #3575, opened 2019-09-09
3d layouts


daniel-karl:
nx_agraph does not process 3d coordinates when setting `dimen=3` in graph before layout

daniel-karl:
#3576 pull request

daniel-karl:
Added tests and updated pull request.

dschult:
Fixed in #3576

PR #3576, opened 2019-09-09
#3575 return coordinates of 3d layouts

daniel-karl:
Now parsing and returning any number of coordinates.


dschult:
Can you add some tests to make sure this feature doesn't get lost in the future?
They can be quite small examples...

daniel-karl:
Added 2 tests (2d and 3d layout).

Issue #3522, opened 2019-07-22
Why is Watts-Strogatz graph generation limited to k<n


tmabraham:
Right now, networkx throws an error if k<n for the `watts_strogatz_graph` function. However, it is actually possible for it to return a complete graph for k>=n. Can this error raising be removed?

dschult:
I guess that k>n would usually be a mistake that should raise an exception so the user can correct the mistake. If we make the code return a complete graph for this case, it could hide such a mistake. You can remove it for your case by handling the exception and assigning a complete graph. I know that is more work for the user, but perhaps better than hiding a mistake for someone else.

tmabraham:
@dschult Then wouldn't a warning be more appropriate?

dschult:
Apparently warnings are often turned off in the default environment. I don't know how true that is and have not checked the details, but we don't use warnings very often because they don't seem effective in the sense you are speaking of. (They are more effective for warning developers.) 

I don't know why the case k == n raises an exception -- perhaps because it is not really a Watts-Strogatz process in that case?

tmabraham:
Why was this closed? Can networkx be updated to allow for complete graphs at k==n?

dschult:
But when ```k==n``` it isn't a random graph... so it isn't a Watts Strogatz graph.
Can you use? 
    nx.watts_strogatz_graph(n, k) if k < n else nx.complete_graph(n)

tmabraham:
The same is true when `p=1` for E-R graphs... Then why is it returning a complete graph? Shouldn't it return an error just like this function. I think including the `k==n` case is probably best for completeness, no?

dschult:
Sounds good -- let's check for ```k==n``` explicitly in the code and return ```complete_graph``` in that case to avoid swapping edges.

KanishkT123:
I made a PR for this issue here #3578 

Let me know if it needs any changes or if I've misunderstood the issue, since it's my first PR!

PR #3579, opened 2019-09-11
Handle k==n within the Watts-Strogatz graph generator

KanishkT123:
It is now compatible with k==n scenario
additionally changed the unit tests for the same effect

This is my first pull request, relating to issue #3522

Please let me know if I need to make any changes!


dschult:
Looks good!  Thanks!

PR #3585, opened 2019-09-14
Christofides

LucaCappelletti94:
Implemented [Christofides](https://en.wikipedia.org/wiki/Christofides_algorithm) and its test suite. I have no idea how to test if the approximation of the TSP is correct, but for certain the cycles returned are Hamiltonian.


jarrodmillman:
See #4083.

dschult:
Looks like this PR #3685 should be combined with #4083 to put the functions into a single module and combine the tests and such.  Should we do that as two PRs with one going first? Or do we merge in a PR and then accept one PR?

Also, this PR has things like tests for the import structure of the package that aren't directly related. Should they get separated out?

dschult:
This PR is now contained in #4607

PR #3598, opened 2019-09-20
remove cyclical references from OutEdgeDataView

AaronOpfer:
We noticed that our garbage collection runs were frequently sweeping OutEdgeDataView objects. Turns out this was because `self._nodes_nbrs`, being stored on `self`, was also referring to `self` inside of its lambda, creating a cycle that required garbage collection to clear.

Fix this by instead holding onto the relevant dictionary object inside the lambda.


dschult:
Thanks very much!  This looks great!

PR #3599, opened 2019-09-23
fix to_directed function

salym:
The `as_view=True` argument is missing so the function returns a deepcopy instead of a view.


dschult:
Thanks!

PR #3600, opened 2019-09-24
Breadth traversal fix

artiomn:
# Possible bug in the graph traversal

Used graph class - `OrderedDiGraph`.

Result of the `list(networkx.edges(graph))`:

```
[('hdf_reader', 'buffer1'), ('buffer1', 'filter1'), ('buffer1', 'filter2'), ('filter1', 'extractor_dest'), ('filter2', 'extractor_dest')]
```

Incorrect result of the `list(bfs_successors(graph, 'buffer1'))`:

```
[('buffer1', ['filter1', 'filter2']), ('filter1', ['extractor_dest'])]
```

Correct result of the `list(bfs_successors(graph, 'buffer1'))`:

```
[('buffer1', ['filter1', 'filter2']), ('filter1', ['extractor_dest']), ('filter2', ['extractor_dest'])]
```

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -81,8 +81,9 @@ def generic_bfs_edges(G, source, neighbors=None, depth_limit=None):
         parent, depth_now, children = queue[0]
         try:
             child = next(children)
+            yield parent, child
+
             if child not in visited:
-                yield parent, child
                 visited.add(child)
                 if depth_now > 1:
                     queue.append((child, depth_now - 1, neighbors(child)))


dschult:
I don't think your correct result is a correct breadth first search -- at least the way we are defining it. In particular, you have a cycle in your list of edges. The edge filter2 -> extractor_dest ends at a node already discovered from node filter1, so it does not belong to a breadth first search and should not appear in the list of successors.

artiomn:
I don't have cycles. This is `OrderedDiGraph`, as I mentioned before, where `extractor_dest` is the endpoint.
Also I have checked graph before using, with `nx.find_cycle()`.
Please, look at the `edges()` result.

dschult:
I know your graph doesn't have cycles. I mean that your proposed "correct" result has cycles.
In other words. The thing you think is correct is not correct.

artiomn:
Ok, let it be incorrect.
How can I get all paths from the root element to the endpoint, using NX?
I want to get the result, like the last example in my first post.

dschult:
You probably don't want "all" paths....(they could retrace their steps in a cycle forever).
But maybe ```nx.all_simple_paths()``` would work.
[More options here.](https://networkx.github.io/documentation/latest/reference/algorithms/shortest_paths.html)

artiomn:
Ok, thank you.

Issue #3157, opened 2018-09-13
possible bug in read_gml?


joelmiller:
If I go to Mark Newman's webpage and download the karate club graph: http://www-personal.umich.edu/~mejn/netdata/karate.zip

I do the command

    nx.read_gml('karate.gml')

and get the error 

    >NetworkXError: node #0 has no 'label' attribute

I get a different error when reading the example on the Wikipedia page for gml format.

https://en.wikipedia.org/wiki/Graph_Modelling_Language

    >NetworkXError: expected an int, float, string or '[', found 'n42' at (4, 5)

Am I misunderstanding something?

dschult:
The [technical report with specs for GML](http://www.fim.uni-passau.de/fileadmin/files/lehrstuhl/brandenburg/projekte/gml/gml-technical-report.pdf) linked to in the docs states that values need to be int, float, string or lists. And ```n42``` is not a string like ```"n42"``` would be.

It also looks like each node/edge should have a ```label``` attribute and the file karate.gml does not.

Perhaps our reader is being too demanding to 1) require quotes around a string and 2) require every node and edge to have a ```label```. But it is following the spec.

ankitameht:
Hey, can i take a look at this PR, if no else is looking.

dschult:
Sounds good!   Thanks!

We want a flexible reader, but it should write the official spec. :)

ankitameht:
Ok. I will have a look at the issue, although I am trying this for the first time!

ankitameht:
Hi Dan, 

Sorry for delay in responding. I suspect the error is coming from gml.py file at the following line:
https://github.com/networkx/networkx/blob/be7c67553d2d27db4c66ce20942e6e443c39bef5/networkx/readwrite/gml.py#L424

Now I have a question here - according to the logic, since `id` is like a unique identifier and `label` is its string name, why are both these supposed to be unique. Current issue could be resolved by creating a default label name (say using its `label-<id>`) if id is mentioned. 

Please let me know if I am looking in the right direction.

dschult:
I think the issue is that the reader gets to the token n42 (without quotes) in the gml file and creates this error:

    >NetworkXError: expected an int, float, string or '[', found 'n42' at (4, 5)

It expects n42 to be surrounded by quotes so it can be identified as a string. That would follow the official spec.  We want it to handle n42 by checking if it is an int, float, string or '[' and if not, to treat it like a string even though it doesn't have quotes around it.

[...]

PR #3603, opened 2019-09-25
Bugfix: Added flexibility in reading values for label and id

KanishkT123:
As per issue #3157 

Added in support for string convertible objects to be used as ID's and labels, as well as source and targets for the GML reader. This adds increased flexibility in reading, even if the file is technically out of spec.


dschult:
Thanks!

PR #3620, opened 2019-10-02
Removed unused import of random module

yaph:


dschult:
Thanks!  
The functionality was replaced by ```py_random_state``` and we forgot to remove ```random```

PR #3621, opened 2019-10-02
Fix typo within incode documentation

AntoineSebert:
Modify *'strategy_saturation_largest_first'* into *'saturation_largest_first'*

### DIFF
--- a/networkx/algorithms/coloring/greedy_coloring.py
+++ b/networkx/algorithms/coloring/greedy_coloring.py
@@ -286,7 +286,7 @@ def greedy_color(G, strategy='largest_first', interchange=False):
        * ``'connected_sequential_bfs'``
        * ``'connected_sequential_dfs'``
        * ``'connected_sequential'`` (alias for the previous strategy)
-       * ``'strategy_saturation_largest_first'``
+       * ``'saturation_largest_first'``
        * ``'DSATUR'`` (alias for the previous strategy)
 
     interchange: bool


dschult:
Hmm.. It looks like the old name is used in the code -- not just the documentation. 
And I don't think the new name is in the code. Why do you think it is a Typo?

AntoineSebert:
The following code:
```python3
greedy_color(G, 'strategy_saturation_largest_first')
```

Produces the folowing error:
```
  File "src/main.py", line 193, in <module>
    main()
  File "src/main.py", line 173, in main
    greedy_color(G, 'strategy_saturation_largest_first')
  File "...\Python\Python37\lib\site-packages\networkx\algorithms\coloring\greedy_coloring.py", line 339, in greedy_color 
    '{0} not valid.'.format(strategy))
networkx.exception.NetworkXError: strategy must be callable or a valid string. strategy_saturation_largest_first not valid.
```

While the following code works fine:
```python3
greedy_color(G, 'saturation_largest_first')
```

We prefix *"strategy_"* should not be included in the strategy name given to the `strategy` parameter of `greedy_color()`

Link to the documentation page :
https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.coloring.greedy_color.html#networkx.algorithms.coloring.greedy_color

dschult:
Got it!...  Sorry -- I missed the part of the code where they switch from strings to function names. Thought it was simply a transfer, but it's a lookup where it does change the name.

Issue #3625, opened 2019-10-03
Add Interrupt functionality to graph traversals


AnjoMan:
One graph traversal technique I've used is an interrupt, essentially a handler function that decides if the branch in question should be traversed. This is important in applications such as energy distribution networks, where you might want to iterate over the circuit in a depth-first fashion but stop if you encounter an open switch that disconnects part of the network.

I have a basic idea for how to implement this and am happy to work on it, but would like to get feedback on whether this is a welcome idea and how people might want it to look.

PR #3627, opened 2019-10-03
Document `subgraph_view` function with support for filter functions

AnjoMan:
This is functionality I've developed external to networkx and thought it might be a welcome feature.

An example of how I've used it is in an energy distribution circuit, where I want to stop iterating a series of conducting elements when I encounter an open switch. For example, in the following circuit i have several conductors (a,b), (b, c), (b, d) and (c, e) -- one of which is a switch. Depending on the status of the switch, i would want to stop traversing the graph so that I don't build a model of the circuit where there is a demand for power that can't be satisfied, which would probably cause my model to fail to solve.

```
    a---------b---------c./ .e->L
               \
                \
                 d->L
```

I'm open to suggestions about how this could work , but my vision for it is that traversal functions can accept a callable which takes a standard set of arguments -- for example, a depth-first-search edge traversal interrupt would be parametrized by the graph and the parent/child, so that within the interrupt i can perform any analysis on the graph and/or edge data that I want before deciding if the branch should be interrupted, which I can do by raising a `StopIteration` error.

I've created an issue, https://github.com/networkx/networkx/issues/3625, related to this PR, which may be a better place to discuss this at a high-level. 

I'll probably work on adding this interface to the `edge_dfs` function and the bfs functions as well, which is why this PR is marked WIP for now.

### DIFF
--- a/doc/reference/functions.rst
+++ b/doc/reference/functions.rst
@@ -22,6 +22,7 @@ Graph
    add_path
    add_cycle
    subgraph
+   subgraph_view
    induced_subgraph
    restricted_view
    reverse_view
--- a/networkx/classes/function.py
+++ b/networkx/classes/function.py
@@ -21,11 +21,14 @@ except ImportError:
 import networkx as nx
 from networkx.utils import pairwise, not_implemented_for
 
+from networkx.classes.graphviews import subgraph_view, reverse_view
+
 __all__ = ['nodes', 'edges', 'degree', 'degree_histogram', 'neighbors',
            'number_of_nodes', 'number_of_edges', 'density',
-           'is_directed', 'info', 'freeze', 'is_frozen', 'subgraph',
-           'induced_subgraph', 'edge_subgraph', 'restricted_view',
-           'reverse_view', 'to_directed', 'to_undirected',
+           'is_directed', 'info', 'freeze', 'is_frozen',
+           'subgraph', 'subgraph_view', 'induced_subgraph', 'reverse_view',
+           'edge_subgraph', 'restricted_view',
+           'to_directed', 'to_undirected',
            'add_star', 'add_path', 'add_cycle',
            'create_empty_copy', 'set_node_attributes',
            'get_node_attributes', 'set_edge_attributes',
@@ -490,15 +493,6 @@ def restricted_view(G, nodes, edges):
         else:
             hide_edges = nxf.hide_edges(edges)
     return nx.graphviews.subgraph_view(G, hide_nodes, hide_edges)
-
-
-@not_implemented_for('undirected')
-def reverse_view(digraph):
-    """Provide a reverse view of the digraph with edges reversed.
-
-    Identical to digraph.reverse(copy=False)
-    """
-    return nx.graphviews.reverse_view(digraph)
 
 
 def to_directed(graph):
--- a/networkx/classes/graphviews.py
+++ b/networkx/classes/graphviews.py
@@ -36,7 +36,8 @@ Often it is easiest to use .copy() to avoid chains.
 from networkx.classes.coreviews import UnionAdjacency, UnionMultiAdjacency, \
     FilterAtlas, FilterAdjacency, FilterMultiAdjacency
 from networkx.classes.filters import no_filter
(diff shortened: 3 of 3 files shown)


dschult:
If I understand you correctly, the dfs_edges function already does this.  It iterates over the edges so after yielding an edge it waits for you to process it before continuing to the next edge. This saves memory and also allows you to stop whenever you like.

    for (u, v) in nx.dfs_edges(G, n):
        if switch_open(u, v):
            print('Stopping DFS')
            break

AnjoMan:
Ah, good observation. The difference in what I've proposed, which I didn't really explain, is that my interrupt doesn't terminate the dfs search. It only terminates traversal *of the interrupted branch*. 

If you look at the test I wrote, you'll see the sample network is a tree with two branches (1 > 2 > 3 > 4 > 5 as well as 1> 6), and after I interrupt traversal of one of the branches, the dfs algorithm yields items from the *other* branch. If I just stop consuming items after encountering edge (3, 4) I'm not guaranteed to ever get (1, 6) because dfs may not have yielded it yet.

dschult:
OK... I think I understand...  But just to check:

- is this different from [the restricted_view approach](https://networkx.github.io/documentation/latest/reference/generated/networkx.classes.function.restricted_view.html#networkx.classes.function.restricted_view) where you can hide nodes or edges depending on a function output?
- I'm perhaps not up to the latest on StopIteration, but doesn't raising a StopIteration inside interrupt_func mean it will be raised in dfs_edges where interrupt_func is used, in which case it would not continue to run the dfs_edges?  Clearly my understanding is incomplete, I recall something changed with how StopIteration is handled inside nested functions/generators...
- We need more clarity in the docs on what is required of the interrupt function.

AnjoMan:
* There is some similarity here with `restricted_view` -- if i were to bulid my own `dfs_edges_interruptable` function around this i think it might look like 

   ```python
  def dfs_edges_interruptable(graph, should_traverse):
      filtered_graph = nx.restricted_view(
           graph, 
           [], 
           (edge for edge in nx.edges() if not should_traverse(edge))
      )
      yield from nx.dfs_edges(filtered_graph)
  ```
  the disadvantage of this approach is that I may have to iterate every edge in the graph before I can make the restricted view, which can be expensive (e.g. suppose a graph 1 > 2 -> ...  > 1000 and the edge that should interrupt my traversal is  (1, 2))
* Agree that `StopIteration` is maybe a bit too obscure. Maybe it would make more sense to return `True` if the edge meets the criteria for stopping
* I'll be sure to add a description in the docs to make it more clear how to use this.

dschult:
Thanks for that.  There is a relatively[ poorly documented feature](https://networkx.github.io/documentation/latest/reference/classes/index.html#module-networkx.classes.graphviews) behind the hood for these "filtered views" of a graph. 

def dfs_edges_interruptable(graph, should_traverse):
   filtered_graph = nx.subgraph_view(graph, filter_edge=should_traverse)
   yield from nx.dfs_edges(filtered_graph)

Or perhaps use the idiom:  ```dfs_edges(subgraph_view(graph, should_traverse))```

AnjoMan:
Ok, thanks for this! I'll explore it a bit and maybe i can just add documentation of that function instead of making this change.

dschult:
This looks good. Thanks!  I especially like that you've moved subgraph_view into the main networkx namespace.

- The ```EdgeView``` output in the examples should be OutEdgeView for a directed graph's view.
- The 1st 3 lines of code in reverse_view (which raise an exception if not directed) should be removed now the you've got ```not_implemented_for``` in place.
- Can you move ```reverse_view``` to the same line as ```subgraph_view``` in ```__all__```. Just to even out the lines there -- and they are both imported into this module, so makes sense to keep together.
- Can you remove blank lines after ```Examples``` and ```Parameters``` headings?
- Change ```nx.graphviews.subgraph_view``` to ```nx.subgraph_view``` in examples. Same for ```nx.reverse_view```. 
Thanks!

AnjoMan:
@dschult thanks for your guidance on this MR. I've made the changes you've requested -- also fixed a bad line in one of the docstring examples which was failing to run. I think it is good to merge.

PR #3630, opened 2019-10-03
Add missing code block

AntoineSebert:


jarrodmillman:
We use reStructuredText and Sphinx for documentation.  So you need to use

```
.. code-block:: python

    # instead of SG = G.subgraph(ordered_nodes)
    SG=nx.OrderedGraph()
    SG.add_nodes_from(ordered_nodes)
    SG.add_edges_from((u, v) for (u, v) in G.edges() if u in SG if v in SG)

```
instead of the markdown markup.

AntoineSebert:
Done !

PR #3632, opened 2019-10-03
Fix typo + factorize `if` statement

AntoineSebert:
* replace `strategy_saturation_largest_first` and `strategy_independent_set` by respectively `saturation_largest_first` and `independent_set` (similar to issue  #3621).

* replace
```python
if strategy is strategy_independent_set:
    msg = 'interchange cannot be used with strategy_independent_set'
    raise nx.NetworkXPointlessConcept(msg)
if strategy is strategy_saturation_largest_first:
    msg = ('interchange cannot be used with'
           ' strategy_saturation_largest_first')
    raise nx.NetworkXPointlessConcept(msg)
```
by
```python
if strategy is strategy_independent_set or strategy is strategy_saturation_largest_first:
    msg = 'interchange cannot be used with ' + strategy
    raise nx.NetworkXPointlessConcept(msg)
```

### DIFF
--- a/networkx/algorithms/coloring/greedy_coloring.py
+++ b/networkx/algorithms/coloring/greedy_coloring.py
@@ -293,7 +293,7 @@ def greedy_color(G, strategy='largest_first', interchange=False):
        Will use the color interchange algorithm described by [3]_ if set
        to ``True``.
 
-       Note that ``strategy_saturation_largest_first`` and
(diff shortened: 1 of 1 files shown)


dschult:
See #3631 which I think handles these cases. I don't think the refactoring of the "if" is needed.
Thanks for the help.

Issue #3644, opened 2019-10-11
cliques documentation notes examples


AdamRosenthal:
in functions documentation notes of `networkx/algorithms/clique.py`, the complete graph is given as an example for worst-case length of cliques list. To my understanding, the complete graph has only 1 maximal clique and it is not the worst-case.

dschult:
Ahh. I understand better now. But the PR removes the example for ```enumerate_all_cliques``` as well. And that does increase exponentially in number of cliques as number of nodes grows. So, maybe only remove it from the two ```find_cliques(_recursive)``` functions.  Is there a good example of a type of graph that increases the number of maximal cliques exponentially as the size of the graph grows?

AdamRosenthal:
Iv'e updated the PR to only remove the example from the two `find_cliques(_recursive)` functions.
The only valid examples I could find were:
1. The complete graph where all perfect matching edges were removed (2^n/2 maximal cliques)
2. The complete graph where all sets of disjoint triangles edges were removed (3^n/3 maximal cliques)

These types of graphs will be worst-case scenarios.

PR #3645, opened 2019-10-11
fix documentation notes (#3644)

AdamRosenthal:
Removed example of worst case scenario for finding maximal cliques on a complete graph


dschult:
Why?

Issue #3607, opened 2019-09-26
Provide a function arg to contracted_nodes to prevent copying


nanodan:
For large graphs, with a lot of contraction steps, the `H = G.copy()` step is by far the bottleneck in speed. It would be nice to have an option to pass an argument like copy=False to the function to operate directly on G instead of making a copy for each contract step.

dschult:
This seems like a good idea and easy to implement as you describe.
Thanks!

KanishkT123:
@dschult @nanodan is anyone working on this? I have some downtime this week and could probably give it a shot if not.

dschult:
That's be great!

MridulS:
Should be closed, PR merged in https://github.com/networkx/networkx/pull/3646

PR #3646, opened 2019-10-11
`contracted_nodes` copy flag added

KanishkT123:
Added a copy switch to the function contracted_nodes that allows the passed in graph G to be changed in place as compared to having it be copied into a new graph first. 

This should significantly allow for some more efficient memory utilization with larger graphs, the caveat being that contracted_nodes now needs to store the edge values from the graph as a list in case of the original graph being changed in place, as the generators reflect the current state of the graph at all times.


KanishkT123:
This is for #3607

KanishkT123:
@dschult I also wanted to ask: The unit tests for contracted_nodes() seem to be using the complete_graph(3) instead of cyclic_graph(3). While this is correct theoretically, would it maybe be better practice to have the test still make comparisons between two cyclic graphs instead, to make it more clear what the test is doing? Just a thought, please ignore if I'm wrong.

dschult:
Seems reasonable to change the expected test result to ```cyclic_graph(3)``` instead of the complete graph. Both are right, but cyclic shows what is being tested.

dschult:
This looks good.  Can you check the PEP8 style guides using pycodestyle?
It mostly shows spacing irregularities: <80 chars per line, spaces after commas and around operators.

Thanks for this!

KanishkT123:
Done!

KanishkT123:
@dschult I think this is ready for merge, unless there are other changes I need to make? Let me know!

dschult:
Yup -- ready to go...  Been a busy week.  Thanks for the gentle nudge. :)

PR #3647, opened 2019-10-12
Optimize _single_shortest_path_length

jonatanwestholm:
The helper function 
_networkx.algorithms.shortest_path.unweighted.**_single_shortest_path_length**_ 
can be optimized.

**Current implementation**: add neighbours of visited nodes to "nextlevel" set. Later, iterate through "nextlevel" set to expand their unseen neighbours. Iterate until "nextlevel" is empty.

**Suggested optimization 1**: keep track of nodes to be expanded in a queue (collections.deque) that remains sorted by distance from source node. Do not add already visited nodes to any set.

**Suggested optimization 2**: terminate immediately if all nodes have been visited. This saves a lot of work for dense, strongly connected graphs.

Investigation and analysis shows the effectiveness of the optimizations. Measured speedup is **at least a factor of 2**, for graphs where python overhead is not dominating. For dense strongly connected graphs, the speedup factor can be up to O(n), where n is the number of nodes.

Below are some speed tests with random graphs (the first 5 graphs), and some special structure graphs. 
```
time / ms                    current         optimized           
n=1e1, m=1e1                    0.15              0.14             
n=1e1, m=3e1                    0.34              0.18            
n=1e2, m=1e2                    2.04              1.35           
n=1e3, m=1e3                   56.76             31.89         
n=1e3, m=1e5                27857.46            612.96        
caveman(100,10)              3373.64           1168.82       
windmill(10, 10)               32.08              4.09          
nary_tree(3, 1000)           1886.88            810.51       
complete(100)                 205.25              3.48          
```

I have published a more thorough post on my blog [garageofcode: optimizing shortest path length](https://garageofcode.blogspot.com/2019/10/optimizing-shortestpathlength-in.html). This includes an analysis of why we observe this speedup. 

Since this is my first pull request here, I would like to say thank you for a great package! 
Also, please make comments about style and content of this pull request, including this description itself. 

//Jonatan

### DIFF
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -10,6 +10,7 @@
 """
 Shortest path algorithms for unweighted graphs.
 """
+from collections import deque
 import networkx as nx
 
 __all__ = ['bidirectional_shortest_path',
@@ -79,20 +80,23 @@ def _single_shortest_path_length(adj, firstlevel, cutoff):
         cutoff : int or float
             level at which we stop the process
     """
-    seen = {}                  # level (number of hops) when seen in BFS
-    level = 0                  # the current level
-    nextlevel = firstlevel     # dict of nodes to check at next level
-
-    while nextlevel and cutoff >= level:
-        thislevel = nextlevel  # advance to next level
-        nextlevel = {}         # and start a new list (fringe)
-        for v in thislevel:
-            if v not in seen:
-                seen[v] = level  # set the level of vertex v
-                nextlevel.update(adj[v])  # add neighbors of v
-                yield (v, level)
-        level += 1
-    del seen
+    visited = set(firstlevel)
+    queue = deque(firstlevel.items())  # sorted by level, ascending
+    for u, level in queue:
+        yield u, level-1  # level 1 means distance 0 to source nodes
+    unvisited = len(adj) - 1
+    while queue:
+        node, level = queue.popleft()
+        if level > cutoff:
+            return
+        for v in adj[node]:
+            if v not in visited:
+                visited.add(v)
+                queue.append((v, level+1))
+                yield v, level
+                unvisited -= 1
+                if not unvisited:  # all nodes seen: terminate
+                    return
 
 
 def single_target_shortest_path_length(G, target, cutoff=None):
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -10,7 +10,6 @@
 """
 Shortest path algorithms for unweighted graphs.
 """
-from collections import deque
 import networkx as nx
 
 __all__ = ['bidirectional_shortest_path',
@@ -80,23 +79,25 @@ def _single_shortest_path_length(adj, firstlevel, cutoff):
         cutoff : int or float
             level at which we stop the process
     """
-    visited = set(firstlevel)
-    queue = deque(firstlevel.items())  # sorted by level, ascending
-    for u, level in queue:
-        yield u, level-1  # level 1 means distance 0 to source nodes
-    unvisited = len(adj) - 1
-    while queue:
-        node, level = queue.popleft()
-        if level > cutoff:
+    seen = {}  # level (number of hops) when seen in BFS
+    level = 0  # the current level
+    nextlevel = set(firstlevel)  # set of nodes to check at next level
+    n = len(adj)
+    while nextlevel and cutoff >= level:
+        thislevel = nextlevel  # advance to next level
+        nextlevel = set([])  # and start a new set (fringe)
+        found = []
+        for v in thislevel:
+            if v not in seen:
+                seen[v] = level  # set the level of vertex v
+                found.append(v)
+                yield (v, level)
+        if len(seen) == n:
             return
-        for v in adj[node]:
-            if v not in visited:
-                visited.add(v)
-                queue.append((v, level+1))
-                yield v, level
-                unvisited -= 1
-                if not unvisited:  # all nodes seen: terminate
-                    return
+        for v in found:
+            nextlevel.update(adj[v])
+        level += 1
+    del seen
 
 
 def single_target_shortest_path_length(G, target, cutoff=None):


dschult:
Thanks for this!  It's about time we revisit this important algorithm.

Can you run your test codes against a version that changes nextlevel to be a set instead of a dict?  It only requires changing the two lines where nextlevel is assigned to. Lines 3 and 6 below.
I obtained speed up by factor of 2 on 100 nodes and factor of 16 on 1000 nodes. The use of dicts here is a throw-back to before sets were a python data structure.  I just wonder what your speed testing code would show using sets vs queues.

    seen = {}                  # level (number of hops) when seen in BFS
    level = 0                  # the current level
    nextlevel = set(firstlevel)     # set of nodes to check at next level
    while nextlevel and cutoff >= level:
        thislevel = nextlevel  # advance to next level
        nextlevel = set([])         # and start a new set (fringe)
        for v in thislevel:
            if v not in seen:
                seen[v] = level  # set the level of vertex v
                nextlevel.update(adj[v])  # add neighbors of v
                yield (v, level)
        level += 1
    del seen

jonatanwestholm:
Absolutely. Pasting in your code above verbatim, called 'set', I get the following results for **all_pairs_shortest_paths**, single run:

```
time / ms                    current               set      pull request           
n=1e1, m=1e1                    0.19              0.10              0.11             
n=1e1, m=3e1                    0.27              0.16              0.17            
n=1e2, m=1e2                    2.09              1.33              1.30           
n=1e3, m=1.4e3                  8.66              7.50              5.70           
n=1e3, m=1e3                   22.46             13.09             14.84          
n=1e3, m=1e4                 3932.15           1194.50           1042.33        
n=1e3, m=1e5                28663.99           3479.59            666.18        
caveman(100,10)              3437.24           1080.76           1210.02       
windmill(10, 10)               33.19              8.87              3.49          
nary_tree(3, 1000)           1960.57            882.87            826.90       
complete(100)                 205.05             16.63              3.43           
mlp(10, 10)                    15.25              4.42              5.13          
mlp(10, 100)                 8069.08           1110.82           1619.68
amnesia(100, 100)             264.08             65.13             41.02         
```
Just the small graphs, 1000 times:
```
time / ms            current           set  pull request
n=1e1, m=1e1           65.62         50.53         48.71
n=1e1, m=3e1          212.20        106.64         85.02
n=1e2, m=1e2         2122.77       1266.92       1292.34
```

My conclusion. For dense, connected graphs (eg complete), the pull request is much better. For sparse graphs, "set" is marginally better. For dense, unconnected graphs (eg mlp(10, 100)), "set" is significantly better.

Just an explanation of what I mean by "MLP": `mlp(k, n)` is a "multilayer perceptron graph" with `k` layers with `n` nodes in each. All nodes in layer i is are connected to all nodes in layer i+1 with directed edges. So mlp(10, 100) has 1000 nodes and 90000 edges.

jonatanwestholm:
Another note about the results: The `amnesia` graph 

image

is actually dense and unconnected, and the PR is faster. However, I designed it to demonstrate a bottleneck in the current algorithm. I am actually surprised that the "set" implementation is so fast for `amnesia`! 
Does python cache old sets or something?

dschult:
No cacheing with old sets -- but sets (and dicts) are wicked fast!! They are the heart of functions and have been optimized and then optimized.

The multilayer perceptron is a great special structure to test! The layers can make it tricky to be fast too. 

One more change to set:  Can you run a version of set with a check for all nodes being seen after each layer?  Replace the while statement with:

    n = len(G)
    while nextlevel and cutoff >= level and len(seen) < n:

That implements your 2nd idea that we should quit when all the nodes are seen. So the resulting "set" code is doing your ideas: 1) replace the nextlevel data structure, and 2) stop after all nodes seen. But with ```sets``` and ```len(seen)```.  Hope to get one best version by combining good parts of the rest. :}

jonatanwestholm:
I'll single out the graphs where PR was much better than set. If you'll take my word for it, 'set+check' is as fast as 'set' for the other graphs. 

```
time / ms            current           set     set+check  pull request
n=1e3, m=1e5        28467.40       3257.46       3234.22        598.96
complete(100)         205.42         16.48         17.60          3.34
```

So the extra check doesn't make it noticeably faster. I think I know why: putting the check in the main loop only avoids the O(n) work of checking that all nodes in thislevel are in seen. But it doesn't avoid the up to O(n^2) work of adding all the neighbours of the penultimate level to nextlevel. 

Now, I did try to put the check in the inner loop with an int counter (as in the pull request), but this actually doesn't make it faster either. It gives pretty much the exact same speed. 

However, we can check if all are seen before adding the found node's neighbours, to avoid potential O(n^2) work in the penultimate layer. 

```python
seen = {}                  # level (number of hops) when seen in BFS
level = 0                  # the current level
nextlevel = set(firstlevel)     # set of nodes to check at next level
n = len(adj)
while nextlevel and cutoff >= level:
    thislevel = nextlevel  # advance to next level
    nextlevel = set([])         # and start a new set (fringe)
    found = set([])
    for v in thislevel:
        if v not in seen:
            seen[v] = level  # set the level of vertex v
            found.add(v)
            yield (v, level)
    if len(seen) == n:
        return
    for v in found:
        nextlevel.update(adj[v])
    level += 1
del seen
```

This seems to make the best of both worlds:

```
time / ms                        set         set+found      pull request
n=1e1, m=1e1                    0.12              0.09              0.12
n=1e1, m=3e1                    0.23              0.16              0.14
n=1e2, m=1e2                    1.27              1.67              1.31
n=1e3, m=1.4e3                  5.34              8.53              6.19
n=1e3, m=1e3                   15.27             17.01             13.30
n=1e3, m=1e4                 1219.37           1137.89           1040.01
n=1e3, m=1e5                 3329.02            682.24            603.17
caveman(100,10)              1053.32           1130.63           1135.58
windmill(10, 10)                8.02              3.55              4.40
nary_tree(3, 1000)            879.00            785.34            820.36
complete(200)                 106.96             10.87             12.32
mlp(10, 10)                     4.50              5.50              5.62
mlp(10, 100)                 1117.21           1116.44           1617.68
amnesia(100, 100)              68.38             69.03             40.23
```

And the small graphs 1000 times:
```
time / ms                set     set+found  pull request
n=1e1, m=1e1           72.05         85.88         72.69
n=1e1, m=3e1          130.15        135.83        106.35
n=1e2, m=1e2         2070.71       2457.97       2067.36
```

So the "found" method does introduce a bit of overhead that is noticeable for small graphs.

jonatanwestholm:
The overhead from the variable "found" can be reduced by using a list instead of a set:

```
time / ms                set     set+found  pull request
n=1e1, m=1e1           76.05         82.56         69.25
n=1e1, m=3e1          132.49        118.93         98.05
n=1e2, m=1e2         2035.55       2225.16       2017.59
```

dschult:
This looks good -- thanks for all the timing.
Another potential difficulty with the PR is that deque seems to require that nodes be sortable.
Is that true?

jonatanwestholm:
Yes, precisely, the comment is a bit misleading. The deque is only implicitly sorted, not actively. 

Now the way I see it, 'set+found' performs at least as well as PR on the tests (at least not much worse). It's also more dogmatic to the style of networkx, there is a very similar implementation in **networkx.algorithms.components.connected._plain_bfs**. So with your recommendation, I'll change the commit to:

```python
seen = {}                  # level (number of hops) when seen in BFS
level = 0                  # the current level
nextlevel = set(firstlevel)     # set of nodes to check at next level
n = len(adj)
while nextlevel and cutoff >= level:
    thislevel = nextlevel  # advance to next level
    nextlevel = set([])         # and start a new set (fringe)
    found = []
    for v in thislevel:
        if v not in seen:
            seen[v] = level  # set the level of vertex v
            found.append(v)
            yield (v, level)
    if len(seen) == n:
        return
    for v in found:
        nextlevel.update(adj[v])
    level += 1
del seen
```
(that is to say, to 'set+found')

dschult:
That sounds like a good plan.  Any of these options are better than the current code. :)

jonatanwestholm:
Cool, thanks for all the guidance! 
networkx is very useful to me so it's a pleasure to contribute :)

PR #3666, opened 2019-10-18
Lukes algorithm implementation

f-rosato:
Introduces the Lukes optimal tree partitioning algorithm in networkx.algorithms.community. This algorithm partitions a tree with float weighted edges and integer weighted nodes, in such a way that the total node weight of the block does not exceed a maximum size and the edge cuts are minimum. It is an exact algorithm and it runs in pseudo-polynomial time.

Reference: Lukes, J. A. (1974). Efficient Algorithm for the Partitioning of Trees. IBM Journal of Research and Development, 18(3), 217–224. https://doi.org/10.1147/rd.183.0217

### DIFF
--- a/doc/reference/algorithms/community.rst
+++ b/doc/reference/algorithms/community.rst
@@ -30,6 +30,14 @@ Modularity-based communities
    
    greedy_modularity_communities
 
+Tree partitioning
+-----------------
+.. automodule:: networkx.algorithms.community.lukes
+.. autosummary::
+   :toctree: generated/
+
+   lukes_partitioning
+
 Label propagation
 -----------------
 .. automodule:: networkx.algorithms.community.label_propagation
--- a/networkx/algorithms/community/__init__.py
+++ b/networkx/algorithms/community/__init__.py
@@ -20,6 +20,7 @@ from networkx.algorithms.community.centrality import *
 from networkx.algorithms.community.kclique import *
 from networkx.algorithms.community.kernighan_lin import *
 from networkx.algorithms.community.label_propagation import *
+from networkx.algorithms.community.lukes import *
 from networkx.algorithms.community.modularity_max import *
 from networkx.algorithms.community.quality import *
 from networkx.algorithms.community.community_utils import *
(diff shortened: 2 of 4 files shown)


dschult:
This looks good. Thank You!
Things still to do:
- check PEP8 code style using pycodestyle (can install using pip or conda) or flake8.

    pycodestyle lukes.py tests/test_lukes.py

- add the new module to the sphinx documentation via doc/reference/algorithms/community.rst

- format docstrings so the first line is <80 chars and the second line is blank.
This may seem restrictive, but that first line is used wherever the function is linked to as a short description. It seems to enforce compact descriptions too. So, put all the restrictions involved with Lukes algorithm into the paragraph after the blank line. Make the first line short and informative. :)

f-rosato:
Thank you for the feedback Dan, I'll try to give it a look. I'm more than happy to learn to use pytest, it doesn't seem too difficult. Just give me a few days if that's not a problem (today I won't be able to work much on this).

dschult:
This error is something unrelated to what you are putting in the PR. It showed up in the last day or two. I think it has to do with a change in how matplotlib behaves, but I haven't tracked it down yet.

Don't worry about the failure in the third Travis-CI environment. 

I'm ready to merge this.  Anything else you know of before we do that?

f-rosato:
Thank you for the confirmation, I tried to parse the Travis log and I figured it was something unrelated.

I just left a #todo [here](https://github.com/networkx/networkx/blob/7448427da78248d12e3686cad41da85aa074a677/networkx/algorithms/community/lukes.py#L119) regarding a sanity check on the values of the edge attribute used, but I can't think of a safe, general and/or fast way to do it. The problem at hand is that if edge attributes are used having one or more values that do not support `<=` and `sum` among them, a `TypeError` will be raised somewhere down the road. Afterall, I think that this potential situation should be self-explicative enough without the need of writing a cumbersome check at the beginning. Let me know if you agree and, if so, if you prefer to leave the #todo there anyways or you want me to get rid of it.

Once this is settled, for me the branch is pullable.

Thank you!

dschult:
I agree that the TypeError down the road is better than a cumbersome check at the beginning.
So I would say the "todo" comment can be removed.
Thanks!

PR #3671, opened 2019-10-19
Run doctest without optional dependencies

jarrodmillman:
Pytest isn't running fixtures like
```
# fixture for pytest
def setup_module(module):
    import pytest
    numpy = pytest.importorskip('numpy')
    scipy = pytest.importorskip('scipy')
    pandas = pytest.importorskip('pandas')
```
in `networkx/convert_matrix.py` or
```
# fixture for pytest
def setup_module(module):
    import pytest
    yaml = pytest.importorskip('yaml')


def teardown_module(module):
    import os
    os.unlink('test.yaml')
```
in `networkx/readwrite/nx_yaml.py`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -47,7 +47,7 @@ test_script:
   # Change to a non-source folder to make sure we run the tests on the
   # installed library.
   - "pushd %TEMP%"
-  - "pytest --pyargs networkx"
+  - "pytest --doctest-modules --pyargs networkx"
 
 cache:
   # Avoid re-downloading large packages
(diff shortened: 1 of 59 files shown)


jarrodmillman:
@dschult I won't have time to look into this for a bit.  I mainly created the PR to make sure we didn't lose track of the issue.  I am not sure what the correct way to do this is.  Maybe it is a bug in pytest.  If you want to look into this, feel free.  Otherwise, I will get to it before the 2.5 release.

jarrodmillman:
@dschult You can see that pytest doesn't even run the fixtures in the coverage report:
- https://codecov.io/gh/networkx/networkx/src/master/networkx/convert_matrix.py#L1242
- https://codecov.io/gh/networkx/networkx/src/master/networkx/readwrite/nx_yaml.py#L110

PR #3678, opened 2019-10-21
Enable tests

jarrodmillman:


jarrodmillman:
@dschult  Do you have any suggestions for how to fix the `classes/tests/test_special.py` tests.  Pytest wasn't running them, so I changed
```
class ThinGraphTester(TestGraph):
    def setUp(self):
```
to
```
class TestThinGraph(TestGraph):
    def setup_method(self):
```
and
```
class ThinDiGraphTester(TestDiGraph):
    def setUp(self)
```
to
```
class TestThinDiGraph(TestDiGraph):
    def setup_method(self):
```

in `networkx/classes/tests/test_special.py`.

You should see the new failures, if you checkout my  [jarrodmillman:enable-tests](https://github.com/jarrodmillman/networkx/tree/enable-tests) branch and run
```
pytest networkx/classes/tests/test_special.py
```

PR #3680, opened 2019-10-21
Deprecate make_str

jarrodmillman:


jarrodmillman:
@dschult I am deprecating `make_str` and `is_string_like`.  They were left over from handling Python 2/3, I believe.  Does this seem reasonable?

dschult:
Yes, this seems reasonable.  I think the history there was more complicated than just the transition from 2/3, but basically dealt with warts of Python2 string handling that have now been fixed.
Yay!

PR #3681, opened 2019-10-22
Enable ThinGraph tests

jarrodmillman:
@dschult I mentioned this problem in a previous PR.  I decided to repackage the issue in this PR and merge all the other changes in the previous PR.

I made the class names start with Test so pytest collects it to run.  I also changed from nose style `setUp` to pytest style `setup_method`.

Any idea how to fix this?

### DIFF
--- a/networkx/classes/tests/test_digraph.py
+++ b/networkx/classes/tests/test_digraph.py
@@ -40,15 +40,6 @@ class BaseDiGraphTester(BaseGraphTester):
         with pytest.raises(nx.NetworkXError):
             G.edges(-1)
 
-    def test_edges_data(self):
-        G = self.K3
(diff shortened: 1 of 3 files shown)


dschult:
Good idea to split these out into separate PR.  I'll take a look...

dschult:
I think the TestThinDiGraph setUp was setting up an undirected set of graphs to test instead of directed graphs. I also had trouble with self.K3 not being set up the way I thought it should be.

I'm not sure I need the ```.copy()``` in every place I made it ```G = self.K3.copy()```.  But I figured, if ```G``` is going to be mutated, we might as well act on a copy to make it clear that self.K3 is not getting changed.

The main change is to use the base class ```BaseGraphTester``` instead of ```TestGraph``` as the base class for ```TestThinGraph```.  Similarly for ```TestThinDiGraph```.  That removes all the tests that check attributes. The ThinGraph doesn't update attributes.

jarrodmillman:
Looks good to me.  Is this ready to merge?

PR #3684, opened 2019-10-23
avoid duplicate tests due to imports

dschult:
Changing the name of ```TestGraph``` when importing it avoids having the tests run in the importing module.  This should speed up CI testing

I rename TestGraph as ttttGraph.  There must be a better name...?


jarrodmillman:
How about `_TestGraph`, `_GraphEdgeSubgraph`, etc.

PR #3691, opened 2019-10-25
Add NumFOCUS CoC

stefanv:
/cc @dschult @jarrodmillman @hagberg 

To bring NetworkX in line with requirements for NumFOCUS affiliation.

I'm busy publishing a formal markdown translation of the NumFOCUS CoC [here](https://github.com/numfocus/numfocus/pull/31) that we can use once it has been merged.

### DIFF
--- a/CODE_OF_CONDUCT.rst
+++ b/CODE_OF_CONDUCT.rst
@@ -1,0 +1,154 @@
+Code of Conduct
+===============
+
+
+Introduction
+------------
+
+This code of conduct applies to all spaces managed by the NetworkX project,
+including all public and private mailing lists, issue trackers, wikis, and
+any other communication channel used by our community.
+
+This code of conduct should be honored by everyone who participates in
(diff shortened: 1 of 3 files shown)


jarrodmillman:
I think the numfocus one is too long and includes stuff that isn't completely relevant.  I was imaging something closer to this:
- https://github.com/networkx/grave/pull/23/files

stefanv:
I think go ahead and merge it, because the board will want to see a CoC in place before they give approval.

I like the SciPy one—we used that for scikit-image too, and I am happy to serve on the committee.  I used groups.io because it takes all of 5 seconds to set up a group, and nothing is restricted about it.  Importantly, you can set the group to allow posting from non-members.

stefanv:
@jarrodmillman Note that you are missing the follow-up guide.

jarrodmillman:
I spoke with Stefan about the follow-up guide and decided to leave this as is for now.  We can add the follow-up guide once we have more experience.

Issue #3687, opened 2019-10-24
Speed-up of Graph.__getitem__


sudeshkagrawal:
The documentation (at  https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.Graph.neighbors.html) says:  
"It is usually more convenient (and faster) to access the adjacency dictionary as G[n]".

However, my testing using %timeit showed that is not true.

```python
%timeit G[2]
%timeit list(G.neighbors(2))
```

```
1.08 µs ± 42.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
685 ns ± 37.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

dschult:
We can easily speed up G[u] to be as fast as it used to be. 
(in G.__getitem__ use G._adj, not G.adj) 
This Issue should lead to a PR which does that. That is a much better fix than changing the docs.  
Thank you for bringing it up!  

```python
    %timeit G[2]
    %timeit G.neighbors(2)
    %timeit G._adj[2]        # the fastest: direct access to the data

    %timeit [n for n in G[2]]
    %timeit [n for n in G.neighbors(2)]
    %timeit [n for n in G._adj[2]]
```

With the resulting times

    630 ns ± 6.83 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
    238 ns ± 3.86 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
    62.7 ns ± 0.626 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

    13.1 µs ± 104 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
    12.6 µs ± 123 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
    11.9 µs ± 65.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

sudeshkagrawal:
What was the reason of switching to G.adj if G._adj is faster?

dschult:
G.adj is read-only as is G.adj[u] and G.adj[u][v] but G.adj[u][v]['attr'] is a read-write dict.
They all use G._adj to look up the info, but they protect against corrupting the data structure.

dschult:
Whoops...  We can't rewrite ```G.__getitem__``` to return ```G._adj[n]``` because while that is faster, it no longer protects the data structure. 

We'll have to change the docs instead...  :{

PR #3695, opened 2019-10-29
Speed-up of graph/multi-graph methods 

manesioz:
Addresses #3687


manesioz:
No worries, I assumed you thought that the performance improvement would be a better tradeoff to the introduced insecurity of the graphs, but if you don't think so you can close this PR.

PR #3705, opened 2019-11-03
Add an implementation for interval_graph and its unit tests

jangwon-yie:
I add an implementation for interval_graph(https://en.wikipedia.org/wiki/Interval_graph) and its unit tests. Please review the code and give me any comments for that it will be merged to main branch. Thanks.


dschult:
Thanks for this!

I have some comments/suggestions...
- it is probably better to raise an exception for an invalid interval than ignoring it.
- checking for a tuple restricts the intervals to be tuples. We've found it better to allow any object of length 2 so people can use lists, numpy.arrays, etc.  
- to check that length is 2, it is often more effective to use something like ```min1, max1 = i1```
- shouldn't you check that the interval is (min, max), So (3, 2) will raise an exception?
```
for i1 in intervals:
    min1, max1 = i1
    if min1 >= max1:  # not sure about = case here
        raise ValueError("Each interval must be a 2-tuple (min, max). Got {}".format(i1))
```
- To avoid the sort on the intervals, can you use the criteria: 
```
if max1 >= min2 and max2 >= min1:
    G.add_edge(i1, i2)
```
Then a double loop over intervals will suffice
(though you have to check ```if i1 == i2: continue``` to skip that case).  

- You need to make this function available in the networkx namespace by adding an import of this module to networkx/generators/__init__.py
- You might consider whether other Graph classes like DiGraph or MultiGraph can be used. If so, use an argument like ```create_using``` and process similar to other generator functions. If not, then rule then out using the ```@not_implemented_for``` decorator.
- You'll need to add this new module to the docs by editing doc/reference/generators.rst
- You can add yourself as a contributor to CONTRIBUTORS.rst

- For docstring parameters, put the type of input after the ":" for the parameter. Then on the next line start your description of the input parameter. This allows Sphinx to format the type and description correctly.  Take a look at other code if you want an example.

jangwon-yie:
Dear, Mr.Shult

I just made a submission which contains updates reflecting your valuable
comments.

Thanks for the guide.

Jangwon Yie

2019년 11월 4일 (월) 오후 12:48, Dan Schult <notifications@github.com>님이 작성:

PR #3706, opened 2019-11-05
Add an implementation of wheel_graph and its unit tests

jangwon-yie:


dschult:
There is already a ```wheel_graph``` function in generators/classic.py

jangwon-yie:
I did not catch that it is already contained. The pull request should be closed.

PR #3707, opened 2019-11-05
Use canonical url to help search engines

jarrodmillman:
This should help search engines return the stable docs as the canonical source.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -10,20 +10,11 @@
 #
 # All configuration values have a default value; values that are commented out
 # serve to show the default value.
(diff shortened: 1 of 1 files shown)


dschult:
Very glad to see this!                    Maybe it fixes #2730

jarrodmillman:
I am not sure this is the correct approach (I don't have any experience with search engine optimization stuff).  It will probably take a bit before we see any changes in search results.  So we should wait a few weeks after it's merged and reassess the situation.  Hopefully, this will help.

PR #3713, opened 2019-11-08
K-truss is defined for edges being in (k-2) triangles and not for k triangles

ogreen:
NX is pruning edges incorrectly for k-truss. Specifically, NX is pruning edges that belong to fewer than k triangles. The correct definition is to remove edges that belong to fewer than (k-2) triangles. As such. NX is pruning edges in an aggressive manner.
The following links point to several verified sources that have the definition of k-truss (including the original paper by J. Cohen).
- https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.505.7006&rep=rep1&type=pdf  (page 6)
- https://graphchallenge.mit.edu/sites/default/files/documents/SubGraphChallenge-2017-02-09.pdf
- https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.505.7006&rep=rep1&type=pdf (page 2 on the bottom right)

This PR also fixes the problem.

### DIFF
--- a/networkx/algorithms/core.py
+++ b/networkx/algorithms/core.py
@@ -370,7 +370,7 @@ def k_truss(G, k):
     """Returns the k-truss of `G`.
 
     The k-truss is the maximal subgraph of `G` which contains at least three
-    vertices where every edge is incident to at least `k` triangles.
+    vertices where every edge is incident to at least `k-2` triangles.
 
     Parameters
     ----------
--- a/networkx/algorithms/core.py
+++ b/networkx/algorithms/core.py
@@ -369,7 +369,7 @@ def k_corona(G, k, core_number=None):
 def k_truss(G, k):
     """Returns the k-truss of `G`.
 
-    The k-truss is the maximal subgraph of `G` which contains at least three
+    The k-truss is the maximal induced subgraph of `G` which contains at least three
     vertices where every edge is incident to at least `k-2` triangles.
 
     Parameters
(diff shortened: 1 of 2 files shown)


dschult:
@DrPiranoid can you take a look at this?

DrPiranoid:
Hi All,

@ogreen  is correct that the original definition was a k-truss is the induced subgraph from edges which are part of (k-2) or more triangles. I cited that paper in the references as well. It's likely Cohen chose that definition so that the k-clique was a k-truss. (This would have to be changed in this PR documentation to be consistent.)

I chose a different convention because i think the k-2 confuses things. This convention is followed in the first reference I cited. I think this notation gives a better understanding of the thing being counted - triangles.

For example the triangle is a 1-truss in my notation and a 3-truss in Cohen's. With that in mind they're otherwise identical.

Either way you decide we should expand the docstring to make the convention more explicit. My description is correct but very terse and doesn't acknowledge the other convention.

Let me know what you decide.

ogreen:
I agree with @DrPiranoid view that k-truss definition is confusing. Yet, I think that it might be better to go with the standard definition\representation as this is what users might expect.

dschult:
I found k-truss in a book: 
[Community Search over Big Graphs By Xin Huang, Laks V.S. Lakshmanan, Jianliang Xu](https://books.google.com/books?id=FdCnDwAAQBAJ&pg=PA24&lpg=PA24&dq=clique+truss&source=bl&ots=s9N2KBRqUw&sig=ACfU3U0qDkNmBE-_5sbrRRyo9i8kM3lCxw&hl=en&sa=X&ved=2ahUKEwizwsj3juPlAhXIMd8KHeYHDWEQ6AEwBHoECAgQAQ#v=onepage&q=clique%20truss&f=false)
that also uses Cohen's definition.

So, I think we need to go with the established definition unless there is a reason stronger than preference.

ogreen:
@dschult , the code in this PR already addresses the fix. Its really a one line change.

dschult:
@ogreen Understood.  Could you add to the description in the doc_string some more detail.
- make it clear whether the subgraph in an induced subgraph.
- explain that some papers (can refer to the reference 1 paper) use a different definition that is a simple change of the original Cohen definition (refer to 2) and that the implemented definition is Cohen's.
Thanks!

ogreen:
@dschult , done.

dschult:
Thanks!

PR #3714, opened 2019-11-09
WIP: Remove boiler plate from top of modules

jarrodmillman:
The copyright and author stuff is not necessary, out-of-date,
and inconsistent.  It takes up visual space and is a pain to
police everyone doing the same thing on the top of the module.

Git handles authorship in a comprehensive and authoritative way.
The LICENSE.txt file applies to all project code.


jarrodmillman:
@dschult  As far as I can tell we no longer need `# -*- coding: utf-8 -*-` anywhere.  It was needed for Python 2.  It also looks like Emacs may have needed it before version 23, which was released in 2009.  I believe I have removed it from all files now.

The only place we need the copyright information (other than in the LICENSE) is in cases like `networkx/algorithms/centrality/second_order.py`.  I removed it from the `examples`.  I also simplified `networkx/__init__.py`.

I removed the unused `#!/usr/bin/env python` from the test files.

I compared the following files and their generated html versions:
```
networkx/algorithms/node_classification/lgc.py
reference/algorithms/node_classification.html#module-networkx.algorithms.node_classification.lgc

networkx/algorithms/non_randomness.py
reference/algorithms/non_randomness.html#module-networkx.algorithms.non_randomness

networkx/algorithms/operators/all.py
reference/algorithms/operators.html#module-networkx.algorithms.operators.all

networkx/algorithms/operators/binary.py
reference/algorithms/operators.html#module-networkx.algorithms.operators.binary

networkx/algorithms/operators/unary.py
reference/algorithms/operators.html#module-networkx.algorithms.operators.unary

```
I can't see what difference (if any) the variations in the modules' docstring format make in the generated html.  So I will leave it alone for now.

Please let me know if you have any more thoughts or questions.

dschult:
That's great!  Thanks!!

jarrodmillman:
I added anyone whose name was removed from a comment to the list of contributors (CONTRIBUTORS.rst) if there name was missing from the list.  Their contributions are also recorded in git.

PR #3715, opened 2019-11-09
Remove old Python 2 code

jarrodmillman:
This removes some of the last remaining bits of Python 2 support.

@dschult Should we deprecate `literal_stringizer` and `literal_destringizer` in `networkx/readwrite/gml.py` (and remove it in nx 2.6)?  According to the documentation,
```
For better interoperability of data generated by Python 2 and Python 3,
we've provided `literal_stringizer` and `literal_destringizer`.
```

### DIFF
--- a/networkx/algorithms/clique.py
+++ b/networkx/algorithms/clique.py
@@ -17,10 +17,6 @@ from collections import deque
 from itertools import chain
 from itertools import combinations
 from itertools import islice
-try:
-    from itertools import ifilter as filter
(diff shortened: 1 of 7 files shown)


dschult:
Yes, let's deprecate those two functions. That'll simplify that code some I hope...

jarrodmillman:
I will create a second PR to deprecate  `literal_stringizer` and `literal_destringizer`.

PR #3719, opened 2019-11-12
Py36+ updates

jarrodmillman:
This upgrades the syntax to Python 3.6 using two automated tools: [fstringify](https://github.com/jacktasia/fstringify) and [pyupgrade](https://github.com/asottile/pyupgrade).

@dschult We have a bunch of code that uses older syntax.  This upgrades the code to use new syntax.  I know we normally don't do this type of thing, but moving from Python 2 to 3.6 is a bit unusual.  I am fairly confident that this doesn't break anything.  It also makes the code shorter, more readable, and more modern.  It should also make the code slightly faster.

It doesn't get everything, since the automated tools are a bit conservative.  But it greatly reduces the number of things that we will need to fix by hand.  If this seems reasonable to you, I will also try to take care of most of the remaining updates (mostly fstring updates) by hand.

### DIFF
--- a/doc/release/contribs.py
+++ b/doc/release/contribs.py
@@ -1,5 +1,5 @@
 # https://github.com/scikit-image/scikit-image/blob/master/doc/release/contribs.py
-import subprocess
+from subprocess import check_output
 import sys
 import string
 import shlex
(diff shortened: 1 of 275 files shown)


dschult:
I'm in favor of updating the code to take advantage of py36 features.
I think this approach is a good start. Looks like mostly switching to set comprehensions and fstrings.
I'll take a closer look soon. Keep it coming! :)

jarrodmillman:
@dschult OK, I am done.  Let me know if you see any potential issues.

dschult:
I've looked through it -- very nice. I learned a couple of things too. Nice to clean up the code. I know such a big commit will probably make it slightly harder to trace back the history of some lines of code, but I think it's worth it.  I don't see any potential issues with this. I can merge it you'd like, or you can when you're ready.

PR #3732, opened 2019-11-22
Update voronoi.py

steph-en-m:
Shouldn't function args also be within back-ticks? I see in some parts of the docstring they are referred to with back-ticks around them but defined without them.


steph-en-m:
Closing this as documentation style is totally different from what I thought.

Issue #3737, opened 2019-12-04
readthedocs error with setup.py


joelmiller:
I've got a package that uses networkx.  When I use readthedocs to build the documentation of my package, I get an error message when it tries to run networkx's setup.py:

> Obtaining networkx from git+https://github.com/networkx/networkx/#egg=networkx (from -r requirements.txt (line 1))
  Updating /home/docs/checkouts/readthedocs.org/user_builds/epidemicsonnetworks/envs/latest/src/networkx clone
  Running command git fetch -q --tags
  Running command git reset --hard -q ab4d8c65759b51fa615837faa61e9436d0350107
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/docs/checkouts/readthedocs.org/user_builds/epidemicsonnetworks/envs/latest/src/networkx/setup.py", line 22
        python_version = f"{sys.version_info[0]}.{sys.version_info[1]}"
                                                                      ^
    SyntaxError: invalid syntax
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /home/docs/checkouts/readthedocs.org/user_builds/epidemicsonnetworks/envs/latest/src/networkx/

Not sure what the issue is.

MridulS:
Make sure you have python 3.6+ as the setup.py file in master branch uses f strings (which were introduced in 3.6)

joelmiller:
That line needs to be fixed then.

You only reach that line if the python version is less than 3.6, and it comes just before the error message explaining that 3.6 is needed.

joelmiller:
it looks like the error message also uses f strings.

MridulS:
Yeah, fixed it https://github.com/networkx/networkx/pull/3738

PR #3738, opened 2019-12-04
remove f strings from setup.py for clear error message < py3.6

MridulS:
fixes https://github.com/networkx/networkx/issues/3737


MridulS:
@jarrodmillman

Issue #3740, opened 2019-12-05
BibTeX link broken in the docs


iliakur:
The following bibtex citation link on [this page of the docs](https://networkx.github.io/documentation/stable/citing.html) is broken:
http://math.lanl.gov/~hagberg/Publications/hagberg-2008-exploring.shtml

I tried to fix it, but Aric Hagberg's website now uses some JS magic to display the BibTeX entries for publications, I don't really know how to turn them into links. I'd be willing to fix this if someone were to give me pointers on how :)

MridulS:
You can change the link to points towards the Scipy website instead for the BibTex and pdf entry 
BibTeX: http://conference.scipy.org/proceedings/SciPy2008/paper_2/reference.bib
PDF: http://conference.scipy.org/proceedings/SciPy2008/paper_2/full_text.pdf

CsatiZoltan:
On the [documentation page](https://networkx.github.io/documentation/stable/citing.html), the BibTeX and the PDF links are still broken.

iliakur:
That page refers to the latest release (2.4) which was published in October 2019 and thus predates this fix.
I'd expect the links to work after the next version.

PR #3741, opened 2019-12-06
fix citation links

iliakur:
Fix for #3740. Thanks @MridulS for the pointer!


dschult:
Can you fix the link to the pdf also? Thanks for this! and Thanks @MridulS

iliakur:
Sorry, somehow that link seemed to have worked for me (but it doesn't now, confusing...)

Thanks for the catch, all fixed now :)

Issue #3727, opened 2019-11-19
Adding triadic analysis from sociology


bakerwho:
I'm proposing to augment NetworkX with functionality for triadic analysis from sociology. I welcome comments and suggestions about how to organize this as a potential PR. This is for a course project, but I would love to actually contribute my code to NetworkX.

##### Introduction to sociological triads
Sociology has long dealt with the concept of triadic interactions. As conceptualized by late 20th century sociologist Georg Simmel, a dyad (or interaction between two parties) is simple - it can be positive or negative, strong or weak. However, studying triadic interactions unearths a significant degree of complexity. A simple example is Simmel's idea of triadic closure - that a strong positive relation between A and B, and a strong positive relation between B and C, imply that there should be a strong positive relation between A and C. Consider a set of three friends, or three members of a family, or three allies in global politics - the rule is strangely pervasive.

This has led to several notions and theories of behavior based of these triads, constituting a type of analysis that is now known as structural interactionism. A huge body of work analyses various mechanisms and patterns that can be identified from simple triadic analysis.

##### Python for triadic analysis
NetworkX has limited provisions for triadic analysis. The only functionality I could find was `triadic_census()`, which simply counts the number of triads in a given directed graph (see [here](https://networkx.github.io/documentation/networkx-2.2/reference/algorithms/triads.html)).

Triadic analysis will likely get very expensive for large networks. However, the goal is first to have a proof of concept that it can be done in conjunction with NetworkX.

##### Proposal
I propose to extend and adapt NetworkX code to simple sociological triadic analysis. I can begin by adding the following functions on a graph `G` (to be expanded):
 - `G.get_triadic_closures()`  - list of closed triads within the graph
 - `G.get_focal_closures(attr_name)` - list the triads which are closed around a node attribute `attr_name`
 - `G.get_stable_triads(stability_func)` - count triads that fulfill a given criterion of stability (a function that takes the graph, 3 nodes, and a criterion on them, eg: focal closure). The idea is to count unstable triads. These could be expected to stabilize either by formation or deletion of existing edges

The following are more crudely thought-out functions for sociological analysis that can be implemented using NetworkX:
 - `G.predict_edge_deletion()`
 - `G.predict_edge_formation()`
 - `G.simulate_triadic_mechanism()`

ericmjl:
@bakerwho I'm super in favour of this! Though I do also want to hear your thoughts: what about making it part of the `nx.___` API as a function, rather than as a class method? 

[...]

PR #3742, opened 2019-12-07
Adding triadic analysis functions

bakerwho:
Re: #3727
**Methods added:**
in `networkx/algorithms/triads.py`:
- `all_triplets()`
- `all_triads()`
- `triad_type()`
- `triads_by_type()`

**Tests added in `networkx/algorithms/tests/test_triads.py` for all the above**

**Notes**
- `triad_type()` implements a (perhaps inefficient) algorithm to identify the triadic type of a triad (an order-3 DiGraph) (good resource on this [here](http://www.stats.ox.ac.uk/~snijders/Trans_Triads_ha.pdf)).
- `all_triplets` and `all_triads` are generator functions. As the number of possible triads in a graph grows like O(n^3), I was wondering if I should print a warning for larger than some number of nodes (say 100) in the DiGraph.

**TODO**
Empty functions that I will soon write are:
- `random_triad()`
- `triadic_closures()`
- `focal_closures()`
- `balanced_triads()`

@ericmjl @dschult @jarrodmillman @MridulS happy to hear your feedback and suggestions.


bakerwho:
Not sure what went wrong with this AppVeyor build (relevant console output below).
```
utils/tests/test_misc.py::test_make_str_with_bytes
22034utils/tests/test_misc.py::test_make_str_with_unicode
22035  c:\python36-x64\lib\site-packages\networkx\utils\misc.py:109: DeprecationWarning: make_str is deprecated and will be removed in 2.6. Use str instead.
22036    warnings.warn(msg, DeprecationWarning)
22037
22038-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

Also unable to diagnose the Travis CI build failure. Can anyone help?

networkx/algorithms/triads.py
```diff
@@ -5,6 +5,8 @@
 """Functions for analyzing triads of a graph."""
 
 from networkx.utils import not_implemented_for
+from itertools import combinations, permutations
+from collections import defaultdict
 
 __all__ = ['triadic_census']
```
MridulS: Travis fails because the `__all__` var isn't updated with the new functions in the file.
should look something like this
``` python
__all__ = ['triadic_census', 'all_triplets', 'all_triads', 'triads_by_type', 'triad_type']

```

bakerwho: Thanks, will fix.

networkx/algorithms/triads.py
```diff
@@ -54,7 +56,7 @@ def triadic_census(G):
     Returns
     -------
     census : dict
-       Dictionary with triad names as keys and number of occurrences as values.
+       Dictionary with triad type as keys and number of occurrences as values.
```
ericmjl: @bakerwho I'm wondering if there might be a good opportunity for educating end-users using the docs by adding  a very brief summary of what the 16 possible types are? To be clear, I don't mean a complete elucidation of them, but more along the lines of the semantic meaning of each of those triad types? Providing this in the docs might be helpful for newcomers to network science, and make NetworkX an even better resource for learning.

bakerwho: @ericmjl this is a great idea. As linked in the original PR, Tom Snijders's slides ([here](http://www.stats.ox.ac.uk/~snijders/Trans_Triads_ha.pdf)) on this are a great resource.

Here's a TLDR version:
There can be 6 unique edges in a triad (order-3 DiGraph) (so 2^^6=64 unique triads given 3 nodes). These 64 triads each display exactly 1 of 16 topologies of triads (topologies can be permuted). These topologies are identified by the following notation:
{m}{a}{n}{type} (for example: 111D, 210, 102), where
- {m} = number of mutual ties (takes 0, 1, 2, 3); a mutual tie is (0,1) AND (1,0)
- {a} = number of assymmetric ties (takes 0, 1, 2, 3); an assymmetric tie is (0,1) BUT NOT (1,0) or vice versa
- {n} = number of null ties (takes 0, 1, 2, 3); a null tie is NEITHER (0,1) NOR (1,0)
- {type} = a letter (takes U, D, C, T) corresponding to up, down, cyclical and transitive. This is only used for topologies that can have more than one form (eg: 021D and 021U). Usage is better understood from the following diagram (taken from Snijders' slides):

Screen Shot 2019-12-07 at 6 01 27 PM

ericmjl: Wonderful! I think the tl;dr you provided is a very nice and concise description. What do you think about adding it in, then? I propose we go for it! :smile:

bakerwho: Sure, I'll add it to the docstring! Also, @ericmjl could you maybe take a look at the Appveyor and Travis reports and help me fix it? They're both failing. Thanks!

bakerwho: Sure, I'll add it in the docstring!

ericmjl:
@dschult sorry to bother you about this, but I'm finding it hard to debug why the CI builds have been failing. Do you have any insight here?

bakerwho:
@ericmjl thanks for that! I clearly forgot to initialize `expected` as a `defaultdict` instead of a normal `dict`! Will fix and try again.

networkx/algorithms/tests/test_triads.py
```diff
@@ -13,3 +14,90 @@ def test_triadic_census():
...
+    for triad in all_triads:
+        name = nx.triad_type(triad)
+        expected[name].append(triad)
+    actual = nx.triads_by_type(G)
+    assert set(actual.keys()) == set(expected.keys())
+    for tri_type, actual_Gs in actual.items():
+        expected_Gs = expected[tri_type]
+        for a in actual_Gs:
+            assert any(nx.is_isomorphic(a, e) for e in expected_Gs)
+    pass
```
MridulS: @bakerwho can you remove this dangling instance of `pass`

dschult:
Clicking on the codecov /patch details and then on the "Diff" tab, shows the code with each line colored by whether that line was run during a test. Most lines are green (tested), some are red (never ran during a test) and some are yellow (conditional statements that ran with one outcome but not all possible outcomes). 

It looks like it wants you to add tests for bad inputs. For example, test what happens when the input graph is not a valid triad in ```triad_type()```.   You can use the ```pytest``` function ```raises()``` to test that a function call raises an exception.

I think you should also comment out the functions that aren't implemented yet (and remove them from ```__all__```). That will improve the code coverage, but more importantly won't lead to confusion for people using the latest version. You can uncomment them when you implement them.

I would suggest you fix any errors and get docs as you like, then we merge this PR before you start working on the other functions. Ask again if I didn't explain clearly.

networkx/algorithms/tests/test_triads.py
bakerwho: Done.

dschult:
One other comment -- you have long strings of if statements but they don't cover all cases. I mean, they might cover all the possible cases, but the code doesn't use "else"... it finishes with ```elif```.  What should the code do if it falls through all the ```elif``` clauses? A brief look at the code without thinking hard makes it look like it could return ```None```.

bakerwho:
@dschult thanks for the comments. Addressing them in order:

- "test what happens when the input graph is not a valid triad in triad_type()"
Any order-3 DiGraph will necessarily be one of the 16 triad types. The only way this fails is if the DiGraph does not have 3 nodes (ie it is not a triad). Currently, I'm using an `assert` statement in triad_type() to ensure that this is the case. How do you suggest I do it instead? If I'm raising an exception, which one should I raise (I'm guessing ValueError?).

- "comment out the functions that aren't implemented yet (and remove them from __all__)"
This makes perfect sense, doing it right away.

- "you have long strings of if statements but they don't cover all cases."
As far as I have checked, they do cover all cases. Firstly, the number of edges in a triad can take {0, 1, 2.., 5, 6} and **no other number**. So the `elif` should be appropriate outside it.
Secondly, I don't have an more efficient algorithm for detecting triad types than matching topologies. So, for 3 or more edges, I have to check all permutations of the edgeset, eg. (e1, e2, e3), (e1, e3, e2) and so on. If for **any** of these permutations, **any** of the conditions holds - that would slot the triad immediately into that triad type. 

Consider the code below:
```
    elif num_edges == 3:
        for (e1, e2, e3) in permutations(G.edges(), 3):
            if set(e1) == set(e2):
                if e3[0] in e1:
                    return "111U"
                if e3[1] in e1:
                    return "111D"
            elif set(e1).symmetric_difference(set(e2)) == set(e3):
                if {e1[0], e2[0], e3[0]} == {e1[0], e2[0],
                                             e3[0]} == set(G.nodes()):
                    return "030C"
                if e3 == (e1[0], e2[1]) and e2 == (e1[1], e3[1]):
                    return "030T"
```
This is equivalent to saying that:
for 3 edges
   for any permutation of edges
      if first and second are symmetric
          if third has a source in first: return '111U'
          if third has a sink in first: return '111D'
     else if third edge is made of nodes that are the symmetric difference of the first and second: 
          if cyclical: return "030C"
          if transitive: return "030T"

As you can see, each of the nested ifs has to be evaluated for each permutation. Evaluating to True implies necessary and sufficient conditions for exit.

I am fairly confident that my implementation and logic are correct. However, getting None as a return value would be a sure sign of an error somewhere.

ericmjl:
@bakerwho I think I have an idea for this one:

> Any order-3 DiGraph will necessarily be one of the 16 triad types. The only way this fails is if the DiGraph does not have 3 nodes (ie it is not a triad). Currently, I'm using an assert statement in triad_type() to ensure that this is the case. How do you suggest I do it instead? If I'm raising an exception, which one should I raise (I'm guessing ValueError?).

Yes, instead of an assert statement in `triad_type()`, I would use an if/raise check, briefly:

```python
if not some_condition:
    raise SomeError()
```

It's better than asserts, because asserts can be disabled in Python at the command line (though I haven't run into any real-world scenario yet). In any case, it's idiomatic to use the if/raise pattern.

Perhaps raising one of the errors defined [here](https://github.com/networkx/networkx/blob/aeeeef92f2229b0cc8ebbbbfcfc87128f546d55f/networkx/exception.py) might be good? One proposal I have here is `NetworkXAlgorithmError`, what do you think?

@dschult I just looked through the details of the `triad_type()` function once more, and I think it's correct as well. At least, in `test_triads.py`, all 16 possible cases. It might not be a fruitful avenue to test further, except maybe for checking that an appropriate error is raised if an invalid graph is entered? I can think of at least the following:

- Graph that doesn't have 3 nodes.
- Graph with self-loops.

We're getting there :tada:!

dschult:
Great!  I figured it was something like that.
To make it clear to future coders looking at your stuff. 
Can you remove the last if statement in each of these cases?
Maybe leave a comment as to which case it is.

```python
for (e1, e2, e3) in permutations(G.edges(), 3):
            if set(e1) == set(e2):
                if e3[0] in e1:
                    return "111U"
                if e3[1] in e1:
                    return "111D"
```
becomes
```python
for (e1, e2, e3) in permutations(G.edges(), 3):
            if set(e1) == set(e2):
                if e3[0] in e1:
                    return "111U"
                # e3[1] in e1:
                return "111D"
```

bakerwho:
@ericmjl @dschult Happy holidays! Got some time to work on this, and the new commit implements:

1. `is_triad()` to check if an input G is a triad
2. `random_triad()` to generate a random triad from a DiGraph G 
3. Error handling (`NetworkXAlgorithmError` raised when a triadic algo is called on a non-triad)
4. Tests for functions in 1 and 2
5. Style fixes (such as commenting out the last if-else parts of `triad_type`)

An input to `is_triad()` needs to satisfy 3 conditions:

- is an instance of nx.Graph
- is a *directed* graph containing 3 nodes
- contains no self edges (technically, it could contain these, but those self-edges would have to be ignored)

ericmjl:
@dschult - just marked approval on the PR. I think it's comprehensively tested enough, and the docs are definitely well-written and clear.

bakerwho:
@ericmjl of course! Happy to help! And thanks for all the help and feedback!

dschult:
The code as written doesn't run. Can you change the imports at the top to something like:

```python
"""Functions for analyzing triads of a graph."""

from itertools import combinations, permutations
from random import sample

import networkx as nx
from networkx.utils import not_implemented_for

__all__ = ['triadic_census', 'is_triad', 'all_triplets', 'all_triads',
           'triads_by_type', 'triad_type', 'random_triad']
```

Then using ```pytest tests/test_triads.py``` shows a number of errors many due to checking ```is_triad``` on the input graph which doesn't have to be a triad!  Simple fixes I believe... :)
Thanks!

dschult:
Thanks for this and Thanks @ericmjl for working with @bakerwho

Issue #3746, opened 2019-12-10
Linestyle not set for directed graph with errors at draw_networkx_edges


tombeek111:
When plotting a directed graph, linestyle is not set

    import networkx as nx
    from matplotlib import pyplot as plt

    G = nx.DiGraph()
    G.add_nodes_from([0,1])
    G.add_edge(0,1)

    pos = {0 : (0,0), 1 : (1,0)}

    plt.figure()
    nx.draw_networkx_nodes(G,pos=pos)
    nx.draw_networkx_edges(G,pos=pos,style='dashed',arrows=True)

    plt.show()
    
This results in a solid line between two nodes, instead of a dashed one. If arrows is set to False, the line is dashed, but of course there is no arrow anymore.

tombeek111:
Adding the argument fixed it:

https://github.com/tombeek111/networkx/tree/bugfix-for-issue-3746

PR #3747, opened 2019-12-11
added linestyle as argument

tombeek111:
When plotting a directed graph, linestyle is not set

    import networkx as nx
    from matplotlib import pyplot as plt

    G = nx.DiGraph()
    G.add_nodes_from([0,1])
    G.add_edge(0,1)

    pos = {0 : (0,0), 1 : (1,0)}

    plt.figure()
    nx.draw_networkx_nodes(G,pos=pos)
    nx.draw_networkx_edges(G,pos=pos,style='dashed',arrows=True)

    plt.show()
    
This results in a solid line between two nodes, instead of a dashed one. If arrows is set to False, the line is dashed, but of course there is no arrow anymore.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -674,6 +674,7 @@ def draw_networkx_edges(G, pos,
                                     color=arrow_color,
                                     linewidth=line_width,
                                     connectionstyle=connectionstyle,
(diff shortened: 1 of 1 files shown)


dschult:
Fixes #3746 
Thanks very much for this! :)

Issue #3470, opened 2019-06-06
GraphML output not using <edge id="XXX">


peterjc:
The NetworkX GraphML output seems to demote edge ``id`` to a named attribute.

The GraphML primer has an example with node colors encoded as attributes of selected nodes, and weights on edges:

http://graphml.graphdrawing.org/primer/graphml-primer.html#Attributes

Example on macOS with Python 3.7 and NetworkX 2.3

```
$ curl -L -O http://graphml.graphdrawing.org/primer/attributes.graphml
$ python3 --version
Python 3.7.0rc1
$ python3 -c "import networkx as nx; print(nx.__version__)" 
2.3
$ python3 -c "import networkx as nx; nx.readwrite.graphml.write_graphml(nx.readwrite.graphml.read_graphml('attributes.graphml'), 'attributes2.graphml')"
```

Input ``attributes.graphml``:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- This file was written by the JAVA GraphML Library.-->
<graphml xmlns="http://graphml.graphdrawing.org/xmlns"  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
  <key id="d0" for="node" attr.name="color" attr.type="string">
    <default>yellow</default>
  </key>
  <key id="d1" for="edge" attr.name="weight" attr.type="double"/>
  <graph id="G" edgedefault="undirected">
    <node id="n0">
      <data key="d0">green</data>
    </node>
    <node id="n1"/>
    <node id="n2">
      <data key="d0">blue</data>
    </node>
    <node id="n3">
      <data key="d0">red</data>
    </node>
    <node id="n4"/>
    <node id="n5">
      <data key="d0">turquoise</data>
    </node>
    <edge id="e0" source="n0" target="n2">
      <data key="d1">1.0</data>
    </edge>
    <edge id="e1" source="n0" target="n1">
      <data key="d1">1.0</data>
    </edge>
    <edge id="e2" source="n1" target="n3">
      <data key="d1">2.0</data>
    </edge>
    <edge id="e3" source="n3" target="n2"/>
    <edge id="e4" source="n2" target="n4"/>
    <edge id="e5" source="n3" target="n5"/>
    <edge id="e6" source="n5" target="n4">
      <data key="d1">1.1</data>
    </edge>
[...]

PR #3763, opened 2019-12-27
MultiGraph from graphml with explicit edge ids #3470

kiryph:
Prior this commit, a graph with explicit edge ids
but no parallel edges stored in a graphml was converted to 
a Graph and the edge ids were moved to the edge data.
This behaviour has been removed.


dschult:
This looks like a good suggestion to me!

Can you also add tests to make sure we check what happens with each case?
The tests that are failing were checking that it did one thing and now that we have changed it we need to check the new behavior.
Thanks!

kiryph:
@dschult I am glad to hear that you think this is a good suggestion.

As requested I have updated and added tests:

- Updated three existing test data:
  1. `simple_directed_data` by removing edge `id=foo`
  2. `attribute_data` by removing ids `e0`, ... `e6` for a directed non-multigraph
  3. `simple_undirected_data` by removing edge `id=foo`
- Updated the existing test:
  1. `test_yfiles_extension` where the yfile data contains an edge id
- Removed the test which explicitly tests the removed behaviour
  1. `test_multigraph_to_graph`
- Added tests based on newly added test data
  1. `test_read_undirected_multigraph_graphml`
  2. `test_read_undirected_multigraph_no_multiedge_graphml`
  3. `test_read_undirected_multigraph_only_ids_for_multiedges_graphml`

Could you give me feedback, if there is something else which should be covered?

dschult:
Thanks for all this -- especially the tests...
I think I misunderstood something -- but maybe I'm just missing it at this point.
How do we get the old behavior? That is, how do we read a graphml file with edge_ids but no parallel edges and have it create a Graph instead of a MultiGraph?    (I'm ignoring directed in this response for simplicity.)

Old system:    
- parallel edges -> MultiGraph
- no parallel edges, no edge_ids -> Graph
- no parallel edges, edge_ids -> Graph with edge_ids stored as edge attribute "id"

New system:  third case -> MultiGraph

This makes it possible to keep it a multigraph even if there are no parallel edges so long as edge_id doesn't exist.
But I think it's not possible to make it create a Graph with edge_ids stored as edge attribute "id".

Perhaps we can include a keyword argument to the reader: self.force_multigraph which makes the output a multigraph no matter what.  To be clear, I'm suggesting using the old system (so third option is possible) but allow an override flag to always make it MultiGraph whatever the input.

What do you think? Why did you need this PR, and would this alternative allow what you needed?

kiryph:
> I think I misunderstood something -- but maybe I'm just missing it at this point. ...

You did *not* misunderstand anything when I read your comment.


> How do we get the old behavior? 

This would not be possible. As soon as someone introduces explicit edge ids into the graphml file, the graphml_reader would create a MultiGraph. This would break backwards compatibility.

In your list the case

> no parallel edges, edge_ids -> Graph with edge_ids stored as edge attribute "id"

would not exist anymore. Since this has been introduced into networkx, there might be users who use this behavior. Therefore my proposal is troublesome.

Your suggestion is better.

> What do you think? Why did you need this PR, and would this alternative allow what you needed?

Yes, this alternative would allow me to use graphml as an exchange format between different python scripts.

For my case it is important that the networkx type does not depend on the actual existence of parallel edges in the graphml file. When I operate on the graph parallel edges can be introduced or exist from the beginning but do not necessarily exist always. However, the scripts are much clearer when I do not have to post-process/special-case/convert graph types. I simply have always MultiGraph (actually MultiDigraph) and I am happy.

Should I write a new PR which introduces `self.force_multigraph` as an option for the graphml_reader?

kiryph:
Here is a diff introducing `force_multigraph` for `nx.read_graphml(<file>, force_multigraph=True)`
```diff
diff --git a/networkx/readwrite/graphml.py b/networkx/readwrite/graphml.py
index df4982c4..a0adfc17 100644
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -196,7 +196,7 @@ def generate_graphml(G, encoding="utf-8", prettyprint=True, named_key_ids=False)


 @open_file(0, mode="rb")
-def read_graphml(path, node_type=str, edge_key_type=int):
+def read_graphml(path, node_type=str, edge_key_type=int, force_multigraph=False):
     """Read graph in GraphML format from path.

     Parameters
@@ -211,12 +211,16 @@ def read_graphml(path, node_type=str, edge_key_type=int):
     edge_key_type: Python type (default: int)
        Convert graphml edge ids to this type as key of multi-edges

+    force_multigraph: bool (optional)
+        If True returns always either MultiGraph or MultiDigraph.

     Returns
     -------
     graph: NetworkX graph
         If no parallel edges are found a Graph or DiGraph is returned.
         Otherwise a MultiGraph or MultiDiGraph is returned.
+        If force_multigraph is True, a  MultiGraph or MultiDiGraph is always
+        returned.

     Notes
     -----
@@ -248,7 +252,11 @@ def read_graphml(path, node_type=str, edge_key_type=int):
     the file to "file.graphml.gz".

     """
-    reader = GraphMLReader(node_type=node_type, edge_key_type=edge_key_type)
+    reader = GraphMLReader(
+        node_type=node_type,
+        edge_key_type=edge_key_type,
+        force_multigraph=force_multigraph
+    )
     # need to check for multiple graphs
     glist = list(reader(path=path))
     if len(glist) == 0:
@@ -743,10 +751,11 @@ else:
 class GraphMLReader(GraphML):
     """Read a GraphML document.  Produces NetworkX graph objects."""

-    def __init__(self, node_type=str, edge_key_type=int):
+    def __init__(self, node_type=str, edge_key_type=int, force_multigraph=False):
         self.node_type = node_type
         self.edge_key_type = edge_key_type
-        self.multigraph = False  # assume multigraph and test for multiedges
+        self.multigraph = True if force_multigraph else False  # assume multigraph
+        # (except user forces multigraph) and test for multiedge
         self.edge_ids = {}  # dict mapping (u,v) tuples to id edge attributes

     def __call__(self, path=None, string=None):
@@ -853,6 +862,8 @@ class GraphMLReader(GraphML):
                 pass
         else:
             edge_id = data.get("key")
+            if edge_id is None and self.force_multigraph:
+                edge_id = G.size() # use edge count as id

         if G.has_edge(source, target):
             # mark this as a multigraph
```

PR #3766, opened 2019-12-28
shortest paths: use internal nodes

bforte:
This implements the changes proposed in #3765 and also adds some more sanity checks.

Regarding the newly introduced function `.get_node` on graphs, you guys probably must decide how to name it and whether you want to make it publicly documented or a dunder method, rename it etc.

### DIFF
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -215,8 +215,8 @@ def bidirectional_shortest_path(G, source, target):
     """
 
     # make sure both source and target are contained in G
-    source = G.get_node(source, what='Source')    
-    target = G.get_node(target, what='Target')    
+    source = G.get_node(source, what='Source')
+    target = G.get_node(target, what='Target')
 
     # call helper to do the real work
     results = _bidirectional_pred_succ(G, source, target)
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -214,9 +214,9 @@ def bidirectional_shortest_path(G, source, target):
     This algorithm is used by shortest_path(G, source, target).
     """
 
-    if source not in G or target not in G:
-        msg = 'Either source {} or target {} is not in G'
-        raise nx.NodeNotFound(msg.format(source, target))
+    # make sure both source and target are contained in G
+    source = G.get_node(source, what='Source')    
+    target = G.get_node(target, what='Target')    
 
     # call helper to do the real work
     results = _bidirectional_pred_succ(G, source, target)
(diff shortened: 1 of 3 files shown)


dschult:
Thanks for this!

I think the problem occurs in bellman_ford when a target is used to build the path.
The path ends with the input target object rather than the matching target object stored on the graph. Changing [one line](https://github.com/networkx/networkx/blob/b59935c8b114a3fbef5a5e6da3e3028d42cfc076/networkx/algorithms/shortest_paths/weighted.py#L1305) of the bellmen_ford function could fix things.
```python
dsts = {n for n in G if n == target} if target is not None else pred
```

Another workaround is to call bellman_ford with the object you want as the target:
```python
nx.shortest_path(G, Foo(0), {n for n in G if n == Foo(1)}.pop(), method="bellman-ford", weight=w) 
```

But this is probably better solved by "fixing" the equality checker of your objects than messing with the shortest_path routines. The bellman_ford code assumes that equality of the nodes implies that returning either object is equivalent. And we know that might not be true if you do some fancy custom object. But shouldn't that imply that you correct your definition of equality?  The correction I am suggesting is to make ```__eq__``` test everything that is "important" about the object to make sure they are the same before returning True.  You get to determine what is important. And if you have different meanings of that in different situations then you should probably rethink your object structure... (?)

bforte:
I agree that it is not proper to define such equality, the "problem" arose in the following scenario:

Say you have a structure of objects that is a network/graph where objects are uniquely identified by a number (eg. phone number, ip address, uuid or similar). And you will do this (sticking to the phone number example):

 - receive the information of the structure with some meta data (say last name of the participant)
 - build the graph with these objects
 - later you want to query shortest paths in the structure, but you only have access to the phone number not the name anymore

At the moment, this will require to keep the meta information separate from the graph. For example you will need to populate the graph with phone numbers only, and you will need to keep a structure mapping phone numbers to names (or access `G._node`).
When you retrieve the shortest path you will then need to query the other structure at every node for the name which gets awkward when the domain basically is just a graph imo.

When I was using the library, to me the documentation read that the library only depends on the hash of the objects. So I thought that I don't need to maintain such a mapping from number to meta data, but just define the hash as this unique number and will receive all the information I will need in the shortest path (it is stored in the graph anyway).

I am not using the library at the moment since I processed all the data, but thought I'd put this here anyway since it's not a big change and won't change things for usual use cases. Another way, that probably incentivises better engineering practices but would be more work (more methods and documentation that face the clients), would be to maintain methods that allow querying by hashes directly (eg. `nx.shortest_path_by_hash(hash_of_source, hash_of_target[, ...])`).

dschult:
NetworkX definitely does *not* only use the hash of a node.  It stores the node in a dict, so it uses a hash for sure. But dicts don't just use the hash. When there is a "collision" (two keys with the same hash) it checks if they are the same key using ```__eq__```.  That's why we say that nodes must be hashable. But we don't say they only need ```__hash__```. I believe they need ```__hash__``` and ```__eq__``` and nothing else, but I might be wrong about that. It's whatever Python dicts use.

PR #3768, opened 2020-01-02
WIP: Fix scipy deprecation warnings

jarrodmillman:
@dschult This PR replaces things like `scipy.array` with `numpy.array`.  SciPy will remove these when they eventually release 2.0 and they are making noise in our test suite now.  Everything is straightforward except `scipy.sqrt`, which maps to `numpy.lib.scimath.sqrt` instead of `numpy.sqrt`.  Here is the difference:
```
>>> import numpy as np  
>>> import scipy as sp      
>>> sp.sqrt(-1)
1j
>>> np.lib.scimath.sqrt(-1)   
1j
>>> np.sqrt(-1)  
__main__:1: RuntimeWarning: invalid value encountered in sqrt
nan
```

Should we continue allowing negative square roots in this context?  I suspect not and am tempted to use `numpy.sqrt`.  I highlighted the two instances below.

### DIFF
--- a/networkx/linalg/laplacianmatrix.py
+++ b/networkx/linalg/laplacianmatrix.py
@@ -3,14 +3,16 @@
 import networkx as nx
 from networkx.utils import not_implemented_for
 
-__all__ = ['laplacian_matrix',
-           'normalized_laplacian_matrix',
-           'directed_laplacian_matrix',
-           'directed_combinatorial_laplacian_matrix']
+__all__ = [
+    "laplacian_matrix",
+    "normalized_laplacian_matrix",
+    "directed_laplacian_matrix",
+    "directed_combinatorial_laplacian_matrix",
+]
 
 
-@not_implemented_for('directed')
-def laplacian_matrix(G, nodelist=None, weight='weight'):
+@not_implemented_for("directed")
+def laplacian_matrix(G, nodelist=None, weight="weight"):
     """Returns the Laplacian matrix of G.
 
     The graph Laplacian is the matrix L = D - A, where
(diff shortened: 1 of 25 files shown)


networkx/linalg/laplacianmatrix.py
```diff
@@ -118,8 +119,8 @@ def normalized_laplacian_matrix(G, nodelist=None, weight='weight'):
     D = scipy.sparse.spdiags(diags, [0], m, n, format='csr')
     L = D - A
     with scipy.errstate(divide='ignore'):
-        diags_sqrt = 1.0 / scipy.sqrt(diags)
-    diags_sqrt[scipy.isinf(diags_sqrt)] = 0
+        diags_sqrt = 1.0 / np.lib.scimath.sqrt(diags)
```
jarrodmillman: `np.lib.scimath.sqrt` or `np.sqrt`?

networkx/linalg/laplacianmatrix.py
```diff
@@ -198,9 +199,9 @@ def directed_laplacian_matrix(G, nodelist=None, weight='weight',
     evals, evecs = linalg.eigs(P.T, k=1)
     v = evecs.flatten().real
     p = v / v.sum()
-    sqrtp = sp.sqrt(p)
+    sqrtp = np.lib.scimath.sqrt(p)
```
jarrodmillman: `np.lib.scimath.sqrt` or `np.sqrt`?

dschult:
Yes, let's use ```np.sqrt```!

I believe this could raise an exception if someone was using negative edge weights that are sufficiently negative to make the weighted degree be negative for some node. But none of the theory with normalized laplacians consider that case. Even the very rare negative edge weights examples I can find make sure that the weighted degree of the nodes is positive. With the more simple ```np.sqrt``` such a user pushing the boundaries of theory will be obviously flagged in a way that they probably should be. Much better than introducing complex values into their spectrum without notifying them. If they are purposely doing that kind of work, they will know to look out for that issue. If not on purpose, they should be stopped from further confusion. :}

Thanks!

PR #3769, opened 2020-01-04
Update year

jarrodmillman:


dschult:
Yay!!  This is much better than previous years where we had to change every .py file!
:)

PR #3773, opened 2020-01-09
add a check before del operation

DongShuaike:
Hi, I'm using angr-management, a cross-platform binary analysis platform which uses networkx and found sometimes the `DiGraph.remove_node()` runs into errors. I added a check before `del self._pred[u][n]` to fix this.

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -521,7 +521,8 @@ class DiGraph(Graph):
         except KeyError:  # NetworkXError if n not in self
             raise NetworkXError(f"The node {n} is not in the digraph.")
         for u in nbrs:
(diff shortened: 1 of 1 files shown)


dschult:
That check should be unnecessary because the for loop iterates over ```self._succ[n]```. If there is an entry ```self._succ[n][u]``` then there better be an entry ```self._pred[u][n]```.  Can you say more about why this error might be happening? Are you manipulting the data structure outside the usual API in any way? Can you come up with a small example that shows the error?

DongShuaike:
Dear Sir, thank you for the quick response, I have contacted the author of `angr-management` which takes advantages of `networkx` and he/she told me it was a bug in their project. I will close the pull request now.

Issue #3658, opened 2019-10-16
Kamada Kawai layout not outputting 3rd dimension


milcent:
When playing with positions for 3d graphs, contrary to the usual srping layout, when I set it to be 
`pos = nx.kamada_kawai_layout(G, dim=3)`, it doesn't work.
An inspection in the resulting dict will show the same as below, with zeros in the third dimension.
`In [40]: pos
Out[40]: 
{'A': array([0.73517262, 0.0116334 , 0.        ]),
 'D': array([0.99999997, 1.        , 0.        ]),
 'E': array([0.01163338, 0.73517261, 0.        ]),
 'F': array([-0.25083676,  0.27272334,  0.        ]),
 'G': array([-0.91055632, -0.0634198 ,  0.        ]),
 'H': array([-0.79471643, -0.79471641,  0.        ]),
 'I': array([-0.06341983, -0.91055634,  0.        ]),
 'J': array([ 0.27272335, -0.2508368 ,  0.        ])}`

dschult:
I have verified this problem. It stems from using ```circular_layout``` as the initial positions for the nodes when ```dim >= 2```.  But "circular" layouts are two-D with zeros in all dimensions above 2.

Perhaps we can find a good initial position that works for ```dim>2```.

milcent:
ThaT would be nice.

Em qua, 16 de out de 2019 às 23:34, Dan Schult <notifications@github.com>
escreveu:

milcent:
I've checking the code to see if I could help.
Is it True to say that the circular_layout is the initialization for the kamada_kaway, and hence that implementing the third dim to the circular_layout would automatically fix the kamada_kaway??

milcent:
Or maybe initialize the positions with the spring layout, which is already force-directed, and hope that the kamada_kaway_solver will converge appropriately.

dschult:
It is true to say that the circular_layout is the initialization for the kamada_kawai_layout. But it isn't clear how to implement the third dimension for circular_layout.  I mean, it already does work for 3-D, it is still a circle... just in 3D.

Your idea of force-directed as init may work, but check that spring_layout does what you expect in 3D first... :)

stefanhannie:
I am just coming across this issue as well. Would it be appropriate to initialize the vertices on a sphere? This is an analogue of the 2D circle.

dschult:
[...]

PR #3782, opened 2020-01-16
fix initializer for kamada_kawai_layout (networkx #3658)

boothby:
See discussion at #3658


dschult:
This looks good.  Anything else before we merge?

The smoke tests could all use removing ```vpos = ``` for each test. That will improve the pycodestyle for the module too. If you don't get to it, I'll do that.
Thanks!

boothby:
Absolutely; I just went along with the existing pattern, but I'll update the PR today or tomorrow.

Issue #3685, opened 2019-10-23
all_shortest_paths inconsistent in undirected graph


nycnighthawk:
Steps to reproduce below
>>> from networkx import nx
>>> g = nx.Graph()
>>> g.add_path([1, 2, 3, 5, 7, 8])
>>> g.add_path([1, 2, 3, 4, 6, 7, 8])
>>> g.edges[3,4]['weight'] = 0
>>> list(nx.all_shortest_paths(g, 1, 8, weight='weight'))
[[1, 2, 3, 5, 7, 8], [1, 2, 3, 4, 6, 7, 8]]
>>> list(nx.all_shortest_paths(g, 8, 1, weight='weight'))
[[8, 7, 5, 3, 2, 1]]

the result is correct for 1, 8, but incorrect for 8, 1 when associate the edge [3, 4] with weight=0, effectively means both paths are equal from calculation point of view.

dschult:
Thanks for this very simple example!!  Looks like a bug.

boothby:
I traced this issue, and the root cause is that `all_shortest_paths` defaults to using Dijkstra's algorithm, which is only correct for positive weights.  Bellman-Ford does not have this limitation, but the runtime is worse.  I see two options here:

1) default to Bellman-Ford and suffer the runtime hit, or
2) check if all weights are positive to determine which algorithm to dispatch.

dschult:
I believe Dijkstra is correct for non-negative weights.
It should be Ok with zero weights or positive weights if I understand correctly.

boothby:
Okay, I stand corrected.  I have a modification to Dijkstra's algorithm that handles zero-weight edges.  This does kick the problem down the road a little.  Minifying @nycnighthawk's example, I add the test:

```
    def test_all_shortest_paths_zero_weight_edge(self):
        g = nx.Graph()
        nx.add_path(g, [0, 1, 3])
        nx.add_path(g, [0, 1, 2, 3])
        g.edges[1, 2]['weight'] = 0
        paths03 = list(nx.all_shortest_paths(g, 0, 3, weight='weight'))
        paths30 = list(nx.all_shortest_paths(g, 3, 0, weight='weight'))
        assert sorted(paths03) == sorted(p[::-1] for p in paths30)
```

Inside `all_shortest_paths`, we compute predecessors,

    >>> nx.dijkstra_predecessor_and_distance(g, 0, 3, weight='weight')
    {0: [], 1: [0], 3: [1, 2], 2: [1]}
[...]

PR #3783, opened 2020-01-19
Correct handling of zero-weight edges in all_shortest_paths

boothby:
Discussion in #3685.

### DIFF
--- a/networkx/algorithms/shortest_paths/generic.py
+++ b/networkx/algorithms/shortest_paths/generic.py
@@ -489,6 +489,7 @@ def all_shortest_paths(G, source, target, weight=None, method='dijkstra'):
 
     return _all_shortest_simple_paths(G, {source}, target, pred)
 
+
 def _all_shortest_simple_paths(G, sources, target, pred):
     """Compute all shortest simple paths in the graph, from any source in
     sources to target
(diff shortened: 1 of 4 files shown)


dschult:
This all looks good -- and I see you've got a fixme to look at goldberg_radzik too. Also good.

I have some minor comments/questions for you:
- the internal function doesn't seem to need input G. Could we remove it?
- it's name is also very close to ```all_shortest_paths```. What do you think of ```_build_paths_from_predecessors```?
- we currently only call this function using a single source. Should we make the input a single source? (I mean, is there a case you anticipate that uses multiple sources?)
- The name of the set of processed nodes is "hits". Is there an origin of this name? It's not a big deal, but would "seen" be better? we have a function "hits" in NX for link analysis (finding hubs and authorities based on links) and we use "seen" in other similar.

Very nice work!!

boothby:
>     * the internal function doesn't seem to need input G. Could we remove it?

Good catch!

>     * it's name is also very close to `all_shortest_paths`. What do you think of `_build_paths_from_predecessors`?

Sure, I didn't like that name much either. 

>     * we currently only call this function using a single source. Should we make the input a single source? (I mean, is there a case you anticipate that uses multiple sources?)

`weighted._bellman_ford` takes multiple sources, and uses this function already.

>     * The name of the set of processed nodes is "hits". Is there an origin of this name? It's not a big deal, but would "seen" be better? we have a function "hits" in NX for link analysis (finding hubs and authorities based on links) and we use "seen" in other similar.

I frequently use the word "hits" for that purpose in my own projects, but I'm happy to conform to local standards.

> Very nice work!!

Thanks!  It's a pleasure to give back to a project that I've been using for years.

PR #3788, opened 2020-01-22
Adding moral and threshold packages to main namespace

salym:
I made sure that the public function names don't conflict with others in the namespace.
There are interesting functions which are not exposed public interface by the __all__ attribute.
May be someone could figure out which one worth being made public.


dschult:
At some point we decided to leave the threshold module out of the networkx namespace but leave it in the codebase. Looks like ```moral``` was left out by an oversight.

Any comments?

salym:
This makes sense. The choice of relevant functions from the threshold module if highly arbitrary so it makes sense to leaving it it out of the namespace to avoid any future surreptitious conflict. Thanks for letting it know. My 8yo kid is using your wonderful library and learning from your source code so it's good to have such a consistency.

PR #3792, opened 2020-01-23
Implemented topo_layout for DAGs

KOLANICH:


KOLANICH:
index

dschult:
Can you describe what this is and when a user would want to consider it?
Also... any citations/references?

KOLANICH:
No, it's a self-invented heuristics. Completely ineffective. And defective:
* it doesn't address the situation when edges overlap. For example an edge (0, 0) -> (3, 3) completely occludes (1, 1) -> (2, 2), but I don't know how to implement detection of such pairs of edges efficiently.
* it doesn't eliminate unneeded edge intersections

Though the algo is split into 2 functions each of them can be improved separately. But I am not going to do that, for my purposes (drawing a small enough dependency tree of packages) the algo works.

Its only good point is that it takes not so many lines of code. It was the main goal - not to code the whole Sugiyama algo which is more than 1000 lines of code (in C++).

dschult:
That's cool. Please put that into a "Notes" section in the doc_string. It'll help the users avoid having to figure out what it does and when to use it.  (and maybe prod someone to implement the more complex algorithms.)

Issue #3775, opened 2020-01-11
subgraph_is_monomorphic throws an error


kanghj:
`subgraph_is_monomorphic` throws a KeyError in some cases. Here is an example for reproduction.

```
import networkx as nx
import networkx.algorithms.isomorphism as iso

G = nx.MultiDiGraph()
G.add_node(1)
G.add_node(2)
G.add_edge(1,2, label="A")
G.add_edge(2,1, label="B")

SG = nx.MultiDiGraph()

SG.add_node(5)
SG.add_node(6)
SG.add_edge(5,6, label="A")

gm = iso.MultiDiGraphMatcher(G, SG, edge_match=iso.categorical_edge_match('label', None))
gm.subgraph_is_monomorphic()
```

Let me know if I can help to fix this. My guess is that the semantic feasibility check in `algorithms/isomorphism/vf2userfunc.py` should do something different when test='mono', but I will need more pointers on this. 

Running `nx.__version__` returns '2.4'

Edit:
Perhaps a fix would be to iterate the edges of G2 (the subgraph) instead of G1 in the semantic feasibility check: 

```
         if self.test == 'mono':
            core_2 = self.core_2
            # 
            for neighbor in G2_adj[G2_node]:
                # G1_node is not in core_1, so we must handle R_self separately
                if neighbor == G2_node:
                    if not edge_match(G1_adj[G1_node][G1_node],
                                      G2_adj[G2_node][G2_node]):
                        return False
                elif neighbor in core_2:
                    if not edge_match(G1_adj[G1_node][core_2[neighbor]], G2_adj[G2_node][neighbor]):
                        return False
```

dschult:
Thanks for the Issue...  and for the Edit--Proposed Fix. 
 
I can verify that this is happening. Haven't been able to look at it hard yet. 
Seems like a bug apparently not covered by any tests...

dschult:
That'd be great! You might start by adding your example as a test... make sure it fails and then put in the fix. :)

PR #3798, opened 2020-01-25
Prevent KeyError on subgraph_is_monomorphic

kanghj:
For #3775, checks that core_1[neighbor] is in G2's adj before checking edge_match for semantic feasibility.

Checking for monomorphism for MultiDiGraph seems like it may be more subtle than I thought, so this just handles DiGraphs.

### DIFF
--- a/networkx/algorithms/isomorphism/tests/test_isomorphvf2.py
+++ b/networkx/algorithms/isomorphism/tests/test_isomorphvf2.py
@@ -318,3 +318,20 @@ def test_noncomparable_nodes():
     assert dgm.is_isomorphic()
     # Just testing some cases
     assert gm.subgraph_is_monomorphic()
+
+def test_monomorphism_edge_match():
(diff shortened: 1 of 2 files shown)


dschult:
Not sure I've thought through this all the way, but shouldn't we return False also when the edge doesn't exist in G2?  Maybe something like: (but please check I'm thinking about this right.)

```python
if G2_node not in G2_adj[G2_node] or \
                        not edge_match(G1_adj[G1_node][G1_node],
                                       G2_adj[G2_node][G2_node]):
                    return False
```

kanghj:
Hmm, that doesn't seem right to me. Based on my (limited) understanding of subgraph monomorphism, if G2 is a monomorphic subgraph of G1, G1 may contain additional edges that are not contained in G2. i.e. G2 isn't necessarily a node-induced subgraph of G1.

dschult:
Thanks very much for this!

Issue #3774, opened 2020-01-10
Feature request: possibility to pass weight function to A* shortest path search


esidebo:
Is there a reason the A* path finder `astar.astar_path(G, source, target, heuristic=None, weight='weight')` does not allow for a weight function like is possible for e.g. dijkstra? I'd like to hide specific edges depending on their attributes by setting their weight to `None` like is exemplified under "Notes" at https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.weighted.single_source_dijkstra.html#networkx.algorithms.shortest_paths.weighted.single_source_dijkstra. 

 I just tried myself to allow for a function by manually changing after line https://github.com/networkx/networkx/blob/b8460be7b055dc18e2fdde3b8739d41f0c0064a3/networkx/algorithms/shortest_paths/astar.py#L114
to instead of calling `w.get(weight, 1)` do:
```
wgt = weight(curnode, neighbor, w)
if wgt is None:
    continue # ignore edge
ncost = dist + wgt
```
I didn't test this thoroughly but seems to work nicely on my small example (i.e. edges with `None` are avoided in the solution). If I'm not missing something, perhaps this could be added?

dschult:
See #3754
We have a number of places where we should be using a weight function (or string) and currently we're only using is as a string. We're definitely fixing this for next version.

dschult:
See also #3797 for a related feature request.
I'm reopening this until fixed along with #3754 and #3797

cmkpl:
I can see from the docs of 2.5rc1.dev_20200420161240, it seems that the function astar_path still only accept string for the parameter weight

Issue #3754, opened 2019-12-18
`bidirectional_dijkstra` does not use the `weight` function


mohamed--abdel-maksoud:
I'm on 2.4:
```
In [16]: nx.__version__
Out[16]: '2.4'
```

Here is a minimal example to reproduce:
```python
def cost(a,b,*_):
    print('cost is called')
    return 1

g = nx.grid_graph(dim=[10, 10])
nx.bidirectional_dijkstra(g, (0,0), (9, 1), weight=cost)
```

"cost is called" is never printed.

dschult:
This is a bug -- that function uses "weight" as if it is a string -- not a function. The documentation clearly says it can be either a string or a function. We need to update it to use either a string or a function. Not sure on the history here...
Thanks!

Issue #717, opened 2012-07-20
Adding multigraph handling to astar_path function (migrated from Trac #731)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/731
Reported 2012-06-10 by rtk08, assigned to @hagberg.

Version 1.6 astar_path function does not support shortest path calculation for multigraphs - a NetworkXError exception is thrown.

Finding the shortest path with an A Star algorithm for a multigraph currently requires reducing it to a graph with single (directed or not) edges between any two nodes. This is very inefficient, particularly if the graph in question is large.

I will modify the the astar_path function to handle path finding in multigraphs.

PR #3799, opened 2020-01-26
Add weight functions to bidirectional_dijkstra and astar

boothby:
See discussion in #3754 and #3774.

Knock-on benefits: 
1. Uncovered/fixed an inconsistency in Bellman-Ford (easy 1-line fix)
2. Added support for multigraphs in `astar`

### DIFF
--- a/networkx/algorithms/shortest_paths/astar.py
+++ b/networkx/algorithms/shortest_paths/astar.py
@@ -70,7 +70,6 @@ def astar_path(G, source, target, heuristic=None, weight='weight'):
 
     push = heappush
     pop = heappop
-    G_succ = G._succ if G.is_directed() else G._adj
     weight = _weight_function(G, weight)
 
     # The queue stores priority, node, cost to reach, and parent.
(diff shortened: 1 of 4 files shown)


dschult:
small tweak -- I think this came from bellman-ford, but it is not needed.

```G_succ = G._succ if G.is_directed() else G._adj``` . Should be just ```G_succ = G._adj```

It's not needed because ```G._adj``` is always ```G._succ``` for directed graphs. It was designed that way so you can just use ```G._adj``` for successors whether G is directed or undirected.

boothby:
Thanks for the heads-up!  I'll fix that sometime tonight or tomorrow.

dschult:
This looks good.  Anything else before its merged?

boothby:
Thanks!  I've got nothing else to add

boothby:
Trawling old tickets, I found #717, which is coincidentally addressed by commit d5aa0e8

cmkpl:
I can see from the docs of 2.5rc1.dev_20200420161240, it seems that the function astar_path still only accept string for the parameter weight

Issue #252, opened 2012-07-20
comments in read_multiline_adjlist (migrated from Trac #252)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/252
Reported 2009-07-21 by @chebee7i, assigned to @hagberg.

read_multiline_adjlist doesn't ignore comment lines once the source node has been read.  As reported here:

   http://groups.google.com/group/networkx-discuss/browse_thread/thread/2f302594f990de56?hl=en

PR #3804, opened 2020-01-30
Trophic Levels #3736

### LINKED ISSUE #3736: Trophic levels (feature suggestion)
Would it be interesting to include an algorithm to calculate trophic levels?

Related notions which could be included are trophic difference and trophic coherence - see wikipedia [1] and [2]. 

This may be a duplicate of #577 from 2012, which was closed due to inactivity.

I've made an initial attempt over at [3], adding a module at `algorithms.centrality.trophic`.

1. https://en.wikipedia.org/wiki/Trophic_coherence
2. Stephen Levine (1980) Several measures of trophic structure applicable to complex food webs. J. theor. Biol. 83, 2, 195-207. https://doi.org/10.1016/0022-5193(80)90288-X
3. https://github.com/tomalrussell/networkx/commit/e715def4abe0a184c0934aa7d6a886b6d5ffe705

tomalrussell:
Work with @chasmani aiming to close #3736 

This adds `networkx.algorithms.centrality.trophic` with `trophic_levels` (which does most of the work), `trophic_differences` and `trophic_coherence`.


dschult:
Yes -- there are many modules that use numpy. Any that do are set up so that if numpy is not installed you can still import networkx. Take a look at a module that uses numpy. e.g. ```drawing/layout.py``` (and there may be better examples -- this is one I was looking at recently).  Mostly, you just import within the function that uses numpy instead of at the module level.

Testing is a little trickier because you don't want to run the tests that require numpy if numpy isn't available. But pytest actually makes it fairly straightforward.  Again, take a look at an existing test module, e.g. test_layout.py .  Ask if you have questions

tomalrussell:
Thanks @dschult - that's working since https://github.com/networkx/networkx/pull/3804/commits/dfb792652dbcb4cd812a85611eed5d59550a7f3e with `pytest.importorskip` in the tests and try/except around `import numpy` within functions.

I'm not sure why codecov/patch is complaining - the only changes in this PR are adding new files, and they should be 100% covered. Is there anything I should do about this?

dschult:
When I click on the codecov "Details" link and then on the "Diff" label, it shows which code has been run during a test and which hasn't. The error handling of no numpy obviously haven't been tested when the environment includes numpy, so that's not a worry. 

The only suggested change is to test a case when cannibalism is False and there are not self-loops.
The "Avoid copy otherwise" comment code doesn't get run during any tests. Minor for sure. But perhaps easy to fix??

dschult:
Looks like [Line #252](https://github.com/networkx/networkx/blob/4a6008938dfe4f19515a733dd8d3271d171ba0f9/networkx/algorithms/centrality/tests/test_trophic.py#L252) should be changed to use ```matrix_d instead of matrix_c```

Anything else planned for this PR? We can add other trophic functions later if they are planned.
Thanks very much!!

tomalrussell:
Thanks - good spot!

That's all we have for now - let me know if there's anything more that should be fixed up for this PR.

dschult:
This looks good.  I only have one thing.
For the docstrings, you include inline math using $stuff$. I think we need that to be :math:`stuff` for the [sphinx documentation](http://www.sphinx-doc.org/es/stable/ext/math.html) system.

Issue #3586, opened 2019-09-14
Feature request - Lattice - periodic on a single axis? 


MrBenGriffin:
Could it be possible to extend the 'periodic' argument to be a dimension-sized tuple of bools, so that periods could be limited to a subset of axes? This would enable eg. frieze groups more easily....

dschult:
I guess the logic would check if 'periodic' was True or False and if neither, then assume it is a list of length 'dim'.  It needs to be backwards compatible and not slow down the usual case too much (I don't think it will).

PR #3807, opened 2020-02-01
option for partially-periodic lattices (networkx #3586)

boothby:
I'm not super happy with how I wrote those docstrings; suggestions are welcome.


dschult:
Thanks for this!!

How about for doc_strings:
```python
    If `periodic` is True, both dimensions are periodic. If False, none are periodic.
    If `periodic` is iterable, it should yield 2 bool values indicating whether the
    1st and 2nd axes, respectively, are periodic.
```

And
```python
    If `periodic` is True, all dimensions are periodic. If False all 
    dimensions are not periodic. If `periodic` is iterable, it should
    yield `dim` bool values each of which indicates whether the
    corresponding axis is periodic.
```

PR #3817, opened 2020-02-07
Remove position offset from periodic lattice tilings

Matt-HJ-Bailey:
The coordinates provided in the 2D periodic 2D graph tilings feature a `0.01*i*i` term. This means that they do not tile in the plane, and instead form a parabola. At large values of `i`, the hexagons or triangles created are no longer regular. 

I've resolved this issue simply by removing the `0.01*i*i` term because it is not mentioned in the documentation.

### DIFF
--- a/networkx/generators/lattice.py
+++ b/networkx/generators/lattice.py
@@ -255,10 +255,7 @@ def triangular_lattice_graph(m, n, periodic=False, with_positions=True,
         jj = (j for i in cols for j in rows)
         xx = (0.5 * (j % 2) + i for i in cols for j in rows)
         h = sqrt(3) / 2
-        if periodic:
-            yy = (h * j + .01 * i * i for i in cols for j in rows)
-        else:
-            yy = (h * j for i in cols for j in rows)
+        yy = (h * j for i in cols for j in rows)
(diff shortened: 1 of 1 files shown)


dschult:
Hmmm....  So you want the graph to be periodic, but the positions to be like they are created for a non-periodic version.  Thanks for pointing out that the docs don't mention the small parabolic shift. That was put in a LONG time ago because drawing it with positions that don't have the shift makes the periodic edges align with other edges. The shifted position does give a kind of 3-D effect that makes the image edge lengths differ (not regular). 

Is there a reason you want the unshifted positions?
Is there a way to make that work well with drawing the graph?

Matt-HJ-Bailey:
Thanks for looking at this, Dan. Let me know if you'd like me to move this discussion to the issue tracker instead. 

I want a periodic graph with non periodic boundaries to study the rings in chemicals like graphene, and the best way to do that is by keeping track of both the graph and its dual.

I haven't found an elegant way to make drawing periodic graphs work, and edges lining up is a problem. In my own work, I just draw the "periodic edges" dotted off towards the edge of the figure, like scipy voronoi_plot_2d does. That probably isn't the best solution here.

PR #3822, opened 2020-02-16
reconsider the lobster generator

### LINKED ISSUE #3701: Wrong random_lobster implementation?
Hi, it seems that [networkx.random_lobster's implementation logic](https://github.com/networkx/networkx/blob/4e9771f04192e94a5cbdd71249a983d124a56593/networkx/generators/random_graphs.py#L1009) is not aligned with the common definition as given in [wolfram mathworld](http://mathworld.wolfram.com/LobsterGraph.html) or [Wikipedia](https://en.wikipedia.org/wiki/List_of_graphs#Lobster).
For example, it can not produce the simplest lobster graph examples such as lobster_example.

austinorr:
## The Issue
In it's present form, the `random_lobster` generator function produces valid lobsters with multiple leaves, but it's only capable of producing an extremely small subset of the allowable lobsters which fit the general definition of "a tree having the property that the removal of leaf nodes leaves a caterpillar graph" and where a caterpillar graph is on in which all leafs are only one edge from a continuous path graph.

To illustrate the present limitation I've created a 15 node lobster graph with the original implementation:
``` python
g = nx.random_lobster(15, 0.6, 0.5, seed=42)
pos = nx.nx_pydot.pydot_layout(g)
nx.draw(g, pos=pos)
```
old_lobster

This graph is clearly a valid lobster, and clearly has multiple leaves. However, none of the branches from the main path will ever fork even though it's allowed by the definition. 

This limitation means that lobsters built with the current implementation will never have a node with degree > 3 even though both the 'backbone' nodes and the 'caterpillar' nodes of the lobster are allowed to have multiple connections as long as the additional connections don't violate the definition.

## Proposed solution requirements
An acceptable solution should:
* not change the current api, or the meaning of any of the existing parameters
* pass all existing tests

## Proposed changes to codebase
This PR closes #3701 by allowing multiple edges to be added at each depth traversed during graph construction. Note that the change converts `if` statements to `while` and adds some safety checks at the top of the function to catch risky user inputs before they start an infinite loop. 

This means we can return valid lobsters that have nodes with degree > 3. The following plot was made with the code submitted in this PR:

```python
g = nx.random_lobster(10, 0.6, 0.5, seed=42)
pos = nx.nx_pydot.pydot_layout(g)
nx.draw(g, pos=pos)
```
new_lobster

## Proposed additional tests
Previous tests for the lobster function checked only that the function ran without throwing an exception. This PR adds a couple of checks to assert that the graph is a lobster according to the definition, and that the subgraph formed by dropping its leaves is a caterpillar graph. I've also added test to assert that if `p2=0` then the graph is a caterpillar. 

To be clear, the existing implementation will pass those tests since it's a valid, if narrow, implementation for lobster creation. I've therefore also added an assertion to check that the max degree is larger than 3 for the test graph. This assertion would fail in the existing implementation but passes in the proposed one.

## Additional discussion
It'd be nice to see if this PR addresses the concerns voiced by @silent567 and @dschult. I think in this case it turns out to be pretty straight-forward to patch the algorithm to generate nearly any graph in the set of valid lobsters.

In fact, I don't think the original implementation meant to limit the breadth of available lobsters so severely, and I'd suggest this PR be considered as a lowly 'bug fix' rather than as a feature change.

### DIFF
--- a/networkx/generators/random_graphs.py
+++ b/networkx/generators/random_graphs.py
@@ -1011,19 +1011,29 @@ def random_lobster(n, p1, p2, seed=None):
     seed : integer, random_state, or None (default)
         Indicator of random number generation state.
         See :ref:`Randomness<randomness>`.
+
+    Raises
+    ------
+    NetworkXError
+        If `p1` or `p2` parameters are >= 1 the while loops will never finish.
     """
+    p1, p2 = abs(p1), abs(p2)
+    if any([p >= 1 for p in [p1, p2]]):
+        raise nx.NetworkXError(f"Probability value of `p1` and `p2` must be < 1.")
+
     # a necessary ingredient in any self-respecting graph library
     llen = int(2 * seed.random() * n + 0.5)
     L = path_graph(llen)
     # build caterpillar: add edges to path graph with probability p1
     current_node = llen - 1
     for n in range(llen):
-        if seed.random() < p1:  # add fuzzy caterpillar parts
+        while seed.random() < p1:  # add fuzzy caterpillar parts
             current_node += 1
             L.add_edge(n, current_node)
-            if seed.random() < p2:  # add crunchy lobster bits
+            cat_node = current_node
+            while seed.random() < p2:  # add crunchy lobster bits
                 current_node += 1
-                L.add_edge(current_node - 1, current_node)
+                L.add_edge(cat_node, current_node)
     return L  # voila, un lobster!
 
 
--- a/networkx/generators/random_graphs.py
+++ b/networkx/generators/random_graphs.py
@@ -1000,6 +1000,11 @@ def random_lobster(n, p1, p2, seed=None):
     leaf nodes. A caterpillar is a tree that reduces to a path graph
     when pruning all leaf nodes; setting `p2` to zero produces a caterpillar.
 
+    This implementation iterates on the probabilities `p1` and `p2` to add
+    edges at levels 1 and 2, respectively. Graphs are therefore constructed
+    iteratively with uniform randomness at each level rather than being selected
+    uniformly at random from the set of all possible lobsters.
+
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


dschult:
Could you add a line or two to the docs saying that this method doesn't select uniformly randomly from all possible lobsters. And maybe some language that p1 and p2 are used repeatedly until failure to add an edge which stops that branch of the process. Then @silent567 should be happy with this too.

austinorr:
Yep, good idea, I’ll push an edit shortly. You’re hoping to document this entirely within the function docstring right?

dschult:
I was thinking of just within the docstring for that function yes. Open to other ideas, but that seems the most simple.

austinorr:
Ok great. While I’m in there, any interest in the `is_caterpillar` or the `is_lobster` functions in the tests? Should they stay inside the test scope, or be moved to a reusable utility module?

austinorr:
@dschult docstring changes are ready to review.

dschult:
Thanks! This looks good.  

As for the ```is_lobster``` functions, they could be moved to a module in the algorithms section of the code. But I think that only makes sense if someone implements more algorithms for lobsters and caterpillars. But I must say I don't know much about the literature on this topic. 

Let's commit it as is and then open a new PR if other algorithms are desirable.
Thanks!!

silent567:
Hi all, thanks for taking care of this. It seems that the generator can generate any lobster now. It's great! 

For myself, I just utilized my simple toy codes for generating lobsters and the project has almost finished. In my case, neural networks can very easily overfit the datasets as the original version can only generate very limited  (actually finite right? 3^n at most?) graphs. The degree distribution is also critical in my problem. Anyway, it's great to see this PR and wish networkx a good future~

austinorr:
Glad to help, good luck with your project!

PR #3829, opened 2020-02-22
WIP: Change EdgeView contains feature

iboates:
This PR is a WIP to get feedback

It is addressing issue 3819: https://github.com/networkx/networkx/issues/3819

After mucking around in the `EdgeView` classes, it seems to me that the simplest way to fix the `__contains__` bug is actually to just constrain the `self._adjdict` variable in the `__init__` function by comparing it to the `nbunch` param (if there is one).

I put a big test in `TestEdgeView` because I noticed it gets executed for all the other test classes due to inheritance. I'm not sure if this is the best way to do this and while I have some experience in pytest, I am not really an expert in all its inner workings so I am open to feedback on how to better structure it.

This seems to work for `Graph` and `DiGraph` objects, but fails for `MultiGraph` and `DiMultiGraph` objects, where I am now stuck. I constrain `self._adjdict` just as before, but for some reason, `__contains__` is now always returning False. You can see this by looking at the `inclusion` dict in the new test in the debugger, it marks the edge `(0, 1, 0)` as being False even though it should be included in both the directed and the undirected graph.

Can I get some feedback on this? Am I going in the right direction or does it need a different approach?

### DIFF
--- a/networkx/classes/reportviews.py
+++ b/networkx/classes/reportviews.py
@@ -766,6 +766,7 @@ class OutMultiEdgeDataView(OutEdgeDataView):
         else:
             nbunch = list(viewer._graph.nbunch_iter(nbunch))
             self._nodes_nbrs = lambda: [(n, adjdict[n]) for n in nbunch]
+            self._adjdict = {u: self._adjdict[u] for u in self._adjdict if u in nbunch}
         self._nbunch = nbunch
         self._data = data
         self._default = default
--- a/networkx/classes/reportviews.py
+++ b/networkx/classes/reportviews.py
@@ -636,6 +636,7 @@ class OutEdgeDataView:
         else:
             nbunch = list(viewer._graph.nbunch_iter(nbunch))
             self._nodes_nbrs = lambda: [(n, adjdict[n]) for n in nbunch]
+            self._adjdict = {u: self._adjdict[u] for u in self._adjdict if u in nbunch}
         self._nbunch = nbunch
         self._data = data
         self._default = default
--- a/networkx/classes/tests/test_reportviews.py
+++ b/networkx/classes/tests/test_reportviews.py
@@ -608,10 +608,15 @@ class TestEdgeView:
         for edge in graph.edges:
             inclusion[edge] = edge in end_edges
 
-        for edge_tuple in [(0, 1), (7, 8)]:
-            assert inclusion[edge_tuple]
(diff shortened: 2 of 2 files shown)


dschult:
Thanks for this!

The idea of the EdgeView classes is to make a low memory -- fast to create object that does useful things with edge information. We tried not to create large objects inside . the EdgeView itself, instead referring to the original graph data structure when needed. We do of course have to store which nodes in nbunch to restrict the view to. And we store a function on self._nodes_nbrs that gets the nodes and the neighbors from the original graph data structure.

By creating a dict to hold the ```self._adjdict``` information, we would be duplicating the info in the graph data structure. Can we find a way to do it without creating a dict on the EdgeView object?  Duplicating the graph data structure is also why this doesn't work with MultiGraphs. That data structure is different from Graphs so you would need to copy the MultiGraph data structure instead of the Graph data structure.

I suggest trying the following approach instead. Currently nbunch is stored as a list on the EdgeDataView object. 
1) Change that to a set (since we really only use it for node lookup, this is faster). 
2) Add a check to the ```__contains__``` method that checks whether the edge should be reported based on nbunch.  Something like (for OutEdgeDataView):
```python
        try:
            u, v = e[:2]
            if u not in self._nbunch:
                return False  # this edge doesn't start in nbunch
            ddict = self._adjdict[u][v]
```
Each of the other EdgeDataView objects would need a different condition. For directed out edge views, check that ```u``` is in nbunch. For undirected edge views, either u or v could be in nbunch. For directed in-edge views, check that ```v``` is in nbunch.

Does this make sense?

Comments on the test code:
- you start with ```graph = self.G``` but  go back to ```self.G``` later. Maybe just stick with ```self.G```?
- to construct ```end_nodes```, use the ```self.G.degree()``` feature of iterating over nodes, degree tuples:  ```[n for n, deg in self.G.degree() if deg == 1]```
- the ```inclusion``` dict simply copies ```end_edges```.  Instead just use ```end_edges``` in your assertions. So, ```assert inclusion[(0, 1)]``` becomes ```assert (0,1) in end_edges```
- instead of checking the type of the graph, use the ```G.is_multigraph()``` method. It is similar to ```G.is_directed()```.  And the way it is written, you don't actually need to check ```is_multigraph``` because the tests are the same either way...  Just check for ```is_directed```

Ask again if questions.

dschult:
Let's close this in favor of #3845

Issue #3834, opened 2020-02-24
Add normalized adjacency matrix and normalized adjacency spectrum


cshjin:
A normalized adjacency matrix is defined as 

`D^{-1/2} A D^{-1/2}`

Add support of normalized adjacency matrix to `graphmatrix.py` and 
add support of normalized spectrum to `spectrum.py`.

PR #3835, opened 2020-02-24
fix-for-issue-3834

cshjin:
Resolve issue #3834 

* `normalized_adjacency_matrix` added in `graphmatrix.py`
* `normalized_adjacency_spectrum` added in `spectrum.py`
* updated document accordingly

### DIFF
--- a/networkx/linalg/graphmatrix.py
+++ b/networkx/linalg/graphmatrix.py
@@ -3,7 +3,12 @@ Adjacency matrix and incidence matrix of graphs.
 """
 import networkx as nx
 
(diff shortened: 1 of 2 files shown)


dschult:
Can you add a paragraph about when you might want to use normalized adjacency instead of adjacency?  Also a reference -- even just to wikipedia -- would be helpful.

Thanks!

dschult:
Thanks for the docs info.  It should be put into the docstring.
For references, take a look at the format from other functions. 

If you have questions ask -- if you don't have time now, I'll get to it eventually.

Issue #3657, opened 2019-10-16
Add support of 'signless' Laplacian


cshjin:
Networkx has the Laplacian matrix supported, moreover, I would like to see the `signless` Laplacian.

According to [__Spectra of Graphs__](https://www.win.tue.nl/~aeb/2WF02/spectra.pdf) by Brouwer and Haemers, the signless Laplacian is defined:
Q = D+A.

It is widely used in bipartite graph analysis and subgraphs ref: [1], [2]. 
And it has similar properties like the Laplacian.

Proposal:
* add function `signless_laplacian_matrix`
* add function `normalized_signless_laplacian_matrix`
* add function `signless_laplacian_spectrum`
* (optional) add function `direct_signless_laplacian_matrix`

[1]: https://www.sciencedirect.com/science/article/pii/S0024379507000316
[2]: http://elib.mi.sanu.ac.rs/files/journals/publ/101/n095p011.pdf

cshjin:
Either update the API of `laplacian_matrix` with 

`laplacian_matrix(G, nodelist=None, weight="weight", signless=False)`

Or kick in with another function `signless_laplacian_matrix`

Any suggestions? I can implement it as long as a proper API is chosen.

dschult:
Hmmm...  a growth in the number of types of laplacian matrices could get tricky.
My initial thought is that this should be a separate set of functions -- like you first proposed.
But I think we should think about that a little bit in terms of the literature and how many more different styles of laplacian matrix are likely to come along.

Perhaps a good way to decide is whether the algorithms that use these matrices depend on which type of laplacian is being used. Are we going to want other functionality than either creating the matrix or its spectrum?

cshjin:
Hmm, I don't see particular functions related to signless Laplacian. 
As far as I know, only the standard functionalities like the spectrum and normalization will be used in some analyses. 

Yes, I agree we should also consider other Laplacians if we want to expand the family of Laplacians, including the signless and deformed cases. 

Hold the PR until a strong need in the future and the proper design of the API.

Ref: 
* signless Laplacian: https://en.wikipedia.org/wiki/Laplacian_matrix#Signless_Laplacian
* deformed Laplacian: https://en.wikipedia.org/wiki/Laplacian_matrix#Deformed_Laplacian

PR #3836, opened 2020-02-25
Add Signless Laplacian algorithms: FIX issue 3657

cshjin:
Update the API for the Laplacian matrix and its spectrum with the new parameter __signless__, which indicates to use the `L = D+A`.

Enhance for issue #3657 .

Feel free to reject this PR if the API is not properly designed.


networkx/linalg/spectrum.py
```diff
@@ -12,7 +12,8 @@
 ]
 
 
-def laplacian_spectrum(G, weight="weight"):
+def laplacian_spectrum(G, weight="weight", 
+                       =False):
```
dschult, line 16: looks like a syntax /typo error here at top of spectrum.py.

Issue #3837, opened 2020-02-25
Quotient graphs - option to label block nodes via a mapping on creation


sr-murthy:
There doesn't seem to be an option to set the labels of the block nodes of a quotient graph via a mapping passed in to the creation method [`networkx.quotient_graph`](https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.minors.quotient_graph.html#networkx.algorithms.minors.quotient_graph). There is a `relabel` boolean that labels the block nodes as non-negative integers, but there is no way of controlling this if the block nodes have to be labelled according to some mapping.

This is an example to illustrate the problem. I have a (directed) strongly connected graph `G` with 21 nodes

    [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30, 32, 34, 36, 38, 40]

There is an equivalence relation `nrel` on these nodes which produces the following six blocks

    [0, 2, 4, 6, 8]      #B1
    [16, 10, 12, 14]     #B2
    [24, 18, 20, 22]     #B3
    [32, 26, 28, 30]     #B4
    [34, 36]             #B5
    [40, 38]             #B6

and an edge relation `brel` on the blocks given by

    >>> brel = lambda B, C: any(e in G.edges for e in product(B, C))

(A simple extension of the edge relation of `G` to the blocks of `Q` - there is an edge from block `B` to `C` if there is an edge in `G` from a node in `B` to a node in `C`).

I create the quotient `Q` using

    >>> Q = nx.quotient_graph(G, nrel, edge_relation=brel)

The nodes of `Q` are now these `frozenset` objects

    >>>Q.nodes
    >>> NodeView((frozenset({34, 36}), frozenset({16, 10, 12, 14}), frozenset({32, 26, 28, 30}), frozenset({40, 38}), frozenset({0, 2, 4, 6, 8}), frozenset({24, 18, 20, 22})))

But I would like to pass a mapping to `nx.quotient_graph` which sets the labels of the block nodes according to this mapping

    >>> block_relabelling = {frozenset({0, 2, 4, 6, 8}): 2, frozenset({10, 12, 14, 16}): 3, frozenset({18, 20, 22, 24}): 4, frozenset({[26, 28, 30, 32]}): 5, frozenset({34, 36}): 6, frozenset({38, 40}): 7}
    >>> nx.relabel_nodes(Q, block_relabelling, copy=False)
    >>> sorted(Q.nodes)
[...]

PR #3839, opened 2020-02-26
User-defined labelling of block nodes when creating quotient graphs

sr-murthy:
* Tweak quotient graph method (`networkx.algorithms.minors.quotient_graph` - allow user-defined mapping of blocks (`frozenset` objects) to labels via dicts and functions via an optional argument with default of `None`; no labels dict or function will cause blocks to be labelled/represented by themselves as `frozenset` objects
* update all tests for quotient graphs involving block relabelling (`networkx.algorithms.tests.test_minors`).

### DIFF
--- a/examples/algorithms/plot_blockmodel.py
+++ b/examples/algorithms/plot_blockmodel.py
@@ -58,7 +58,8 @@ H = nx.convert_node_labels_to_integers(H)
 # Create parititions with hierarchical clustering
 partitions = create_hc(H)
 # Build blockmodel graph
-BM = nx.quotient_graph(H, partitions, relabel=True)
+labels = {frozenset(b): i for i, b in enumerate(partitions)}
+BM = nx.quotient_graph(H, partitions, labels=labels)
 
 # Draw original graph
 pos = nx.spring_layout(H, iterations=100)
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -3,6 +3,8 @@ from itertools import chain
 from itertools import combinations
 from itertools import permutations
 from itertools import product
+from itertools import tee
+from typing import Callable
 
 import networkx as nx
 from networkx import density
@@ -50,7 +52,7 @@ def equivalence_classes(iterable, relation):
 
 
 def quotient_graph(G, partition, edge_relation=None, node_data=None,
-                   edge_data=None, labels=False, create_using=None):
+                   edge_data=None, labels=None, create_using=None):
     """Returns the quotient graph of `G` under the specified equivalence
     relation on nodes.
 
@@ -108,12 +110,14 @@ def quotient_graph(G, partition, edge_relation=None, node_data=None,
         * 'density', the density of the subgraph of `G` that this
           block represents.
 
-    labels : dict or function
-        If a ``dict`` it must be a mapping of the blocks, as ``frozenset``
-        objects, to block labels. If a function then it must be the function
-        on the nodes of G associated with the equivalence relation used to
-        create the quotient, i.e. the function ``f`` such that the
-        equivalence relation given by ``x <-> y`` if and only if ``f(x) = f(y)``
+    labels : None or dict or function
+        If ``None`` the blocks are labelled as themselves, namely
+        ``frozenset`` objects. If a ``dict`` it must be a mapping of
+        the blocks, as ``frozenset`` objects, to block labels. If a
+        function then it must be the function on the nodes of G
+        associated with the equivalence relation used to create the
+        quotient, i.e. the function ``f`` such that the equivalence
+        relation given by ``x <-> y`` if and only if ``f(x) = f(y)``
         on the nodes of G.
 
     create_using : NetworkX graph constructor, optional (default=nx.Graph)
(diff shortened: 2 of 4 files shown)


sr-murthy:
@dschult I've made the changes to illustrate how user-defined block relabelling can work when creating quotient graphs, either via dicts or functions associated with the equivalence relation. I had to change `relabel` from a bool to an optional `label` which is either `None` (labelling as frozenset objects is retained, which is the existing behaviour when `relabel=False`) or a `dict` mapping of blocks to user-defined labels, or the function associated with the equivalence relation used to create the quotient.

sr-murthy:
@dschult Is this PR likely to reviewed at some point?

dschult:
I am absolutely swamped with the pandemic disruption at the moment and am unlikely to get to any meaningful review done before mid May. I've put the 2.5 release tag on this PR to be sure it gets done in time for that.   Thanks for the gentle nudge. :}

sr-murthy:
@dschult No problem at all, absolutely understood. Thanks.

sr-murthy:
@dschult Are you able to merge this now?

dschult:
I have been looking at it occasionally over the last couple weeks. I am having trouble with the overall API for the function -- especially with the ```partitions``` and ```labels``` inputs. It feels like they are providing the same information in a different form. The examples all seem to construct ```labels``` by reformatting ```partitions```.  I want to think about a different call signature.

The ability to have ```partitions``` be an equivalence relation function should probs be given to the user by making ```equivalence_classes``` a public function. Then partitions could be a dict keyed by label to containers holding the nodes for that partition. Partitions that are containers of sets could easily be converted to dicts using ```partitions = dict(enumerate(partitions))```.  I think a function from node to partition label should also be supported too -- perhaps with examples in the docstring for how to create the partitions input from such a function. Both that construction and equivalence_classes``` could be shown side-by-side. Having that outside the function simplifies the parsing somewhat and it is fairly easy to do:  ```p = {n: f(n) for n in G}``` for a labels function and ```p = equivalence_classes(G)```.

I would say the default behavior should be to use labels of the partitions dict as the new nodes. If labels not provided then we use numbers assigned by ```dict(enumerate(partitions))``` and explain/demonstrate in the docstring how they can obtain that mapping from new nodes to old nodes. We'd also need to explain that using partitions as an equiv. relation function will lead to new nodes that are assigned numbers based on the order constructed from ```equivalence_classes```.

Backward compatibility:
We can identify old-style partitions: lists of sets or lists of lists would not have an ```items()``` attribute.
Callable ```partitions``` have ```__call__```. They can be handled as described above.

The input ```relabel``` when set to "False" should construct the new nodes to be ```frozenset(p)```. It should probably also show a deprecation warning where we plan to remove this argument in nx v3.0.   ```relabel``` left as default can switch to the new default behavior so long as the docstring has a "Note" that old-style default behavior can be fixed in both old and new versions by explicitly setting ```relabel=False```.  ```relabel=True``` can be ignored because it is the new default.

I'm not completely sure this will all be smooth -- I need to look at ```equivalence_classes``` more to see if it does the job. Before anything more is implemented I'd like some feedback from you --- a user dedicated enough to think about all this. What is the best way to represent these partitions and get the info to the quotient_graph function?  How would you naturally think about a partition? A labeled partition? What is someone from this field going to be looking for?

sr-murthy:
@dschult THanks for the review and comments, appreciated. I will think about these things, and get back to you. It's been a while since I touched base on this, with examples.

sr-murthy:
@dschult Apologies for the delay, but I will reply in more detail this weekend.

In reply to your general question about the best way of representing partitions (of a set X), this can be either as the set of blocks of the partition, or as the function `f` which assigns to each element `x` of `X` its block `B(x)`, with respect to `rel`. It is natural to choose labels for the blocks, namely, representative elements, one from each block.

I would also go back to one of my original posts in the issue discussion

https://github.com/networkx/networkx/issues/3837#issuecomment-591277009

In that post, I gave as an example, the quotient graph `Q` of a graph `G`, with nodes `{0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30, 32, 34, 36, 38, 40}`, whose blocks are
 
    {34, 36}
    {10, 12, 14, 16}
    {26, 28, 30, 32}
    {38, 40}
    {0, 2, 4, 6, 8}
    {18, 20, 22, 24}

I observed that `quotient_graph` these blocks were being labelled as follows

    {34, 36} -> 0
    {10, 12, 14, 16} -> 1
    {26, 28, 30, 32} -> 2
    {38, 40} -> 3
    {0, 2, 4, 6, 8} -> 4
    {18, 20, 22, 24} -> 5

My original question was, what is the reason for chosing this labelling, over a more natural labelling, given by

    {34, 36} -> 34
    {10, 12, 14, 16} -> 10
    {26, 28, 30, 32} -> 26
    {38, 40} -> 38
    {0, 2, 4, 6, 8} -> 0
    {18, 20, 22, 24} -> 18

Here the labels 34, 10, 26, 38, 0, 18 are simply the starting nodes of the subgraphs of nodes equivalent under the quotient relation. I could also have chosen 36, 12, 28, 38, 2, 20 etc., but it doesn't matter.

The purpose of the changes I've submitted are to allow

dschult:
The idea for the currently used labeling system when ```relabel is True``` is that the labels (0,1,2,3,4,5) are the indexes of each partition in the input ```partitions```. This is useful for looking up the nodes in the original graph from the partition node in the quotient graph:   ```partitions[label]```.  It is potentially confusing to have the new graph nodes be one of the nodes from the original graph. Then you can't tell what 36 is referring to.

In terms of code then, let's refer to a list-of-lists as a dict-of-list via ```enumerate```. Also, I'm using the term "list" here but "set" could be used if lookup is desirable. Also I'm treating tuple like frozenset.
I think we have various ways to store partitions:
1) a dict-of-lists keyed by partition label to lists of nodes
2) a dict keyed by tuples of nodes to partition label.
3) a function from node to partition label.
4) an equivalence relation stored as a function rel(node1, node2) that returns True if related.

Hopefully we can accept any of these and convert to a convenient one easily.  I'd like to avoid having 2 forms of the partition info input to the function in the same call.  In fact, since it is useful to have the mapping from new_node to sets of old nodes, I think it is simplest to have the user enter case 1) with the help of the docstring.

My preference is to allow case 1): either a dict-of-lists or a list-of-lists.  Then:
- we use the docstring to guide users to the ```equivalence_classes()``` function to convert case 4) to case 1). 
- we provide an example of switching keys and values to get from 2) to 1). Something like: ```{label: part for part, label in labels_dict.items()}```.
- we have an example with a function where conversion from 3) to 1) happens with code like:
```python
partitions = defaultdict(list)
for node in G:
    partitions[f(node)].append(node)
```

I think this allow code improvement by avoiding the relabeling at the end. I think the existing code remains unchanged otherwise except for changing ```product(b, c)``` to ```product(partitions[b], partitions[c])``` everywhere.  This would definitely involve some deprecation because the API changes -- but the old input of list of partitions would work. 

Thoughts?

sr-murthy:
I'll update the PR tomorrow.

The advantage I described might be specific to my problem, but it's a question of perspective - if the nodes in each block are totally ordered, so that there is an entry node and an exit node for each block, and edges between two blocks represent transitions from the exit node of one to the entry node of the second, then it is convenient to have the nodes in the quotient graph (the blocks) be labelled by the entry or exit nodes (as appropriate) in the blocks.

sr-murthy:
@dschult I think it's better to create a new PR for these changes, so closing this one.

sr-murthy:
Created new PR https://github.com/networkx/networkx/pull/4183. Will update with docstring changes.

PR #3841, opened 2020-02-28
Fixed Bug in generate_gml(G, stringizer=None)

Abhi5h3k:
network-X-bug-fix

generate_gml(G, stringizer=None) function used by write_gml(G, path, stringizer=None)

raise NetworkXError('%r is not a valid key' % (key,))

This error is caused when key contains "_" underscore character. Like "first_name" will raise this error.

Fix: updatd the regular expression to allow "_"


dschult:
We need to check that "_" is allowed by the GML standard.
We can read GML with _ in it, but we should be writing standard GML

ankitarima:
> network-X-bug-fix
> 
> generate_gml(G, stringizer=None) function used by write_gml(G, path, stringizer=None)
> 
> raise NetworkXError('%r is not a valid key' % (key,))
> 
> This error is caused when key contains "_" underscore character. Like "first_name" will raise this error.
> 
> Fix: updatd the regular expression to allow "_"

Thanks for the update I was facing the same issue @Abhi5h3k

dschult:
Can either of you check that the gml standard does or doesn't allow underscores in the strings? 
Perhaps we should provide a "strict=True" argument? Making that False could allow things like this.

dschult:
The original GML standard has been removed from the creator's webpage. It seems that this markup language is now bing supported and evolves with the network analysis community, especially drawing and analysis packages. I'm going to merge this change which allows WRITING a GML file that includes underscores in the keys of the key-value pairs.

PR #3845, opened 2020-03-01
WIP: Change EdgeDataView __contains__ feature (2nd attempt)

iboates:
I took the advice of last time and this time I think I am close, but I am still having trouble with the MultiEdge class family tests.

From what I can tell, it seems that as soon as you pass an `nbunch` parameter to the `EdgeView` cosntructor (and those of its family), it automatically redirects to an (associated) `EdgeDataView`. For this reason I didn't implement any nbunch check logic because it seems to already be handled in the associated EdgeDataViews.

However now it seems to be failing whenever it runs the `test_contains` function in the context of any `MultiEdge` class. It seems to me that this function is just run in all of its subclasses and can no longer handle the new restriction of edges via the `nbunch`. Could it be that this test was only passing due to the presence of the bug, and its fix now causes the test to fail? As I said before I am no big expert on pytest so I don't want to presume anything so someone else should look at this. But if my guess is correct then I think this test has to be split down into the respective subclasses to take into account the changes to the various `__contains__` methods

### DIFF
--- a/networkx/classes/reportviews.py
+++ b/networkx/classes/reportviews.py
@@ -634,7 +634,8 @@ class OutEdgeDataView:
         if nbunch is None:
             self._nodes_nbrs = adjdict.items
         else:
-            nbunch = set(viewer._graph.nbunch_iter(nbunch))
+            # dict retains order of nodes but acts like a set
+            nbunch = dict.fromkeys(viewer._graph.nbunch_iter(nbunch))
             self._nodes_nbrs = lambda: [(n, adjdict[n]) for n in nbunch]
         self._nbunch = nbunch
         self._data = data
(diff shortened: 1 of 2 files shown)


dschult:
Most of the time, ```G.edges``` is called without ```data``` or ```nbunch```. So there is a separate set of classes for that common, easy to speed up case.

In the ```DataEdgeView``` family, most of the time, ```nbunch is None```. Only ```data``` is specified. So, you can't assume that ```self._nbunch``` is a set.  That is what is causing the error.

You will need to check: ```if self._nbunch is not None and u is not in self._nbunch:``` or similar.

iboates:
That seems to have done it. All the tests are now passing, including the ones I added for testing `__contains__` with an nbunch

dschult:
This fixes #3829 and #3819

PR #3848, opened 2020-03-04
make return value of minimum_st_node_cut consistently a set

### LINKED ISSUE #3846: `minimum_st_node_cut` returns empty list instead of set for adjacent nodes
https://github.com/networkx/networkx/blob/3f4f9c3379a5d70fc58852154aab7b1051ff96d6/networkx/algorithms/connectivity/cuts.py#L284

Should read `return {}`. Was questioning my sanity for a bit there 😉

PalAditya:
Fixes #3846 

As given in the issue description, replaced the list with set and changed the test that was expecting a list to make it expect a set. Please review and tell if anything else is to be done.


dschult:
This looks ready to go. I changed the title to be more informative.
I don't think this affects doc_strings or tests other than the one you changed.
Thanks!

PR #3854, opened 2020-03-05
Fix return values when drawing empty nodes and edges  #3833

sanghack81:
drawing an empty collection of nodes returns an empty PathCollection.
drawing an empty collection of edges returns (i) an empty LineCollection if neither a given graph is a directed graph nor `arrows` is enabled, (ii) an empty list, otherwise.


dschult:
Thanks!

Issue #3330, opened 2019-02-28
kernighan lin optimization


zcollins0:
I have noticed that the NetworkX implementation of Kernighan-Lin is missing the optimizations suggested by here: https://dl.acm.org/citation.cfm?id=259855 They are proven to produce exactly the same result, and contribute a massive speedup to KL. In particular, the column scanning method would be very easy to implement. The NetworkX KL implementation is too slow to use for any sizable graph, and the column scanning optimization should increase its usability to graphs of 10x the current size.

I intend to make a pull request adding column scanning to KL. Would this be more appropriate as an additional algorithm, or is it ok as a change to KL?

dschult:
This sounds great -- As you dive into creating the new data structure for Quick Cut, I would suggest that you try to build it out of native Python structures (like dict) wherever possible. They are quite fast. As for how to name them, why don't you (for now) rename the existing as ```original_ker...``` and use the new methods with the current name. That way we can test against the older method easily and users have both available if they need/want to compare.

zcollins0:
Sure! I'm prepping for my thesis defense the next few weeks, so I won't have a whole lot of time. Hopefully I can have a good deal of this done by the end of March.

boothby:
@zcollins0 are you still planning to work on this?  I recently became annoyed with how slow the K-L implementation is, and I can take this over if you're busy.

boothby:
I implemented the column and diagonal scanning algorithms, as well as an even simpler sweep which alternates between sides to move single nodes to/from.  This "best node first" strategy appears to be common practice.  I tried a bunch of examples, dense and sparse, and didn't observe a significant difference in the distribution of cuts produced by these methods.  The best-first algorithm has the simplest code by far, and ran the fastest overall.

As per @dschult 's comment, I used pre-existing data structures.  I did limited testing of both the PairingHeap and BinaryHeap (with a handful of dense and sparse graphs), and the BinaryHeap won out by about 10%.

dschult:
This looks very nice @boothby :)

Any comments @zcollins0 ?

PR #3858, opened 2020-03-08
implemented faster sweep algorithm for kernighan_lin_bisection

boothby:
Discussion in #3330

### DIFF
--- a/networkx/algorithms/community/kernighan_lin.py
+++ b/networkx/algorithms/community/kernighan_lin.py
@@ -45,7 +45,10 @@ def kernighan_lin_bisection(G, partition=None, max_iter=10, weight='weight',
     algorithm.
 
     This algorithm partitions a network into two sets by iteratively
-    swapping pairs of nodes to reduce the edge cut between the two sets.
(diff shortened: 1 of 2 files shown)


dschult:
This looks very nice!  Thanks!

I have a question about the second test change inside the test_multigraph function. You add a test using 100 different random seeds. Does this test the seed argument or tests all possible different orderings of the labels/nodes? I'd prefer not to have many tests of a random process in our test suite...  Depending on the goal, could we instead either: go through all permutations of the labels (I think there are 24 of them as opposed to 100), or: check that seed is working using some other method?

Do you want to add part of your paragraph from #3330 describing the "best node first" strategy to the docstring?

Any other additions expected to this PR? It's very close to ready as far as I can see.

boothby:
While I didn't intend to leave those 100 iterations in... it was motivated by a bug caused by a typo in computing update costs in my early experiments (wherein certain seeds would get the right answer), perhaps we should go the 24-labelings route. 

Updating that and the docstring; otherwise I think this is good to go.  Thanks a lot!

PR #3861, opened 2020-03-09
Fix reversed args in gaussian_random_partition_graph

jamesjer:
This was discovered while building networkx with python 3.9.0 alpha 4 in Fedora Rawhide.  The reversed arguments cause the decorator function to receive False as the seed argument, which leads to a very strange error down inside python's random code.


dschult:
Yikes!   That's pretty bad...   and looks like might have been caused by our inconsistent ordering of those arguments in this module. 

Thanks very much!

dschult:
Fixed in #3952

PR #3866, opened 2020-03-14
Fix example in docstring

JarnoRFB:


dschult:
Hmm...  strange that this didn't get flagged by our doctests testing.
Maybe pytest uses the module namespace?

Issue #3871, opened 2020-03-20
gaussian_random_partition_graph "directed" issue.


jsal13:
There is a mismatch in the function for `gaussian_random_partition_graph` where `directed` is being sent to `seed` and vice versa.  This makes the default version of the graph directed instead of undirected.

dschult:
Fixed in #3952

PR #3872, opened 2020-03-20
fix: gaussian_random_partition_graph kwargs fix [#3871];

jsal13:
Description: Previously, there was a mismatch where `directed` was being sent to `seed` and vice versa on this line.  This makes it explicit which arg is passed where.

Tested locally and passes.


dschult:
See #3861. And #3871 of course...
 
Thank you very much!

dschult:
I'm going to reopen this....   Is there anything wrong with merging it?
Any further work still on the way, or is this enough to fix the problem.
Thanks!

dschult:
Fixed in #3952

Issue #3849, opened 2020-03-04
Panther Algorithm for Fast top-k Similarity Search


storopoli:
Hello,

Could we have an implementation of the Panther Algorithm (Zhang et al., 2015)[^1] ?

According to the authors, the algorithm is very fast to search for similarity when compared to other implementations. SimRank has a complexity of $mathrm{O}\left(I|V|^{2} \bar{d}^{2}\right)$, where $|V|$ is the number of vertices, $\bar{d}$ is the average degree, $I$ is the number of iterations. And Panther has a complexity of $\mathrm{O}\left(|V| \bar{R}T + |V| \bar{M}\right)$, where $\bar{R}$ is the average number of paths that contain a vertex and is proportional to the average degree $\bar{d}$, $T$ is path length, $\bar{M$} represents the average number of co-occurred vertices with a vertex and is proportional to $\bar{d}$.

![comparison](https://i.imgur.com/Fq7Hy5o.png?1)
---
[^1]: Zhang, J., Tang, J., Ma, C., Tong, H., Jing, Y., & Li, J. (2015). Panther: Fast top-k similarity search on large networks. In Proceedings of the ACM SIGKDD International Conference on Knowledge Discovery and Data Mining (Vol. 2015-August, pp. 1445–1454). Association for Computing Machinery. https://doi.org/10.1145/2783258.2783267

mrecachinas:
@dschult @storopoli Let me know if #3886 is sufficient.

storopoli:
Wow amazing! I will test it right away. Thanks!

mrecachinas:
@storopoli - Note: I didn't implement Panther++ (i.e., vector similarity), just Panther (i.e., path similarity).

storopoli:
Oh ok I will be monitoring #3886

storopoli:
Any news guys?

mrecachinas:
Sorry @storopoli -- see #4400 🙇

PR #3886, opened 2020-03-31
Add Panther algorithm per #3849

mrecachinas:
This addresses #3849.

A couple remaining questions:

* Is there a better home for `generate_random_paths`? It’s a generic Markovian random walk that could be used elsewhere. Alternatively, if this already exists or there's a more efficient approach, I'm all ears.
* I used `numpy` for a bit of this. Is there a desire to have a pure Python version as well (like `simrank_similarity` vs `simrank_similarity_numpy`)?
* Should `n_choose_k` be elsewhere or perhaps should we use `scipy.special.binom` instead?
* Do we want Panther++ added (requires a KD-tree, which we can use from `scipy.spatial`)?


mrecachinas:
There's an issue where node names must be the node's index... fixing this currently.

dschult:
```n_choose_k``` is nicely expressed in Python 3.8+ as ```math.comb(n, k)```.
But we probably need to keep something around until we stop supporting Python 3.7.

There is only a reason to keep a ```numpy``` version and plain python if there is an advantage to each approach. For example, the algorithm used might be different with one using matrix manipulations and one using graph search techniques, or similar.

Thanks for this!  I think there could be demand for Panther++. Let me think about where ```generate_random_paths``` might fit better.  

small tweaks -- typo "ictionary" in docs for ```generate_random_paths```.  And could the example do soemthing with random_path to show that it has two outputs and what someone might want to do with them?  Perhaps you could return just the paths and have an optional argument ```index_map``` that could be filled if supplied. ??

**Comment by mrecachinas** [comment]
@dschult Fixed the typo and moved `index_map` to be an optional kwarg that, if provided, will be populated. I also added an example in the docstring illustrating that.

Let me know if there's anything else. I'll finish my implementation of Panther++.

File: networkx/algorithms/similarity.py [hunk 1]
```diff
@@ -1386,3 +1393,249 @@ def simrank_similarity_numpy(
...
+    G : NetworkX graph
+        A NetworkX graph
+    v : node
+        Source node for whom to find the top ``k`` similar other nodes
+    k : int
+        The number of most similar nodes to return
+    path_length : int
+        How long the randomly generated paths should be (``T`` in [1]_)
+    c : float
+        A universal positive constant. Defaults to 0.5 per [1]_
```
dschult, line 1454: c: float (default = 0.5)
    A universal positive constant used to scale the number of sample random paths to generate.

networkx/algorithms/similarity.py
```diff
@@ -1386,3 +1393,249 @@ def simrank_similarity_numpy(
...
+def panther_similarity(G, v, k=5, path_length=5, c=0.5, delta=0.1, eps=None):
+    """Returns the Panther similarity of nodes in the graph ``G`` to node ``v``.
+    Panther is a similarity metric that says "two objects are considered
+    to be similar if they frequently appear on the same paths." [1]_.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+        A NetworkX graph
+    v : node
```
dschult, line 1447: Call the node source instead of v?

networkx/algorithms/similarity.py
```diff
@@ -1386,3 +1393,249 @@ def simrank_similarity_numpy(
...
+        return 1
+    elif k < n - k:
+        return reduce(mul, range(n - k + 1, n + 1)) // math.factorial(k)
+    else:
+        return reduce(mul, range(k + 1, n + 1)) // math.factorial(n - k)
+
+
+def panther_similarity(G, v, k=5, path_length=5, c=0.5, delta=0.1, eps=None):
+    """Returns the Panther similarity of nodes in the graph ``G`` to node ``v``.
+    Panther is a similarity metric that says "two objects are considered
```
dschult, line 1440: add blank line after single-line function description (used for links to this function in Sphinx doc creation).

networkx/algorithms/similarity.py
```diff
@@ -1386,3 +1393,249 @@ def simrank_similarity_numpy(
...
+    top_k_sorted_names = map(lambda n: node_map[n], top_k_sorted)
+    top_k_with_val = dict(zip(top_k_sorted_names, S[top_k_sorted]))
+
+    # Remove the self-similarity
+    top_k_with_val.pop(v, None)
+    return top_k_with_val
+
+
+def generate_random_paths(G, sample_size, path_length=5, index_map=None):
+    """Randomly generate ``sample_size`` paths of length ``path_length``.
```
dschult, line 1546: Would ```generate_random_paths``` be better as a generator yielding paths?

networkx/algorithms/similarity.py
mrecachinas: Addressed in 81e2307.

networkx/algorithms/similarity.py
mrecachinas: Addressed in 221da98.

networkx/algorithms/similarity.py
mrecachinas: Addressed in 81e2307.

networkx/algorithms/similarity.py
mrecachinas: That makes sense to me. Addressed in 49302de.

storopoli:
Any news guys? Will it be released in the next version?

dschult (commented):
I just have a few nits to pick. And, yes...  This will definitely get into the next release.

- Could we keep n_choose_k an unadvertised function by taking it out of ```__all__```? It really should be ```math.comb(n, k)``` but we have to wait for Python v3.8 to make that work. Perhaps a note in the doc_string to that effect is reasonable too.

- Add hooks to the docs in ```doc/reference/algorithms/similarity.rst``` by just adding the names of the new functions to that file.  That'll pull the doc_strings into the sphinx doc system automagically.

- I believe the ```setdefault``` method is advised against. Can you replace that idiom with something like:
```python
if node in index_map:
    index_map[node].add(path_index)
else:
    indexmap[node] = {path_index}
```

- You add ```1/ sample_size``` repeatedly does it make any difference in speed if you store ```1/sample_size``` once snd then use it vs recomputing in the loop?  I assume the difference is small, but my ancient fortran background is making me twitch. :}  You can leave it or change... 

That's all I've got.  Nice job -- very clean and clear code. :}

dschult:
Rats -- I didn't notice....  two things:  
1) you don't need the ```>>> import networkx as nx``` in the doc_string examples. That is run automagically by the testing software before the example code.
2) you can use ```n_choose_k``` without reference to ```nx.``` because it is in the current module.

So I think the errors go away by removing ```nx.``` from the ```nx.n_choose_k``` and for looks, please take out ```import networkx as nx```

Thanks!

mrecachinas:
@dschult Think I addressed everything. Let me know if there's anything else you would like changed!

Also, I'll submit another PR for Panther++ once I finish it.

storopoli:
@mrecachinas great job! Waiting for the panther++ algorithm also...

dschult:
Quick question:  In the docs, delta is:  
The probability that 𝑆 is not an epsilon-approximation to (R, phi)

What are R and phi?

You can answer here-- but if you prefer, you can add more to that doc string.

storopoli:
@dschult according to the original Panther article, $R$ is the number of random paths and $\phi$  is the probability that an element sampled from a set A $\subseteq$ D, where D is the domain.

ref: Zhang, J., Tang, J., Ma, C., Tong, H., Jing, Y., & Li, J. (2015). Panther: Fast Top-k Similarity Search on Large Networks. Proceedings of the 21th ACM SIGKDD International Conference on Knowledge Discovery and Data Mining, 1445–1454. https://doi.org/10.1145/2783258.2783267

dschult:
Good Thanks!   

But what is S... ?   I don't think these symbols are defined in the doc_string.... only in the article.
The code refers to "S", but it is an array of values, which is hard to see how to be epsilon close to a 2-tuple (R, phi)...

mrecachinas:
Sorry - S is just the similarity.

jarrodmillman (changes requested):

dschult (approved):

mrecachinas:
@jarrodmillman Let me know if I've addressed your change requests.

jarrodmillman:
@mrecachinas Sorry for the delay.  Ross or I will take a look very soon!  We are planning to release 2.6 soonish and I want to get this in first.  Thanks for your patience.

Despite our slow response time, we would also like to get Panther++ in soon.  So I hope you are still interested in contributing it!

rossbar (approved):
I went ahead an made some small touch-ups in rossbar/networkx@0973646c. I tried to push these to your branch (to spare you the spam via the github suggestion feature) but was unable to do so, likely because it seems like your fork originated from somewhere other than `networkx/networkx`. I wasn't able to PR against your fork either, so if you could apply those changes via cherry-pick that'd be great. If you're not comfortable doing so, you can either:
 1) Add rossbar as a collaborator to your fork and I'll open a PR against your fork or
 2) We just put this in and I can make a follow-on PR

Generally this LGTM - I'm not very familiar with the algorithm so my opinion shouldn't be weighted too heavily but the implementation seems good! I too was struck by the question of whether there is a better place for `generate_random_paths`, but I don't have any ideas and I don't think that should be a sticking point. Thanks @mrecachinas !

mrecachinas:
@jarrodmillman I'll dust off and finish the Panther++ I started a few months ago.

@rossbar Ahh, I think I forked someone else's fork for a previous PR (adding SimRank, IIRC), so yeah that's probably the issue. I'll cherry-pick if you're good with it.

PR #3891, opened 2020-04-03
added default arg for json dumps for jit_data func

lookbeat:


dschult:
Hi!  Thanks for this!

Can you add "default" to the documentation of the parameters for this function?
You can just say that it is passed to the ```json.dumps``` function and have them look up that function to figure out what values are possible if that is easiest. But there should be something describing what the ```default``` argument does.

lookbeat:
sure thing

PR #3893, opened 2020-04-04
Fixing docs for nx.info(), along with necessary tests

raamana:


networkx/classes/function.py
```diff
@@ -525,14 +525,28 @@ def create_empty_copy(G, with_data=True):
 
 
 def info(G, n=None):
-    """Print short summary of information for the graph G or the node n.
+    """
+    Return a short summary of information for the graph G or the specified node n.
```
dschult, line 529: Can you make the first line of the doc_string a short one-line description?
Then leave a blank line and then include a paragraph (like what you've got).
That way Sphinx can identify the one-line description for any time the function 
is referred to.    Thanks!

PR #3895, opened 2020-04-05
Made CONTRIBUTING.rst more clearer

Mackyboy12:
I saw that in CONTRIBUTING.rst that is wasn't clear that you had to navigate to the folder before adding the upstream repo. So I added that to make sure that that was clearer for newbies like me.


dschult:
Thanks!

PR #3897, opened 2020-04-06
added basic uniform cost search listing edges

aijanai:
Added support for Uniform Cost Search over tree-like graphs.

Dijkstra shortest path is not able to generalize a visit of a graph guided by the weight of edges, so I came up with a simple implementation based on a heap and a check that avoids visiting same node again.

Thanks


aijanai:
Fixes #3899

dschult:
Can you define briefly what a uniform cost search is in the paragraph after the blank line near the top of the doc_string?   Thanks!

Can you explain how/why this doesn't generalize from djikstra? Do we need yet another search? :)

aijanai:
Hi,

this is not a shortest path algorithm. This is a traversal, it just returns a list of edges.: it is the equivalent of bfs_edges, or dfs_edges, but with node expansion policy guided by the weights of edges.
I'll add the description to the code as per your request.

Many thanks!

dschult:
This is a tirade on naming with the final result that we stick with what you've got...

tl;dr :   uniform cost search is a "bad name" for this type of search. But we'll use it anyway...

long version:
Unfortunately, the term "uniform cost search" is misleading/confusing at best, even if the AI world has adapted it. The word "uniform" seems to mean that the paths on the fringe are approximately uniform. Yet, the distinction between them is precisely what the search is based upon... so why bring focus on their uniformity when the algorithm is exploiting their non-uniformity.  

I tried to look for a better term -- and indeed, it is often used interchangeably with Dijkstra based searches. A search (it's even the same as your link) on [Wikipedia](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm#Practical_optimizations_and_infinite_graphs) for uniform cost search comes up with Dijkstra on infinite graphs...  and [stackoverflow](https://stackoverflow.com/questions/12806452/whats-the-difference-between-uniform-cost-search-and-dijkstras-algorithm) suggests that a uniform cost search is a dijkstra search with specified source and target, even though that's the problem Dijkstra originally formulated.   

But, in the end, I have to admit that the term "Graph Theory" is a "bad name" too. A graph is better described as a network. The word graph is usually associated with visualization of the relationship between 2 continuous variables.

Future work based on this idea may include ```ucs_successors, ```ucs_predecessors``` and maybe even ```edge_ucs```, where the naming conventions could follow those of the breadth_first_search functions:

```The difference is that ‘edge_bfs’ yields edges even if they extend back to an already explored node while ‘bfs_edges’ yields the edges of the tree that results from a breadth-first-search (BFS) so no edges are reported if they extend to already explored nodes. That means ‘edge_bfs’ reports all edges while ‘bfs_edges’ only report those traversed by a node-based BFS.```

dschult:
This code should be improved slightly in the following ways:
- get the weight argument to match function-interface of the dijkstra code
- make sure it work with nodes that can't be sorted by include a counter in the priority queue -- like ```_multisource_dijkstra```

Documentation:
- explain in docs that edges back to seen nodes are not yielded
- add "see also" links to the bfs_edges/dfs_edges functions as well as to dijkstra

More Tests?:
- think about what happens with negative weights/cycles,  MultiGraphs, DiGraphs, MultiDiGraphs and other edge cases.
- add a test to link this with dijkstra path reporting on a simple graph to ensure they report the same thing

That's a fair amount of futzing around -- fix what you can and then I'll help with the rest.
Thanks!

PR #3900, opened 2020-04-07
add paley graph

sauxpa:
Hi,

I'm a young researcher in math with a keen interest in graphs. I've been using networkx a lot recently so I'd like to start contributing.

I've been doing some work around expanders, a class of regular graphs for which there are only two generators currently available in networkx. This is a proposal to add the class of Paley graphs, a family of graphs on p nodes obtained by matching x and y iif y-x is a square mod p. 

Said graphs are indeed regular when p is a prime number (there are (p-1)/2 distinct squares in this case).

I tried to match the style of existing functions in expanders.py and the associated test, but any feedback would be greatly appreciated.

Thanks,

Patrick

P.S : below is a visualization of Paley graphs for a few increasing values of p:
paley

### DIFF
--- a/networkx/generators/expanders.py
+++ b/networkx/generators/expanders.py
@@ -4,7 +4,7 @@
 import itertools
 import networkx as nx
 
-__all__ = ['margulis_gabber_galil_graph', 'chordal_cycle_graph']
+__all__ = ['margulis_gabber_galil_graph', 'chordal_cycle_graph', 'paley_graph']
 
 
 # Other discrete torus expanders can be constructed by using the following edge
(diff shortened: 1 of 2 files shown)


dschult:
This looks good!  Thanks very much!
I'm ready to merge unless something else is coming.  
We can always do another PR for other things too.

sauxpa:
Thanks! Ready to merge, I'll open another PR for the rest when it's ready.

dschult:
Rats -- I forgot to ask you to add the new function to the expanders section of ```doc/reference/generators.rst```.  Otherwise it won't show up in the documentation.
Can you make a very short PR with that?
Thanks for this!

sauxpa:
No worries, would you mind taking a look at https://github.com/networkx/networkx/pull/3927 ?
Cheers

PR #3907, opened 2020-04-11
allow set of edge nodes

### LINKED ISSUE #3884: Add support for sets of graph edges
Currently it looks like only tuples and lists of graph edgeds are supported, however with a small change I think it would be easy to add support for any Iterable, providing greater flexibilty to users of the library. Are there any considerations I'd need to take before submitting a PR?

Mattwmaster58:
Fixes #3884

### DIFF
--- a/networkx/convert.py
+++ b/networkx/convert.py
@@ -50,7 +50,7 @@ def to_networkx_graph(data, create_using=None, multigraph_input=False):
          any NetworkX graph
          dict-of-dicts
          dict-of-lists
-         list of edges
+         iterator yielding edges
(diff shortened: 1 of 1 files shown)


dschult:
This looks good -- Thanks!

Can you change the docs to "container of edges"? 
Strictly speaking a list is an iterable, but not an iterator.
That language is easy to mis-state, so I'd prefer "container".

Mattwmaster58:
Sure. I wasn't aware the difference between an iterable vs iterator, thanks for pointing that out.

Mattwmaster58:
@dschult Could you check it out now?

PR #3916, opened 2020-04-14
In Pajek export add extra information when casting 'id' to int() fails.

### LINKED ISSUE #3910: Pajek export assumes Graph node's id attribute to be int castable
In pajek export it reads https://github.com/networkx/networkx/blob/b0ca6ce081cc6e66d509c5f7fed60728da821c25/networkx/readwrite/pajek.py#L57

I was wondering, since NetworkX supports arbitrary Python objects as Graph nodes where does the assumption come from that a node's `id` attribute is `int` castable? An objects `id` attribute would have nothing to do with Python's builtin `id()`. But I'm not familiar with the Pajek file format so don't know about the relevance and special treatment of `id` in it.

I was working with a Graph whose node's `id` attribute happened to be a string. For it Pajek export failed with
```
ValueError: invalid literal for int() with base 10: 'arn:aws:iam::579763146468:role/aws-service-role/trustedadvisor.amazonaws.com/AWSServiceRoleForTrustedAdvisor'
```

lloesche:
Resolves #3910
Tested with Python 2.7 and 3.8
Preserves the original error message and adds the extra information as an additional argument.

### DIFF
--- a/networkx/readwrite/pajek.py
+++ b/networkx/readwrite/pajek.py
@@ -54,7 +54,12 @@ def generate_pajek(G):
         na = G.nodes.get(n, {}).copy()
         x = na.pop('x', 0.0)
         y = na.pop('y', 0.0)
-        id = int(na.pop('id', nodenumber[n]))
+        try:
+            id = int(na.pop('id', nodenumber[n]))
+        except ValueError as e:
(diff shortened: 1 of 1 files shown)


lloesche:
Updated and squashed for 88 char line length PEP 8 check.

PR #3919, opened 2020-04-15
Correctly infer numpy float types

alexhenrie:
This fixes a bug where np.float64 values are given an integer type in the GraphML file, which leads to parsing errors.

### DIFF
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -382,9 +382,10 @@ class GraphMLWriter(GraphML):
             types = self.attribute_types[(name, scope)]
 
             if len(types) > 1:
-                if str in types:
+                types = set(self.xml_type[t] for t in types)
+                if "string" in types:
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for this!  
Can you remove the deepcopy from the tests?
It should work without deepcopy right?

alexhenrie:
It would work, I just didn't want to interfere with attribute_numeric_type_graph's original purpose. But if you're not concerned about that then I'm happy to change it.

alexhenrie:
I have made the requested changes. Since `test_numpy_float64_inference` is run after `test_write_read_attribute_numeric_type_graphml`, even without the deepcopy there is in practice no change to the tests that `attribute_numeric_type_graph` was originally for.

alexhenrie:
@dschult Could you take another look please?

alexhenrie:
Thank you!

PR #3925, opened 2020-04-18
Add algorithms for regular graphs

camdencheek:
Adds the functions `is_regular`, `is_k_regular`, and `k_factor`.

The functions `is_regular` are fairly self-explanatory. The
`k_factor` function implements an algorithm for computing the k-factor
of a graph. The algorithm is described in "An algorithm for computing simple
k-factors" (Meijer, H., Núñez-Rodríguez, Y., & Rappaport, D. 2009).

### DIFF
--- a/doc/reference/algorithms/index.rst
+++ b/doc/reference/algorithms/index.rst
@@ -55,6 +55,7 @@ Algorithms
    planarity
    planar_drawing
    reciprocity
+   regular
    rich_club
    shortest_paths
    similarity
--- a/doc/reference/algorithms/regular.rst
+++ b/doc/reference/algorithms/regular.rst
@@ -1,0 +1,11 @@
+*******
+Regular
+*******
+
+.. automodule:: networkx.algorithms.regular
+.. autosummary::
+   :toctree: generated/
+
+   is_regular
+   is_k_regular
+   k_factor
--- a/networkx/algorithms/__init__.py
+++ b/networkx/algorithms/__init__.py
@@ -37,6 +37,7 @@ from networkx.algorithms.operators import *
 from networkx.algorithms.planarity import *
 from networkx.algorithms.planar_drawing import *
 from networkx.algorithms.reciprocity import *
+from networkx.algorithms.regular import *
 from networkx.algorithms.richclub import *
 from networkx.algorithms.shortest_paths import *
 from networkx.algorithms.similarity import *
--- a/networkx/algorithms/regular.py
+++ b/networkx/algorithms/regular.py
@@ -59,6 +59,9 @@ def k_factor(G, k, matching_weight='weight'):
     """Compute a k-factor of G
 
     A k-factor of a graph is a spanning k-regular subgraph.
+    A spanning k-regular subgraph of G is a subgraph that contains
+    each vertex of G and a subset of the edges of G such that each
+    vertex has degree k.
 
     Parameters
     ----------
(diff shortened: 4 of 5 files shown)


camdencheek:
Could someone help me out with the travis build? Seems to be complaining that `numpy` does not exist.

dschult:
Actually, could you remove the numpy requirement? 

The ```is_k_regular(G)``` code can be:
```python
all(d == k for n, d in G.degree)
```
And ```is_regular(G)``` is not much different once you get an arbitrary node:
```python
n1=nx.utils.arbitrary_element(G) 
if not G.is_directed(): 
    d1 = len(G.adj[n1])    # could be  d1 = G.degree(n1)
    return all(d1 == d for n, d in G.degree) 
else:
     d1=len(G.pred[n1]) 
     in_regular = all(d1 == d for n, d in G.in_degree) 
     d1=len(G.succ[n1]) 
     out_regular = all(d1 == d for n, d in G.out_degree) 
     return in_regular and out_regular 
```

That said, the way to handle the numpy requirement is to skip the tests when numpy isn't available. You use ```numpy = pytest.importorskip('numpy')``` or similar.
Take a look at the tests for another module that imports numpy.

But I don't think you need numpy and it will be faster not to have to create the (potentially large) numpy array just to compute degree.

camdencheek:
Thanks for taking a look. I removed the numpy requirement in favor of your snippets. I didn't realize there was so much overhead to creating the numpy array.

dschult:
For small graphs the overhead is negligible. But for 10K nodes it is significant and for 100K it probably won't fit in many computers. And we've got the in_degree and out_degree as just the length of a dict so no need to create and sum a matrix. :}

Some other comments/requests:
- you define k-regular and regular in their functions quite well :) For the k-factor code, can you break down "spanning k-regular subgraph" with another sentence or two to make it easy to read. Leave the sentence as is... but add something to explain k-regular and spanning subgraph.
- can you find a more descriptive name than gadget? Also the method ```resolve_node``` takes ```g``` as an argument but doesn't seem to have a node in sight. It makes it look like "g" is the node you are resolving. I think this is only used immediately after creating the gadget. So maybe it can just be called from __init__ with the graph passed into the gadget constructor. But I may be missing some reason to have it separated out to call later...
- you don't need the ```g.add_node(outer)``` or ```inner``` just before you add an edge between them. The add_edge method adds the nodes too. (ok either way really not big deal)
- For the exception for not perfect matching, can you make the message refer somehow to k-factor? As it is, someone calling it may be looking for a k-factor and not know what to do with a response that a perfect matching does not exist. 

This is nice.. Thanks!

camdencheek:
Thanks for the quick and detailed review. I appreciate you taking the time to explain things

With regards to the name "gadget", that's the name for this construct in the paper I'm referencing, so I felt it was best use the terminology from the paper for the sake of anyone using it as a reference.

camdencheek:
I implemented the changes from your feedback and also added another test to bump coverage up. Let me know what you think on the name "gadget". I'm happy to change it if you feel that it's worth a clearer name than that used in the paper.

dschult:
I think the name "gadget" should not have been used in the paper -- but it was... :}  So, it's fine where it is.

dschult:
OK... last bits:  

- add these functions to the docs by creating a section in ```doc/reference/algorithms/regular.rst``` Use ```reciprocity.rst``` or others as a model to work from.  To test locally you have to install sphinx and others in ```requirements/doc.txt```. But Travis-CI will check if it runs without checking locally.
- Add something to ```networkx/algorithms/__init__.py``` to get the function to be included in the main namespace.

Ask if you have questions, or would prefer I do this stuff. I think we're close to merging.

camdencheek:
I think I managed to get the docs right, but I'm not super familiar with sphinx. I built them locally and was able to view them in a browser fine, but there were a few errors that seemed unrelated.

PR #3927, opened 2020-04-20
add paley graph to doc

sauxpa:
Just merged paley graph (https://github.com/networkx/networkx/pull/3900) but forgot to add an entry in the doc, so here goes.


dschult:
Great!  Thanks!

PR #3928, opened 2020-04-21
Fix argument order of call inside gaussian_random_partition_graph 

Songyu-Wang:


dschult:
This adds tests to the fix from #3952

Issue #3930, opened 2020-04-22
nx.to_pandas_edgelist() will not update the "source" and "target" columns if they already exist


iboates:
I discovered this during a workflow where I am transforming between pandas dataframes and networkx graphs rather frequently. The old "source" and "target" columns were hanging around from a previous transformation.

I realize that it is probably bad practice to already have "source" and "target" as attributes in a graph, especially when they are wrong, but there is no mention of this behaviour on the [docs page](https://networkx.github.io/documentation/stable/reference/generated/networkx.convert_matrix.to_pandas_edgelist.html) and it seems to me that a more intuitive behaviour would be for it to overwrite the columns in the dataframe with the actual current graph information, and not what may be hanging around in the attributes list.

See below. In this example, I create an undirected graph with a "backwards" edge, then straighten the graph out by transforming it into a tree [(see here)](https://stackoverflow.com/questions/60742779/creating-a-directed-graph-from-an-undirected-graph-with-dfs-tree-but-keep-attri). However, since "source" and "target" already exist (perhaps from a previous conversion from a graph to dataframe, and then back to a graph again), then they are not updated when writing back to a dataframe again.

```
import networkx as nx

graph = nx.Graph([
    (0, 1, {"source": 0, "target": 1}),
    (1, 2, {"source": 1, "target": 2}),
    (3, 2, {"source": 3, "target": 2})
])
digraph = nx.DiGraph(graph)
digraph.remove_edges_from(digraph.edges - nx.dfs_edges(graph, 0))

df1 = nx.to_pandas_edgelist(digraph)  # Default is to use source="source" and target="target"
print(df1.head())
df2 = nx.to_pandas_edgelist(digraph, source="source2", target="target2")
print(df2.head())
```

Output of ´print(df1.head())´ is:

```
   source  target
0       0       1
1       1       2
2       3       2
```

Output of ´print(df2.head())´ is:

```
   source2  target2  source  target
0        0        1       0       1
1        1        2       1       2
2        2        3       3       2
```

dschult:
Yup -- it looks like the code actually does write "source" and "target" to the edge attributes and then overwrites them with the previously existing data.  Maybe it should raise an exception if the columns already exist. It's not good to overwrite data without telling someone.

iboates:
[...]

PR #3935, opened 2020-04-26
Fix for #3930 (source & target columns not overwritten when converting to pd.DataFrame)

iboates:
`nx.to_pandas_edgelist()` will now check the source and target column names individually and raise a `NetworkXError` if there is a detected conflict.

Two tests also added to verify this. They try to convert a graph with source and target column names that already exist as attributes. They will only assert True if a `NetworkXError` is raised, otherwise they will assert False.

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -259,9 +259,9 @@ def to_pandas_edgelist(
     target_nodes = [t for s, t, d in edgelist]
     all_keys = set().union(*(d.keys() for s, t, d in edgelist))
     if source in all_keys:
-        raise NetworkXError("Specified source column '{}' already exists as an attribute in some edges.".format(source))
+        raise NetworkXError(f"Source name '{source}' is an edge attribute name")
     if target in all_keys:
-        raise NetworkXError("Specified target column '{}' already exists as an attribute in some edges.".format(target))
+        raise NetworkXError(f"Target name '{target}' is an edge attribute name")
     edge_attr = {k: [d.get(k, float("nan")) for s, t, d in edgelist] for k in all_keys}
     edgelistdict = {source: source_nodes, target: target_nodes}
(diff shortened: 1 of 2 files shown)


dschult:
Excellent -- thanks for this! I especially like the tests...

Comments/suggestions:
- The two tests can be nicely combined by using ```pytest.raises``` in a similar way to the ```test_exceptions``` method which appears about line 25 in the testing file. You can also follow that style to remove need for importing ```NetworkXError```.

- Creation of the test graph can be more compact using 
```python
G = nx.path_graph(10)
G.add_weighted_edges_from((u, v, u) for u, v in list(G.edges))
```
- Can you add a test of the NetworkXError using the default source/target names?
```python
nx.set_edge_attributes(G, 0, name="target")
pytest.raises(nx.NetworkXError, nx.to_pandas_edgelist, G)
```
and similarly for ```name="source"``` .

- The error message (in the function itself -- not tests) is rather long. 
And we're slowly switching to [formatted strings](https://docs.python.org/3/tutorial/inputoutput.html).  So, how about (?)
```nx.NetworkXError(f"Source name "{source}" is an edge attribute name")'```

iboates:
Thanks for the feedback.

- Combined the tests used `pytest.raises` but also made default & custom source/target columns separate
- Made graph creation more compact like you suggested
- Shortened exception text as suggested (PEP8 should not complain now)
- I also prefer f-strings but I saw that the library is supposed to support Python 3.5, but f-strings are only compatible with Python 3.6 and forward. I added it anyway as you suggested but that could be an issue for 3.5.

dschult:
You are correct that NetworkX v2.4 supports Python 3.5. But with v2.5 we are moving to support Python 3.6-3.9. So anything that gets merged now will be Python 3.6+.  f-strings are ok for that.
Thanks!

PR #3947, opened 2020-05-07
Update astar.py to fix docstring for the commit  #3799

cmkpl:
Correct the docstring of the function astar_path for the commit  #3799


dschult:
Thanks!

Issue #3948, opened 2020-05-07
Feature request: pass weight functions in class algorithms.simple_paths like in class algorithms.shortest_paths - Code added, pending for merge


cmkpl:
For consistency, if weight function can be passed into the functions in the class algorithms.simple_paths like in the class algorithms.shortest_paths, it would be great.

cmkpl:
In networkx.algorithms.simple_paths,

I tried to replace 

```
def length_func(path):
    return sum(G.adj[u][v][weight] for (u, v) in zip(path, path[1:]))
```

to 

```
if callable(weight):
    def length_func(path):
        return sum(weight(u, v, G.get_edge_data(u, v)) for (u, v) in zip(path, path[1:]))
else:
    def length_func(path):
        return sum(G.adj[u][v][weight] for (u, v) in zip(path, path[1:]))
```

but the result is not so promising as follows -

```
from numpy.random import random

G = nx.grid_graph(dim=[500, 500])
nx.set_edge_attributes(G, {e: random()*100 for e in G.edges()}, 'cost')
from itertools import *  

def cost(u,v,x):
    if u == (0,0) and v == (1,0):
        cost = 100000
    elif u == (0,1) and v == (1,1):
        cost = 100000
    else:
        cost = x["cost"]
    return cost
       
def k_shortest_paths(G, source, target, k, weight):
    return (islice(nx.shortest_simple_paths(G, source, target, weight), k))

for path in k_shortest_paths(G, (0,0), (20,5), 6, weight=cost):
    print(path)
    print(sum(cost(u,v, G.get_edge_data(u, v)) for (u, v) in zip(path, path[1:])))
    print("**********")
```

Result:
```
[(0, 0), (1, 0), (2, 0), (3, 0), (4, 0), (5, 0), (6, 0), (7, 0), (8, 0), (9, 0), (10, 0), (11, 0), (12, 0), (13, 0), (14, 0), (15, 0), (16, 0), (16, 1), (16, 2), (16, 3), (16, 4), (16, 5), (17, 5), (18, 5), (19, 5), (20, 5)]
102400
**********
[...]

PR #3949, opened 2020-05-08
Add weight function for shortest simple paths for #3948

cmkpl:
Add this function for aligning the weight function in shortest paths class.

### DIFF
--- a/networkx/algorithms/simple_paths.py
+++ b/networkx/algorithms/simple_paths.py
@@ -380,8 +380,13 @@ def shortest_simple_paths(G, source, target, weight=None):
         length_func = len
         shortest_path_func = _bidirectional_shortest_path
     else:
-        def length_func(path):
-            return sum(G.adj[u][v][weight] for (u, v) in zip(path, path[1:]))
+        if callable(weight):
+            def length_func(path):
+                return sum(weight(u, v, G.get_edge_data(u, v)) for (u, v) in zip(path, path[1:]))
+        else:
+            def length_func(path):
+                return sum(G.adj[u][v][weight] for (u, v) in zip(path, path[1:]))
+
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for this! 
Can we use the [utility function currently](https://github.com/networkx/networkx/blob/master/networkx/algorithms/shortest_paths/weighted.py#L40) in ```algorithms/shortest_paths/weighted.py``` called ```_weight_function```? 

I think that could remove a lot of the extra code that is deciding which way to compute weights. It might even be faster, though that is hard to predict in Python. The idea would be to call this centrally defined function on the input ```weight``` and then use the returned value as a function that computes the weight for each edge. The checking for multigraph is done once when ```weight``` is defined.  It's also possible it doesn't work for this case, but I think it does work.

I guess you'd have to import with a long name like: 
```python
```

cmkpl:
Thanks for your suggestion.  Please find the updated commit in this request.

dschult:
This looks good... Thanks!!
While I've got your attention, I've been thinking that a better name of this function would be something like ```simple_paths_by_length```. Does this make sense to you?  
I don't want to do it in this PR.  I'll talk to more people and do it as part of prep for v2.5.
But do you think of the old name? of  the proposed name?

cmkpl:
It seems like `simple_paths_by_length` implies that length is a `parameter` to be used which is optional.  How about something like `simple_paths_sorted_by_length`? But it is a bit long and not aligned with the conventions like `lexicographical_topological_sort`.

What concerned me the most is whether it will break the backward compatibility.

dschult:
Yes, I agree that backward compatibility is important.   Maybe when we go to v3.0...
The problem with ```shortest_simple_paths``` is that it yields paths that aren't the shortest.

PR #3960, opened 2020-05-16
Add option for named key ids to GraphML writing.

ryan-duve:
Provide the option to give the `<key>` elements' `id` attribute the same value as the `attr.name`'s attribute value.

```
<?xml version='1.0' encoding='utf-8'?>					<?xml version='1.0' encoding='utf-8'?>
<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:	  |	<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:
<key id="d1" for="node" attr.name="prop2" attr.type="string"/	  |	<key id="prop2" for="node" attr.name="prop2" attr.type="strin
<key id="d0" for="node" attr.name="prop1" attr.type="string"/	  |	<key id="prop1" for="node" attr.name="prop1" attr.type="strin
<graph edgedefault="directed"><node id="0">				<graph edgedefault="directed"><node id="0">
  <data key="d0">val1</data>					  |	  <data key="prop1">val1</data>
  <data key="d1">val2</data>					  |	  <data key="prop2">val2</data>
</node>									</node>
<node id="1">								<node id="1">
  <data key="d0">val_one</data>					  |	  <data key="prop1">val_one</data>
  <data key="d1">val2</data>					  |	  <data key="prop2">val2</data>
</node>									</node>
<edge source="0" target="1">						<edge source="0" target="1">
  <data key="d2">edge_value</data>				  |	  <data key="edge_prop">edge_value</data>
</edge>									</edge>
</graph></graphml>							</graph></graphml>
```

### DIFF
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -409,7 +409,7 @@ class GraphMLWriter(GraphML):
         try:
             return self.keys[keys_key]
         except KeyError:
-            if self.named_key_ids is True:
+            if self.named_key_ids:
                 new_id = name
             else:
                 new_id = f"d{len(list(self.keys))}"
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -61,7 +61,7 @@ __all__ = [
 
 @open_file(1, mode='wb')
 def write_graphml_xml(G, path, encoding='utf-8', prettyprint=True,
-                      infer_numeric_types=False):
+                      infer_numeric_types=False, named_key_ids=False):
     """Write G in GraphML XML format to path
 
     Parameters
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for this idea and implementation. 

Can you motivate this change in a sentence or two? (Why should we do this instead of sticking to the style of the doc examples for GraphML?)

Also, will this impact reading the resulting file? How will it impact the attributes of the read Graph? 

(minor PEP8 style change) can you remove the "is True" when checking ```self.self.named_key_id is True```?

ryan-duve:
Thanks @dschult. I fixed the line for PEP8 compliance and added tests in the last two commits.

The `<key id=d0 attr.name="my_name">`-style in the GraphML examples is fine; this change does not affect that behavior by default.

It is helpful to have an option of alternatively assigning the same value for the `id` attribute as the `<key>` element's `attr.name` attribute, e.g., `<key id="my_name" attr.name="my_name">`.  This removes the need to post-process GraphML, substituting the `d0`, `d1`, ... strings in the `<key>` and `<data>` elements, so they are properly read by Gremlin Server, Amazon Neptune's GraphML conversion tool, and any other implementation that use the `id` this way.

The read graph's nodes and attributes are unchanged when written this way.  One of the tests included in this PR demonstrates this by writing the same graph both ways, reading it in twice and comparing all the nodes and edges.

Please let me know if you think of another way I can improve this contribution.

dschult:
Thanks very much for the code and your responses!

Issue #3953, opened 2020-05-14
Same Graph, different Simrank results.


Carghaez:
Hello there! 
![simrank](https://user-images.githubusercontent.com/297838/81880095-ccabe980-958c-11ea-9058-46bcab40ee11.PNG)

I want to reproduce the simrank of the graph linked above, so I use your FL simrank_similarity and simrank_similarity_numpy and... I noticed that the results are different!

steps to reproduce

```python
import networkx
import numpy

options = {
    'font_weight': 'bold',
    'node_color': 'red',
    'font_color': 'black',
    'edge_color': 'red',
    'node_size': 1000,
    'width': 2,
    'arrowstyle': '-|>',
    'arrowsize': 15
}

Labels = {0:'Univ', 1:'ProfA', 2:'ProfB', 3:'StudentA', 4:'StudentB'}

G = networkx.DiGraph()

G.add_node(0, label='Univ')
G.add_node(1, label='ProfA')
G.add_node(2, label='ProfB')
G.add_node(3, label='StudentA')
G.add_node(4, label='StudentB')

G.add_edges_from([(0, 1), (0, 2), (1, 3), (2, 4), (4, 2), (3, 0)])

# Debug
# networkx.draw_networkx(G, arrows=True, with_labels=True, labels=Labels, **options)

# simrank_similarity
sim = networkx.algorithms.similarity.simrank_similarity(G)
sim_array = numpy.array([[sim[u][v] for v in sorted(sim[u])] for u in sorted(sim)])
print(sim_array)

# simrank_similarity_numpy
sim_array_numpy =  networkx.algorithms.similarity.simrank_similarity_numpy(G)
print(sim_array_numpy)
```

Based on the image above, the simrank_similarity_numpy shows the expected weights as opposed to simrank_similarity which shows unexpected values.

Maybe I am missing something important or somewhat else?

Thanks in advice!

ryan-duve:
I took a look at this and was able to reproduce the problem.  I opened [a PR](https://github.com/networkx/networkx/pull/3961) proposing a fix.

dschult:
Thanks very much for this report!  Nicely put together... (and followed up when a fix was proposed) Thanks...

PR #3961, opened 2020-05-17
Fix simrank_similarity with directed graph input

ryan-duve:
User @Carghaez pointed out the SimRank function is not behaving properly in [Issue 3953](https://github.com/networkx/networkx/issues/3953).  It appears the algorithm was trying to iterate over out-neighbor pairs.  On the [top of page 540 of the original paper](https://dl.acm.org/doi/pdf/10.1145/775047.775126) it says the algorithm should iterate over in-neighbor pairs.

This fix is performed by changing `G[<node>]` to `G.predecessors(<node>)`.

**Note to maintainers:** I had to change several tests after the fix.  A cycle graph using `DiGraph` works, but it gives trivial results since no nodes have common nodes pointing to them.  For a non-trivial choice, I chose to go with the implementation in the paper (the same one @Carghaez used in the issue he raised).  The `expected` strings match results from both the numbers in the paper and the `simrank_similarity_numpy` function.

**Disclaimer**: I just learned about SimRank today and am by no means an expert, so please review this change with a critical eye.


Carghaez:
I also found this bug a few days ago and I confirm that is the right fix!

dschult:
This looks like you changes from using ```G[u]``` to ```G,predecessors(u)``` but that won't work for undirected graphs. Then you removed the tests of undirected graphs and changed them to directed graphs. So, while it is passing all the tests now, it still won't work for undirected graphs. Am I correct, or am I missing an if-clause somewhere that makes this only get used for directed graphs?  I'll get a chance to dig around in this sometime soon. But it would reassure me if you left the current tests of cycle_graphs and just added the new tests of DiGraphs.  It's Ok to have two assert statements in the same test. Just leave the existing one and add the new one.

Let me know if I'm missing something.

Carghaez:
@dschult I already fix that issue here https://github.com/networkx/networkx/pull/3962

ryan-duve:
@dschult Here's what happened:

1. I opened the original PR supporting DiGraphs but breaking tests for undirected graphs because I mistakenly thought undirected graphs weren't supported by SimRank.
1. @Carghaez recognized the error and created a regression PR with a conditional to support both directed and undirected graphs.
1. This morning I pulled in the changes from that PR into this one.
1. This morning I restored the tests for undirected graphs and documentation, putting several assert in the same tests like you suggested.

The tests for the PR are now additive only and directed graphs are supported.

networkx/algorithms/similarity.py
```diff
@@ -1270,7 +1270,11 @@ def avg_sim(s):
         return sum(newsim[w][x] for (w, x) in s) / len(s) if s else 0.0
 
     def sim(u, v):
-        return importance_factor * avg_sim(list(product(G[u], G[v])))
+        if nx.classes.function.is_directed(G):
```
dschult, line 1273: Maybe combine avg_sim and sim using something like:

```python
Gadj = G.pred if G.is_directed() else G.adj
def sim(u, v):
    pairs = list(product(Gadj[u], Gadj[v]))
    if not pairs:
        return 0.0
    return importance_factor * sum(newsim[w][x] for w, x in pairs) / len(pairs)
```

ryan-duve: @dschult Cool to learn about `G.pred` and `G.adj`.  I change the conditional per your recommendation and pushed the change.

At first I also combined the functions, but as a new user of NetworkX my opinion is this made the SimRank function harder to read and understand.  I left them as separate for now so you can review.  If it is more in line with NetworkX style to combine the functions, let me know and I will make that change, too.

Carghaez:
Looks good for me! +1

PR #3962, opened 2020-05-17
Fix regression on Fix issue 3953

Carghaez:
User @ryan-duve accidentally changed the algorithm behavior for non-direct graphs. The goal of the fix is to implement the correct algorithm for direct graphs. 

For indirect graphs it is already correct.


ryan-duve:
@Carghaez, I am new to SimRank, but my understanding is it is only defined for directed graphs.  If that is correct, would it be better to replace line 
```
return importance_factor * avg_sim(list(product(G[u], G[v])))
```
with
```
raise NetworkXAlgorithmError("SimRank requires a directed  graph.")
```
or some other explicit failure?

Carghaez:
> @Carghaez, I am new to SimRank, but my understanding is it is only defined for directed graphs. If that is correct, would it be better to replace line
> 
> ```
> return importance_factor * avg_sim(list(product(G[u], G[v])))
> ```
> 
> with
> 
> ```
> raise NetworkXAlgorithmError("SimRank requires a directed  graph.")
> ```
> 
> or some other explicit failure?

SimRank is a general similarity measure, based on a simple and intuitive graph-theoretic model. SimRank is applicable in any domain with object-to-object relationships, that measures similarity of the structural context in which objects occur, based on their relationships with other objects. Effectively, SimRank is a measure that says "two objects are considered to be similar if they are referenced by similar objects." 

So, it's not only for directed graph

ryan-duve:
@Carghaez Thank you for this fix.  I pulled it into https://github.com/networkx/networkx/pull/3961 and restored the tests per the request from the maintainers.

I'm interested in learning more about applying SimRank to undirected graphs.  I only see applications for directed graphs in the paper referenced in the NetworkX docs.  Do you have a paper you recommend where I can learn more about this?

PR #3963, opened 2020-05-18
Return empty generator instead of empty list

### LINKED ISSUE #3911: Incorrectly returning [] when a generator is expected
This return value (`[]`) is not consistent with the intention of the `all_simple_paths` function:

```py
if source in targets:
    return []
```

From: https://github.com/networkx/networkx/blob/b0ca6ce081cc6e66d509c5f7fed60728da821c25/networkx/algorithms/simple_paths.py#L230

Context:

```py
def all_simple_paths(G, source, target, cutoff=None):
    """Generate all simple paths in the graph G from source to target.
    A simple path is a path with no repeated nodes.

    [...]

    Returns
    -------
    path_generator: generator
       A generator that produces lists of simple paths.  If there are no paths
       between the source and target within the given cutoff the generator
       produces no output.
```

I recommend, instead, that NetworkX provide a reusable empty generator function, such as:

```py
def empty_generator():
    yield from ()
```

Then, in `all_simple_paths` (and almost certainly, other places in the library), call `return empty_generator()`

muratgu:
Fixes #3911

### DIFF
--- a/doc/reference/generators.rst
+++ b/doc/reference/generators.rst
@@ -52,7 +52,7 @@ Expanders
 
    margulis_gabber_galil_graph
    chordal_cycle_graph
-
+   paley_graph
 
 Lattice
 -------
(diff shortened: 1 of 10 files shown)


dschult:
Can you separate this branch from the branch that fixes the documentation? 
Probably a ```git rebase -i HEAD~6``` and then remove the commits that fixed the documentation. That'll make this PR restricted to only code relevant to this issue.

Thanks!

muratgu:
It appears I made it worse now. I will close and recreate it.

PR #3967, opened 2020-05-20
return empty generator instead of empty list

### LINKED ISSUE #3911: Incorrectly returning [] when a generator is expected

muratgu:
Fixes #3911

### DIFF
--- a/networkx/algorithms/simple_paths.py
+++ b/networkx/algorithms/simple_paths.py
@@ -6,6 +6,7 @@ import networkx as nx
 from networkx.utils import not_implemented_for
 from networkx.utils import pairwise
 from networkx.utils import empty_generator
+from networkx.algorithms.shortest_paths.weighted import _weight_function
 
 __all__ = [
     'all_simple_paths',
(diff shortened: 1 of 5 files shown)


dschult:
This may seem a little bit picky, but the original question was being a little picky too... :}

The old code passes your new tests.  That is ```list(path) == []``` is True whether path is an empty generator or an empty list. If we're going to bother making a generator object that just returns an empty iterator, then shouldn't we either test for that or forget the whole thing?  I looked into how to test for that, and it doesn't appear to be easy. So, I'm thinking we shouldn't spend time on this and spend time on algorithms instead.  

I'm going to merge this to avoid spending more time on it. :}

Issue #3956, opened 2020-05-14
Add useful `__repr__` methods


cool-RR:
For example, instead of `<networkx.classes.digraph.DiGraph object at 0x0000000003D6DE80>` we could have `<networkx.DiGraph: 7 nodes, 5 edges>`

ryan-duve:
I think it might be very verbose to unambiguously define the graph with `__repr__`.  Would it serve better to give classes a human-readable `__str__` instead?

dschult:
```__repr__``` is supposed to create a string that could be used as code to generate the object. That is very hard to do well for a graph object with arbitrary attributes. It's better to use the 
```nx.info(G)``` function is you are interested...

cool-RR:
I submitted #3968 , if you have any comments please post them there.

dschult:
We won't be implementing this.

PR #3968, opened 2020-05-21
Implement Graph.__repr__ for #3956

cool-RR:


cool-RR:
@dschult Nudge

dschult:
I'll just repeat my comment from #3956 
We won't be implementing this.

```__repr__``` is supposed to create a string that could be used as code to generate the object. That is very hard to do well for a graph object with arbitrary attributes. It's better to use the
nx.info(G) function is you are interested...

**Comment by cool-RR** [comment]
> `__repr__` is supposed to create a string that could be used as code to generate the object. 

There are so many counterexamples to what you just said, including from core Python...

```
In [1]: import math               
                                  
In [2]: math                      
Out[2]: <module 'math' (built-in)>
```

**Comment by MridulS** [comment]
> ```
> In [1]: import math               
>                                   
> In [2]: math                      
> Out[2]: <module 'math' (built-in)>
> ```

This is exactly what it will be from networkx too? `<module networkx from SomePath>`

Talking about the graph object, `__repr__` should be unambiguous and possibly passing the `eval(object.__repr()) == object` test and there isn't an easy mapping from arbitrary attributes in  a Graph object to a string.

Numpy example ->
```
eval(array([1, 2, 3]).__repr__()) == array([1, 2, 3])
```

The string information added here is accessed from `nx.info(graph_object)`. Maybe it could be argued to have `__str__` method for Graph but not really `__repr__`.

PR #3972, opened 2020-05-25
Enhancement for voterank

manasjoshi14:
1. Implemented the VoteRank algorithm for directed graphs as described in the reference paper.
2. Removed the max_iter parameter. In each iteration, a new node is selected as an influential spreader. The max_iter parameter should be the same as the number of nodes to extract.
3. The code returns only nodes with positive votes as per the previous author. Updated the document for the same.
4. Added unit test for voterank algorithm on directed graphs.

### DIFF
--- a/networkx/algorithms/centrality/tests/test_voterank.py
+++ b/networkx/algorithms/centrality/tests/test_voterank.py
@@ -7,21 +7,41 @@ import networkx as nx
 
 
 class TestVoteRankCentrality:
+    # Example Graph present in reference paper
     def test_voterank_centrality_1(self):
         G = nx.Graph()
         G.add_edges_from([(7, 8), (7, 5), (7, 9), (5, 0), (0, 1), (0, 2),
                           (0, 3), (0, 4), (1, 6), (2, 6), (3, 6), (4, 6)])
         assert [0, 7, 6] == nx.voterank(G)
 
+    # Graph unit test
     def test_voterank_centrality_2(self):
         G = nx.florentine_families_graph()
         d = nx.voterank(G, 4)
         exact = ['Medici', 'Strozzi', 'Guadagni', 'Castellani']
         assert exact == d
 
-    # Directed graph unit test
+    # DiGraph unit test
     def test_voterank_centrality_3(self):
         G = nx.gnc_graph(10, seed=7)
         d = nx.voterank(G, 4)
         exact = [3, 6, 8]
         assert exact == d
+
+    # MultiGraph unit test
+    def test_voterank_centrality_4(self):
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for this!

I have some questions:
- Why doesn't this work for multigraphs? It seems like it would.
- This algorithm as originally written changes the graph G by adding a node attribute 'voterank' which looks like the voterank when the algorithm completes, but I'm pretty sure that isn't a useful value. The point is to select the nodes, not figure out the voterank of those nodes not selected after all selected nodes are selected.
If I am correct that this isn't useful information, I would prefer that it not overwrite any potential node attribute already stored in G. Thus, could we change the algorithm to use an internal dict keyed by node to the voterank. That would probably faster as well. 
- Could we remove the word "seed" in the docstring. And phrase the process as "Select `number_of_nodes` nodes in `G` using VoteRank"  using "select" instead of "compute".
- Could you remove extraneous parens, e.g. ```if G.is_directed():```?
- I think you don't need to set ```AvgDegree = 0``` before setting it, and you can remove ```float``` when getting the average (we only support Python3.x now).

I realize many of these relate to updating the original code rather than to your changes to the code, but while we're changing things, let make is as good as possible. Thanks!

manasjoshi14:
Thanks for the review. I will implement these changes soon. 

The paper does not talk about the working of the algorithm on multigraphs. I believe we can modify it to work on multigraphs by merging multiple edges between two nodes to a single edge and then running the algorithm. 
For eg. let there exist N edges from node "u" to node "v". Information transfer from "u" to "v" requires only one edge, so "u" should receive a vote from "v" only once rather than N times(one for each edge).
Should I go ahead with this approach?

dschult:
I think that approach to multigraphs is something the user could implement easily.
I was thinking that you could process each edge as a vote. Then multiple edges give multiple votes for the same node.   Both the votes and decrease in voting power can work with a multiedge acting this way.  You are correct that the paper doesn't include either method though. I wonder if anyone has done that and studied its effectiveness.

manasjoshi14:
I have implemented the approach suggested by you along with a note in the docstring and unit tests. 
The paper also doesn't mention anything about self-loops. I believe we should remove self-loops since voting for itself or information transfer to itself doesn't make much sense.

dschult:
It looks like there is a rogue print statement in voterank. :}
This looks great! Thanks for taking on the documentation and tests and expansion to multigraphs. I think this is a good setp forward to the voterank part of NetworkX. :}

manasjoshi14:
Removed the print statement. Anything else you would like me to update here?

PR #3974, opened 2020-05-26
add d-separation algorithm

arunwise:
This PR implements an algorithm for testing _d-separation_ in DAGs. _d_separation_ is a central concept in probabilistic graphical models and it is used for testing conditional independence between random variables. 

Given that many libraries for probabilistic graphical models (like Pomegranate, Pyro etc) use NetworkX, it is somewhat surprising that this hasn't made its way into NetworkX so far. While the application is for probabilistic graphical models, the concept is purely graph theoretical. So, I hope that it gets added to the NetworkX library.

### DIFF
--- a/networkx/algorithms/d_separation.py
+++ b/networkx/algorithms/d_separation.py
@@ -4,12 +4,45 @@ Algorithm for testing d-separation in DAGs.
 *d-separation* is a test for conditional independence in probability
 distributions that can be factorized using DAGs.  It is a purely
 graphical test that uses the underlying graph and makes no reference
-to the actual distribution parameters.  For a detailed treatment see
-[1]_, [2]_.
+to the actual distribution parameters.  See [1]_ for a formal
+definition.
+
+The implementation is based on the conceptually simple linear time
+algorithm presented in [2]_.  Refer to [3]_, [4]_ for a couple of
+alternative algorithms.
+
+
+Examples
+--------
+
+>>> import networkx as nx
+>>> 
+>>> # HMM graph with five states and observation nodes
+... g= nx.DiGraph()
+>>> g.add_edges_from([('S1', 'S2'), ('S2', 'S3'), ('S3', 'S4'), ('S4', 'S5'),
+...                   ('S1', 'O1'), ('S2', 'O2'), ('S3', 'O3'), ('S4', 'O4'),
+...                   ('S5', 'O5')])
+>>> 
+>>> # states/obs before 'S3' are d-separated from states/obs after 'S3'
+... nx.d_separated(g, {'S1', 'S2', 'O1', 'O2'}, {'S4', 'S5', 'O4', 'O5'}, {'S3'})
+True
+
+
+References
+----------
+
+..  [1] Pearl, J.  (2009).  Causality.  Cambridge: Cambridge University Press.
+
+..  [2] Darwiche, A.  (2009).  Modeling and reasoning with Bayesian networks.  Cambridge: Cambridge University Press.
+
+..  [3] Shachter, R.  D.  (1998).  Bayes-ball: rational pastime (for determining irrelevance and requisite information in belief networks and influence diagrams).  In , Proceedings of the Fourteenth Conference on Uncertainty in Artificial Intelligence (pp.  480–487).  San Francisco, CA, USA: Morgan Kaufmann Publishers Inc.
+
+..  [4] Koller, D., & Friedman, N. (2009). Probabilistic graphical models: principles and techniques. The MIT Press.
+
 """
 
 from collections import deque
-from typing import Hashable, AbstractSet
+from typing import AbstractSet
 
 import networkx as nx
 from networkx.utils import not_implemented_for, UnionFind
@@ -18,10 +51,10 @@ __all__ = ["d_separated"]
 
 
 @not_implemented_for("undirected")
-def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
-                y: AbstractSet[Hashable], z: AbstractSet[Hashable]) -> bool:
+def d_separated(G: nx.DiGraph, x: AbstractSet, y: AbstractSet,
+                z: AbstractSet) -> bool:
     """
-    Return whether node sets ``x`` and ``y` are separated by ``z``.
+    Return whether node sets ``x`` and ``y`` are d-separated by ``z``.
 
     Parameters
     ----------
@@ -39,7 +72,8 @@ def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
 
     Returns
     -------
-    bool : True if ``x`` is d-separated from ``y`` given ``z`` in ``G``.
+    b : bool
+        A boolean that is true if ``x`` is d-separated from ``y`` given ``z`` in ``G``.
 
     Raises
     ------
@@ -51,32 +85,6 @@ def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
     NodeNotFound
         If any of the input nodes are not found in the graph,
         a :exc:`NodeNotFound` exception is raised.
-
-    Examples
-    --------
-    >>> import networkx as nx
-    >>> g = nx.path_graph(10, nx.DiGraph)
-    >>> assert nx.d_separated(g, {0,1,2,3}, {5,6,7,8,9}, {4})
-
-    Notes
-    -----
-    While there are other algorithms for testing *d-separation* (for
-    e.g. [3]_), the algorithm presented in [1]_ is perhaps the simplest
-    linear time algorithm and is therefore used in this implementation.
-
-
-    References
-    ----------
-    ..  [1] Darwiche, A.  (2009).  Modeling and reasoning with
-        Bayesian networks.  Cambridge: Cambridge University Press.
-
-    ..  [2] Pearl, J. (2009). Causality. Cambridge: Cambridge University Press.
-
-    ..  [2] Shachter, R. D. (1998). Bayes-ball: rational pastime (for
-        determining irrelevance and requisite information in belief networks
-        and influence diagrams). In , Proceedings of the Fourteenth Conference
-        on Uncertainty in Artificial Intelligence (pp. 480–487). San
-        Francisco, CA, USA: Morgan Kaufmann Publishers Inc.
 
     """
 
--- a/networkx/algorithms/d_separation.py
+++ b/networkx/algorithms/d_separation.py
@@ -35,7 +35,7 @@ def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
         Second set of nodes in ``G``.
 
     z : set
-        Set of conditioning nodes in `G`.
+        Set of conditioning nodes in ``G``. Can be empty set.
 
     Returns
     -------
--- a/networkx/algorithms/d_separation.py
+++ b/networkx/algorithms/d_separation.py
@@ -1,0 +1,122 @@
+"""
+Algorithm for testing d-separation in DAGs.
+
+*d-separation* is a test for conditional independence in probability
+distributions that can be factorized using DAGs.  It is a purely
+graphical test that uses the underlying graph and makes no reference
+to the actual distribution parameters.  For a detailed treatment see
+[1]_, [2]_.
+"""
+
+from collections import deque
+from typing import Hashable, AbstractSet
+
+import networkx as nx
+from networkx.utils import not_implemented_for, UnionFind
+
+__all__ = ["d_separated"]
+
+
+@not_implemented_for("undirected")
+def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
+                y: AbstractSet[Hashable], z: AbstractSet[Hashable]) -> bool:
+    """
+    Test whether two sets of nodes are d-separated in a DAG by a third
+    set of conditioning nodes.
+
+    Parameters
+    ----------
+    G : graph
+        A NetworkX DAG.
+
+    x : set
+        First set of nodes in `G`.
+
+    y : set
+        Second set of nodes in `G`.
+
+    z : set
+        Set of conditioning nodes in `G`.
+
+    Returns
+    -------
+    bool : True if `x` is d-separated from `y` given `z` in `G`.
+
+    Raises
+    ------
+    NetworkXError
+        The *d-separation* test is commonly used with directed
+        graphical models which are acyclic.  Accordingly, the algorithm
+        raises a :exc:`NetworkXError` if the input graph is not a DAG.
+
+    NodeNotFound
+        If any of the input nodes are not found in the graph,
+        a :exc:`NodeNotFound` exception is raised.
+
+    Examples
+    --------
+    >>> import networkx as nx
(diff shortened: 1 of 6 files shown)


ericmjl:
@arunwise thank you for your PR! I really enjoyed reading through the code and seeing what you've put in. I have not yet gone through a detailed code review yet, but I will be blocking out an hour this Saturday to do so. Mostly I'll just be checking through the tests to make sure they are correct, which I trust they are, just need to have a second pair of eyes.

@dschult please let me know if there's something else I should look out for apart from correctness of the algorithm.

dschult:
Thanks @ericmjl for taking a look at this. 

The other things to look for are 
- proper documentation (add the module to the rst file, use one line short summary at top of docstring followed by blank line, list parameters and return values)
- tests that use random numbers should specify a seed and not repeatedly test many random inputs in these tests (though that code could remain as commented code).
- The API should be simple and where relevant match current API.
- The algorithm should not be already implemented somewhere else in NetworkX and if there is overlap, it should be determined whether to refactor common parts or leave them separate (and duplicate). Duplicate code is not always bad, but often it is better to avoid it.

Can't think of anything else to look for other than correct implementation at this time.
Thanks!

arunwise:
Thanks @ericmjl @dschult 
I am looking forward to the feedback/comments from you.

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,122 @@
...
+
+__all__ = ["d_separated"]
+
+
+@not_implemented_for("undirected")
+def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
+                y: AbstractSet[Hashable], z: AbstractSet[Hashable]) -> bool:
+    """
+    Test whether two sets of nodes are d-separated in a DAG by a third
+    set of conditioning nodes.
```
ericmjl, line 25: To fulfill the 1-line summary requirement, here is a strawman suggestion.

```suggestion
    Return whether node sets ``x`` and ``y` are separated by ``z``.
```

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Parameters
+    ----------
+    G : graph
+        A NetworkX DAG.
+
+    x : set
+        First set of nodes in `G`.
```
ericmjl, line 33: Sphinx-style syntax will render single backticks as italicized.

```suggestion
        First set of nodes in ``G``.
```

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,122 @@
...
+
+    y : set
+        Second set of nodes in `G`.
```
ericmjl, line 36: (copy/pasted) please change backticks.

```suggestion
        Second set of nodes in ``G``.
```

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,122 @@
...
+
+    z : set
+        Set of conditioning nodes in `G`.
```
ericmjl, line 39: (copy/pasted) please change backticks.

```suggestion
        Set of conditioning nodes in ``G``. Can be empty set.
```

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Returns
+    -------
+    bool : True if `x` is d-separated from `y` given `z` in `G`.
```
ericmjl, line 43:
```suggestion
    bool : True if ``x`` is d-separated from ``y`` given ``z`` in ``G``.
```

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,122 @@
+"""
```
ericmjl, line 1: @arunwise please update the docs with a reference to this module.

networkx/algorithms/tests/test_d_separation.py
```diff
@@ -0,0 +1,79 @@
...
+        cls.fork_G.add_edges_from([(0, 1), (0, 2)])
+        nx.freeze(cls.fork_G)
+
+        cls.collider_G = nx.DiGraph(name="collider")
+        cls.collider_G.add_edges_from([(0, 2), (1, 2)])
+        nx.freeze(cls.collider_G)
+
+        cls.naive_bayes_G = nx.DiGraph(name="naive_bayes")
+        cls.naive_bayes_G.add_edges_from([(0, 1), (0, 2), (0, 3), (0, 4)])
+        nx.freeze(cls.naive_bayes_G)
```
ericmjl, line 23: This one was the hard one for me; the others I could catch. @arunwise, can you enlighten me here? 

`0` refers to the predictor (non-plural) vector `x`, and `1, 2, 3, 4` refer to the classes `y` - is this correct? I had always thought the NB model was the other way around - a whole bunch of predictors `1, 2, ... n` pointing to the class label output `0`.

networkx/algorithms/tests/test_d_separation.py
```diff
@@ -0,0 +1,79 @@
+from itertools import combinations
+import pytest
```
ericmjl, line 2: @dschult @arunwise I noticed we're using `pytest` nowadays. Any chance that we could start having new tests use the `def test_...` style instead of the old unittest class-style? Or are you planning on keeping the old unittest style, @dschult? I just think it the isolated functions would make isolating the tests more easy, and prevent accidental future cross-talk between test examples because of human error. (We also wouldn't have to deal with boilerplate repeat!) In this paradigm, we would isolate each of the graphs (path, fork, collider, NB and asia) into their own test examples, rather than have them all encapsulated in a single class.

ericmjl:
Some notes for myself.

From @dschult's checklist:

- proper documentation:
    - [ ] add the module to the rst file
    - [x] use one line short summary at top of docstring followed by blank line
    - [x] list parameters and return values
- [x] tests that use random numbers should specify a seed and not repeatedly test many random inputs in these tests (though that code could remain as commented code).
- [x] The API should be simple and where relevant match current API.
- [x] The algorithm should not be already implemented somewhere else in NetworkX and if there is overlap, it should be determined whether to refactor common parts or leave them separate (and duplicate). Duplicate code is not always bad, but often it is better to avoid it.

networkx/algorithms/tests/test_d_separation.py
dschult: We are definitely using pytest style now. But pytest allows both Class and Function interfaces. The Function interface is often easier to read, but the class interface allows a single setup method to create the graphs and then each of the test methods use those graphs. In that sense this implementation is probably good style because the setup method creates multiple graphs and the first two test methods use those multiple graphs.

So, the class style for tests is fine. And there is a choice whether to split out all tests for one graph into its own function vs make a class for all graphs with fewer test methods. What do you think @ericmjl ? Would it be more clear with a function for each graph, testing all aspects for that graph? Or is it better to have a method for each type of test and the graphs all get created in the setup method?  Either is perfectly good pytest syntax. pytest allows both functions or classes.

ericmjl: @dschult I'll eventually defer to your preference (being the lead maintainer) :smile:, but in case my opinion counts for anything :laugh:, I definitely lean towards fixtures + functions rather than classes.

I know it's not good to prematurely optimize, but I think in this case, having each of the four graphs available as a fixture to be used in other places may be helpful as more graph-based causality algorithms get implemented.

Also, I think it's possibly beneficial to keep test examples as isolated as possible from one another, because of the reasons I argued for above. 

By separating out the tests, we make reading the output of any errors and isolating the exact test a bit more easy. Also, doing so makes possible adding docstrings to each of the tests (highly recommended practice!) which explain _why_ a test exists and what to expect from it, in plain English.

Of course, this isn't one of those hills I'm willing to die on, so I don't mind if this suggestion isn't accepted - no hard feelings! That said, I did also take the liberty of reworking some of the tests in a [gist](https://gist.github.com/ericmjl/43d14da2d2cde5e632e78cdcd816d5c0), to show how the test suite could be organized, with all of the goodies I mentioned above: fixtures, docstrings, etc. It should also lower the barrier for @arunwise to change the test suite style, if he also chooses to do so.

networkx/algorithms/tests/test_d_separation.py
arunwise: In the DAG of a NB model the root is the target variable and the features are its children. So in this example `0` is the node corresponding to target variable and `1,2,3,4` are the nodes corresponding to the features. The details like number of classes is not part of the graph. I hope that answers your question.

ericmjl: Got it, thanks!

networkx/algorithms/d_separation.py
arunwise: @ericmjl Will do. Since I am new to this repo, I was under the impression that documentation gets automatically generated from the docstrings in the source file. 

I will add the necessary `rst` file docs and make other changes that are necessary.

networkx/algorithms/d_separation.py
arunwise: This is actually clear and concise.

networkx/algorithms/d_separation.py
arunwise: The implementation allows all three sets to be empty. If `x` or `y` is empty the result is trivially True, since there are no conditional independence tests to be made. 

So, I feel that the elaboration (that the set can be empty) should be made for all or none. Your thoughts ?

networkx/algorithms/d_separation.py
ericmjl: > I was under the impression that documentation gets automatically generated from the docstrings in the source file.

You are correct! It's just that the reference needs to be made, that's all 😄.

networkx/algorithms/d_separation.py
ericmjl: That is true. The trivial result is that "nothing is d-separated from nothing by nothing", which is quite hilarious now that I think about it.

That said, thinking from the perspective of how causal inference tasks usually go, I _think_ it is true that most of the time `x` and `y` will not be empty sets, while `z` might or might not be. The intent here is to nudge in the correct pedagogical direction, especially in a situation where we don't want to be _too_ strict with the restrictions. That is why I placed the `can be empty set` elaboration only at the docstring part for `z`.

networkx/algorithms/d_separation.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import Hashable, AbstractSet
+
+import networkx as nx
+from networkx.utils import not_implemented_for, UnionFind
+
+__all__ = ["d_separated"]
+
+
+@not_implemented_for("undirected")
+def d_separated(G: nx.DiGraph, x: AbstractSet[Hashable],
```
arunwise, line 21: @ericmjl While these type hints may be useful, the function signature in the documentation ends up being cluttered. I am leaning towards removing the type hints since the types are mostly documented in the docstring below. I would like to hear your thoughts on that.

ericmjl: @arunwise they're actually a very good practice, and I was super happy you put them in there in the first place! 😄 

This is because there are static code checkers like mypy that leverage type hints to check that the code written is correct. For example, with mypy on, I can (in source `.py` files) know whether I've passed in an object that is not of the categories accepted by a function. I'd be inclined to leave them in here, and if you really would like to remove some clutter, taking them out from the docstring could be a way out, though I'd leave those in there as well, because they are the "human-readable" form of the docs.

networkx/algorithms/d_separation.py
arunwise: @ericmjl Updated the docs.

networkx/algorithms/tests/test_d_separation.py
arunwise: @ericmjl Thanks for the [gist](https://gist.github.com/ericmjl/43d14da2d2cde5e632e78cdcd816d5c0)

@ericmjl @dschult I personally use the pytest fixture approach when writing tests. However, I noticed that almost all other modules used unittest approach. So, to keep things uniform I wrote the tests in unittest style. I don't have a strong opinion in this particular case, but generally I lean towards fixtures for reasons mentioned by @ericmjl

dschult: This all sounds good. pytest fixtures are good.  And FWIW, any setup function put in a class can be created as a function and then called from the test_function as well. :}

ericmjl: Looks like we have convergence on adopting the contemporary code style for tests. @arunwise would you be kind enough to finish up the PR with the reworked test suite from the gist? (Just a heads-up, it may be imperfect, as I didn’t do a full checkout of the repository.)

arunwise: @ericmjl I reworked the unit-tests to use fixtures. Most of it is based on your gist. One minor issue was that pytest has deprecated calling fixtures directly. Since, the fixtures are used as plain functions in some of the tests, I had to resort to the approach described [here](https://docs.pytest.org/en/latest/deprecations.html#id24). 

Please let me know your thoughts on the reworked test suite.

arunwise:
> @arunwise wonderful work! I think your contribution will be well-used going forward.
> 
> I'm going to request one final change: Can you add yourself to the contributors list? The place to do this is `CONTRIBUTORS.rst`. I'd love to make sure that you're properly credited on there!
> 
> Other than that, everything looks great to me! @dschult, may I have the pleasure of hitting the big green button later? 😄

Done.

I enjoyed contributing this PR and I would like to thank you and @dschult for the constructive feedback.

ericmjl:
Wonderful, @arunwise! Thank you too for your PR :smile:.

Some final reminders - once you're done, don't forget to delete your fork and re-fork the repo for any new PR. I noticed that you had PR-ed in from `arunwise/master`, which isn't recommended practice, but in any case, I didn't want to stop the PR from moving forward, so the solution around potentially future messy git logs is to delete and re-fork. 

@dschult, if you're ok with the PR, please let me know, I'd love to hit the big green button! Also proposing merging by squashing, so that the commit history is clean (I do this with pyjanitor), unless you have other preferences.

ericmjl:
@arunwise sorry to bother you about this, but I hadn't heard back from @dschult and was about to just hit the merge PR button when I saw the conflict.

It appears I can't resolve the conflicts, as I might not have the necessary permissions.

Could you pull in the latest networkx master into your branch, resolve the conflict in `CONTRIBUTORS.rst` (keep both names), and push up the changes? I will go ahead and hit the merge button as soon as everything is green.

arunwise:
> @arunwise sorry to bother you about this, but I hadn't heard back from @dschult and was about to just hit the merge PR button when I saw the conflict.
> 
> It appears I can't resolve the conflicts, as I might not have the necessary permissions.
> 
> Could you pull in the latest networkx master into your branch, resolve the conflict in `CONTRIBUTORS.rst` (keep both names), and push up the changes? I will go ahead and hit the merge button as soon as everything is green.

@ericmjl No problem. I fixed the merge conflict.

ericmjl:
@arunwise and @dschult, I took the liberty of merging! Thank you @arunwise for all the hard work!

dschult:
Yes!  Thanks @arunwise, and also Thanks @ericmjl for your work with this too. :}

PR #3978, opened 2020-05-28
Add distinctiveness centrality

iandreafc:
I recreated this pull request after I did some mess with the branch. Now hopefully everything is right.

We recently developed a Python package to calculate Distinctiveness Centrality.
Here is our code: https://github.com/iandreafc/distinctiveness
also available on Pypi and already built over Networkx: https://pypi.org/project/distinctiveness/
Here is the paper that describes the metric: https://doi.org/10.1371/journal.pone.0233276

I am not an expert programmer and these are the first times using GitHub, so please be patient ❤️
I created a pull request, hope I have done things correctly.
If I did something wrong, some help would be very much appreaciated 😃

This is a pull request for issue #3969

### DIFF
--- a/doc/reference/algorithms/centrality.rst
+++ b/doc/reference/algorithms/centrality.rst
@@ -109,6 +109,13 @@ Dispersion
    :toctree: generated/
 
    dispersion
+   
+Distinctiveness
+----------
+.. autosummary::
+   :toctree: generated/
+
+   distinctiveness
 
 Reaching
 --------
--- a/networkx/algorithms/centrality/__init__.py
+++ b/networkx/algorithms/centrality/__init__.py
@@ -7,6 +7,7 @@ from .current_flow_betweenness import *
 from .current_flow_betweenness_subset import *
 from .degree_alg import *
 from .dispersion import *
+from .distinctiveness import distinctiveness
 from .eigenvector import *
 from .group import *
 from .harmonic import *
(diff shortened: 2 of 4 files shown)


dschult:
Could you add (at least) another paragraph in the docstring to state what "distinctiveness" refers to or means without using the english term distinctive? I guess what I'm looking for is a bigger picture description of what this measure intends to provide, why someone might choose to use it. It'd be great to also describe how it could/should be used.  Yet another framing of the discussion could be: why is this in the centrality part of the code and how does distinctiveness have anything to do with centrality?


It is unusual to have so many values computed for the nodes of a graph in one function, can you give me some feel for what the advantage of this API is over, e.g. a function for each of the measures? Would someone want one of the measures and have to compute them all? Would they only ever want all the measures? When would they want what?

Thanks for this!

iandreafc:
Many thanks @dschult for this new suggestions!

I have extended the documentation, with the hope everything is clearer now. This is a centrality metric as it all starts from degree and weighted degree, but these measures are rescaled, similarly to what you would do in text mining while using TF-IDF. For extensive use cases there is the paper, but I tried to follow you suggestion and be much more descriptive in the docstring. 

Once you calculate one metric, the additional effort required to calculate the others is almost negligible, in terms of computational time. This is why I usually calculate them all. They are all different flavors of the same concept (it is just the weighting scheme that changes). This is very useful for forecasting excercises, for example, when you want to test which metric works better. However, I also understand your suggestion and that this is less typical. Therefore, I have revised the function. You can now calculate all the metrics, or just select some of them.

What do you think? Thanks!

PR #3982, opened 2020-05-31
Update CONTRIBUTORS.rst

Carghaez:
Added Gaetano Pietro Paolo Carpinato as contributor from PR #3961 and #3954

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -171,6 +171,7 @@ is partially historical, and now, mostly arbitrary.
 - Alessandro Luongo
 - Huston Hedinger
 - Oleguer Sagarra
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.rst
```diff
@@ -171,6 +171,7 @@ is partially historical, and now, mostly arbitrary.
 - Alessandro Luongo
 - Huston Hedinger
 - Oleguer Sagarra
+- Gaetano Pietro Paolo Carpinato, Github `<https://github.com/Carghaez>`_, LinkedIn `<https://linkedin.com/in/gaetanocarpinato/>`_
```
ericmjl, line 174: @Carghaez thanks for your contributions! I think the others don't put in links though, so can we reduce the listing to just your name please?

cc: @dschult, who leads maintenance and might have a different opinion from me.

dschult: I don't think the list is too long. Let's go ahead and include links.

ericmjl: Thanks for the guidance, @dschult!

PR #3985, opened 2020-06-04
fix bug for to_scipy_sparse_matrix function

haochenucr:
The nx.to_scipy_sparse_matrix(G, nodelist=None, dtype=None, weight='weight', format='csr') function will raise a ValueError if the the input graph G contains selfloops, but the subgraph induced on the input nodelist does not contains any selfloop. For example:

>>> g = nx.Graph()
>>> g.add_edges_from([[0,0],[1,2],[1,3]])
>>> nx.to_scipy_sparse_matrix(g, [1,2])
  File "XXX/networkx/convert_matrix.py", line 845, in to_scipy_sparse_matrix
    for u, v, d in selfloops
ValueError: not enough values to unpack (expected 2, got 0)

So when we subtract the data on the diagonal to avoid selfloop entries get double counted when symmetrizing, we only judge if the subgraph induced on the input nodelist contains selfloops, otherwise, we do not have to do the subtraction.

### DIFF
--- a/networkx/convert_matrix.py
+++ b/networkx/convert_matrix.py
@@ -836,7 +836,7 @@ def to_scipy_sparse_matrix(G, nodelist=None, dtype=None,
         c = col + row
         # selfloop entries get double counted when symmetrizing
         # so we subtract the data on the diagonal
-        selfloops = list(nx.selfloop_edges(G, data=True))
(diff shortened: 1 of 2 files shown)


dschult:
Please add a test that shows the previous break (and passes with the new code).
Thanks for this!

Issue #1012, opened 2013-11-17
Is there anyway to get some bayesian graph manipulation tools like moralizing and triangulating?


mb3152:
I have been wrapping old matlab code, but networkx stuff would be great!

hagberg:
Closed due to inactivity.

chebee7i:
Good move. Also, at this point there is a mature package dedicated to this: https://github.com/pgmpy/pgmpy

If interest in this picks up again, I suspect conversions between the two libraries would be a good option instead of an all-out implementation.

PR #4004, opened 2020-06-16
junction_tree for #1012

mbruhns:
I have implemented the junction tree algorithm to get the junction tree of a directed or undirected graph. This feature was somewhat already discussed in #1012.


mbruhns:
Thanks for your feedback! I tried to implement the requested changes. Since this is my first contribution to a project of this size I hope I got it right.

examples/algorithms/plot_junction_tree_algorithm.py
```diff
@@ -0,0 +1,19 @@
+import networkx as nx
```
jarrodmillman, line 1: The examples need a docstring at the top of the file.  This is required by https://sphinx-gallery.github.io/stable/index.html, which we use to make the gallery from the examples.

See, for example, `https://github.com/networkx/networkx/blob/master/examples/algorithms/plot_beam_search.py`, which starts:
```
"""
===========
Beam Search
===========

Beam search with dynamic beam width.

The progressive widening beam search repeatedly executes a beam search
with increasing beam width until the target node is found.
"""
```

networkx/algorithms/tree/junction_tree_algorithm.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+#   Copyright (C) 2020 by
+#   Matthias Bruhns <matthias.bruhns@uni-jena.de>
+#   All rights reserved.
+#   BSD license.
+#   Copyright 2016-2020 NetworkX developers.
+#   NetworkX is distributed under a BSD license
+#
+# Authors: Matthias Bruhns <matthias.bruhns@uni-jena.de>
```
jarrodmillman, line 9: Please remove this.

examples/algorithms/plot_junction_tree_algorithm.py
jarrodmillman: This should fix the Travis CI error.

networkx/algorithms/tree/tests/test_junction_tree_algorithm.py
```diff
@@ -0,0 +1,79 @@
...
+def test_junction_tree_directed_confounders():
+    B = nx.DiGraph()
+    B.add_edges_from([("A", "C"), ("B", "C"), ("C", "D"), ("C", "E")])
+
+    G = junction_tree(B)
+    J = nx.Graph()
+    J.add_edges_from(
+        [
+            (("C", "E"), ("C",)),
+            (("C",), ("A", "B", "C")),
```
dschult, line 14: I'm confused by this test example. I don't know much about junction trees so I may be missing something.
One node in the junction tree is ```("A","B","C")```. But in this example, A-B-C is not a clique of the original graph.
Also, you have two edges in line 14 and 15 which seem to be opposite direction of the same nodes. J is undirected, so you should only add one of these two lines?   But still, it seems to me that these lines should be:
```(("C",),  ("A", "C")),
      (("C"),  ("B", "C"))
```

networkx/algorithms/tree/junction_tree_algorithm.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Returns
+    -------
+    junction_tree : NetworkX graph
+                    The corresponding junction tree of G.
+
+    Raises
+    ------
+    NetworkXNotImplemented
+        The algorithm does not support Graph, MultiGraph and MultiDiGraph.
```
dschult, line 45: I believe ```Graph``` should not be included in this list.

networkx/algorithms/tree/tests/test_junction_tree_algorithm.py
dschult: I see now that the junction tree is not using cliques of the original graph, but cliques of the moralized, triangulated graph.  The docstring should be corrected to accurately describe a junction tree.  :)

networkx/algorithms/tree/junction_tree_algorithm.py
mbruhns: Correct, I removed it.

networkx/algorithms/tree/tests/test_junction_tree_algorithm.py
mbruhns: You're right, I updated the docstring accordingly :)

rossbar:
I took the liberty of rebasing this on `master` to enable the doc-building CI service, which will make it easier to review the changes to the documentation. I will take another pass once the CI is done!

networkx/algorithms/tree/junction_tree_algorithm.py
```diff
@@ -0,0 +1,81 @@
...
+       conference on Uncertainty in artificial intelligence (UAI’94).
+       Morgan Kaufmann Publishers Inc., San Francisco, CA, USA, 360–366.
+    """
+
+    clique_graph = nx.Graph()
+    if G.is_directed():
+        moralized = moral.moral_graph(G)
+        chordal_graph, _ = complete_to_chordal_graph(moralized)
+    else:
+        chordal_graph, _ = complete_to_chordal_graph(G)
```
rossbar, line 61:
```suggestion
    if G.is_directed():
        G = moral.moral_graph(G)
    chordal_graph, _ = complete_to_chordal_graph(G)
```

networkx/algorithms/tree/junction_tree_algorithm.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@not_implemented_for("multigraph", "MultiDiGraph")
+def junction_tree(G):
+    r"""Returns a junction tree of a given graph.
+
+    Parameters
+    ----------
+    G : networkx.Graph
+        Directed graph
```
rossbar, line 18: The input can be an undirected graph as well, correct? The description here isn't consistent with the implementation.

rossbar:
I should also mention: 

Git will be confused on your local machine because I changed history
when I force pushed to your remote branch. So you will have to run
a few commands. If you aren't familiar with what to do, please take
a look at this:
https://stackoverflow.com/questions/9813816/git-pull-after-forced-update

If you aren't sure after looking at that, please don't hesitate to ask!

networkx/algorithms/tree/junction_tree_algorithm.py
mbruhns: I'm glad you mentioned the naming! I had a similar idea a few days ago but somehow forgot about it. I choose "decomposition", which makes sense :) I also implemented your suggested change.

networkx/algorithms/tree/junction_tree_algorithm.py
mbruhns: You're correct, I updated the docstring accordingly.

mbruhns:
> I should also mention:
> 
> Git will be confused on your local machine because I changed history
> when I force pushed to your remote branch. So you will have to run
> a few commands. If you aren't familiar with what to do, please take
> a look at this:
> https://stackoverflow.com/questions/9813816/git-pull-after-forced-update
> 
> If you aren't sure after looking at that, please don't hesitate to ask!

I used `git pull --rebase` which hopefully did the correct thing :smile:

networkx/algorithms/tree/__init__.py
```diff
@@ -3,3 +3,4 @@
 from .mst import *
 from .recognition import *
 from .operations import *
+from .junction_tree_algorithm import *
```
jarrodmillman, line 6: ``.decomposition``

jarrodmillman: That should fix your CI errors as well.

doc/reference/algorithms/tree.rst
```diff
@@ -61,6 +61,14 @@ Spanning Trees
    minimum_spanning_edges
    maximum_spanning_edges
 
+Junction Tree Algorithm
```
jarrodmillman, line 64: This needs to be fixed as well.

mbruhns: I fixed both entries, CircleCI still seems to fail. It won't show me any details on why the build failed, can you help me out here?

jarrodmillman: That was an issue with CircleCI (unrelated to this PR) that I fixed (hopefully) in #4147.  I merged master into your branch and pushed to your remote.  That should fix the CircleCI failure.

jarrodmillman:
I couldn't really see what was happening with the original example.
old
I also thought it would be helpful to see some of the intermediate steps, so I added the moralized graph.
new

The example still needs improvement.

1. It should have some text explaining what it happening.  For example, if the first subplot is supposed to be a bayes net maybe it should describe the factorization (e.g., $p_A \cdot p_{B|A} \dots $) and so on.  It doesn't need much, but just having a couple of lines of code shouldn't be sufficient for a gallery example.

2. The bayes net and the moralized graph nodes aren't in the same locations.  I thought having the same layout would ensure they line up.  What am I missing?

3. As noted in the docstring, the junction tree isn't unique.  But maybe we should (arbitrarily) make it so (possibly w/ a flag).  I am not sure there is an advantage to having it change everytime it is run.  It certainly makes making nice looking examples more difficult.  It also makes testing more difficult.  I suspect all the tests must use graphs with unique decompositions.

jarrodmillman:
@dschult , @rossbar  Should we use graphviz for the layout?  It would be nice to have the node labels automatically make the node size correct.  It is also annoying that it currently sometimes hides the edges.

jarrodmillman:
It would be nice to have plots that look more like the ones on page 3 and 12 of Mark's tutorial:
https://ai.stanford.edu/~paskin/gm-short-course/lec3.pdf

rossbar:
> The bayes net and the moralized graph nodes aren't in the same locations. I thought having the same layout would ensure they line up. What am I missing?

It looks like the axes limits are inconsistent between the first two axes objects, even though the positions are identical. That's an easy fix, but annoying that the user has to do it manually.

> As noted in the docstring, the junction tree isn't unique. But maybe we should (arbitrarily) make it so (possibly w/ a flag).

Would decorating with one of the seed-setting decorators accomplish this?

> Should we use graphviz for the layout?

It looks like the majority of the [examples](https://networkx.github.io/documentation/stable/auto_examples/index.html) only use `nx_pylab` tools, though there are some that use pygraphviz/pydot. My vote would be to stick with whatever has the least amount of boilerplate. It would definitely be a good idea to revisit the example gallery as the visualiztion tools are improved.

jarrodmillman:
@rossbar The reason the junction tree is not deterministic is that ``chordal_graph_cliques`` in ``algorithms/chordal.py`` returns the **set** of maximal cliques.  I think the easiest way to fix it may be to change
```
cliques = [tuple(sorted(i)) for i in chordal_graph_cliques(chordal_graph)]
```
to
```
cliques = [tuple(sorted(i)) for i in chordal_graph_cliques(chordal_graph)]
cliques.sort()
```
in ``junction_tree`` in ``algorithms/tree/decomposition.py ``.

mbruhns:
> @rossbar The reason the junction tree is not deterministic is that `chordal_graph_cliques` in `algorithms/chordal.py` returns the **set** of maximal cliques. I think the easiest way to fix it may be to change
> 
> ```
> cliques = [tuple(sorted(i)) for i in chordal_graph_cliques(chordal_graph)]
> ```
> 
> to
> 
> ```
> cliques = [tuple(sorted(i)) for i in chordal_graph_cliques(chordal_graph)]
> cliques.sort()
> ```
> 
> in `junction_tree` in `algorithms/tree/decomposition.py `.

Correct, that should work. I think I already had implemented that in a earlier version, but got rid of it for some reason. It should be in the code nonetheless.

dschult:
Be careful with ```sort```ing.  The nodes of the graph are not guaranteed to be sortable. For example, they might be NetworkX Graphs.  So, it's fine to use "sort" in an example but it shouldn't be in the public-facing function.
Perhaps better to see if we can get ```chordal_graph_cliques``` to use/return a different data structure.

It's too bad that we are starting to avoid sets. I might have to go back to using dicts as sets (which was fairly standard before the "set" data structure in python.  Python could fix the temptation by using the same under-the-hood ordered mechanism for sets that they do now for dicts. It is certainly performant. I was surprised that they didn't do that in Python 3.7.

mbruhns:
This might be a little off-topic, but can you give me some insights on the avoidance of sets?

dschult:
Well, my understanding is that we are trying to avoid the non-deterministic nature of the algorithm. It is nondeterministic because the order of iteration over a set can vary from one creation of the set to another. So, the order that the cliques show up in the output is not the same every time. That makes it slightly harder to test whether your function is working.

It's also good practice to make your computations reproducible (sending the code to someone else and having them run it will give the same answer) and that's not the case if the code uses nondeterministic data structures. Am I getting the right gist @jarrodmillman ?

We can still use sets.... but its a little more of a pain than other data structures especially when deterministic behavior is desired.

jarrodmillman:
@dschult Why don't we change ``chordal_graph_cliques`` from
```
    cliques = set()
    for C in (G.subgraph(c).copy() for c in connected_components(G)):
        cliques |= _connected_chordal_graph_cliques(C)

    return cliques
```
to
```
    for C in (G.subgraph(c).copy() for c in connected_components(G)):
        if C.number_of_nodes() == 1:
            yield frozenset(C.nodes())
        else:
            unnumbered = set(C.nodes())
            v = arbitrary_element(C)
            unnumbered.remove(v)
            numbered = {v}
            clique_wanna_be = {v}
            while unnumbered:
                v = _max_cardinality_node(C, unnumbered, numbered)
                unnumbered.remove(v)
                numbered.add(v)
                new_clique_wanna_be = set(C.neighbors(v)) & numbered
                sg = C.subgraph(clique_wanna_be)
                if _is_complete_graph(sg):
                    new_clique_wanna_be.add(v)
                    if not new_clique_wanna_be >= clique_wanna_be:
                        yield frozenset(clique_wanna_be)
                    clique_wanna_be = new_clique_wanna_be
                else:
                    raise nx.NetworkXError("Input graph is not chordal.")
            yield frozenset(clique_wanna_be)
```
and delete ``_connected_chordal_graph_cliques``.

This better matches ``find_cliques`` in ``algorithms/clique.py``.

For now, we could put a deprecation warning in saying that in 3.0 ``chordal_graph_cliques``  will return a generator instead of a set.

mbruhns:
> Well, my understanding is that we are trying to avoid the non-deterministic nature of the algorithm. It is nondeterministic because the order of iteration over a set can vary from one creation of the set to another. So, the order that the cliques show up in the output is not the same every time. That makes it slightly harder to test whether your function is working.
> 
> 
> 
> It's also good practice to make your computations reproducible (sending the code to someone else and having them run it will give the same answer) and that's not the case if the code uses nondeterministic data structures. Am I getting the right gist @jarrodmillman ?
> 
> 
> 
> We can still use sets.... but its a little more of a pain than other data structures especially when deterministic behavior is desired. 

Ahh, sorry, I misunderstood your comment. I thought your point was to avoid sets in general.

jarrodmillman:
I updated the [gallery example](https://117-890377-gh.circle-artifacts.com/0/doc/build/html/auto_examples/algorithms/plot_decomposition.html#sphx-glr-auto-examples-algorithms-plot-decomposition-py) to use graphviz for its layout and I made the first two subplots share axes limits.  We can improve it later.  First, I would like to decide what to do about  #4162.

@dschult I think we should squash and merge this PR.  Decide what to do about #4162.  And then figure out why ``draw_networkx`` is clipping the plot.

dschult:
Thanks @mbruhns !

PR #4007, opened 2020-06-18
Added parallelization option for all_pairs_shortest_path as suggested…

hcars:
Added parallelization option for all_pairs_shortest_path as suggested in the TODO.


MridulS:
@hcars thanks for the PR! We probably will not just depend on multiprocessing for parallelization.

Further discussions in https://github.com/networkx/networkx/pull/6306 and https://github.com/networkx/networkx/issues/4064

PR #4012, opened 2020-06-21
Fix exception causes and messages in 12 modules

cool-RR:
@dschult This is the continuation of #4000

I also noticed a few mistakes in error messages along the way, I corrected these too.


dschult:
Thanks @cool-RR !!

PR #4016, opened 2020-06-23
Add maximum weight clique algorithm

jamestrimble:
I have implemented an algorithm for maximum weight clique that is often faster than enumerating all maximal cliques, which I understand to be the current best way to solve the problem in NetworkX. The new algorithm can also be useful for solving the _unweighted_ max clique problem by calling max_weight_clique(G, None).

It would only require a few more lines of code to also cover the maximum (weight) independent set problem, so perhaps that would be worth considering.

This is my first attempt to commit to a large open source project, so hopefully I've done it roughly right!


networkx/algorithms/clique.py
```diff
@@ -540,3 +540,159 @@ def cliques_containing_node(G, nodes=None, cliques=None):
...
+        self.expand([], 0, vertices)
+
+
+@not_implemented_for('directed')
+def max_weight_clique(G, weight='weight'):
+    """Find a maximum weight clique in G.
+
+    A _clique_ in a graph is a set of vertices such that every two distinct
+    vertices are adjacent.  The _weight_ of a clique is the sum of the weights
+    of its vertices.  A _maximum weight clique_ of graph G is a clique C in G
```
dschult, line 652: The doc isn't clear whether the weight is for edges or nodes here.

jamestrimble: Thanks for spotting this. Referring to edge attributes rather than node attributes was a bad mistake in my documentation, which is now fixed. (I had copied the docstring for the `weight` parameter from another NetworkX function for consistency, but had not noticed that that one used edge weights.) To algorithm is for max node weight clique, rather than the less common max edge weight clique problem.

dschult: Probably just a nit-pick...,  but could we change "vertices" to "nodes" and vertex to node?

jamestrimble: Sorry - I should have done that from the start for consistency with the rest of NetworkX. I have now changed vertex/vertices to node/nodes throughout the MWC code.

dschult: Thanks very much for your quick response.....
This looks ready!

PR #4028, opened 2020-06-28
added new Link Prediction algorithm (CCPA)

itsshavar:
CCPA was published in Nature Scientific Reports -2020. I will add some more link prediction algorithm later.

### DIFF
--- a/networkx/algorithms/link_prediction.py
+++ b/networkx/algorithms/link_prediction.py
@@ -14,7 +14,7 @@ __all__ = ['resource_allocation_index',
            'preferential_attachment',
            'cn_soundarajan_hopcroft',
            'ra_index_soundarajan_hopcroft',
-           'within_inter_cluster']
+           'within_inter_cluster','common_neighbor_centrality']
 
 
 def _apply_prediction(G, func, ebunch=None):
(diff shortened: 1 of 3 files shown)


itsshavar:
it's my first pull request in open source community. Thank you

networkx/algorithms/link_prediction.py
```diff
@@ -198,6 +198,70 @@ def predict(u, v):
...
+    (2, 3) -> 3.4000000000000004
+
+    References
+    ----------
+    .. [1] Ahmad, I., Akhtar, M.U., Noor, S. et al. 
+           Missing Link Prediction using Common Neighbor and Centrality based Parameterized Algorithm. 
+           Sci Rep 10, 364 (2020). 
+           https://doi.org/10.1038/s41598-019-57304-y
+    """
+    sortest_path = nx.shortest_path(G)
```
dschult, line 261: typo: sortest -> shortest?

Then do the common_neighbors need to be sorted before finding their length?
Could be:  
```python
alpha * len(list(nx.common_neighbors(G, u, v))) + (1 - alpha) * (len(G) / (len(shortest_path[u][v]) - 1))
```
That is long enough you might want/need to split across 2 lines.

itsshavar: Thank you Dan. I will do the changes.

Issue #3938, opened 2020-05-01
Add option to BFS_edges to process children in order based on node attribute


tambulkar:
I would like to add the option to pass in a comparator function that will add children to the queue in an order specified by the comparator and their [node attributes](https://networkx.github.io/documentation/stable/reference/generated/networkx.classes.function.set_node_attributes.html)

dschult:
That seems like a nice practice problem: take the code for bfs_edges, read it and understand it sufficiently to change the order of the nodes when added to the queue.  If you can make it easily understandable code and general enough to be useful for others and not detrimental to the time needed for the standard case, please create a PR for adding it to NetworkX. :}

adnanmuttaleb:
#4029

PR #4029, opened 2020-06-28
add the option of sorting node's neighbors during bfs traversal 

adnanmuttaleb:
Allow client to pass a function that customize the order in which neighboring nodes are queued.

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -62,6 +62,7 @@ def generic_bfs_edges(G, source, neighbors=None, depth_limit=None, sort_neighbor
     .. _PADS: http://www.ics.uci.edu/~eppstein/PADS/BFS.py
     .. _Depth-limited-search: https://en.wikipedia.org/wiki/Depth-limited_search
     """
+
     def get_neighbors(node):
         neighbors_list = neighbors(node)
         if callable(sort_neighbors):
@@ -218,7 +219,8 @@ def bfs_tree(G, source, reverse=False, depth_limit=None, sort_neighbors=None):
     """
     T = nx.DiGraph()
     T.add_node(source)
-    edges_gen = bfs_edges(G, source, reverse=reverse, depth_limit=depth_limit, sort_neighbors=sort_neighbors)
+    edges_gen = bfs_edges(G, source, reverse=reverse, depth_limit=depth_limit,
+                          sort_neighbors=sort_neighbors)
     T.add_edges_from(edges_gen)
     return T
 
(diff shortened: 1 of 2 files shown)


adnanmuttaleb:
#3938

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -58,10 +62,16 @@ def generic_bfs_edges(G, source, neighbors=None, depth_limit=None):
     .. _PADS: http://www.ics.uci.edu/~eppstein/PADS/BFS.py
     .. _Depth-limited-search: https://en.wikipedia.org/wiki/Depth-limited_search
     """
+    def get_neighbors(node):
```
dschult, line 65: How about: 
```python
if callable(sort_neighbors):
    neighbors = lambda node: sort_neighbors(neighbors(node))
```
Then you don't have to use get_neighbors and avoid excess processing? Are their other effects?

dschult:
Also, can you fix the PEP8 line-too-long complaints?

And get you add a test that ensure the sorting actually works.

networkx/algorithms/traversal/breadth_first_search.py
adnanmuttaleb: I think using `lambda` function will cause a `RecursionError`, thats why I used a regular functions.

adnanmuttaleb:
@dschult sure I will fix the PEP8 issues and I will add test function.

adnanmuttaleb:
@dschult fixed PEP8 issues, and added `test_bfs_edges_sorting` test case to `test_bfs.py`

networkx/algorithms/traversal/breadth_first_search.py
dschult: The lambda function won't cause a RecursionError because it doesn't use recursion on itself. It is inside a recursive function, but the lambda itself is not recursive. I'm pretty sure that will be OK.  

You could also use regular functions if you prefer with something similar to (lots of ways to do it):
```python
if callable(sort_neighbors):
    def get_neighbors(node):
        return sort_neighbors(neighbors(node))
else:
    get_neighbors = neighbors
```

adnanmuttaleb: when calling a `lambda`  the `neighbors` inside will evaluate to the defined `lambda` (which have the same name `neighbors`) , so there will be a recursion error, minimal example:

```
    def adjecent(x):
      return (x - 1, x + 1)

    adjecent = lambda x: sorted(adjecent(x))

    print(adjecent(10))
```

will Output `RecursionError: maximum recursion depth exceeded`. But I found a way to do it:

```
    if callable(sort_neighbors):
        _neighbors = neighbors
        neighbors = lambda node: iter(sort_neighbors(_neighbors(node)))
```

The `iter` is needed because the `sort_neighbors` may return an iterable instead of an iterator. I will update to this implementation , if you OK with it?

dschult: That's great -- Thanks!

adnanmuttaleb:
@dschult all done.

PR #4051, opened 2020-07-07
Update sphinx

jarrodmillman:
The previous version had an error that was preventing bridges
from being autodocumented.


dschult:
Thanks for finding this -- I couldn't figure out why bridges was failing to work... :}

PR #4066, opened 2020-07-11
relabel_nodes now preserves edges in multigraphs

### LINKED ISSUE #4058: relabel_nodes on MultiGraphs does not preserve both edges when two nodes are replaced by one
When the graph contains edges (0,1) and (0,2), and I relabel both 1 and 2 to 3, I expected two edges from (0,3) but only one node is preserved.

Multi*Graph supports parallel edges between nodes and I expected it to preserve both edges when "merging" the two old nodes together.

Tested on both networkx 2.4 and the latest version on master.

```python
import networkx as nx

G = nx.MultiDiGraph([(0,1),(0,2)])
G[0][1][0]["value"] = "a"
G[0][2][0]["value"] = "b"

print(G[0])
# Output:
# {1: {0: {'value': 'a'}}, 2: {0: {'value': 'b'}}}

mapping = {1:3, 2:3}

nx.relabel_nodes(G, mapping, copy=False)

print(G[0])
# Output:
# {3: {0: {'value': 'b'}}}
# Expected:
# {3: {0: {'value': 'a'}, 1: {'value': 'b'}}}
```

nik0sc:
Fixes #4058.

In a multigraph, relabeling two or more nodes that share predecessors into the same new label will preserve all edges, but will change the edge keys in the process. Previously, edge keys were not changed, but if edges ended up having the same (tail, head, key) tuple after the relabel, only one of those edges would be inserted into the relabeled graph.

### DIFF
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -13,7 +13,7 @@ def relabel_nodes(G, mapping, copy=True):
 
     mapping : dictionary
        A dictionary with the old labels as keys and new labels as values.
-       A partial mapping is allowed.
+       A partial mapping is allowed. Mapping 2 nodes to a single node is allowed.
 
     copy : bool (optional, default=True)
        If True return a copy, or if False relabel the nodes in place.
@@ -64,9 +64,8 @@ def relabel_nodes(G, mapping, copy=True):
     >>> list(H)
     [0, 1, 4]
 
-    In a multigraph, relabeling two or more nodes that share predecessors
-    into the same new label will preserve all edges, but will change the
-    edge keys in the process:
+    In a multigraph, relabeling two or more nodes to the same new node
+    will retain all edges, but may change the edge keys in the process:
 
     >>> G = nx.MultiGraph()
     >>> G.add_edge(0, 1, value="a")  # returns the key for this edge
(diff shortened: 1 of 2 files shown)


dschult:
Hi!  Thanks for this -- very helpful.

I think the logic for finding keys still doesn't work in some funky cases -- in particular, it seems like directed edges for the in-place routine might end up with the same key because it is checking against G so 2 new edges would get the same key.  Perhaps it is just a matter of including a seen_edges set to hold them for the in-place case too.  For the copy-case I think you need to add both directions of each edge to seen_edges when the graph is undirected.

To investigate a little, I tried to extend the tests beyond a structure with one node to many. I added many to one node as well. And I combined the graph type and copy/in-place combinations into a single function with a loop.  What do you think of something like this?
```python
    def test_relabel_multidigraph_inout_merge_nodes(self):
        for MG in (nx.MultiGraph, nx.MultiDiGraph):
            for cc in (True, False):
                G = MG([(0, 4), (1, 4), (4, 2), (4, 3)])
                G[0][4][0]["value"] = "a"
                G[1][4][0]["value"] = "b"
                G[4][2][0]["value"] = "c"
                G[4][3][0]["value"] = "d"
                G.add_edge(0, 4, key="x", value="e")
                G.add_edge(4, 3, key="x", value="f")
                mapping = {0: 9, 1: 9, 2: 9, 3: 9}
                H = nx.relabel_nodes(G, mapping, copy=cc)
                # No ordering on keys enforced
                assert {"value": "a"} in H[9][4].values()
                assert {"value": "b"} in H[9][4].values()
                assert {"value": "c"} in H[4][9].values()
                assert len(H[4][9]) == 3 if G.is_directed() else 6
                assert {"value": "d"} in H[4][9].values()
                assert {"value": "e"} in H[9][4].values()
                assert {"value": "f"} in H[4][9].values()
                assert len(H[9][4]) == 3 if G.is_directed() else 6
```

Another request (which is admittedly minor) is for you to switch variable names from tail-head to source-target to match the surrounding code. And the "current_keys" variable could be removed since it's used only once.

Thanks!

PR #4067, opened 2020-07-11
Add tree isomorphism

craigschmidt:
Add an algorithm for finding if two undirected trees are isomorphic,
and if so returns an isomorphism between the two sets of nodes.

This algorithm uses a routine to tell if two rooted trees (trees with a
specified root node) are isomorphic, which may be independently useful.

Includes tests with 100% coverage that show correctness for all trees up to a certain size (assuming the nx.nonisomorphic_trees routine is correct)


dschult:
Thanks for this!  I have only scanned through the code quickly at this point. Here are some comments. If you're up for learning a little more about how the package is set up, take a stab at fixing them. Otherwise I'll get to it.

Rename the module to tree_isomorph.py to match the other modules' naming system. (similar for test file)

Add a definition of the variable ```__all__``` just after your imports. It should equal a list of all functions you want added to the networkx package when importing.  Look at some nearby modules for examples.

Hook into the namespace:  Add an entry to ```algorithms/isomorphism/__init_.py``` which imports your module into the isomorphisms module (and thus into the main networkx namespace).  Mimic the other entries there.

Check for yourself that importing networkx from your repository allows access to your functions.

Hook your tests into the documentation system by making an entry in ```doc/reference/algorithms/isomorphism.rst``` that mimics the entries there already. I think a subsection for tree isomorphisms is reasonable and it should include the functions you "make public" in the ```__all__``` variable above.

You should install and run "black" (also known as psf/black) on your two modules -- this is a recent addition to networkx and we hope to unify the style and spacing.  Right now the PEP8speaks tells you what needs fixing. But black ought to fix it for you.

Thanks again!

craigschmidt:
@dschult 
Hi Dan, I incorporated the changes you suggested, and ran the code through black.  I wasn't entirely sure how to use the development version in other code, so I didn't verify that `from networkx import *` allows the two public facing functions.

networkx/algorithms/isomorphism/tests/test_tree_isomorphism.py
```diff
@@ -0,0 +1,285 @@
...
+    edges_2 = []
+    for (u, v) in t2.edges():
+        # translate to names for t1
+        u = mapping[u]
+        v = mapping[v]
+        if d2:
+            edges_2.append((u, v))
+        else:
+            if u < v:
+                edges_2.append((u, v))
```
dschult, line 48: The ordering of nodes may be different for the mapped version of the network.
It’s not clear tome that this will work correctly.  Could this be tested by checking an isomorphism that e.g. reverses the order of nodes for some edge?

craigschmidt: I think that's the correct way to test for isomorphism.  Two graphs are isomorphic if and only if they have same set of edges after relabeling one graph the isomorphism.   https://mathworld.wolfram.com/IsomorphicGraphs.html

In the case of a directed graph, the edge order clearly matters. But on an undirected graph, the edge (u,v) is the same as the edge (v,u).  That's why I put the edge in particualr form of u < v, before testing that the set of edges is the same. 

In the routine positive_single_tree, used in testing, I randomly shuffle the nodes before constructing a test graph. You're right that I should probably shuffle the edge order as well, to ensure it is correct both ways.  I've added a random_swap routine to do that. 

There isn't really a good proof of correctness for this algorithm.  The Aho, Hopcroft, and Ullman reference doesn't even really state the algorithm completely.  And once it was in that book, nobody seems to have published about it.  So the testing seems very important.  I tested the positive case for trees up to 14 nodes, and the negative case up to 11.

dschult: I just mean that this ```check_isomorphism``` function uses the order ```u<v``` for t1 and ``` mapping of u < mapping of v``` for t2.
Maybe use ```new_u = mapping[u]``` (and v) and then check ```u<v``` as you do... or save the orig u and v before mapping and check ```orig_u < orig_v```.  It doesn't matter to me what way you arrange the variables-- but the node order for edges in the two edge collections should be the same or the function will report (1,2) as different from (2,1) in an undirected graph.  I feel like I'm not being clear, so ask if something doesn't make sense.

dschult: Ahhh....     My bad!   You **DO** want to check the order of the mapped nodes... NOT the order before mapping.
Sorry about that.

dschult:
This is ready to be merged. :}

Issue #3357, opened 2019-03-26
Add function to calculate weight of a specific path


patsumby-sohonet:
I can create a graph with edges with multiple attirbute weights on the edges (ie $cost and latency).

I can identify the shortest path based on cost or the shortest path based on latency.

What I can't do is find the cost of the shortest latency path or find the latency of lowest cost path.

I've worked around this by writing a helper function as follows:

```python
def cost_of_path(graph,path,metric):
        total = 0
        hops = len(path)
        for i in range(1,hops):
                value = min([v[metric] for v in graph[path[i]][path[i-1]].values()])
                total += value
        return total

cost_of_path( G, [A,B,C], "latency") # returns latency of this path
cost_of_path( G, [A,C], "cost") # returns $ cost of this path
```

[A.B,C] is the lowest $ cost path.
[A,C] is the lowest latency path

Seems this should be easy enough to add in and would be a useful function of the system.

The "min([v[metric] for v in graph[path[i]][path[i-1]].values()])" line is because I'm using a multigraph and there are potentially multiple edges between some pairs of nodes.

Loving networkx though, very useful.

dschult:
This seems like a reasonable addition. Let's make it work for directed/undirected/multiedge 
A good place for it could be in classes/functions.py but I'm open to suggestions...

manesioz:
To generalize this you essentially want a function to calcluate the cost of *any* arbitrary path, not necessarily the shortest one (relative to your weight).

Surprisingly there doesn't seem to be any general function for this (`all_simple_paths()` is close but it generates paths unnecessarily, we already have a single one in mind), but it seems like it would be easy to implement (using the available `traversal` methods). What do you think?

dschult:
No traversal is needed. The path is already known and provided. You just need to loop over the edges and add up the edge attributes along that path.

```python
    def path_weight(G, path, weight):
        for n, nbr in nx.utils.pairwise(path):
            sum G[n][nbr][weight]
```
Obviously this pseudocode doesn't work, but something close to it should work.
Then make it work for multiedge, etc.

manesioz:
[...]

PR #4069, opened 2020-07-12
Developing function to calculate path cost for any user-specified path

DNiquette16:
Hello! 

So I wrote a function called "path_weight" to calculate the cost of a specific path with respect to a specified edge attribute (e.g. cost, weight, distance). I believe I have a function that works properly with graphs as well as multigraphs and throws an error if the path does not exist. However, this is my first time submitting work to a project like this so I would love any feedback as I am sure this is far from being ready to merge. 

Also, this is connected to issue #3357

Thanks!


DNiquette16:
Hey,

Thanks for all the great feedback!

I believe that I address all of the comments except for the use of the _weight_function(). I had a question about that. If my understanding is correct, I don't believe that single underscore functions are imported by default when using `from module import *`. Do you have any advice on how to approach this one? I have tried to modify the __init__.py in networkx/algorithms/shortest_paths and tried various import statements in networkx/classes/function.py but I haven't been able to import the function. I am sure there is an easy fix to this, I just haven't found it.

Thanks!

dschult:
Good question. The underscore functions usually are not imported if you use ```from shortest_path import *```
But so long as shortest_path has been imported, we can still get the "private" functions by explicitly listing them.
The file at ```networkx/algorithms/shortest_path/astar.py``` provides as example:

```python
```
It's a little convoluted, but it works.

dschult:
I'm not sure why the third Appveyor test is failing to start. Hopefully it will be fixed for your next commit -- don't worry about it.

The loop for ```is_path``` creates a potentially long list. So it might be better to switch to a lower footprint approach, something like (not tested):
```python
for node, nbr in nx.utils.pairwise(path):
    if nbr not in G[node]:
        return False
return True
```

DNiquette16:
Hmm, okay that is kind of the track that I was headed down. However when I import the following at the top of networks/classes/functions.py:
`from networkx.algorithms.shortest_paths.weighted import _weight_function`
And then use it with a dummy test in my path_weight function to see if I can call it:
`x = _weight_function(G,weight)`
I get the following error:
image

However, I didn't make any changes to networkx/algorithms/d_separated.py

That's also a very good point about is_path, I will make that change as well!

dschult:
Looks like this is running into an import order issue where ```function.py``` is imported before enough of the classes are imported that it is too soon to import stuff from ```algorithms```.  Looks like it might be time to move _weight_function into a different module (perhaps utils?)

Let's just implement this with weight being a string (attribute name).  There is another issue #4085 for working toward using _weight_function wherever possible in the package.  That's the place to change its location and we'll update this code as part of that anyway.

DNiquette16:
Okay, thanks for helping me figure out what was going wrong with that import. Definitely sounds like a good idea to make that a separate issue. I just pushed my changes to `is_path()` and I believe I incorporated the rest of the changes that you had mentioned initially. Let me know if there is more to fix and I'll be happy to make more adjustments!

jarrodmillman:
@DNiquette16 This branch has conflicts that need to be resolved.  Also if you haven't already, please run black to format the code:
```
pre-commit install
```
before making your final commit and black will be run automatically for you.  You can also run it manually:
```
black .
```

If you don't have time to take care of this, no worries.  I will do it before merging your PR later tonight or tomorrow.

Thanks!

PS.  If you haven't already, you may want to look over ``CONTRIBUTING.rst`` (also available here:  https://networkx.github.io/documentation/latest/developer/contribute.html) at some point.

PR #4070, opened 2020-07-13
Clear pygraphviz object after creating networkx object

PranayAnchuri:
I noticed a memory leak in a downstream application that reads a ton of dot files and converts them to networkx objects using `read_dot` function. My application crashed after reading about 500 dot files (each has around 500 nodes). Upon inspecting the `nx_agraph.py` file, I noticed that the pygraphviz object created during `read_dot` is not cleared. This PR fixes the memory leak issue by first creating networkx object (the one we want to return) and then clearing the intermediate pygraphviz object.

P.S : The reverse function `write_dot` does indeed clear the intermediate object.


dschult:
Nice catch!  Thanks

PR #4072, opened 2020-07-14
Added harmonic centrality subset extension and tests.

ChristopherReinartz:
Extension of the already existing method for for harmonic_centrality. 
The new method allows the definition of subsets for both nodes 'u', for which harmonic centrality is to be calculated and 
nodes 'v', which are the nodes with respect to which the calculation is performed. 
The previous harmonic centrality method sets 'v' to cover all other nodes in the graph, following the classical definition of harmonic_centrality.


jarrodmillman:
To deal with the PEP8 issues, you should install and run black (https://github.com/psf/black).  There is a lot of overlap with this function and ``harmonic_centrality``.  @dschult Maybe it would be better to make this an option for ``harmonic_centrality``, rather than a new function.

ChristopherReinartz:
@jarrodmillman  Should I integrate the subset function as an option into harmonic_centrality, as you suggested? I could extend the function signature of harmonic_centrality from harmonic_centrality(G, nbunch=None, distance=None) to harmonic_centrality(G, nbunch=None, nbunch_v=None, distance=None). If nbunch_v isn't specified, the existing function would be used and otherwise the new one. This shouldn't affect any existing usage of harmonic_centrality, as far as I can see.

dschult:
I think that sounds good -- there is a backward compatibility argument for the order ```harmonic_centrality(G, nbunch=None, distance=None, nbunch_v=None)```
But more than that, is there a better name than nbunch_v?  For example, ```starting_nodes```. Or maybe they should be called sources and targets?

Issue #2852, opened 2018-01-28
MacOS Graphviz MultiGraph tests failing on Travis


dschult:
The Travis Graphviz version for Mac OS is 2.40 while it is 2.3? for linux. But I'm not sure that is the issue.

dschult:
Yes, it looks like the same issue as those issues with pygraphviz. 
I'm turning the tests back to using Graphviz 2.38 instead of 2.40. When 2.42 comes out we can go back to just getting the most recent version.

dschult:
Note that Graphviz v2.42 is out -- once conda and other packages switch to 2.42 this bug should go away.

PR #4075, opened 2020-07-14
Use newer osx on travis

jarrodmillman:
Closes #2614.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,23 +43,17 @@ matrix:
           - libgdal-dev
           - graphviz
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode12
       language: generic
       env:
-      - TRAVIS_PYTHON_VERSION=3.6
+      - TRAVIS_PYTHON_VERSION=3.7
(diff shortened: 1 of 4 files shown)


jarrodmillman:
We are getting two errors:
```
FAILED networkx/drawing/tests/test_agraph.py::TestAGraph::test_multi_undirected
FAILED networkx/drawing/tests/test_agraph.py::TestAGraph::test_multi_directed
```

It looks like conda is still using an old, buggy version of graphviz and we are hitting this problem again #2852.  It looks like brew has been updated, so I will try it.  We will probably want conda to update before the next release.

dschult:
I believe current Graphviz is v2.42, but the buggy version (v2.40) is still installed by conda. 
Once we get v2.42 the tests should pass.

jarrodmillman:
@dschult The tests are passing now.  We no longer use conda and instead use brew to install graphviz-2.44.1 on travis-ci.

dschult:
This looks good -- and a nice solution to the conda graphviz mess. Thanks!

Issue #4065, opened 2020-07-11
Add option to specify a column for edge keys in Multi(Di)Graphs when reading through pandas


MridulS:
When reading in graphs using pandas `from_pandas_edgelist` should we have an option to let users specify the columns from the dataframe for edge keys?

Example
``` python
G = nx.from_pandas_edgelist(dataframe, source=source,
       target=target, edge_attr=[attr1, attr2, attr3], create_using=nx.MultiGraph())
```
currently networkx will create keys for this graph automatically (0, 1, 2, 3 ......) and to access the edge data we would need to do something like `G[source_node][edge_node][0]`

it would be nice to add an option of adding custom edge keys (assuming we want attr1 to be edge key)
``` python
G = nx.from_pandas_edgelist(dataframe, source=source,
       target=target, edge_attr=[attr2, attr3], edge_key=attr1, create_using=nx.MultiGraph())
```
and access the edge information using  `G[source_node][edge_node][attr1_node]`

dschult:
I don't think we've thought carefully about multigraphs, edge keys and uniqueness of edge keys in a pandas data frame. This Issue sounds like a good idea. We should probably also consider what to do when the edge keys match.  Without multiedges we just overwrite existing data when new edges come in. Maybe duplicate edge/edgekey should do that same?  But maybe not...  What are the use cases here? Are there common data frames that represent multigraphs?

Daniel-Davies:
I think this issue could be useful for a project I am currently doing- may I investigate this? 

As you've described, one logical thing to do with multiedges/ duplicates is to overwrite the existing data- but perhaps there could be some user configurable options. For example, a default would be to overwrite, another method could be a 'log' approach (where each G[source_node][edge_node] is a list), or perhaps another option could be to take the most occuring element?

dschult:
Yes -- please investigate this.  I think for starters you can follow the ```G.add_edge``` treatment that if the edge exists, "adding" the edge just updates the data dict. That is the simplest treatment.
Thanks!

Daniel-Davies:
Raised a PR for the basic version at the moment, but I suppose some discussion is needed on how to deal with duplicates. I will take a closer look tomorrow and propose an option as a new issue?

PR #4076, opened 2020-07-14
Allow custom keys for multiedges in from_pandas_edgelist

Daniel-Davies:
Adding PR for feature stated in issue #4065, regarding the ability to provide a custom key for multigraphs in the from_pandas_edgelist function. Code currently (as stated in issue) deals with duplicate keys by overwriting them, and I will look into better mechanisms in future PRs. Tests have been added and code was run through black.

### DIFF
--- a/networkx/classes/multigraph.py
+++ b/networkx/classes/multigraph.py
@@ -94,7 +94,7 @@ class MultiGraph(Graph):
 
     >>> keys = G.add_edges_from([(4,5,{'route':28}), (4,5,{'route':37})])
     >>> G[4]
-    AdjacencyView({3: {0: {}}, 5: {0: {}, 1: {'route': 28}, 2: {'route': 37}}})
+    AdjacencyView({5: {0: {}, 1: {'route': 28}, 2: {'route': 37}}})
 
     **Attributes:**
 
--- a/networkx/classes/multigraph.py
+++ b/networkx/classes/multigraph.py
@@ -94,7 +94,7 @@ class MultiGraph(Graph):
 
     >>> keys = G.add_edges_from([(4,5,{'route':28}), (4,5,{'route':37})])
     >>> G[4]
-    AdjacencyView({5: {0: {}, 1: {'route': 28}, 2: {'route': 37}}})
+    AdjacencyView({3: {0: {}}, 5: {0: {}, 1: {'route': 28}, 2: {'route': 37}}})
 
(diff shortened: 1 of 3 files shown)


jarrodmillman:
We use psf/black to format code now.  So you don't need to manually fix all the pep8 issues.  Instead install and run black and it will fix these issues for you:  https://github.com/psf/black

dschult:
This looks pretty good. Almost ready.  Some comments:

Could you inline the 3 helper functions? They are only used once and the namespace is already pretty big in this module.

Could you add documentation of the new feature into the docstring?

Thanks!

dschult:
The Travis-CI is failing on a new OSX test platform just put in place yesterday. It is not your code so don't worry about getting that test environment to pass. The others are all passing.

This looks good -- here's some last bits to clean up:
- is the ```key = -1``` needed? what is its purpose?
- can you make the msg on line 386 avoid the white space and newline by concating strings like this"
```python
msg = ("first part"
               f"second part {with formating}")
```
- change the name of one of the first two tests (same name) so they both run.
- can you remove the "as excinfo" in the with statement for pytest.raises? PEP8 complains because excinfo is defined but isn't used.

Daniel-Davies:
Hello
Thank you very much for supporting my first pull request to NetworkX- i really appreciate it. I have made the changes requested, and blackened the code. 

I also took a look into the issue of multigraph repeating keys- there wasn't a great deal of information I could find for existing use cases in which multigraph keys can conflict, and after thinking about the issue in general, I don't think NetworkX should handle a user-created duplicate key, for two main reasons:

- Resolving the issue would either need a 'log of changes', which implies a list return type for each edge (conflicting with the existing use cases of the multigraph), or something like a max-value protocol, which could cause the user more trouble in debugging if the duplicated key was an accident.
- In my opinion at least, the NetworkX code base shouldn't grow in complexity (and hence add a barrier to future contributors) by having to deal with these kinds of 'edge cases', which are quite easily configurable on the user side, before entering the dataframe.

On the other hand, if this is a feature that you think would be important, I think the best way to handle it would be by having the user input take the following formats:

edge_key :: string => normal, overwrite behaviour
edge_key :: tuple (string, string) => the first string is the key, and the second string would be the method to deal with duplicate keys (e.g. 'save_all' to output edges as a list, 'most_common' etc)

Thoughts?

Thank you again for all the help!

PR #4079, opened 2020-07-16
Install Python after updating brew

jarrodmillman:
> Formulae that declare an unconditional dependency on the "python" formula are bottled against Homebrew’s Python 3.x and require it to be installed.

``graphviz``  declares an unconditional dependency on the "python", which is now 3.8 by default.  And

> WARNING: When you brew install formulae that provide Python bindings, you should not be in an active virtual environment.

So I moved the brew install before we create the venv.

NOTE:  Quotes are from https://docs.brew.sh/Homebrew-and-Python

### DIFF
--- a/tools/travis/osx_install.sh
+++ b/tools/travis/osx_install.sh
@@ -3,16 +3,16 @@ set -ex
 
 section "OSX install section"
 
-# set up Python and virtualenv on OSX
-git clone https://github.com/matthew-brett/multibuild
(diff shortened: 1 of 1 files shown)


jarrodmillman:
@dschult , @rossbar It turns out this had an easy fix (once looked at the brew instructions).  I need to step away from my machine for a bit.  Once the tests pass, feel free to squash and merge this.

Issue #4085, opened 2020-07-17
Allow weight_function wherever edge weight is used


dschult:
We have a nice treatment of edge weights in the shortest_path sub-package: the weight can be the name of the edge attribute that holds the weight, but it can also be a function of the data dict for an edge. Thus the weight can be static, or it can be computed as needed.

There are a number of other places in the code-base where edge weights are used. It would be good to enable the weight function interface throughout.

dschult:
I'll try to construct a list of functions where a "weight" argument could be upgraded to the ```_weight_function``` approach.  This might be a good topic for a virtual sprint (lots of places with similar code-fixes to implement.

adnanmuttaleb:
@dschult I tried to fix the `reaching.py`  as you suggested, but in order to make it completely consistent you have to call `_weight_function`  in each function that use the `weight` parameter directly in the invocation tree. Or maybe we can assume that `weight` is always a callable of type `Callable(u: Node, v :Node, data:Mapping)` in all functions. To do so we may have something like:

```
from functools import wraps

def _weight_function(G, weight):
    if callable(weight):
        return weight
    if G.is_multigraph():
        return lambda u, v, data: min(attr.get(weight, 1) for attr in data.values())
    return lambda u, v, data: data.get(weight, 1)


def weighted(func):
  @wraps(func)
  def wrap(G, *args, weight='weight', **kwargs):
    weight = _weight_function(G, weight)
    return func(G, *args, weight=weight, **kwargs)
```

usage:

```
@weighted
def single_source_dijkstra_path(G, source, cutoff=None, weight='weight'):
  pass

@weighted
def global_reaching_centrality(G, weight=None, normalized=True):
  pass
```

dschult:
Yes, I think the decorator approach is probably best.  It'd be great if it could work for other weight-like parameters like "cost" or "capacity", but that might make it too complex.

We should move ```_weight_function``` and its decorator into the networkx/utils directory, perhaps in the decorators.py module.

adnanmuttaleb:
Great, I will work on it. We may create a general parameterized decorator that work for any keyword argument `weight`, `capacity`..etc.  I will submit PR soon hopefully.

adnanmuttaleb:
After doing this initial draft a few question raised:

1. what is the meaning of `is_weighted` if weight now is always a function.
2. what is the expected behavior if client passed `weight` as None, (consider the graph as not-weighted or something else).
[...]

PR #4114, opened 2020-07-29
Initial implementation of weight function wherever edge weight is used

adnanmuttaleb:
Add support for either direct access or computable attributes such as, `weight`, `capacity`, ..etc. For more info check a #4085. This PR is in draft phase.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,49 @@
+# See: https://circleci.com/docs/2.0/language-python/
+
+version: 2
+jobs:
+  build:
+    working_directory: ~/repo
+    docker:
+      - image: circleci/python:3.8.3-buster
+
+    steps:
+      - checkout
+
+      - run:
+          name: install_graphviz
+          command: |
+            sudo apt-get install graphviz libgraphviz-dev
+
+      - run:
+          name: install_tex
+          command: |
+            sudo apt-get install texlive texlive-latex-extra latexmk
+
+      - run:
+          name: install_dependencies
+          command: |
+            python3 -m venv venv
+            source venv/bin/activate
+            pip install --upgrade pip
+            pip install -r requirements.txt
+            pip install -r requirements/doc.txt
+            pip install -r requirements/optional.txt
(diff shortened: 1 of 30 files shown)


jarrodmillman:
We use ``black`` for formatting now:
  https://github.com/psf/black

This means you don't need to manually deal with PEP8 issues anymore---just run black.  You can read more in our contributor guide:  https://networkx.github.io/documentation/latest/developer/contribute.html

adnanmuttaleb:
@jarrodmillman thank you for letting me know.

dschult:
I see the ```local_bridges``` function is using ```weight=None``` to indicate a function that always returns 1, but there's got to be a better way to do that. It does always return 1, but edge attribute lookups are usually inner loop items where performance matters. I'll try to figure out a good way to improve local_bridges to avoid this issue.

Similarly for ```shortest_simple_paths```.  
And I think ```_bidirectional_shortest_path``` just ignores the weight argument so no decorator needed.

Looks like the remaining 2 errors are for ```nx.johnson```.

The title of the PR is "initial implementation".  Is there more to be done? How much of the codebase did you look through to find appropriate functions?  Looking elsewhere can be a different PR.

adnanmuttaleb:
@dschult sorry for being late, but I was really busy, I have changed naming to `edge_attribute` which is much better, since we may create something similar as `node_attribute`, `path_attribute` in the future. Also for the functions that has `None` as default for `weight`  but they expecting `_weight_function` to return function, I changed the default to `"weight"`. Still we have to fix the errors related to `nx.johnson`, and decide what to do if the wrapped attribute (e.g `weight`) does not have a default value (e.g `_dijkstra_multisource`), should I raise an error?.

adnanmuttaleb:
Regarding the `is_weighted` check, can we modify it behavior, so that the `weight` parameter is a function, that should return non `None` for all edges in the graph `G`.  And for any function the works only on weighted graph, we could decorate it as follow:

```
@edge_attribute("weight", default=None)
def johnson(G, weight="weight"):
    ...
```

PR #4130, opened 2020-08-04
Added basic cythonized code

matusvalo:
This PR adds cythonization to networkx as Proof of Concept.  This PR focuses on improvement of loading/dumping graphs and one algorithm is picked. As data set is graph with 334863 vertices and 925872 edges. Performance improvements:

* `python -m timeit -n 10 -r 3 -s 'import networkx as nx' 'nx.read_adjlist("data/out.com-amazon")'`
(**28.87%** increase):
cython: 10 loops, best of 3: 3.67 sec per loop
python: 10 loops, best of 3: 5.16 sec per loop

* `python -m timeit -n 10 -r 3 -s 'import networkx as nx; G = nx.read_adjlist("data/out.com-amazon")' 'nx.write_adjlist(G, "data/out.com-amazon_deleteme")'`
(**30.34%** increase):
cython: 10 loops, best of 3: 1.01 sec per loop
python: 10 loops, best of 3: 1.45 sec per loop

* `python -m timeit -n 10 -r 3 -s 'import networkx as nx' 'nx.read_edgelist("data/out.com-amazon.edgelist")'`
(**8.87%** increase):
cython: 10 loops, best of 3: 11.3 sec per loop
python: 10 loops, best of 3: 12.4 sec per loop

* `python -m timeit -n 10 -r 3 -s 'import networkx as nx; G = nx.read_adjlist("data/out.com-amazon")' 'nx.write_edgelist(G, "data/out.com-amazon.edgelist")'`
(**20.84%** increase):
cython: 10 loops, best of 3: 2.05 sec per loop
python: 10 loops, best of 3: 2.59 sec per loop

* python -m timeit -n 10 -r 3 -s 'import networkx as nx; G = nx.read_adjlist("data/out.com-amazon")' 'nx.biconnected_components(G)'
(**8.80%** increase):
cython: 10 loops, best of 3: 2.59 usec per loop
python: 10 loops, best of 3: 2.84 usec per loop

### Installation
Code is written in pure python mode hence it is able to run in pure python with current performance or can be compiled to have performance gains. Hence **the code can be installed and used without no change**. To cythonize and build code the user needs to have cython installed and full build dependencies (GCC, python headers etc). Building is enabled by exporting `NETWORKX_ENABLE_SPEEDUPS` variable - for building and installing networkx with speedups run the following command:
```
NETWORKX_ENABLE_SPEEDUPS=1 python setup.py install
```
For installing networkx code in pure python installation is same as now:
```
python setup.py install
```

In future there is possibility to publish binary wheels with prebuilt cythonized code where users can install the library without execution build.

### Modifications to networkx library
1. The cythonization was done using pure python mode [1] via annotations. 
2. In order to cythonize read/write code there was need of removing dependency to decorator library and manually create decorators.

### Future improvements
Current code is just proof of concept which goal was to introduce smallest possible changes to code base. Further improvements can be introduced which will yield better performance. All of changes will support running as pure python code, but some of them will need more invasive changes to code base:
1. Implementing graph classes as extension classes [2]. This change will boost mainly code calling API of classes. E.g. https://github.com/networkx/networkx/blob/938063c9011a4ecd4f3632a7a59ed63334d821ee/networkx/readwrite/adjlist.py#L213 is calling `add_edges_from` method for each line in input file. This is bottleneck since calling python methods is slow. This migration has unfortunatelly several blockers:
   * static class attributes are not supported by cython
   * `**kwargs` are not supported by cython
   * not all classes can be migrated - e.g. `MultiDiGraph` cannot be migrated since it has 2 parents
2. gml loading/dumping cannot be cythonized since cython does not support typed namedtuples. This can be fixed via migrating to basic namedtuple
3. Introducing separate speedups. Hot paths can be improved by reimplementing them directly in cython with more low level code - e.g. using C++ constructs [3]. Unfortunately this will add maintenance burden to keep 2 versions of the same logic.
4. distribution can be improved by distributing via pre-compiled wheel packages.

But even following simple cythonization seems to have major performance impact as showed before. Is there a will to have this kind of speedups present in the networkx? If yes, how we can start?

[1] https://cython.readthedocs.io/en/latest/src/tutorial/pure.html
[2] https://cython.readthedocs.io/en/latest/src/tutorial/cdef_classes.html
[3] https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html


jarrodmillman:
Could you tell us more about the performance issues you are running into that Cython solves for you?  The small tests you did only shows a few seconds of improvement.  Do you have a real-world problem where this improves things from say hours to minutes for you?  If you tell us more about the bottleneck you are experiencing, we may be able to suggest another way to address the performance issues you are experiencing.

dschult:
I think we should be looking at Cython and Numba and perhaps other things for ways to improve performance.  This is a start -- but shows that more is needed for any significant progress. 

Thanks!

jarrodmillman:
It may be worth making a NXEP about this before deciding whether to go with Cython, Numba, or something else.  This may also impact the release process, since we will want to make optimized wheels.

PR #4132, opened 2020-08-05
Updated draw_networkx to accept numpy array for edgelist

### LINKED ISSUE #4129: edgelist in draw_network does not support arrays
Using numpy 1.18.2 and networkx 2.4, it is not possible to use a numpy array for the `edgelist` argument of [``draw_networkx``](https://networkx.github.io/documentation/stable/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html).
It raises:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

I normally would not have opened an issue for that but I do not see the point of the current implementation, so in case it can be changed harmlessly to support numpy arrays, that might be a win-win.

The issue comes from lines 575-579 on 2.4 ([lines 640-647](https://github.com/networkx/networkx/blob/master/networkx/drawing/nx_pylab.py#L640) on master) that read:

```python
    if edgelist is None:
        edgelist = list(G.edges())

    if not edgelist or len(edgelist) == 0:  # no edges!
        return None
```

I am not sure what the combination of ``not edgelist`` and ``len(edgelist) == 0`` is for, here, so I would propose to change it to

```python
    if edgelist is None:
        edgelist = list(G.edges())
    elif len(edgelist) == 0:  # no edges!
        return None
```

which should still do the job for all valid inputs I can think of and would also work with numpy arrays.

**EDIT** if iterables should be supported, one could replace ``not edgelist`` by ``not isinstance(edgelist, Iterable)``

Silmathoron:
Fixes #4129 
Small update to ``draw_networkx`` to work with latest numpy versions when a numpy array is provided for ``edgelist``.
Added associated test.


Silmathoron:
Mhh... CircleCI tests are failing and I don't think that's on me, if somebody can check...

dschult:
I can verify that the Circle-CI errors are due to inability of the testing environment to load ruby2.5. Not your problem.

Thanks for this!

networkx/drawing/tests/test_pylab.py
```diff
@@ -250,3 +250,10 @@ def test_alpha_iter(self):
     def test_error_invalid_kwds(self):
         with pytest.raises(ValueError, match="Received invalid argument"):
             nx.draw(self.G, foo="bar")
+
+    @pytest.mark.skipif(os.environ.get("OPTIONAL_DEPS") != "1",
```
dschult, line 254: This ```skipif``` fature is something I didn't know about. Thanks for that!
But there is a tool more specifically for this case that I'd rather use.
(I'd rather not have to set my environment variables when testing locally.)

It's called [pytest.importorskip](https://docs.pytest.org/en/stable/reference.html#pytest-importorskip). Basically you try to import and ifthe import doesn't work it skips this test.  So, here you would remove the decorator and replace the import line with:
```python
```

Silmathoron: Done

networkx/drawing/nx_pylab.py
```diff
@@ -639,8 +639,7 @@ def draw_networkx_edges(
 
     if edgelist is None:
         edgelist = list(G.edges())
-
-    if not edgelist or len(edgelist) == 0:  # no edges!
+    elif len(edgelist) == 0:  # no edges!
```
dschult, line 642: Ack -- sorry I didn't see this before.  

I think this ```elif``` should be ```if``` because even after the previous if clause, edgelist could be an empty list (if the graph has no edges).

Silmathoron: Ah, you're right, fixed

PR #2067, opened 2016-04-07
Added dag functions and fixed PEP8 errors

Erotemic:
Added transitive_reduction, source_nodes, and sink_nodes.
Added corresponding tests.
Fixed minor PEP8 Errors.

In my projects I've written a bunch of helper functions for networkx and I'd like to start merging 
some of the more complete functions into the master branch. 

The main part of this pull request is an implementation of transitive_reductions for DAGs. 
I also added in helpers for source and sink nodes. I added the corresponding tests and they all seem to pass. I also cleaned up minor PEP8 formatting issues inconsistent with the rest of the file.

PR #4144, opened 2020-08-09
Add strong connectivity augmentation

tjelinek:
Hello,
I apologize for the mailing list spam. I send the files for review. I would especially like to ask you to review the changes in __init__ files and import in tests.
Modules sink and source are edited copy-paste of isolate. I hope it's ok.

Tomas

### DIFF
--- a/networkx/algorithms/__init__.py
+++ b/networkx/algorithms/__init__.py
@@ -44,8 +44,10 @@ from networkx.algorithms.shortest_paths import *
 from networkx.algorithms.similarity import *
 from networkx.algorithms.graph_hashing import *
 from networkx.algorithms.simple_paths import *
+from networkx.algorithms.sink import *
 from networkx.algorithms.smallworld import *
 from networkx.algorithms.smetric import *
+from networkx.algorithms.source import *
 from networkx.algorithms.structuralholes import *
 from networkx.algorithms.sparsifiers import *
 from networkx.algorithms.swap import *
--- a/networkx/algorithms/connectivity/__init__.py
+++ b/networkx/algorithms/connectivity/__init__.py
@@ -8,6 +8,7 @@ from .disjoint_paths import *
 from .kcomponents import *
 from .kcutsets import *
 from .stoerwagner import *
+from .strong_connectivity_augmentation import *
 from .utils import *
 
 __all__ = sum(
@@ -20,6 +21,7 @@ __all__ = sum(
         kcomponents.__all__,
         kcutsets.__all__,
         stoerwagner.__all__,
+        strong_connectivity_augmentation.__all__,
         utils.__all__,
     ],
     [],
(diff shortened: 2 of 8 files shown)


dschult:
Hello, @tjelinek   Thanks for the gentle nudge on responding to this PR. 

This looks like a good addition to the augmentation routines in NetworkX. Thank you!
I have some big picture suggestions that request a fair amount of refactoring, but hopefully are easy.
I haven't been able to go through the core routine much yet but will work on that.

- The function looks like it will fit nicely in the ```edge_augmentation.py``` module. The test module can similarly be combined with the corresponding test module.
- The function name recognizes the contribution of these two gentlemen, but doesn't describe the function's actions and doesn't match the docs where the term "strongly connectivity augmentation" is repeatedly used to describe the algorithm.  Maybe a different name would be better... perhaps the module name should become the function name when it gets put into the ```edge_augmentation.py``` module.  The two people's names can be made more prominent in the docs.
- The doc_string for the function has no paragraph describing what the function does. By that I mean the standard format is:  1) short, one line description, 2) blank line, 3) paragraph description of the function, 4) parameters... ...    So, the paragraph defines the terms, explains what the function does, maybe talks about who did this work, refers to the citations, and generally makes it clear whether a reader might want to use this function or not.
- the two modules sink.py and source.py should really be combined with isolate.py. Name it something that refers to all three of these ideas (generalize the name of the module). Maybe leaves.py?  There's got ot be something better than that... :} 
- install and run black for python style changes. Most of the suggestions are for lines being too long, so (especially with docs) try to shorten the doc_string lines and let black handle all the code.
- imports can ```import networkx as nx``` and then use ```nx.something``` rather than specifying the path within networkx to that function. Sometimes you may need ```nx.connectivity.something```.
- the sub-function ```search``` is only called in one place. Can the code be placed inline rather than separated as a function?
- add hooks into ```doc/reference/algorithms/connectivity.rst``` to the new function so the docs show your new function.

Thanks again!

Erotemic:
@tjelinek FWIW: I'm the author of the existing edge augmentation routines, which are currently only implemented for undirected graphs. In fact the existing algorithms are implemented from the same reference paper: Eswaran, Kapali P., and R. Endre Tarjan. (1975) Augmentation problems.       http://epubs.siam.org/doi/abs/10.1137/0205044 (does anyone know if there is a freely available version of this paper around? I'm no longer in school so I can't access it anymore). 

It looks like `strong_connectivity_augmentation` is the directed version of `bridge_augmentation` (or `k_edge_augmentation` where `k=1`). 

Does it make sense to expose your algorithms as backends with the existing `k_edge_augmentation` problem for use in the case where `G` is directed and `k=1`? That would solve the 2nd problem cited by @dschult  where the function name is not descriptive. 

Also, I would love having networkx functions for `sources` and `sinks`. I proposed the idea 4 years ago in #2067 but it looked like the reviewers preferred making people use one-liners. I'd love if networkx reconsidered that position. I really hate typing out a list comprehension every time I want the sources of a graph (its much more readable with a function too). 

I agree with the other points that @dschult mentioned. I'm also willing to help review this PR if needed as I have experience in the area.

PR #4152, opened 2020-08-14
Update dependencies

### LINKED ISSUE #3786: why gdal==1.10.0 in requirements?
Hi,

When running the pip command (```pip install networkx[all]```), I ran into GDAL setup error.

I was surprised by the error, because GDAL (version 2.2.2) was already installed on my desktop:
```
pip show gdal
Name: GDAL
Version: 2.2.2
Summary: GDAL: Geospatial Data Abstraction Library
Home-page: http://www.gdal.org
...
```

Looking at [networkx distutils requirements](https://github.com/networkx/networkx/blob/master/requirements/extras.txt), I realized that GDAL requires strictly 1.10.0 version for GDAL:
```
pygraphviz>=1.5
pydot>=1.2.4
gdal==1.10.0
```

**Question**
Is there a particular reason why ```gdal==1.10.0```, instead of ```gdal>=1.10.0```? Having a strict version usually makes the lib install error-prone. Thanks in advance.

jarrodmillman:
Fix #3786

### DIFF
--- a/requirements/extras.txt
+++ b/requirements/extras.txt
@@ -1,3 +1,3 @@
-pygraphviz>=1.5
-pydot>=1.2.4
+pygraphviz>=1.5, <2.0
+pydot>=1.4.1
 gdal==1.10.0
--- a/requirements/extras.txt
+++ b/requirements/extras.txt
@@ -1,3 +1,2 @@
 pygraphviz>=1.5, <2.0
 pydot>=1.4.1
-gdal==1.10.0
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -1,4 +1,4 @@
-numpy>=1.17
+numpy>=1.19
 scipy>=1.4
 pandas>=1.0
 matplotlib>=3.2
--- a/requirements/release.txt
+++ b/requirements/release.txt
@@ -1,2 +1,2 @@
-twine>=2.0.0
-wheel>=0.33.6
+twine>=3.2
+wheel>=0.34
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,3 +1,3 @@
-pytest>=5.4
+pytest>=6.0
 pytest-cov>=2.10
 codecov>=2.1
--- a/setup.py
+++ b/setup.py
@@ -134,7 +134,6 @@ extras_require = {
         "pygraphviz",
         "pydot",
         "pyyaml",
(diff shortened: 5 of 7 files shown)


dschult:
I'm not sure how much this matters, but the conda forges don't have pygraphviz 1.6 yet, so they will have to get that working in order to make networkx installs work.  Maybe we should leave pygraphviz requirement as >=1.5 for now... ???

jarrodmillman:
I changed ``pygraphviz==1.6`` to  ``pygraphviz>=1.5, <2.0``.  That should work for both pip and conda forge.  It should also prevent installing ``pygraphviz==2.0`` when that is released.

For ``gdal``, I just removed it from ``all`` so ``pip install networkx[all]`` won't cause a problem b/c of ``gdal``.  There is no good solution for having an entry in requirements that will work on all (maybe even most?) computers.  If we use,  ``gdal>=1.10.0`` then if you have already correctly installed gdal, it should work.  But if you haven't installed it, then it will likely not work.  I always have to manually update my ``requirements/extra.txt`` file to what it needs to be on each computer I use.

Maybe we should remove ``gdal`` from the requirements file as well.  I took a quick look and couldn't find another project that includes it in its requirements file.  For instance, it isn't include in Fiona:
- https://github.com/Toblerity/Fiona/blob/master/requirements.txt
- https://github.com/Toblerity/Fiona/blob/master/requirements-dev.txt

I found this project:
- https://github.com/nextgis/pygdal

But I am not sure who uses it.  We could try doing something like what Fiona does:
- https://github.com/Toblerity/Fiona/blob/master/setup.py

But that doesn't install the GDAL library.  So I am not sure how much it helps.

@dschult Does it seem reasonable to remove it from the requirements file?

I think that will greatly reduce the number of people running into gdal installation issues.  I suspect most of the people who are running into issues installing it are running into problems b/c they just want to install all optional dependencies without having a specific need for each of them.  So, if we remove it from the automatic install options, hopefully the only people who will run into problems will be people that are specifically trying to use the GDAL functionality.  When they do, they will see that they need to install GDAL.  Moreover, I suspect people who need to use it already have it installed or know how to install it.

If this seems reasonable, I may wait until after 2.5 is released.  Removing it will involve changing our CI systems.  And I have been rethinking how we handle requirement files and our dependencies recently.  There may be a bigger overhaul coming soon.

PR #4160, opened 2020-08-15
Get steiner_tree to work with MultiGraphs by postprocessing

### LINKED ISSUE #3155: steiner_tree should accept MultiGraph
I'm using `steiner_tree` on a road network which may have multiple edges between nodes. 

It looks like `steiner_tree` will fail if passed a `MultiGraph`:
- as a next-to-last step, edges are generated as `(u, v)` tuples pairwise [here](https://github.com/networkx/networkx/blob/master/networkx/algorithms/approximation/steinertree.py#L87)
- before being passed to `G.edge_subgraph` which raises a `ValueError` from `nx.filter.show_multiedges`

This should reproduce the issue:

```python
import networkx as nx
import networkx.algorithms.approximation as nxa

def test_simple_steiner_tree():
    G = nx.Graph()
    G.add_weighted_edges_from([
        (1, 2, 1),
        (2, 3, 1),
        (3, 4, 1),
        (3, 5, 1)
    ])
    terminal_nodes = [2, 4, 5]
    expected_edges = [
        (2, 3),
        (3, 4),
        (3, 5)
    ]
    T = nxa.steiner_tree(G, terminal_nodes)
    assert list(T.edges) == expected_edges

def test_multi_steiner_tree():
    G = nx.MultiGraph()
    G.add_weighted_edges_from([
        (1, 2, 1),
        (2, 3, 1),
        (2, 3, 999),
        (3, 4, 1),
        (3, 5, 1)
    ])
    terminal_nodes = [2, 4, 5]
    expected_edges = [
        (2, 3, 0),
        (3, 4, 0),  # first edge has weight one
        (3, 5, 0)
    ]
    T = nxa.steiner_tree(G, terminal_nodes)

test_simple_steiner_tree()  # passes
test_multi_steiner_tree()  # throws ValueError 
```

The quick fix might be to add `@not_implemented_for('multigraph')`. 

For my current purposes, the following does the trick to handle the `MultiGraph` case:

```python
# get unique links pairwise (u, v)
links = set(chain.from_iterable(
    pairwise(d['path'])
    for u, v, d in mst_edges
))

# for each link in the chain
multi_edges = []
for u, v in links:    
    # consider each edge  between the pair of nodes,
    # keeping track of the one with the minimum weight
    # (there may be a better way - convenience functions/accessors?)
    num_edges = G.number_of_edges(u, v)
    min_k = 0
    min_weight = None
    for k in range(num_edges):
        curr_weight = G.edges[u, v, k][weight]
        if min_weight is None:
            min_weight = curr_weight
        elif curr_weight < min_weight:
            min_weight = curr_weight
            min_k = k
    multi_edges.append((u, v, min_k))

# create subgraph from multi edges - list of (u, v, k)
T = G.edge_subgraph(multi_edges)
```

dschult:
Fixes #3155

### DIFF
--- a/networkx/algorithms/approximation/steinertree.py
+++ b/networkx/algorithms/approximation/steinertree.py
@@ -46,10 +46,22 @@ def metric_closure(G, weight="weight"):
     return M
 
 
-@not_implemented_for("multigraph")
 @not_implemented_for("directed")
 def steiner_tree(G, terminal_nodes, weight="weight"):
     """ Return an approximation to the minimum Steiner tree of a graph.
+
+    The minimum Steiner tree of `G` w.r.t a set of `terminal_nodes`
+    is a tree within `G` that spans those nodes and has minimum size
+    (sum of edge weights) among all such trees.
+
+    The minimum Steiner tree can be approximated by computing the minimum
+    spanning tree of the subgraph of the metric closure of *G* induced by the
+    terminal nodes, where the metric closure of *G* is the complete graph in
+    which each edge is weighted by the shortest path distance between the
+    nodes in *G* .
+    This algorithm produces a tree whose weight is within a (2 - (2 / t))
+    factor of the weight of the optimal Steiner tree where *t* is number of
+    terminal nodes.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


dschult:
This is ready to merge AFAICT

PR #4162, opened 2020-08-15
Prepare for turning chordal_graph_cliques into a generator

jarrodmillman:
See #4004.

### DIFF
--- a/doc/developer/deprecations.rst
+++ b/doc/developer/deprecations.rst
@@ -44,9 +44,10 @@ Make sure to review ``networkx/conftest.py`` after removing deprecated code.
 Version 3.0
 ~~~~~~~~~~~
 
-* In ``networkx/readwrite/gml.py`` remove ``literal_stringizer`` and related tests.
-* In ``networkx/readwrite/gml.py`` remove ``literal_destringizer`` and related tests.
-* In ``networkx/utils/misc.py`` remove ``is_string_like`` and related tests.
-* In ``networkx/utils/misc.py`` remove ``make_str`` and related tests.
-* Remove ``networkx/utils/contextmanagers.py`` and related tests.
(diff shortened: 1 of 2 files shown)


dschult:
I agree that a generator makes the most sense here. 
Code like (from @jarrodmillman in #4004 
```python
for C in (G.subgraph(c).copy() for c in connected_components(G)):
        if C.number_of_nodes() == 1:
            yield frozenset(C.nodes())
        else:
            unnumbered = set(C.nodes())
            v = arbitrary_element(C)
            unnumbered.remove(v)
            numbered = {v}
            clique_wanna_be = {v}
            while unnumbered:
                v = _max_cardinality_node(C, unnumbered, numbered)
                unnumbered.remove(v)
                numbered.add(v)
                new_clique_wanna_be = set(C.neighbors(v)) & numbered
                sg = C.subgraph(clique_wanna_be)
                if _is_complete_graph(sg):
                    new_clique_wanna_be.add(v)
                    if not new_clique_wanna_be >= clique_wanna_be:
                        yield frozenset(clique_wanna_be)
                    clique_wanna_be = new_clique_wanna_be
                else:
                    raise nx.NetworkXError("Input graph is not chordal.")
            yield frozenset(clique_wanna_be)
and delete _connected_chordal_graph_cliques.
```
I believe this also makes the cliques deterministic. The connected components are ordered and while the nodes within each clique are not ordered, the cliques should be yielded in the same order. Nice!!

jarrodmillman:
I already added the code from #4004 to a new internal function ``_chordal_graph_cliques`` and now ``chordal_graph_cliques`` is just the warning and
```
return {c for c in _chordal_graph_cliques(G)}
```
Before 3.0, we just need to replace ``chordal_graph_cliques`` with ``_chordal_graph_cliques``.

@dschult If this looks OK to you, please go ahead and merge.

PR #4182, opened 2020-08-28
Some docstring fixes for draw_networkx_edge_labels() in nx_pylab.py + one typo

alexmalins:
This PR fix some issues with the Parameters section of the docstring for ``draw_networkx_edge_labels()``. Summary of the changes:

- ``clip_on`` was included as docstring parameter, but this isn't actually an argument accepted by the function - entry removed.
- Conversely, ``rotate`` is an acceptable argument, but there was no docstring entry - entry added.
- ``ax`` parameter was included twice - one entry removed.
- Erroneous default for ``font_size`` - fixed.
- Order of docstring parameters changed to match the function arguments order. This improves the readability of the docs.

PR also includes a fix for a typo on line 252.

Some of the other docstrings in nx_pylab.py would benefit from re-aligning the order of the parameters, making explicitly clear which parameters are optional, etc., but I wanted to check that these changes are ok first before venturing any further.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -805,6 +805,7 @@ def draw_networkx_labels(
     horizontalalignment="center",
     verticalalignment="center",
     ax=None,
+    clip_on=True,
 ):
     """Draw node labels on the graph G.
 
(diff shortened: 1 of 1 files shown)


dschult:
Excellent! --- Thanks very much for this.  
I'm not surprised that this needs a close look and your help is much appreciated.

I would say the "clip_on" argument probably should be added with a default of True. We do specify ```clip_on=True``` in our call to ```text()``` and I can imagine a user wanting to set it since it is a feature matplotlib provides.  That change makes this PR no longer only about docs, but I think it is worth crossing the boundary when it improves the package. Putting clip_on as the last parameter doesn't change the API so no deprecations are needed.

I approve these changes

alexmalins:
@dschult - thanks for the comments

Ok, the last commit adds `clip_on` as an argument to both functions where it is used (`draw_networkx_labels()` and `draw_networkx_edge_labels()`), with True as the default.

The other changes are:
- Add `bbox` parameter to the docstring of `draw_networkx_labels()`  (it was missing before), and tried to improve the description of this parameter in both functions.
- Explicitly specify the optional parameters in the docstrings of the two functions.

Unfortunately the last change has created two long docstring lines for `verticalalignment`. Not sure how best to deal with this as moving "optional" to the following line will break the Sphinx rendering of the docstring (I tried using both four and eight indent spaces). I had run black before committing and it didn't complain about these two long lines like @pep8speaks is. I looked at some NumPy source, and it accepts long lines in instances like this (see e.g. L1452 [here](https://github.com/numpy/numpy/blob/master/numpy/core/multiarray.py)). I guess the easy options are either i) delete "optional" for these two cases, or ii) ignore @pep8speaks. Any thoughts?

Finally does anything need to be added to the tests wrt the `clip_on` change? I tested it locally in Jupyter, and supplying True works the same as before (labels clipped). With False labels are not clipped and the axes are resized automatically to show the full graph and labels.

jarrodmillman:
Just run black and ignore pep8speaks.

alexmalins:
@jarrodmillman I did, it leaves everything unchanged.

dschult:
This looks good!  I don't know a good way to test many of the matplotlib parameters unless you can find a programmatic way to look at the image. We can run a "smoketest" where we create an image with both options and if there are no complaints then we assume it is working (no smoke, no fire?).  I'm also fine with leaving ```clip_on``` untested.

I believe we don't have black automatically check doc_strings yet.  I feel like getting sphinx to interpret it correctly is more important than the PEP8speaks flags. 

Thanks very much!  I approve this PR.

PR #4183, opened 2020-08-29
Partitions for quotient graphs

sr-murthy:
Various updates in `algorithms/minors.py`:

* make `equivalence_classes` a public method
* tweak `quotient_graph` to accept partitions as dicts (of block labels as keys and block lists/tuples/sets as values)
* update `quotient_graph` docstring to provide examples to the user of how to use `equivalence_classes` to represent different representations of partitions in a form acceptable to `quotient_graph` (see discussion in https://github.com/networkx/networkx/pull/3839#issuecomment-675217619)

### DIFF
--- a/doc/reference/algorithms/minors.rst
+++ b/doc/reference/algorithms/minors.rst
@@ -9,5 +9,5 @@ Minors
    contracted_edge
    contracted_nodes
    identified_nodes
-   equivalence_relation
+   equivalence_classes
    quotient_graph
--- a/doc/reference/algorithms/minors.rst
+++ b/doc/reference/algorithms/minors.rst
@@ -9,4 +9,5 @@ Minors
    contracted_edge
    contracted_nodes
    identified_nodes
+   equivalence_relation
    quotient_graph
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -21,7 +21,7 @@ chaini = chain.from_iterable
 
 
 def equivalence_classes(iterable, relation):
-    """Returns equivalence classes of `iterable` given by `relation`.
+    """Returns equivalence classes of `relation` when applied to `iterable`.
 
     The equivalence classes, or blocks, consist of objects from `iterable`
     which are all equivalent. They are defined to be equivalent if the
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -9,7 +9,13 @@ from networkx import density
 from networkx.exception import NetworkXException
 from networkx.utils import arbitrary_element
 
-__all__ = ["contracted_edge", "contracted_nodes", "identified_nodes", "quotient_graph"]
+__all__ = [
+    "contracted_edge",
+    "contracted_nodes",
+    "equivalence_classes",
+    "identified_nodes",
+    "quotient_graph"
+]
 
 chaini = chain.from_iterable
 
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -265,6 +265,19 @@ def quotient_graph(
         >>> list(M.edges())
         [(0, 1), (1, 2)]
 
+    Partitions can be represented in various ways:
+    ::
+
+        (0) a list/tuple/set of block lists/tuples/sets
+        (1) a dict with block labels as keys and blocks lists/tuples/sets as values
+        (2) a dict with block lists/tuples/sets as keys and block labels as values
+        (3) a function from nodes in the original iterable to block labels
+        (4) an equivalence relation function on the target iterable
+
+    As `quotient_graph` is designed to accept partitions represented as (0), (1) or
+    (4) only, the `equivalence_classes` function can be used to get the partitions
+    in the right form, in order to call `quotient_graph`.
+
     .. _Strongly connected component: https://en.wikipedia.org/wiki/Strongly_connected_component
 
     References
@@ -285,8 +298,7 @@ def quotient_graph(
         )
 
     # If the partition is a dict, it is assumed to be one where the keys are
-    # user-defined block labels, and values are block lists, i.e. lists of
-    # graph nodes related under the graph equivalence relation
+    # user-defined block labels, and values are block lists, tuples or sets.
     if isinstance(partition, dict):
         partition = [block for block in partition.values()]
 
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -21,17 +21,60 @@ chaini = chain.from_iterable
 
 
 def equivalence_classes(iterable, relation):
-    """Returns the set of equivalence classes of the given `iterable` under
-    the specified equivalence relation.
+    """Returns the set of equivalence classes, or blocks, of the equivalence
+    relation `relation` when applied to the elements of `iterable`.
 
-    `relation` must be a Boolean-valued function that takes two argument. It
-    must represent an equivalence relation (that is, the relation induced by
-    the function must be reflexive, symmetric, and transitive).
+    Parameters
+    ----------
+    iterable : list, tuple, or set
+        An iterable of elements/nodes.
 
-    The return value is a set of sets. It is a partition of the elements of
-    `iterable`; duplicate elements will be ignored so it makes the most sense
-    for `iterable` to be a :class:`set`.
+    relation : function
+        A Boolean-valued function that implements an equivalence relation
+        (reflexive, symmetric, transitive binary relation) on the elements
+        of `iterable` - it must take two elements and return `True` if
+        they are related, or `False` if not.
 
+    Returns
+    -------
+    set of frozensets
+        A set of frozensets representing the partition induced by the equivalence
+        relation function `relation` on the elements of `iterable`. Each
+        member set in the return set represents an equivalence class, or
+        block, of the partition.
+
+        Duplicate elements will be ignored so it makes the most sense for
+        `iterable` to be a :class:`set`.
+
+    Examples
+    --------
+    Let `X` be the ring of integers from `0` to `9`, modulo 3, and consider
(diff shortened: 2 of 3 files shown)


dschult:
You should also add a line in the docs for the "new" function.
```doc/reference/algorithms/minors.rst```

And then the doc_string itself should be updated to standards with
one-line intro, blank line, longer description, sections for Parameters and Returns.

**Comment by sr-murthy** [comment]
@dschult I've updated the docstring for `equivalence_relation` (the docstring update for `quotient_graph` will be a separate commit). I note that when I added `equivalence_relation` in `doc/reference/algorithms/minors.rst` the file change wasn't being detected by Git - this subfolder isn't being Git-ignored, so I don't know why it doesn't show up in staging.

**Comment by sr-murthy** [comment]
OK, my mistake - I was looking at another copy of networkx on my system.

**Comment by sr-murthy** [comment]
Just updated with both changes. I'm working on the docstring changes for `quotient_graph`, will probably update tomorrow.

**Comment by sr-murthy** [comment]
@dschult I'm getting a doctest failure for `minors.py` for the pypy job, because of output format differences. Is there a way to fix this?
```

___________ [doctest] networkx.algorithms.minors.equivalence_classes ___________

056     The equivalence classes of this relation are `{0, 3, 6, 9}`, `{1, 4, 7}`,

057     `{2, 5, 8}`: `0`, `3`, `6`, `9` are all divisible by `3` so leave a zero

058     remainder; `1`, `4`, `7` all leave a remainder of `1`; while `2`, `5`

059     and `8` all leave the remainder `2`. We can see this by calling

060     `equivalence_classes` with `iterable` as X and `relation` as a function

061     implementation of `R`.

062 

063         >>> X = set(range(10))

064         >>> def mod3(x, y): return (x - y) % 3 == 0

065         >>> equivalence_classes(X, mod3)

Expected:

    {frozenset({1, 4, 7}), frozenset({8, 2, 5}), frozenset({0, 9, 3, 6})}

Got:

    {frozenset({0, 3, 6, 9}), frozenset({1, 4, 7}), frozenset({2, 5, 8})}
```

**Comment by dschult** [comment]
One way to avoid the "equivalent, but formatted differently" error in the doctest is to wort the results:
```python
>>> sorted(equivalence_classes(X, mod3))
[frozenset({1, 4, 7}), frozenset({2, 5, 8}), frozenset({0, 3, 6, 9})]

```
It would be nice not to have to convert to a list just for formatting -- it misleads the user to think that we somehow can't work with the sets. But I don't know a better way to arrange the output across platforms. 

I supposed we could create an ```expected``` variable and then print an equality check that says "True". But that looks like a test instead of natural code that someone would write.
```python
>>> expected = {frozenset({1, 4, 7}), frozenset({2, 5, 8}), frozenset({0, 3, 6, 9})}
>>> equivalence_classes(X, mod3) == expected
True

```

Maybe @rossbar has some idea for which is better (or another way)?

sr-murthy:
@dschult I don't see a way around the error, which is still occuring after this change in the docstring

    >>> sorted(equivalence_classes(X, mod3))
    [frozenset({1, 4, 7}), frozenset({2, 5, 8}), frozenset({0, 3, 6, 9})]

The problem here is doctest itself, which should either provide some way of specifying different output formats for different Python platforms, or some other way of asserting equality of outputs. How about skipping doctest for this method?

dschult:
Yes -- put the doc_string back the way it was without sorted and add a comment to skip the test.
If we ever drop pypy from tests we might be able to remove the skip. But this should be fine for this PR.

sr-murthy:
@dschult I think I'm done with the PR - apart from the docstring update in `equivalence_classes`, the only code change in `quotient_graph` is to allow partitions passed in as dicts with block labels as keys and lists/tuples/sets as values. The docstring has been updated to give an example.

sr-murthy:
As I indicated in the `quotient_graph` docstring, there are various ways of representing partitions, namely:

    (0) a list/tuple/set of block lists/tuples/sets
    (1) a dict with block labels as keys and blocks lists/tuples/sets as values
    (2) a dict with block lists/tuples/sets as keys and block labels as values
    (3) a function from nodes in the original iterable to block labels
    (4) an equivalence relation function on the target iterable

Currently, `quotient_graph` is designed to accept partitions in the form (0), (1) or (4). It would seem fairly trivial for the user to figure out how to convert from (2) or (3) into an acceptable form, so I didn't give examples of this. If you think this is necessary, I can do that.

doc/reference/algorithms/minors.rst
```diff
@@ -9,4 +9,5 @@ Minors
    contracted_edge
    contracted_nodes
    identified_nodes
+   equivalence_relation
```
dschult, line 12: I think should be ```equivalence_classes```

networkx/algorithms/minors.py
```diff
@@ -9,22 +9,60 @@
...
+]
 
 chaini = chain.from_iterable
 
 
 def equivalence_classes(iterable, relation):
-    """Returns the set of equivalence classes of the given `iterable` under
-    the specified equivalence relation.
+    """Returns the set of equivalence classes, or blocks, of the equivalence
+    relation `relation` when applied to the elements of `iterable`.
```
dschult, line 25: Sphinx wants a one line, short description followed by a blank line. It then uses this as a description of the function when another page links to this function.  After the blank line you can put all the details and alternate wording. Something like: 
```
"""Returns equivalence classes of `iterable` given by `relation`.

    The equivalence classes, or blocks, consist of objects from `iterable` which
    are all equivalent. They are defined to be equivalent if the `relation` function
    returns `True` when passed any two objects from that class, and `False` otherwise.
   To define an equivalence relation the function must be reflexive, symmetric and transitive.
```

networkx/algorithms/minors.py
```diff
@@ -9,22 +9,60 @@
...
+    The equivalence classes of this relation are `{0, 3, 6, 9}`, `{1, 4, 7}`,
+    `{2, 5, 8}`: `0`, `3`, `6`, `9` are all divisible by `3` and leave zero
+    remainder; `1`, `4`, `7` leave remainder `1`; while `2`, `5` and `8` leave
+    remainder `2`. We can see this by calling `equivalence_classes` with
+    with `X` and a function implementation of `R`.::
+
+        >>> X = set(range(10))
+        >>> def mod3(x, y): return (x - y) % 3 == 0
+        >>> equivalence_classes(X, mod3)    # doctest: +SKIP
+        {frozenset({1, 4, 7}), frozenset({2, 5, 8}), frozenset({0, 3, 6, 9})} # doctest: +SKIP
```
dschult, line 65: Can you put the order of the nodes here back to the order provided by CPython (rather than either sorted or the order produced by pypy). That is the most likely outcome produced...

sr-murthy:
All done, except for the test cases, which I will update later today. I will adapt the existing test cases in `test_minors.py`, to pass in partitions dicts (type (1)).

Note that for the doctest issue, the basic Python 3.7 (3.7.4) shell on my system produces the following output for `equivalence_classes` in the example I gave:
```
Python 3.7.4 (v3.7.4:e09359112e, Jul  8 2019, 14:54:52) 
[Clang 6.0 (clang-600.0.57)] on darwin
>>> from networkx.algorithms.minors import equivalence_classes
>>> equivalence_classes(set(range(10)), lambda x, y: (x - y) %3 == 0)
{frozenset({1, 4, 7}), frozenset({8, 2, 5}), frozenset({0, 9, 3, 6})}
```

networkx/algorithms/minors.py
```diff
@@ -9,22 +9,65 @@
...
+    "quotient_graph"
+]
 
 chaini = chain.from_iterable
 
 
 def equivalence_classes(iterable, relation):
-    """Returns the set of equivalence classes of the given `iterable` under
-    the specified equivalence relation.
+    """Returns equivalence classes of `iterable` given by `relation`.
```
sr-murthy, line 24: It is more correct to say that the equivalence classes are of the relation, not the iterable.

networkx/algorithms/minors.py
dschult: Right -- the output from your Python 3.7 is what I think we should put in the docs. 
I believe these also match with Python 3.6 and 3.8. They don't match with pypy.
The current doc_string values are "sorted", and I think it would be good to have the docs match
precisely what is produced by the mainstream Python versions.

networkx/algorithms/minors.py
dschult: Ahh... good... That's an important grammar that I somehow have missed.  Thanks!

sr-murthy:
@dschult Three test cases have been added to `test_minors.py` - one for each of the three types of partition dicts (dicts of lists, dicts of tuples, dicts of sets). On the doctest output fragment for `equivalence_classes` this was fixed in the previous commit
```
        ...
        >>> equivalence_classes(X, mod3)    # doctest: +SKIP
        {frozenset({1, 4, 7}), frozenset({8, 2, 5}), frozenset({0, 9, 3, 6})} # doctest: +SKIP
```

networkx/algorithms/minors.py
```diff
@@ -9,22 +9,66 @@
...
+]
 
 chaini = chain.from_iterable
 
 
 def equivalence_classes(iterable, relation):
-    """Returns the set of equivalence classes of the given `iterable` under
-    the specified equivalence relation.
+    """Returns equivalence classes of `relation` when applied to the elements
+    of `relation`.
```
dschult, line 25: I'm trying to get this to one line (I know... quotient graph isn't one line either....)
But also, `relation` shouldn't appear twice.

```
"""Returns equivalence classes of `relation` applied to elements of `iterable`

```

sr-murthy: Sorry, this was a typo - it's in one line now.

Issue #4188, opened 2020-09-04
regression: nx 2.5 doesn't support graphml attribute type "long" anymore


josch:
Take this graphml file:

```
<?xml version='1.0' encoding='utf-8'?>
<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
  <key attr.name="cudfversion" attr.type="long" for="node" id="d6" />
  <graph edgedefault="directed">
  </graph>
  </graphml>
```

And this Python code:

    python3 -c "import networkx; networkx.read_graphml('test.xml')"

This worked with networkx 2.4 but with 2.5 I'm getting:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<decorator-gen-692>", line 2, in read_graphml
  File "/tmp/networkx/networkx/utils/decorators.py", line 239, in _open_file
    result = func_to_be_decorated(*new_args, **kwargs)
  File "/tmp/networkx/networkx/readwrite/graphml.py", line 258, in read_graphml
    glist = list(reader(path=path))
  File "/tmp/networkx/networkx/readwrite/graphml.py", line 775, in __call__
    (keys, defaults) = self.find_graphml_keys(self.xml)
  File "/tmp/networkx/networkx/readwrite/graphml.py", line 951, in find_graphml_keys
    "type": self.python_type[attr_type],
KeyError: 'long'
```

The commit that broke this functionality is 857aa81ead611dd05c7445b53bd128131033d50e.

Please consider restoring this. There exists software that produces graphs with attribute type "long" which I cannot read in anymore since the release of 2.5.

dschult:
I see what you mean.  The word "long" used as an attribute type should be translated as an int.

Thank You!!

PR #4189, opened 2020-09-04
graphml: re-add graph attribute type 'long' after 857aa81 removed it

josch:
closes: #4188


jarrodmillman:
Thanks for this.  Would you mind adding a test so that we don't break this again in the future here:
https://github.com/networkx/networkx/blob/master/networkx/readwrite/tests/test_graphml.py

dschult (approved):

PR #4201, opened 2020-09-10
Test mac osx via actions

jarrodmillman:
This moves our testing to GH actions from Travis.  Travis was running too much.  The mac tests were particularly slow (sometimes taking over 30 minutes just to run 3.7 with dependencies).  This will make are CI system faster and also slightly reduce the lines of code.


rossbar (approved):
LGTM, thanks @jarrodmillman !

PR #4206, opened 2020-09-13
Fix UnionFind.union to actually select the heaviest root as the new root

jbradberry:
Previous versions in this area of code used `max()` to select the root of the heaviest set, but a fix for a different issue changed this to pull the first element from the sorted set of roots, i.e. the _lightest_ root, not the heaviest.


dschult:
Thanks for this -- especially the test. :}  

The change in behavior was released in NetworkX v2.5 and will be fixed in v2.6.

dschult (approved):

rossbar (approved):
LGTM, thanks @jbradberry !

PR #4221, opened 2020-09-24
Add max common induced subgraph

jamestrimble:
This PR adds functions for the max common (connected) induced subgraph problem.

I created a new module for these within `nx.algorithms`. Was this the right thing to do? There already exists one max common induced subgraph implementation in `networkx.algorithms.isomorphism.ISMAGS.largest_common_subgraph`. I added a small function wrapping this to the new module.

### DIFF
--- a/doc/reference/algorithms/index.rst
+++ b/doc/reference/algorithms/index.rst
@@ -48,6 +48,7 @@ Algorithms
    link_prediction
    lowest_common_ancestors
    matching
+   maximum_common_subgraph
    minors
    mis
    non_randomness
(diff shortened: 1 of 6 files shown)


Erotemic:
This is somewhat related to a PR I'm working on #4169, which solves a related problem: `maximum_common_ordered_tree_embedding` restricted to ordered directed trees in roughly O(n^4) time. Once that is merged I may work on the `maximum_common_ordered_subtree` variant, which would return subtrees that are induced isomorphisms in addition to being embeddings.

Having a more general version of what I've been working on is definitely something I'm interested in. I'm not a core dev, but this looked interesting the point where I wanted to comment. 

If anyone's interested in my 2cents: would a better location for this be inside the networkx.algorithms.isomorphism subpackage rather than a new algorithm subpackage?

jamestrimble:
Thanks @Erotemic. I had thought that the functions might be easier to find in a new subpackage. (Perhaps common_subgraph would be better?) But the isomorphism subpackage would make sense too. I’m happy to leave this decision to the core maintainers.

PR #4222, opened 2020-09-28
fix "see also" links in json_graph.tree

wimglenn:
one was duplicate and one was self-referential.  https://networkx.github.io/documentation/stable/reference/readwrite/generated/networkx.readwrite.json_graph.tree_data.html


dschult:
Thanks!

Issue #4213, opened 2020-09-16
Deprecation warning with flatten


k2shah:
Since 2.5, when `flatten` calls `is_string_like()` it results in `DeprecationWarning`.

https://github.com/networkx/networkx/blob/dae8b01782ca4fada704d11a1d16c60663f8a90d/networkx/utils/misc.py#L54-L65

`DeprecationWarning`  is from 
https://github.com/networkx/networkx/blob/dae8b01782ca4fada704d11a1d16c60663f8a90d/networkx/utils/misc.py#L28-L35

I'm happy to update `is_string_like(obj)` -> `isinstance(obj, str)` in a PR.

jarrodmillman:
Please do.  Thanks!

k2shah:
@jarrodmillman i made the PR

k2shah:
closed by #4223

PR #4223, opened 2020-09-28
changed is_string_like to isinstance

k2shah:
fixes issue #4213


rossbar (approved):
I'm +1 on the change as a quick fix but while reviewing this I noticed that
 1) `flatten` is untested
 2) It is only called in one place internally
 3) It may fall under the category of "helper functions that aren't actually that helpful"

Therefore IMO it's a strong candidate for deprecation. Thanks for finding/fixing this @k2shah !

PR #4235, opened 2020-10-05
Check formatting of PRs via black

jarrodmillman:
@dschult , @rossbar This replaces ``pep8speaks`` with just checking that black doesn't need to reformat the code.  If it reformats the code the check fails and the changes are shown in the failed action.  If the linting test fails, the PR author will need to run black on their local branch, commit the changes, and push to their remote branch.

I also updated to the newest release of ``black`` and ``pre-commit``.

There are several files that no longer conform to the black defaults.  So the lint test is failing.  I will fix that before this is merged (i.e., once folks have seen what this looks like when it fails).

Hopefully, this will be the end of us accidentally checking in code that isn't formatted by black using the default settings.


rossbar (approved):
LGTM - I like this approach better than the pep8 bot as it's a little less noisy. One thing that would make it nicer is if the diffs from the failed linting action were highlighted.

jarrodmillman:
@rossbar The diffs from failed linting are now hightlighted using red/green.

dschult:
This seems good -- I like that it is obvious to us that its got mis-styled
code. It's too bad that the user still has to figure out 1) that not running
black is why it failed, and 2) how to run black so that it will pass.
Can we find a way that the CI runs black and makes a commit in the PR?
Is there a reason (like black changes we might not want?) we shouldn't do that?

A more helpful message (like: use black) would be nice too. Many may not know the reference to "lint".  Can we make the name of this test: use black for formatting

jarrodmillman:
@dschult I can set it up so that a commit is added to the PR.  I am not sure whether we should do that or not.  It may be confusing to some contributors that their remote branch has commits that they don't have.

Ill change the name tomorrow.'

dschult:
Yes -- I see the problem. The contributor would have to pull from their repo even though they didn't make the changes... :{   that would be confusing.   I'm just trying to find a way that "style" isn't a hurdle to contributing or reviewing code. We should be thinking about the code and the API -- not the code style.

Maybe we could forget about style/black for merging PRs, but once a month(?) run black on the whole repo?  Or, maybe forget about style and add running black to the release process?  In this case, we should find a tool to make the doc_strings styled nicely too... run them both before each release.  I'm just brainstorming here... don't feel pressure to implement any of this.  But it'd be nice to not think about style so much.

jarrodmillman:
+1 on not having to think about style so much!  We should definitely keep brainstorming how to minimize the time we spend on formatting issues.

My concern (at this point) is that if we take care of formatting after merging the PRs, we will introduce a lot of unnecessary churn on master and in the history.  I would prefer telling users who have difficulty running the black command to just ignore the "Lint" (name to be determined) failure and then we can run black before merging.  I suspect most people will be able to run black themselves.  We may need to tell them to do so.  We can also push the changes to their branch while they are still working on the PR.  But then we will at least be able to warn contributors what we are doing and instruct them to pull the new commit to their local branch.

My inclination at this point is to merge this and see how it goes.  This should be less noisy than pep8speaks and will prevent us from merging code that doesn't conform to black (which we have been doing w/ pep8speaks).  If contributors are having difficulty using black, we can revisit this and see if there is a better way to handle it.  Perhaps, we should make a point to revisit this after a month of using the new system to see how it is going.

(I would also like to eventually add additional linting.  That is why I used pre-commit to run black instead of running it manually.  If we add additional linting checks to ``.pre-commit-config.yaml``, then the linting tests will automatically include them.  So we may need to tell users to make additional changes depending on what future linting tests we add.  At this point, we aren't committing to do that.  I am just pointing out that this can be expanded in the future if it works well and reduces the amount of time we require contributors to spend on formatting/style issues.)

dschult:
Yes, I think it is worth trying this system and seeing how it goes. 
I approve this PR.

PR #3856, opened 2020-03-06
adds dedensification and densification functions

dpfens:
Adds dedensification (`networkx.algorithms.dedensification.dedensify`) and densification (`networkx.algorithms.dedensification.densify`) functions as defined in [Scalable Pattern Matching over Compressed Graphs via Dedensification](http://www.cs.umd.edu/~abadi/papers/graph-dedense.pdf) with option for `expansive` or `nonexpansive` results, and option for `inplace` dedensification or densification.

Examples of how functions can be called:
```python
import networkx as nx

graph = nx.DiGraph()
# build structure of graph by adding nodes and edges

dedensified_graph, compression_nodes = nx.algorithms.dedensification.dedensify(graph, threshold=10, prefix='compression-')

# perform business operations on dedensified graph

# convert dedensified graph back into original graph
densified_graph = nx.algorithms.dedensification.densify(dedensified_graph, compression_nodes)
```

PR #4240, opened 2020-10-06
adds dedensification function

dpfens:
**NOTE**:  This change was originally found in pull request #3856, but Github Actions wouldn't work anymore due to changes to the `.github/workflow` files since my initial fork, so I re-forked and re-implemented the changes.

Original content:

Adds dedensification (`networkx.algorithms.dedensification.dedensify`) and densification (`networkx.algorithms.dedensification.densify`) functions as defined in [Scalable Pattern Matching over Compressed Graphs via Dedensification](http://www.cs.umd.edu/~abadi/papers/graph-dedense.pdf) with option for `expansive` or `nonexpansive` results, and option for `inplace` dedensification or densification.

Examples of how functions can be called:
```python
import networkx as nx

graph = nx.DiGraph()
# build structure of graph by adding nodes and edges

dedensified_graph, compression_nodes = nx.algorithms.dedensification.dedensify(graph, threshold=10, prefix='compression-')

# perform business operations on dedensified graph

# convert dedensified graph back into original graph
densified_graph = nx.algorithms.dedensification.densify(dedensified_graph, compression_nodes)
```

### DIFF
--- a/doc/reference/algorithms/index.rst
+++ b/doc/reference/algorithms/index.rst
@@ -66,6 +66,7 @@ Algorithms
    smetric
    sparsifiers
    structuralholes
+   summarization
    swap
    threshold
    tournament
--- a/examples/algorithms/plot_dedensification.py
+++ b/examples/algorithms/plot_dedensification.py
@@ -1,0 +1,93 @@
+"""
+=============
+Dedensification
+=============
+
+Examples of dedensification of a graph.  Dedensification retains the structural
+pattern of the original graph and will only add compressor nodes when doing so
+would result in fewer edges in the compressed graph.
+"""
+import networkx as nx
+from networkx.algorithms import summarization
+import matplotlib.pyplot as plt
+
+plt.suptitle("Dedensification")
+
+original_graph = nx.DiGraph()
+white_nodes = ["1", "2", "3", "4", "5", "6"]
+red_nodes = ["A", "B", "C"]
+node_sizes = [250 for node in white_nodes + red_nodes]
+node_colors = ["white" for n in white_nodes] + ["red" for n in red_nodes]
+
+original_graph.add_nodes_from(white_nodes + red_nodes)
+original_graph.add_edges_from(
+    [
+        ("1", "C"),
+        ("1", "B"),
+        ("2", "C"),
+        ("2", "B"),
+        ("2", "A"),
+        ("3", "B"),
+        ("3", "A"),
+        ("3", "6"),
+        ("4", "C"),
+        ("4", "B"),
+        ("4", "A"),
+        ("5", "B"),
+        ("5", "A"),
+        ("6", "5"),
+        ("A", "6"),
+    ]
+)
+base_options = dict(with_labels=True, edgecolors="black")
+pos = {
+    "3": (0, 1),
+    "2": (0, 2),
+    "1": (0, 3),
+    "6": (1, 0),
+    "A": (1, 1),
+    "B": (1, 2),
+    "C": (1, 3),
+    "4": (2, 3),
+    "5": (2, 1),
+}
+ax1 = plt.subplot(1, 2, 1)
+plt.title("Original (%s edges)" % original_graph.number_of_edges())
+nx.draw_networkx(original_graph, pos=pos, node_color=node_colors, **base_options)
+
+nonexp_graph, compression_nodes = summarization.dedensify(
+    original_graph, threshold=2, copy=False
+)
+nonexp_node_colors = list(node_colors)
+nonexp_node_sizes = list(node_sizes)
+for node in compression_nodes:
+    nonexp_node_colors.append("yellow")
+    nonexp_node_sizes.append(600)
+plt.subplot(1, 2, 2)
+
+plt.title("Dedensified (%s edges)" % nonexp_graph.number_of_edges())
+nonexp_pos = {
+    "5": (0, 0),
+    "B": (0, 2),
+    "1": (0, 3),
+    "6": (1, 0.75),
+    "3": (1.5, 1.5),
+    "A": (2, 0),
+    "C": (2, 3),
+    "4": (3, 1.5),
+    "2": (3, 2.5),
+}
+c_nodes = list(compression_nodes)
+c_nodes.sort()
+for spot, node in enumerate(c_nodes):
+    nonexp_pos[node] = (2, spot + 2)
+nx.draw_networkx(
+    nonexp_graph,
+    pos=nonexp_pos,
+    node_color=nonexp_node_colors,
+    node_size=nonexp_node_sizes,
+    **base_options
+)
+
+plt.tight_layout()
+plt.show()
--- a/networkx/algorithms/summarization.py
+++ b/networkx/algorithms/summarization.py
@@ -1,0 +1,215 @@
+"""
+Graph summarization finds smaller representations of graphs resulting in faster
+runtime of algorithms, reduced storage needs, and noise reduction.
+Summarization has applications in areas such as visualization, pattern mining,
+clustering and community detection, and more.  Core graph summarization
+techniques are grouping/aggregation, bit-compression,
+simplification/sparsification, and influence based. Graph summarization
+algorithms often produce either summary graphs in the form of supergraphs or
+sparsified graphs, or a list of independent structures. Supergraphs are the
+most common product, which consist of supernodes and original nodes and are
+connected by edges and superedges, which represent aggregate edges between
+nodes and supernodes.
+
+Grouping/aggregation based techniques compress graphs by representing
+close/connected nodes and edges in a graph by a single node/edge in a
+supergraph. Nodes can be grouped together into supernodes based on their
+structural similarities or proximity within a graph to reduce the total number
+of nodes in a graph. Edge-grouping techniques group edges into lossy/lossless
+nodes called compressor or virtual nodes to reduce the total number of edges in
+a graph. Edge-grouping techniques can be lossless, meaning that they can be
+used to re-create the original graph, or techniques can be lossy, requiring
+less space to store the summary graph, but at the expense of lower
+recontruction accuracy of the original graph.
+
+Bit-compression techniques minimize the amount of information needed to
+describe the original graph, while revealing structural patterns in the
+original graph.  The two-part minimum description length (MDL) is often used to
+represent the model and the original graph in terms of the model.  A key
+difference between graph compression and graph summarization is that graph
+summarization focuses on finding structural patterns within the original graph,
+whereas graph compression focuses on compressions the original graph to be as
+small as possible.  **NOTE**: Some bit-compression methods exist solely to
+compress a graph without creating a summary graph or finding comprehensible
+structural patterns.
+
+Simplification/Sparsification techniques attempt to create a sparse
+representation of a graph by removing unimportant nodes and edges from the
+graph.  Sparsified graphs differ from supergraphs created by
+grouping/aggregation by only containing a subset of the original nodes and
+edges of the original graph.
+
+Influence based techniques aim to find a high-level description of influence
+propagation in a large graph.  These methods are scarce and have been mostly
+applied to social graphs.
+
+*dedensification* is a grouping/aggregation based technique to compress the
+neighborhoods around high-degree nodes in unweighted graphs by adding
+compressor nodes that summarize multiple edges of the same type to
+high-degree nodes (nodes with a degree greater than a given threshold).
+Dedensification was developed for the purpose of increasing performance of
+query processing around high-degree nodes in graph databases and enables direct
+operations on the compressed graph.  The structural patterns surrounding
+high-degree nodes in the original is preserved while using fewer edges and
+adding a small number of compressor nodes.  The degree of nodes present in the
+original graph is also preserved. The current implementation of dedensification
+supports graphs with one edge type.
+
+For more information on graph summarization, see `Graph Summarization Methods
+and Applications: A Survey <https://dl.acm.org/doi/abs/10.1145/3186727>`_
+"""
+import networkx as nx
+
+__all__ = [
+    "dedensify",
+]
+
+
+def dedensify(G, threshold, prefix=None, copy=True):
+    """Compresses neighborhoods around high-degree nodes
+
+    Reduces the number of edges to high-degree nodes by adding compressor nodes
+    that summarize multiple edges of the same type to high-degree nodes (nodes
+    with a degree greater than a given threshold).  Dedensification also has
+    the added benefit of reducing the number of edges around high-degree nodes.
+    The implementation currently supports graphs with a single edge type.
+
+    Parameters
+    ----------
+    G: graph
+       A networkx graph
+    threshold: int
+       Minimum degree threshold of a node to be considered a high degree node.
+       The threshold must be greater than or equal to 2.
+    prefix: str or None, optional (default: None)
+       An optional prefix for denoting compressor nodes
+    copy: bool, optional (default: True)
+       Indicates if dedensification should be done inplace
+
+    Returns
+    -------
+    dedensified networkx graph : (graph, set)
+        2-tuple of the dedensified graph and set of compressor nodes
+
+    Notes
+    -----
+    According to the algorithm in [1]_, removes edges in a graph by
+    compressing/decompressing the neighborhoods around high degree nodes by
+    adding compressor nodes that summarize multiple edges of the same type
+    to high-degree nodes.  Dedensification will only add a compressor node when
+    doing so will reduce the total number of edges in the given graph. This
+    implementation currently supports graphs with a single edge type.
+
+    Examples
+    --------
+    Dedensification will only add compressor nodes when doing so would result
+    in fewer edges::
+
+        >>> original_graph = nx.DiGraph()
+        >>> original_graph.add_nodes_from(
+        ...     ["1", "2", "3", "4", "5", "6", "A", "B", "C"]
+        ... )
+        >>> original_graph.add_edges_from(
+        ...     [
+        ...         ("1", "C"), ("1", "B"),
+        ...         ("2", "C"), ("2", "B"), ("2", "A"),
+        ...         ("3", "B"), ("3", "A"), ("3", "6"),
+        ...         ("4", "C"), ("4", "B"), ("4", "A"),
+        ...         ("5", "B"), ("5", "A"),
+        ...         ("6", "5"),
+        ...         ("A", "6")
+        ...     ]
+        ... )
+        >>> c_graph, c_nodes = nx.dedensify(original_graph, threshold=2)
+        >>> original_graph.number_of_edges()
+        15
+        >>> c_graph.number_of_edges()
+        14
+
+    A dedensified, directed graph can be "densified" to reconstruct the
+    original graph::
+
+        >>> original_graph = nx.DiGraph()
+        >>> original_graph.add_nodes_from(
+        ...     ["1", "2", "3", "4", "5", "6", "A", "B", "C"]
+        ... )
+        >>> original_graph.add_edges_from(
+        ...     [
+        ...         ("1", "C"), ("1", "B"),
+        ...         ("2", "C"), ("2", "B"), ("2", "A"),
+        ...         ("3", "B"), ("3", "A"), ("3", "6"),
+        ...         ("4", "C"), ("4", "B"), ("4", "A"),
+        ...         ("5", "B"), ("5", "A"),
+        ...         ("6", "5"),
+        ...         ("A", "6")
+        ...     ]
+        ... )
+        >>> c_graph, c_nodes = nx.dedensify(original_graph, threshold=2)
+        >>> # re-densifies the compressed graph into the original graph
+        >>> for c_node in c_nodes:
+        ...     all_neighbors = set(nx.all_neighbors(c_graph, c_node))
+        ...     out_neighbors = set(c_graph.neighbors(c_node))
+        ...     for out_neighbor in out_neighbors:
+        ...         c_graph.remove_edge(c_node, out_neighbor)
+        ...     in_neighbors = all_neighbors - out_neighbors
+        ...     for in_neighbor in in_neighbors:
+        ...         c_graph.remove_edge(in_neighbor, c_node)
+        ...         for out_neighbor in out_neighbors:
+        ...             c_graph.add_edge(in_neighbor, out_neighbor)
+        ...     c_graph.remove_node(c_node)
+        ...
+        >>> nx.is_isomorphic(original_graph, c_graph)
+        True
+
+    References
+    ----------
+    .. [1] Maccioni, A., & Abadi, D. J. (2016, August).
+       Scalable pattern matching over compressed graphs via dedensification.
+       In Proceedings of the 22nd ACM SIGKDD International Conference on
+       Knowledge Discovery and Data Mining (pp. 1755-1764).
+       http://www.cs.umd.edu/~abadi/papers/graph-dedense.pdf
+    """
+    if threshold < 2:
+        raise nx.NetworkXError("The degree threshold must be >= 2")
+
+    high_degree_nodes = set()
+    low_degree_nodes = set()
+
+    degrees = G.in_degree if G.is_directed() else G.degree
+    # Group nodes based on degree threshold
+    high_degree_nodes = set([n for n, d in degrees if d > threshold])
+    low_degree_nodes = G.nodes() - high_degree_nodes
+
+    auxillary = {}
+    for node in G:
+        high_degree_neighbors = frozenset(high_degree_nodes & set(G[node]))
+        if high_degree_neighbors:
+            if high_degree_neighbors in auxillary:
+                auxillary[high_degree_neighbors].add(node)
+            else:
+                auxillary[high_degree_neighbors] = {node}
+
+    if copy:
+        G = G.copy()
+
+    compressor_nodes = set()
+    for index, (high_degree_nodes, low_degree_nodes) in enumerate(auxillary.items()):
+        low_degree_node_count = len(low_degree_nodes)
+        high_degree_node_count = len(high_degree_nodes)
+        old_edges = high_degree_node_count * low_degree_node_count
+        new_edges = high_degree_node_count + low_degree_node_count
+        if old_edges <= new_edges:
+            continue
+        compression_node = "".join(str(node) for node in high_degree_nodes)
+        if prefix:
+            compression_node = str(prefix) + compression_node
+        for node in low_degree_nodes:
+            for high_node in high_degree_nodes:
+                if G.has_edge(node, high_node):
+                    G.remove_edge(node, high_node)
+
+            G.add_edge(node, compression_node)
+        for node in high_degree_nodes:
+            G.add_edge(compression_node, node)
+        compressor_nodes.add(compression_node)
+    return G, compressor_nodes
--- a/networkx/algorithms/summarization.py
+++ b/networkx/algorithms/summarization.py
@@ -172,9 +172,6 @@ def dedensify(G, threshold, prefix=None, copy=True):
     if threshold < 2:
         raise nx.NetworkXError("The degree threshold must be >= 2")
 
-    high_degree_nodes = set()
-    low_degree_nodes = set()
-
     degrees = G.in_degree if G.is_directed() else G.degree
     # Group nodes based on degree threshold
     high_degree_nodes = set([n for n, d in degrees if d > threshold])
--- a/doc/reference/algorithms/summarization.rst
+++ b/doc/reference/algorithms/summarization.rst
@@ -1,0 +1,9 @@
+*************
+Summarization
+*************
+
+.. automodule:: networkx.algorithms.summarization
+.. autosummary::
+   :toctree: generated/
+
+   dedensify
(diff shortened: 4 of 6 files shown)


dschult:
Sorry that you had to refork and copy the changes over.
And I don't know why I didn't nogtice this before, but you should also add your functions to ```networkx.algorithms.__init__.py``` so that they can be called using e.g. ```nx.dedensify()```

Also, the top comment of the main module says that these function remove edges from a graph. Doesn't densify add edges to the graph? Is there a better way to describe what these functions could be used for?

Thanks!

rossbar:
> Also, the top comment of the main module says that these function remove edges from a graph. Doesn't densify add edges to the graph? Is there a better way to describe what these functions could be used for?

Related to this, it struck me that perhaps the module name should be `densification` instead of `dedensification`? It makes more sense to me to have the module name after the "forward" procedure rather than the "reverse" (in the same way that `if positive` is usually preferable to `if not negative`). I'm not too familiar with the terminology though, so if "dedensification" is the more prevalent term in network science then please ignore this comment!

dpfens:
@dschult My fault, I should have caught that, and I should have done a much better job describing the overall functionality of the module.  I just made another attempt at describing the contents of the module.

@rossbar Good idea, I hadn't thought about that.  A [glance](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C47&q=densification+graph+theory&btnG=) at Google Scholar suggests that `densification` is the more prevalent term and has been around longer, and your point about using the "forward" procedure seems more intuitive to me.  I just changed the module name to `densification`.

Also, if anyone is interested, pages 4-6 of [this PDF](https://web.eecs.umich.edu/~dkoutra/tut/18-icdm-part1.pdf#page=4) give a pretty quick intuition of dedensification.

jarrodmillman:
I am not sure about the terminology here.  But I suspect ``densification`` means something else.  This is more like a compression method.  So sparsification would have been better maybe, but it was suspect it was already taken.

Sparsification involves sampling edges to reduce the graph, but in a way that preserves certain properties.  Dedensification involves grouping edges around high degree nodes to reduce the number of edges.  I suspect densification involves dynamic graphs and involves the evolution of edge density.  I am not sure there should be a densification function in here.  How would people use it?  The reason for dedensification is to reduce a big graph to run some algorithms on it.  The densification just reconstructs the original graph, I think.  Normally you wouldn't delete the original graph.  The original large graph would be your input data normally.

jarrodmillman:
It would be good to add an example to the gallery.  Could you reproduce the figure?

Something like this:
https://github.com/networkx/networkx/blob/master/examples/algorithms/plot_decomposition.py

Except you would create the original graph, call dedesification on it, then plot it using nx and plt.

dpfens:
@jarrodmillman  I don't know for sure about the actual definition of `densification` either.  According to some research papers I just found, the definitions of graph densification is fairly consistent:

*  [Graph Densification](https://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.225.4341):
    > Given a graph G the goal of graph densification is to come up with another graph H that has significantly more edges than G but nevertheless approximates G well with respect to some set of test functions

*  [On the Interplay between Strong Regularity and Graph Densification](https://arxiv.org/pdf/1703.07107.pdf#page=6) - first paragraph of the 6th page
    > *graph densification* aims to significantly increase the number of edges in G while preserving its properties as much as possible

I think the `densify` function meets that criteria, but like you said, it may not be as useful.  When I wrote it my thought was that someone may `dedensify` a really big graph prior to saving it to persistent storage to save space/money.  Later they might read the compressed graph from persistent storage and `densify` it when running algorithms that require the original graph.  That said, `dedensify` is only ~15 lines of code, so it would be fairly easy for someone to implement if they needed it.

If I change the module to `summarization`, would you want it structured to have submodules (making it a folder with an `__init__.py` file) to account for future additions to it?

I can reproduce the original graph(s) from the papers. Thanks for including the example and for offering to help.  I've done very little in `networkx` with visualizations so I might need the help.

jarrodmillman:
Let's see what Dan and Ross think about the name and structure.

You could include the code to un-dedensify in a doc example.  Create original graph, a dedensified graph, and then you could show that un-dedensify the dedensified graph reconstructs the original graph.

For the visualization part, you may find it easiest to just look at the gallery:
https://networkx.github.io/documentation/latest/auto_examples/index.html

You can click on any of the examples to see the code that generated the figure.  Some of the examples will have parts of what you want to draw.  I often use the examples to cobble together new plots.

jarrodmillman:
For style and formatting, we use:
https://github.com/psf/black/

There are instructions for setting up pre-commit to automatically run black for you:
https://networkx.github.io/documentation/latest/developer/contribute.html

Basically, you need to do the following in your virtual environment:
```
pre-commit install
```

Then when you type ``git commit``, pre-commit will run black for you.

You can do it manually as well:
```
pip install black==20.8b1
black examples/algorithms/plot_dedensification.py
```

doc/reference/algorithms/index.rst
```diff
@@ -31,6 +31,7 @@ Algorithms
    cuts
    d_separation
    dag
+   densification
```
jarrodmillman, line 34: You probably want summarization here.

dpfens:
I just created the example script and changes the module to `summarization`, here is the visualization it currently produces
dedensification_example

I attempted to replicate the format of the original and expansive graphs (left and center figures) in the paper as closely as possible using the `multipartite_layout` as you suggested, with the exception of the non-expansive graph on the right, as the layout in the paper did not resemble a multipartite graph.

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,187 @@
+"""
+Algorithms for reducing the number of nodes/edges in a graph
```
jarrodmillman, line 2: If you have the energy and time, it would nice to add a bit more context here about graph summarization.  

Maybe a few sentences about graph summarization methods and applications w/ a link to:
https://dl.acm.org/doi/abs/10.1145/3186727

I would also add something about the core techniques:

    grouping methods: node and edge based
    bit compression methods: lossy and lossless
    sparsification methods
    influence based methods

And dedensification note where dedensification falls in this categorization.

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,187 @@
...
+    When expansiveness is not used, dedensification will only be applied when
+    it would reduce the total number of edges in the given graph.
+
+    Examples
+    --------
+    Non-expansive dedensification will only add compressor nodes when doing so
+    would result in fewer edges.  Nonexpansive is the default for
+    ```expansive``::
+
+        >>> import networkx as nx
```
jarrodmillman, line 67: Please remove all the
```
>>> import networkx as nx
```

That is always assumed in our docstring examples.

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,187 @@
...
+    When expansiveness is used, dedensification is performed on all sets of
+    low/high degree nodes, even when it will not reduce the number of edges.
+
+    When expansiveness is not used, dedensification will only be applied when
+    it would reduce the total number of edges in the given graph.
+
+    Examples
+    --------
+    Non-expansive dedensification will only add compressor nodes when doing so
+    would result in fewer edges.  Nonexpansive is the default for
```
jarrodmillman, line 64: Is nonexpansive really the default for ``expansive``?

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,187 @@
...
+        >>> original_graph.number_of_edges()
+        15
+        >>> c_graph.number_of_edges()
+        14
+
+    A dedensified, directed graph can be "densified" to reconstruct the
+    original graph::
+
+        >>> import networkx as nx
+        >>> original_graph = nx.DiGraph()
```
jarrodmillman, line 93: I would do something like this:
```
        >>> original_graph = nx.DiGraph()
        >>> original_graph.add_nodes_from(
        ...     ["1", "2", "3", "4", "5", "6", "A", "B", "C"]
        ... )
        >>> original_graph.add_edges_from(
        ...     [
        ...         ("1", "C"), ("1", "B"),
        ...         ("2", "C"), ("2", "B"), ("2", "A"),
        ...         ("3", "B"), ("3", "A"), ("3", "6"),
        ...         ("4", "C"), ("4", "B"), ("4", "A"),
        ...         ("5", "B"), ("5", "A"),
        ...         ("6", "5"),
        ...         ("A", "6")
        ...     ]
        ... )
        >>> c_graph, c_nodes = nx.dedensify(original_graph)
        >>> for c_node in c_nodes:
        ...         all_neighbors = set(nx.all_neighbors(c_graph, c_node))
        ...         out_neighbors = set(c_graph.neighbors(c_node))
        ...         for out_neighbor in out_neighbors:
        ...             c_graph.remove_edge(c_node, out_neighbor)
        ...         in_neighbors = all_neighbors - out_neighbors
        ...         for in_neighbor in in_neighbors:
        ...             c_graph.remove_edge(in_neighbor, c_node)
        ...             for out_neighbor in out_neighbors:
        ...                 c_graph.add_edge(in_neighbor, out_neighbor)
        ...         c_graph.remove_node(c_node)
        ...
        >>> nx.is_isomorphic(original_graph, c_graph)
        True
```

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,187 @@
+"""
+Algorithms for reducing the number of nodes/edges in a graph
+
+*dedensification* reduces the number of edges in unweighted graphs by adding
```
jarrodmillman, line 4: Could you say something about what properties this graph summary preserves.  It obviously changes the number of nodes and edges.  It changes the node degrees.  It also changes things like the number of distinct edges paths.  What does it preserve?  What types of algorithms can you still run on it (and get an answer that tells you something about the original graph)?

networkx/algorithms/summarization.py
dpfens: Yes, `expansive` currently defaults to `False`.  If the function is executed with `expansive=True`, the function will add a compressor node to the graph even when doing so will not result in fewer edges in the graph.  When `expansive=True`, the function only adds a compressor node when it will reduce the number of edges in the graph, which I think would be the most generally desired outcome [the paper covers this at the end of page 3, and the top of page 4](https://www.cs.umd.edu/~abadi/papers/graph-dedense.pdf#page=3).

I chose to include the option of running the algorithm as `expansive`, as the expansive/non-expansive approaches can change how graph pattern matching algorithms need to be implemented and would need to be accounted for if/when those pattern-matching queries are added to `networkx` in the future.  The implementation differences are referenced at [the top of page 4 of the paper](https://www.cs.umd.edu/~abadi/papers/graph-dedense.pdf#page=4).

That said, should the default value be changed to `expansive=True`?

networkx/algorithms/summarization.py
dpfens: Makes sense. Done

networkx/algorithms/summarization.py
dpfens: sorry.  Done

networkx/algorithms/summarization.py
dpfens: Yes, definitely. I'll do this shortly.

networkx/algorithms/summarization.py
dpfens: Yes, I will add this.

networkx/algorithms/summarization.py
dpfens: Sorry for the delay.  I have made a first attempt at giving a summary of graph summarization

networkx/algorithms/summarization.py
dpfens: I have given this a first attempt.  I was not sure how detailed to get with it, so let me know what needs to be changed

examples/algorithms/plot_dedensification.py
```diff
@@ -0,0 +1,102 @@
...
+    ]
+)
+
+layers = (("1", "2", "3"), ("C", "B", "A", "6"), ("4", "5"))
+for index, layer in enumerate(layers):
+    for node in layer:
+        original_graph.nodes[node]["layer"] = index
+
+base_options = dict(with_labels=True, edgecolors="black")
+pos = nx.multipartite_layout(original_graph, subset_key="layer")
```
dschult, line 53: The positions might look better if we're more prescriptive on the position. These are integer locations, but might be good enough.
```suggestion
# position nodes in a 3x3 grid 
lines = (("1", "2", "3"), ("A", "B", "C"), ("4", "5", "6"))
orig_pos = {n: (line, spot) for line, nodes in enumerate(lines) for spot, n in enumerate(nodes)}
```

examples/algorithms/plot_dedensification.py
```diff
@@ -0,0 +1,102 @@
...
+    exp_node_sizes.append(600)
+    layers[1].append(node)
+
+for key in layers:
+    for node in layers[key]:
+        exp_graph.nodes[node]["layer"] = key
+plt.subplot(1, 3, 2)
+
+plt.title("Expansive (%s edges)" % exp_graph.number_of_edges())
+pos = nx.multipartite_layout(exp_graph, subset_key="layer")
```
dschult, line 77:
```suggestion
# position nodes in a grid 
lines = (("1", "2", "3", "4", "5", "6"), c_nodes, ("A", "B", "C"))
orig_pos = {n: (line, spot) for line, nodes in enumerate(lines) for spot, n in enumerate(nodes)}
```

examples/algorithms/plot_dedensification.py
```diff
@@ -0,0 +1,102 @@
...
+nonexp_graph, compression_nodes = summarization.dedensify(original_graph)
+nonexp_node_colors = list(node_colors)
+nonexp_node_sizes = list(node_sizes)
+for node in compression_nodes:
+    nonexp_node_colors.append("yellow")
+    nonexp_node_sizes.append(600)
+plt.subplot(1, 3, 3)
+
+plt.title("Non-Expansive (%s edges)" % nonexp_graph.number_of_edges())
+pos = layout(nonexp_graph, prog="fdp")
```
dschult, line 94:
```suggestion
# position nodes in a grid
lines = (("1", "2", "3", "4", "5", "6"), compression_nodes, ("A", "B", "C"))
orig_pos = {n: (line, spot) for line, nodes in enumerate(lines) for spot, n in enumerate(nodes)}
```

examples/algorithms/plot_dedensification.py
dpfens: @dschult Thank you for the grid code.  I tried it and the figure produced tended to have lots of overlapping edges which made the figure harder to read.  Here are those results:

dedensification_example_suggested

 I got the best results (least number of overlapping edges) by hard-coding the positions to match those in the paper, so I am going to include those `pos` values.  Here 
dedensification_example_paper_replicated

networkx/algorithms/summarization.py
dpfens: @jarrodmillman when you get a chance, could you review the module documentation?

dschult:
I like the docs for the module. And it's great to have the example in the gallery. The top docstring for the gallery example could use a paragraph describing what the example is all about... -> what it does and why that might be interesting.

I have a question about the ```expansive``` flag input:  If dedensification is used to reduce the total number of edges, why would someone want to use the ```expansive``` option?

I think this is a good addition and it's close to being ready to merge.

dpfens:
@dschult Thank you for the feedback, I just added a short description to the example.

`expansive=True`, would be used when the developer poorly chooses a `threshold` value and/or to allow developers to simplify algorithms whose performance could be impacted by a compressor potentially not existing in the neighborhood of a high degree node.  According to the [original paper](https://www.cs.umd.edu/~abadi/papers/graph-dedense.pdf#page=3): 

> The dedensification in Algorithm 1 (`expansive` algorithm) does not necessarily reduce the number of edges in the graph. This happens when the number of compressor nodes is extremely high due to a poor tuning of [`threshold`] (page 3)
> ...
> in order to get the nonexpansive strategy to work, slight modifications are necessary to the query algorithms presented in the next section.  Nonetheless, we found these modifications caused only small differences in query performance, and we focus on the simpler version of the algorithms in this paper (page 4).

Since it will likely not be very apparent to other developers,  I just modified the Notes section of the `dedensify` function (changes in **bold**):

> When expansiveness is used, dedensification is performed on all sets of low/high degree nodes, even when doing so will not reduce the number of edges.  **Expansiveness can result in a high number of compressor nodes when the threshold is poorly tuned, so expansiveness is not recommended**. When expansiveness is not used, dedensification will only be applied when it would reduce the total number of edges in the given graph.

jarrodmillman:
I am not sure it makes sense to provide the ``expansive=True`` option.  It seems to me the authors focused on the simpler version of the algorithm (which has the **problem** that it doesn't always  produce the desired goal of reducing the number of edges) because simpler algorithms are easier to discuss and analyze.  They justified focusing on the simpler (but wrong) algorithm because the modifications to make the algorithm correct (in the sense that it didn't make the graph larger)  "caused only small differences in query performance."

In practice, I don't see the use case for introducing nodes and not reducing the number of edges.  Could you explain your use cases in more detail?  Do other implementations of this algorithm implement both the nonexpansive version and the simpler one.

By the way, I will have more time later this week to review this more carefully.  But I think Dan is right that this is close to merging.

dpfens:
@jarrodmillman All good points about the algorithm described in the original paper, so I went ahead and removed the `expansive` option.

I have never needed to implement an algorithm specifically to account for expansive dedensification, and I cannot think of a scenario where added more nodes without removing edges would be beneficial.  Including the `expansive` option only   All of the papers I've found referencing dedensification are concerned with pattern matching/querying algorithms but only mention the approach described in the original paper.

I was not able to find any other implementations of this algorithm on GitHub.  Graph summarization algorithms appear to be less common.

Sounds good about the review.  Thanks again for taking a look.

examples/algorithms/plot_dedensification.py
```diff
@@ -0,0 +1,104 @@
+"""
+=============
+Dedensification
+=============
+
+Examples of dedensification of a graph.  Dedensification retain the structural
+pattern of the original graph will only add compressor nodes when doing so
+would result in fewer edges in the compressed graph. Both methods retain the
+structural pattern of the original graph.
```
dschult, line 9: Grammar check this paragraph.

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,219 @@
...
+import networkx as nx
+
+__all__ = [
+    "dedensify",
+]
+
+
+def dedensify(G, threshold=2, prefix=None, inplace=False):
+    """
+    Compresses neighborhoods around high-degree nodes
```
dschult, line 70: start doc_string immediately after """ (not on a new line)

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,219 @@
...
+        ...         ("2", "C"), ("2", "B"), ("2", "A"),
+        ...         ("3", "B"), ("3", "A"), ("3", "6"),
+        ...         ("4", "C"), ("4", "B"), ("4", "A"),
+        ...         ("5", "B"), ("5", "A"),
+        ...         ("6", "5"),
+        ...         ("A", "6")
+        ...     ]
+        ... )
+        >>> c_graph, c_nodes = nx.dedensify(original_graph)
+        >>> for c_node in c_nodes:
```
dschult, line 149: Maybe add a comment line that this part of the example re-densifies the graph back to its original state.

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,219 @@
...
+    for node in high_degree_nodes | low_degree_nodes:
+        node_neighbors = set(G.neighbors(node))
+        high_degree_neighbors = frozenset(node_neighbors & high_degree_nodes)
+        if high_degree_neighbors:
+            low_degree_neighbors = auxillary.get(high_degree_neighbors)
+            if not low_degree_neighbors:
+                auxillary[high_degree_neighbors] = set([node])
+            else:
+                low_degree_neighbors.add(node)
+
```
dschult, line 194:
```suggestion
    auxillary = {}
    for node in G:
        high_degree_neighbors = frozenset(high_degree_nodes & set(G[node]))
        if high_degree_neighbors:
            if high_degree_neighbors in auxillary:
                auxillary[high_degree_neighbors].add(node)
            else:
                auxillary[high_degree_neighbors] = {node}
```

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,219 @@
...
+    low_degree_nodes = set()
+    for node in G.nodes():
+        if isinstance(G, nx.DiGraph):
+            node_degrees = G.in_degree(node)
+        else:
+            node_degrees = G.degree(node)
+        if node_degrees > threshold:
+            high_degree_nodes.add(node)
+        else:
+            low_degree_nodes.add(node)
```
dschult, line 182:
```suggestion
    degrees = G.in_degree if G.is_directed() else G.degree
    for node, node_degree in degrees:
        if node_degree > threshold:
            high_degree_nodes.add(node)
        else:
            low_degree_nodes.add(node)
```

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,219 @@
...
+                auxillary[high_degree_neighbors] = set([node])
+            else:
+                low_degree_neighbors.add(node)
+
+    if not inplace:
+        G = G.copy()
+
+    compressor_nodes = set()
+    for index, nodes in enumerate(auxillary.items()):
+        high_degree_nodes, low_degree_nodes = nodes
```
dschult, line 200:
```suggestion
    for index, (high_degree_nodes, low_degree_nodes)  in enumerate(auxillary.items()):
```

dschult (commented):
I've made a few suggestions. If my suggestions for code don't make sense or don't work, or aren't your style that's fine, but I think they take advantage of some NetworkX and some Python features. They are suggestions...

networkx/algorithms/summarization.py
dpfens: Done

networkx/algorithms/summarization.py
dpfens: Done

networkx/algorithms/summarization.py
dpfens: @dschult  I like how this looks and its actually how I originally had written it, but it will bring the line length beyond 80 characters which would raise PEP8 issues.  If those can be ignored, I'll apply this change

examples/algorithms/plot_dedensification.py
dpfens: You're right, this needed to be changed.  It's now been updated to:

> Examples of dedensification of a graph.  Dedensification retains the structural pattern of the original graph and will only add compressor nodes when doing so would result in fewer edges in the compressed graph.

networkx/algorithms/summarization.py
jarrodmillman: We use black for our style and it uses 88 char line lengths.  You can learn more about our recommended setup here:

To run black manually
```
black <file>
```
or you can set up the pre-commit hook as we recommend in the developer guide.

jarrodmillman: We recently disabled pep8speaks, but if you see it complaining just ignore it.

dpfens:
What changes still need to be made to this?

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,214 @@
...
+    to high-degree nodes.  Dedensification will only add a compressor node when
+    doing so will reduce the total number of edges in the given graph. This
+    implementation currently supports graphs with a single edge type.
+
+    Examples
+    --------
+    Dedensification will only add compressor nodes when doing so would result
+    in fewer edges::
+
+        >>> import networkx as nx
```
dschult, line 107: The line ```import networkx as nx``` is automatically run for any doctests of the examples in the docstring.
So you don't need this line.

examples/algorithms/plot_dedensification.py
```diff
@@ -0,0 +1,103 @@
...
+pos = {
+    "3": (0, 1),
+    "2": (
+        0,
+        2,
+    ),
+    "1": (
+        0,
+        3,
+    ),
```
dschult, line 57: I'm not sure why these lines are formatted this way -- seems like they could be one line per node like the others.
If `black` did this I think you could put them on one line and black would not re-split them.  Worth a try anyway.

dschult (approved):
This looks ready to go for me!

Thanks very much!

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,214 @@
...
+    Parameters
+    ----------
+    G: graph
+       A networkx graph
+    threshold: int, optional (default=2)
+       Minimal number of degrees to be considered a high degree node
+    prefix: str or None, optional (default: None)
+       An optional prefix for denoting compressor nodes
+    inplace: bool, optional (default: False)
+       Indicates if dedensification should be done inplace
```
rossbar, line 86: This appears to be the only function where `inplace` is used as a kwarg. The releases notes from `api_0.99.rst`  state:

 > The argument inplace=False|True has been replaced with copy=True|False.

So maybe it'd be better to use a copy kwarg here instead?

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    high_degree_nodes = set()
+    low_degree_nodes = set()
+
+    degrees = G.in_degree if G.is_directed() else G.degree
+    for node, node_degree in degrees:
+        if node_degree > threshold:
+            high_degree_nodes.add(node)
+        else:
+            low_degree_nodes.add(node)
```
rossbar, line 180:
```suggestion
    degrees = G.in_degree if G.is_directed() else G.degree
    # Group nodes based on degree threshold
    high_degree_nodes = set([n for n, d in degrees if d > threshold])
    low_degree_nods = G.nodes() - high_degree_nodes
```

Using set difference would make this more concise, but please double-check me to make sure the suggestion is consistent with the original behavior!

rossbar (commented):
I had a few questions about the API related to the `threshold` and `inplace` arguments:
  - should `threshold` be an kwarg with a default value?
  - should `inplace` -> `copy`, which I *think* is more consistent with the rest of NX.

I left a smattering of other suggestions --- feel free to take/leave them!

The added tests are quite heavy and I found them difficult to follow, but coming up with good unit tests for dedensification seems like it's challenging, so I don't think it should be a blocker.

networkx/algorithms/summarization.py
```diff
@@ -0,0 +1,214 @@
...
+    with a degree greater than a given threshold).  Dedensification also has
+    the added benefit of reducing the number of edges around high-degree nodes.
+    The implementation currently supports graphs with a single edge type.
+
+    Parameters
+    ----------
+    G: graph
+       A networkx graph
+    threshold: int, optional (default=2)
+       Minimal number of degrees to be considered a high degree node
```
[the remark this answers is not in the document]
dpfens: @rossbar I couldn't think of a reason to ever pick a value less than 2, so I left it as a default value.  I just changed it to a positional argument with no default value.

networkx/algorithms/summarization.py
dpfens: @rossbar I just double-checked this, your code would produce the original behavior, and is much much more concise.  I just applied the change.

networkx/algorithms/summarization.py
dpfens: @rossbar Sorry about that.  I just changed the kwarg to be `copy=False`.

dpfens:
@rossbar Thank you for taking the time to look through this, I applied all of your suggestions.

Good point about the tests.  I added comments to the tests to summarize regarding what each test is checking for.

networkx/algorithms/summarization.py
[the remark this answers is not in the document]
rossbar: I think the positional argument makes it more clear that choosing the threshold is an important part of the algorithm.

It's pretty clear from the algorithm description that a threshold of 1 makes no sense, but if you wanted to be explicit about it you could add a check to enforce the constraint near the top of the function, something like:

```python
if threshold < 2:
    raise nx.NetworkXError("The degree threshold must be >= 2")
```

Or mention it in the `threshold` parameter description in the docstring. Not really necessary but a possibility if you think it helps!

networkx/algorithms/summarization.py
rossbar: The only remaining issue with this is what the default value for `copy=` should be. On the one hand, a default of `copy=True` is more in keeping with the functional-style that inputs aren't modified by the functions, but OTOH this function is likely to be used primarily on large graphs, where there's a good reason (memory usage) not to produce copies.

The results of a quick grep make it seem like the default for all other functions that have a `copy=` kwarg is `copy=True`, so I tend to lean that way. Thoughts @jarrodmillman @dschult ?

rossbar (commented):
Just one outstanding question about what the default value of the `copy` kwarg should be, but other than that this LGTM! Thanks @dpfens for all the hard work and the nice new feature!

networkx/algorithms/summarization.py
dschult: I think it produces fewer "surprises" to have the default be `copy=True`.
When people are trying it on smaller cases to get used to it this default makes sense. When they get serious and want to use it for a big graph and save memory, they should know enough by then to be able to turn it off.
But maybe I am naive. :}

networkx/algorithms/summarization.py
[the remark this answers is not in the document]
dpfens: @rossbar I just added it as an enforced constraint, as you suggested.

networkx/algorithms/summarization.py
dpfens: Sorry about that...again.  I just updated it to be `copy=True`.

dpfens:
@rossbar Thanks for all the help from everyone so I could get it right.  I think I have a better idea of how to do these now, so I should able to write a better contribution next time.

rossbar (approved):
Thanks @dpfens! This LGTM now.

PR #4257, opened 2020-10-15
Fix betweenness_centrality doc paper links

j6k4m8:
The links in the betweenness_centrality documentation no longer point to live paper references. This PR updates the docstrings with evergreen DOI links instead! :)


dschult:
Excellent!  Thank you so much...  
They aren't pdf links anymore but that's the point I suppose...
We should really be using the universal locator DOI links...

PR #4259, opened 2020-10-15
Make degree rank plot legible

jarrodmillman:


rossbar (approved):
Replacing the log scaling is definitely an improvement. I also made a slight adjustment to the inset axes to (hopefully) prevent nodes of the graph from being draw directly on top of the parent axes.

PR #4269, opened 2020-10-18
Add gene-gene network

jarrodmillman:
@ericmjl @rossbar @dschult  I decided to add a biological network example to the PyData global slidedeck.  I want all the figures to be from the gallery, so I am adding this example.  I found the dataset online and I don't know much about it.  It would be great if we could improve it before tomorrow evening when I will record the talk.

For example, I know biologists have used degree and betweenness centrality in network analysis, but I don't know if it makes sense here.  It doesn't have to make sense since we aren't claiming anything; but it would be nice.

Would it make more sense to have
```
  size = degree
  color = betweenness centrality
```
I tried and it didn't look great.  But maybe with different settings, it would.

Any other ideas?  Is there a better color gradient or way to plot this?


ericmjl (approved):

ericmjl:
@jarrodmillman this is definitely a nice example to include! In terms of principles of viz, I would opt to prioritize the node position first rather than use a spring-like layout, but since the purpose is not scientific inference, I guess anything goes here :)

rossbar (approved):
I think this is a really nice example, especially in terms of illustrating centrality.

The figure itself is quite busy, but I'm not sure how it could be improved. I tried an approach in jarrodmillman/networkx#5 based on highlighting the `n` most central nodes (and the edges that touch them) but I'm not sure that that approach is really any better:

wormbrain_10central

examples/algorithms/plot_betweenness_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+largest_component = max(components, key=len)
+H = G.subgraph(largest_component)
+
+# compute centrality
+centrality = nx.betweenness_centrality(H, k=10, endpoints=True)
+
+#### draw graph ####
+fig, ax = plt.subplots(figsize=(20, 15))
+pos = nx.spring_layout(H, k=0.15, seed=4552321)
+node_color = [d * 10 for n, d in H.degree()]
```
dschult, line 30: You could try coloring by community structure, with size by betweenness.
It takes a couple minutes on my machine to compute the community. 
But the node colors nicely show the communities that are already obvious from positions using spring_layout.

```suggestion
# compute community structure
lpc = nx.community.label_propagation_communities(H)
community_index = {n: i for i, com in enumerate(lpc) for n in com}

#### draw graph ####
fig, ax = plt.subplots(figsize=(20, 15))
pos = nx.spring_layout(H, k=0.15, seed=4552321)
node_color = [community_index[n] for n in H]
```

PR #4273, opened 2020-10-19
Fixed issue where edge attributes were being silently overwritten during node contraction

jtimmons:
This PR fixes issue #4150 where remapped edges during node contraction were silently overwriting existing edges. Take the example from the original reported issue:

Given a Graph or DiGraph with the shape `(src1) --> (dest) <-- (src2)`; the node `src2` is contracted into the node `src1` to give the resulting graph `(src1) --> (dest)`. Currently, the "disappearing" edge `(src2, dest)` _overwrites_ the existing edge `(src1, dest)` including all of the edge's attributes.

This PR adds a check for existing edges while remapping edges from the "disappearing" node to the existing one. If we would be overwriting an edge, we instead append the attributes of the "disappearing" edge into a `"contraction"` attribute on the existing edge - this mirrors how the contracting node attributes are handled. In the case outlined in the original issue, we now end up with the following attributes:

```
{
  "value": "src1-value",
  "contraction": {
    ("src2", "dest"): {
      "value": "src2-value"
    }
  }
}
```

Note that this is a break from the existing API (even if it's pretty esoteric) - does this warrant an entry in the release notes?

### DIFF
--- a/networkx/algorithms/minors.py
+++ b/networkx/algorithms/minors.py
@@ -508,7 +508,7 @@ def contracted_nodes(G, u, v, self_loops=True, copy=True):
             continue
 
         if not H.has_edge(w, x) or G.is_multigraph():
-            H.add_edge(w, x, *d)
+            H.add_edge(w, x, **d)
         else:
             if "contraction" in H.edges[(w, x)]:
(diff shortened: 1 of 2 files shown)


dschult (approved):
I approve this change

jarrodmillman (approved):

Issue #3139, opened 2018-09-03
contracted_edge() unpacks MultiGraph edges incorrectly when calling contracted_nodes()


Cerebus:
The edge key value gets unpacked into self_loops as a positional argument, but the call also has self_loops as a kwarg.
# Example
```
import networkx as nx
import random
g = nx.complete_graph(5, create_using=nx.MultiGraph())
nx.contracted_edge(g, random.choice(list(g.edges)))
```
## Results
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-220-465f568dfaa5> in <module>()
      1 g = nx.complete_graph(5, create_using=nx.MultiGraph())
----> 2 nx.contracted_edge(g, random.choice(list(g.edges)))

~/.virtualenvs/bernoulli/lib/python3.6/site-packages/networkx/algorithms/minors.py in contracted_edge(G, edge, self_loops)
    451         raise ValueError('Edge {0} does not exist in graph G; cannot contract'
    452                          ' it'.format(edge))
--> 453     return contracted_nodes(G, *edge, self_loops=self_loops)

TypeError: contracted_nodes() got multiple values for argument 'self_loops'
```
# Partial workaround
Slicing the edge key out of the edge in the call to contracted_edge() allows the function to work:
```
>>> nx.contracted_edge(g, random.choice(list(g.edges))[:-1])
<networkx.classes.multigraph.MultiGraph at 0x73c460e3d5f8>
```
But this loses the edge identity, which might be important for some algorithms.
# More complete workaround
The equivalent operation using contracted_nodes() works:
```
u, v, k = random.choice(list(g.edges))
g = nx.contracted_nodes(g, u, v, self_loops=False)
```
But IMHO both cases should work.  :)

dschult:
It seems to me this is a bug in ```nx.contracted_edge``` which should not assume that ```edge``` is length 2. It should use ```*edge[:2]``` as the argument, or add a line with ```u, v = edge[:2]``` before using ```nx.contracted_nodes(G, u, v, self_loops=self_loops)```

Thanks for the report!

PR #4274, opened 2020-10-19
Fixed issue where edge contraction was unpacking edges incorrectly

jtimmons:
Fixes issue #3139 where edge contraction had the incorrect assumption that edges were always a 2-tuple such as `('u', 'v')`. This was causing failures when contracting multigraph edges which can be referenced explicitly with their index such as `('u', 'v', 0)`.

Updated the edge contraction logic to use at most the first two elements in the tuple


dschult (approved):
Looks good!  Thanks. I approve this PR.

PR #4283, opened 2020-10-24
Add readwrite section to tutorial

jarrodmillman:
As part of deprecating nx_yaml #4281 and gpickle #4282 I am drafting a readwrite section for the tutorial.  I am planning something short, but I would like it to be somewhat prescriptive.

So far I just have a table from the pydata talk and a short section on using yaml and pickle to save NX graphs.

Here are some resources I am looking at:
- https://gephi.org/users/supported-graph-formats/
- https://www.researchgate.net/publication/330185503_Overview_of_Standard_Graph_File_Formats
- https://arxiv.org/pdf/1503.02781.pdf
- Pitas, Ioannis (2016), "1.5 Graph storage formats and visualization", Graph-Based Social Media Analysis

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -597,4 +597,46 @@ node positions, or write the graph in dot format for further processing.
 
 See :doc:`/reference/drawing` for additional details.
 
+Reading and writing graphs
+--------------------------
+
+.. csv-table:: Supported file formats
+   :header: "Format", "Description"
+   :widths: 20, 40
+
+   "Adjacency list", "Each row 1st column as out-degree"
+   "Edge list", "2 column, source$\rightarrow$ target"
+   "DOT", "Graphviz format"
+   "GEXF", "XML implementation"
+   "GML", "Similar to DOT"
+   "GraphML", "XML implementation"
+   "LEDA", "Between edge list and Pajek"
+   "Pajek", "Edge list + node and edge attr"
+   "SparseGraph6", "Adjacency list variant"
+
+You can read and write NetworkX graphs as Python pickles.
+
+>>> import pickle
+>>> G = nx.path_graph(4)
+>>> with open('test.gpickle', 'wb') as f:
+...     pickle.dump(G, f, pickle.HIGHEST_PROTOCOL)
+... 
+>>> with open('test.gpickle', 'rb') as f:
+...     G = pickle.load(f)
+... 
+
+You can also read and write NetworkX graphs in YAML format
+using pyyaml.
+
+>>> import yaml
+>>> G = nx.path_graph(4)
+>>> with open('test.yaml', 'w') as f:
+...     yaml.dump(G, f)
+... 
+>>> with open('test.yaml', 'r') as f:
+...     G = yaml.load(f, Loader=yaml.FullLoader)
+... 
+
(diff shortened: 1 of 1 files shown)


dschult (approved):

rossbar (approved):
This looks good, thanks @jarrodmillman 

I fixed an rst link and had to apply the change from #4548 here as well to get the doc build working after the recent yaml change.

doc/tutorial.rst
```diff
@@ -597,4 +597,46 @@ node positions, or write the graph in dot format for further processing.
...
+   :header: "Format", "Description"
+   :widths: 20, 40
+
+   "Adjacency list", "Each row 1st column as out-degree"
+   "Edge list", "2 column, source$\rightarrow$ target"
+   "DOT", "Graphviz format"
+   "GEXF", "XML implementation"
+   "GML", "Similar to DOT"
+   "GraphML", "XML implementation"
+   "LEDA", "Between edge list and Pajek"
```
stefanv, line 613: Can you move this down so that there is no forward reference to Pajek?

doc/tutorial.rst
```diff
@@ -597,4 +597,46 @@ node positions, or write the graph in dot format for further processing.
...
+   "Pajek", "Edge list + node and edge attr"
+   "SparseGraph6", "Adjacency list variant"
+
+You can read and write NetworkX graphs as Python pickles.
```
stefanv, line 617: I'd start with the example you'd like users to follow.  Pickles aren't compatible across versions typically, so perhaps this isn't it.

jarrodmillman: I am not sure what to do here.  I didn't put either the yaml or pickle example here because I thought people should do it.  We used to have functions to write and read yaml and pickle files.  I deprecated them because they didn't seem very useful and were basically one-liners.  Without thinking much more than that, I thought they deserved a mention in the docs at best.  But I am not sure they deserve even that.  Maybe we should just leave them both out of the documentation.  We could just leave the table and the link to the docs.  Thoughts?

@dschult Is there a preferred method for saving a NetworkX graph?

stefanv: People may forget that NX graphs are dicts-of-dicts, so I think a reminder that they can be written to disk is good.  I'd simply put the "best obvious" way at the top.

jarrodmillman: Would yaml be a better format?

stefanv: Yes, I'd think yaml and json are both fine (and approximately equivalent).

dschult: There is not a preferred method for saving a NetworkX Graph. 
I think historically most of us have used pickle.

stefanv: Understood, but if we had to give an example, what would we choose? I don't think there is any advantage to using pickle over the others.

jarrodmillman: I don't have time to look into this now, so I just moved this to the 2.7 milestone.  I will go back to my original plan of learning more about graph file formats.  Once I understand the issues better, I will finish this section of the tutorial.

I have a few links to resources I found online in the first comment on this PR.  If anyone knows of other resources, please add them.

PR #4291, opened 2020-10-26
Add simple graph w/ manual layout

jarrodmillman:
Adding simple figure I made for the pydata global talk.


rossbar (approved):
LGTM. It seems like this example might be a good candidate for the "basic" category, though I think it's fine in "drawing" as well. Just a thought!

jarrodmillman:
I put it in basic and then moved it to drawing.  I am not sure where it should go.  I am also not sure that our current categories make sense.  It may make sense to reorganize the directory layout.  I'll think about it.  If anyone has suggestions for the directory (category) structure, please let me know.

examples/drawing/plot_simple_graph.py
```diff
@@ -0,0 +1,38 @@
...
+G = nx.Graph()
+G.add_edge(1, 2)
+G.add_edge(1, 3)
+G.add_edge(1, 5)
+G.add_edge(2, 3)
+G.add_edge(3, 4)
+G.add_edge(4, 5)
+
+# explicitly set positions
+pos = {1: (0, 0), 2: (-0.1, 0.03), 3: (0.2, 0.017), 4: (0.4, 0.0255), 5: (0.5, 0.003)}
```
dschult, line 21: This is a very good example to have in the Gallery.  
Just the kind of thing someone would want to do and not know how to do it.

Perhaps they wouldn't want so many decimals in their positions though...
Would they more naturally use integers (or half integers).  
I usually use manually entered values when I want a grid of nodes... Like what is used for the dedensify example.
For example (integers for left columns then need 0.5 to center the other columns vertically)

```python
G=nx.DiGraph([(0, 3), (1, 3), (2, 4), (3, 5), (3, 6), (4, 6), (5, 6)])

# group nodes by column
left_nodes = [0, 1, 2]
middle_nodes = [3, 4]
right_nodes = [5, 6]

# set the position according to column (x-coord)
pos = {n: (0, i) for i, n in enumerate(left_nodes)}
pos.update({n: (1, i + 0.5) for i, n in enumerate(middle_nodes)})
pos.update({n: (2, i + 0.5) for i, n in enumerate(right_nodes)})
```

Not a big deal, and maybe it doesn't look as nice.  I haven't tried drawing it.

dschult (approved):

PR #4296, opened 2020-10-28
Move a few imports inside functions to improve import speed of the library

### LINKED ISSUE #2900: Long import times
Using [**import_profiler**](https://pypi.python.org/pypi/import-profiler/0.0.3), I've traced slow imports to **networkx**, especially:

* `networkx.generators` (300ms)
* `networkx.linalg` <- `.algebraicconnectivity` <- `numpy` (200ms)
* `networkx.readwrite` <- `.graphml` <- `xml.etree.cElementTree` (100ms)

I'm not using `linalg` at all, and I rarely need `readwrite`. I don't think I need `generators` either. These three trees represent 60% of the import-time for **networkx**.

When my filesystem is slow, the import-times can be much worse.

I have a couple of ideas for you:
* If possible, avoid importing these unless they are needed.
* Or, depend on and use [**lazy-import**](https://pypi.python.org/pypi/lazy-import) for at least these 3 sub-trees.

dschult:
Some of the imports are at the module level still -- instead of inside a function.
This negates our efforts at lazy importing, thus increasing import time (by a very small amount but users are still complaining).

Fixes #2900 

I will take another look at this in light of #4277 before it should be merged.

Feedback? @pb-cDunn

### DIFF
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -4,34 +4,6 @@ import math
 import networkx as nx
 from networkx.utils import py_random_state
 
-# Accommodates for both SciPy and non-SciPy implementations..
-try:
-    from scipy.special import zeta as _zeta
-
-    def zeta(x, q, tolerance):
-        return _zeta(x, q)
-
-
-except ImportError:
-
-    def zeta(x, q, tolerance):
-        """The Hurwitz zeta function, or the Riemann zeta function of two
-        arguments.
-
-        ``x`` must be greater than one and ``q`` must be positive.
-
-        This function repeatedly computes subsequent partial sums until
-        convergence, as decided by ``tolerance``.
-        """
-        z = 0
-        z_prev = -float("inf")
-        k = 0
-        while abs(z - z_prev) > tolerance:
-            z_prev = z
-            z += 1 / ((k + q) ** x)
-            k += 1
-        return z
-
 
 __all__ = [
     "caveman_graph",
@@ -725,9 +697,33 @@ def _powerlaw_sequence(gamma, low, high, condition, length, max_iters, seed):
     raise nx.ExceededMaxIterations("Could not create power law sequence")
 
 
-# TODO Needs documentation.
+def _hurwitz_zeta(x, q, tolerance):
+    """The Hurwitz zeta function, or the Riemann zeta function of two arguments.
+
+    ``x`` must be greater than one and ``q`` must be positive.
+
+    This function repeatedly computes subsequent partial sums until
+    convergence, as decided by ``tolerance``.
+    """
+    z = 0
+    z_prev = -float("inf")
+    k = 0
+    while abs(z - z_prev) > tolerance:
+        z_prev = z
+        z += 1 / ((k + q) ** x)
+        k += 1
+    return z
+
+
 def _generate_min_degree(gamma, average_degree, max_degree, tolerance, max_iters):
     """Returns a minimum degree from the given average degree."""
+    # Defines zeta function whether or not Scipy is available
+    try:
+        from scipy.special import zeta
+    except ImportError:
+        def zeta(x, q):
+            return _hurwitz_zeta(x, q, tolerance)
+
     min_deg_top = max_degree
     min_deg_bot = 1
     min_deg_mid = (min_deg_top - min_deg_bot) / 2 + min_deg_bot
@@ -738,7 +734,7 @@ def _generate_min_degree(gamma, average_degree, max_degree, tolerance, max_iters
             raise nx.ExceededMaxIterations("Could not match average_degree")
         mid_avg_deg = 0
         for x in range(int(min_deg_mid), max_degree + 1):
-            mid_avg_deg += (x ** (-gamma + 1)) / zeta(gamma, min_deg_mid, tolerance)
+            mid_avg_deg += (x ** (-gamma + 1)) / zeta(gamma, min_deg_mid)
         if mid_avg_deg > average_degree:
             min_deg_top = min_deg_mid
             min_deg_mid = (min_deg_top - min_deg_bot) / 2 + min_deg_bot
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -721,6 +721,7 @@ def _generate_min_degree(gamma, average_degree, max_degree, tolerance, max_iters
     try:
         from scipy.special import zeta
     except ImportError:
+
         def zeta(x, q):
             return _hurwitz_zeta(x, q, tolerance)
 
--- a/networkx/generators/geometric.py
+++ b/networkx/generators/geometric.py
@@ -35,23 +35,22 @@ def geometric_edges(G, radius, p):
     Radius uses Minkowski distance metric `p`.
     If scipy available, use scipy KDTree to speed computation.
     """
+    nodes_pos = G.nodes(data="pos")
     try:
         from scipy.spatial import cKDTree as KDTree
     except ImportError:
-        KDTree = None
-    nodes_pos = G.nodes(data="pos")
-    if KDTree is not None:
-        nodes, coords = list(zip(*nodes_pos))
-        kdtree = KDTree(coords)  # Cannot provide generator.
-        edge_indexes = kdtree.query_pairs(radius, p)
-        edges = ((nodes[u], nodes[v]) for u, v in edge_indexes)
+        # no scipy KDTree so compute by for-loop
+        radius_p = radius ** p
+        edges = [
+            (u, v) for (u, pu), (v, pv) in combinations(nodes_pos, 2)
+            if sum(abs(a - b) ** p for a, b in zip(pu, pv)) <= radius_p
+        ]
         return edges
-    # no scipy KDTree so compute by for-loop
-    edges = [
-        (u, v)
-        for (u, pu), (v, pv) in combinations(nodes_pos, 2)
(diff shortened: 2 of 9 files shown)


networkx/generators/community.py
```diff
@@ -725,9 +697,34 @@ def _powerlaw_sequence(gamma, low, high, condition, length, max_iters, seed):
...
+
+
 def _generate_min_degree(gamma, average_degree, max_degree, tolerance, max_iters):
     """Returns a minimum degree from the given average degree."""
+    # Defines zeta function whether or not Scipy is available
+    try:
+        from scipy.special import zeta
+    except ImportError:
+
+        def zeta(x, q):
```
Erotemic, line 725: Should there be a warning if scipy isn't found? I imagine the pure python implementation will be much slower, and a weird slowdown may be difficult to debug if the user accidentally uninstalls scipy.

networkx/linalg/algebraicconnectivity.py
```diff
@@ -378,6 +374,9 @@ def algebraic_connectivity(
     --------
     laplacian_matrix
     """
+    global np
```
Erotemic, line 377: This is some pretty sneaky code. I had to do a double take to figure out why you were doing this. I think I'd prefer if the `import numpy as np` line was just in every function rather than hoping that these functions are always called first. Its not the end of the world, but I place a sizable amount of weight on avoiding global variables whenever possible.

networkx/readwrite/gexf.py
```diff
@@ -193,47 +193,48 @@ class GEXF:
...
+                (np.uint16, "int"),
+                (np.uint32, "int"),
+                (np.uint64, "int"),
+                (np.int_, "int"),
+                (np.intc, "int"),
+                (np.intp, "int"),
+            ] + types
+
+        self.xml_type = dict(types)
+        self.python_type = dict(reversed(a) for a in types)
```
Erotemic, line 237: This code seems to rely on getting the items in `types` in a particular order. Several keys here will not be unique and you are relying on the CPython implementation to always overwrite the duplicated values. 

This also has an issue that `'int'` maps to `numpy.int64` iff numpy is present. Did you mean to use "integer" there like you did with the builtin types? I would recommend changing the first `types` variable to `builtin_types` and then have the second one be `numpy_types`. Then conditionally build `types` as `builtin_types + numpy_types` or just `builtin_types` (this has the added benefit of making the code idempotent).

Then I would rewrite the `python_type` line as

```python
self.python_type = {v: k for k, v in builtin_types}
```

which avoids any ambiguity. Biggest issue though is that "int" versus "integer" key. Not sure if you meant to do that.

networkx/generators/geometric.py
```diff
@@ -36,31 +29,29 @@ def euclidean(x, y):
...
+    except ImportError:
+        KDTree = None
+    nodes_pos = G.nodes(data="pos")
+    if KDTree is not None:
+        nodes, coords = list(zip(*nodes_pos))
+        kdtree = KDTree(coords)  # Cannot provide generator.
+        edge_indexes = kdtree.query_pairs(radius, p)
+        edges = ((nodes[u], nodes[v]) for u, v in edge_indexes)
+        return edges
+    # no scipy KDTree so compute by for-loop
```
Erotemic, line 49: I have a few comments here: 

First, small issue, but I would have put this in an else block. I like to avoid early returns when possible (makes it easier to copy-paste in IPython) and it also makes it clear that this is an alternative to the above code. 

Second, computing `radius ** p` will be executed on every iteration of this loop. It would be more efficient to do this once `p_thresh = radius ** p` and then using `p_thresh` instead. 

Demo showing how python does not precompute this value:

```python
class MyValue:
    def __init__(self, value):
        self.value = value

    def __pow__(self, arg):
        print('EXECUTE OP')
        return self.__class__(self.value ** arg)


def main():
    x = MyValue(3)
    [i for i in range(10) if x ** 3]

if __name__ == '__main__':
    main()
```

The `print('EXECUTE OP')` executes 10 times.

Lastly, this definition of `edges` is a list whereas the one in the KD-Tree version is a generator. It should likely be changed to be consistent.

Erotemic (commented):
I made a few comments. Hope these help.

networkx/generators/geometric.py
dschult: These are good suggestions..  Thanks!

networkx/generators/community.py
pb-cdunn: It's no different than the existing code. Warnings for missing scipy could be addressed in a separate PR if that's an issue.

networkx/linalg/algebraicconnectivity.py
pb-cdunn: I don't have a strong opinion on this. If we can avoid `np` global altogether, great. But if that's too much effort, this way is not terrible.

I wonder about the cost of this pattern though. With what we had, `import scipy` (and also `numpy`) was tried only once (per module). If you import it inside a function, then it is imported multiple times. If found, then the extra imports are nearly free, just dict lookup. But if _not_ found, then the filesystem will be checked at the start of each function, which could be onerous.

pb-cdunn (commented):
I guess all I want is to avoid importing the most expensive packages (scipy and numpy?) when they are not needed.

But I think the pattern in this PR could be problematic. (See my other comment.)

Erotemic:
@pb-cdunn Nested imports do not import multiple times. There is a global `sys.modules` dictionary that maps module names to loaded modules. The second time around its as fast as a dict lookup (one of the fastest things in Python). 

You are right about the not found case. This can be avoided by having some cached function like:

```python
import functools

@functools.lru_cache(None)
def _scipy_module():
    try:
        import scipy
    except Exception:
        scipy = None
    return scipy
```

```python
In [4]: with ubelt.Timer('foo'): 
   ...:     _scipy_module() 
   ...:                                                                                                                                

tic('foo')
...toc('foo')=0.0151s

In [5]: with ubelt.Timer('foo'): 
   ...:     _scipy_module() 
   ...:                                                                                                                                

tic('foo')
...toc('foo')=0.0000s
```

This avoids the issue with the not-found case. However, how many times are you calling this in the not found case when you don't have the ubiquitous scipy or numpy installed? And how large of graphs are those pure-python methods able to handle? Does this actually matter in the case where graph size is going to bottleneck you before file-system searches? If the answer is No, I think the current `import` style might be simpler and faster in the most common use case. 

On numpy, I guess avoiding it is nice in some applications, but parallel computation on long lists of numbers is near-universally helpful. The numpy packages is also significantly smaller and more ubiquitous (a difficult feat to accomplish) than scipy. I typically put import scipy inside functions, but I think there is an argument for assuming numpy globally (because avoiding that dict lookup does still help in critical sections).

networkx/linalg/algebraicconnectivity.py
dschult: The cost of a failed import will not be a limiting factor for speed with any of these functions. If speed is an issue they will already have installed numpy or scipy. The goal here is to make it fast to ignore these functions.  If you never call these functions, this will speed it up.

The problem with making a global np is only if someone is using np as a variable name in the `__main__` namespace. At least that is my understanding. I agree that global namespace interactions are to be avoided, but so are imports within a function instead of at the module level. Such is the nature of these import-time optimizations.

networkx/readwrite/gexf.py
dschult: This comment should be addressed in a different PR

networkx/linalg/algebraicconnectivity.py
dschult: I removed global np in favor of more import numpy as np calls inside the functions.
This only affects algebraicconnectivity.py

pb-cdunn:
> I typically put import scipy inside functions, but I think there is an argument for assuming numpy globally...

I completely disagree with that. This is a graph library, not a numerics library. Anyway, I'll look at this again someday after this is released.

Erotemic:
> This is a graph library, not a numerics library

There are deep connections between graph theory and linear algebra. In some cases I imagine its faster to compute the answer to a graph question by converting to a matrix representation, solving the reduced problem and converting back. My point is that graphs and numerics are not mutually exclusive. I'm not saying that numpy must be assumed to exist and globally imported, but I am saying there is an argument for it.

pb-cdunn:
True. But it's unfortunate that we are forced to load things we do not need.

The real problem was that, long ago, all components were star-imported into the top module, and many others recursively, for convenience. That was a mistake, I think inarguably. Now we're looking for work-arounds. If **scipy** is no longer load, that should help.

Erotemic:
Yeah, I agree that was a mistake. I used to do that with my libraries as well. 

Sometimes, I still do like to have something close to that functionality, but I never use `import *` anymore. Instead I use a package I wrote called `mkinit` (https://github.com/Erotemic/mkinit) that autogenerates explicit `__init__.py` files that have a similar effect to `import *` (it statically introspects module attributes and includes them depending on command line arguments and special dunder variables in the `__init__.py` itself).

If you are looking for workarounds, you could try using `mkinit` to generate the `__init__.py` files, which would give you backwards compatible behavior that you could gradually move away from. In PR #4349 I create two new submodules and there is a comment that documents how I used mkinit to generate each of them: (e.g. https://github.com/networkx/networkx/blob/b6e6c4633f3c1b80c7084d8976edf8386fcf9e67/networkx/algorithms/minors/__init__.py#L20)

PR #4318, opened 2020-11-05
Drop Py3.6 support per NEP 29

jarrodmillman:


jarrodmillman:
I'll do this after transitioning to GH actions.

PR #4319, opened 2020-11-05
Make nx.pagerank a wrapper around different implementations, use scipy one by default

MridulS:
TODO:
- [x] The original python implementation doesn't support multigraphs but scipy and numpy does. How to align that? Merge all the edges?
- [x] Update tests to check for 'impl' arguments.
- [x] Add deprecation warnings for 2.6


networkx/algorithms/link_analysis/pagerank_alg.py
```diff
@@ -231,6 +265,8 @@ def google_matrix(
         p = np.repeat(1.0 / N, N)
     else:
         p = np.array([personalization.get(n, 0) for n in nodelist], dtype=float)
+        if p.sum() == 0:
```
jarrodmillman, line 268: Won't trying to divide by 0, raise a ZeroDivisionError?

jarrodmillman:
I am not sure I like ``impl``.  I was initially imagining that we would check whether it was available or not.  But now that scipy is a default dependency, I suppose it is a question whether it makes sense to check.  Maybe it would be better to always use the scipy implementation.  And then if people want to use a slow implementation than can manually access one of the private functions directly.

I like leaving the slow versions in the library for educational purposes, but I don't know if we should make it easy for people to access them.

Maybe we should rename ``pagerank_scipy`` as ``pagerank`` and tell pytest that the tests are known failures on pypy.

networkx/algorithms/link_analysis/pagerank_alg.py
MridulS: Surprisingly no! A numpy array will be created with nan.
```
>>> arr = np.array([0., 0., 0., 0.])
>>> arr /= arr.sum()
>>> arr
array([nan, nan, nan, nan])
```

MridulS:
I am not too attached to the way we do this, but I think we should make sure that we don't have the same thing (pagerank in this case) explained 3 times in the documentation. With private functions we may have to write up something about it somewhere in the documentation (that they exist). I think with the 'impl' way we can document it in a subtle way.

rossbar (commented):
After thinking about this more, I agree with @MridulS that it is important to be able to have some mechanism to select which implementation is used. For example, if a user has a graph that is very dense, then the scipy implementation might actually be less performant/use more memory than the numpy implementation. It would be good to allow users control over selecting the approach based on a-priori knowledge of the structure of the graph.

I still tend to lean more towards type-based dispatching rather than a kwarg switch for this type of selection, e.g.

```python
def pagerank(G, *args, **kwargs):
    if isinstance(G, scipy.sparse.csr_matrix):
        return _pagerank_scipy(G, *args, **kwargs)
    elif isinstance(G, numpy.ndarray):
        return _pagerank_numpy(G, *args, **kwargs)
    else:
        return _pagerank_python(G, *args, **kwargs)
```

This approach gives the user the ability to select implementation and has the advantage that any format conversions are done explicitly outside of the function rather than internally. Downsides include discoverability and the need to potentially convert back-and-forth between formats all the time.

But I'm getting too far afield --- this is stuff I'd like to put in an NXEP. I do think the `impl=` is a way better approach than asking users to explicitly call "private" functions (e.g. _pagerank_python), and it would be nice to have a scipy/numpy default as they should be a lot faster.

MridulS:
@rossbar I didn't go with type-based dispatching as I guess we should have a NXEP about it first and I didn't want to introduce a surprise functionality for 2.6 (maybe we could to test it out 🤷‍♂️ ).
We can always change the internal wrapper implementation to be type-based.
PS: Even if a networkx object is passed `pagerank_scipy` will be faster than `pagerank_python`(including the time to convert it to scipy sparse object).

rossbar:
> I guess we should have a NXEP about it first and I didn't want to introduce a surprise functionality for 2.6

I totally agree.

> Even if a networkx object is passed pagerank_scipy will be faster than pagerank_python

Yeah I think that's the big advantage here - making the scipy implementation the default.

networkx/algorithms/link_analysis/pagerank_alg.py
jarrodmillman: Thanks.

dschult:
I thought that the python version of pagerank works with multigraphs.  It uses `stochastic_graph` which is supposed to work with multidigraphs.  Maybe we need to have a test showing whether it works with multiGraphs or not.

MridulS:
In the current implementation of pagerank it does assume that it's a not a MultiGraph.

This line needs to be fixed (to sum of all the weights for MultiGraphs?) (W is the stochastic graph)
https://github.com/networkx/networkx/blob/4c38d8868c9cfd93ac7eebacf336226976b95130/networkx/algorithms/link_analysis/pagerank_alg.py#L151

dschult:
My understanding is that the stochastic_graph transformation (that creates W) handles the MultiGraph aspect of the problem. So W is NOT a multigraph. The code has already taken care of the multigraph nature of G by computing the weighted degrees of each nodes as it constructs W. Those weighted degrees are obtained by summing the multiedge attributes.

MridulS:
Looks like stochastic_graph doesn't transform the incoming graph to a normal directed graph. It keeps the multi edges.
```
In [1]: import networkx as nx

In [2]: G = nx.MultiDiGraph()

In [3]: G.add_edges_from([(1, 2), (1, 2), (1, 2), (2, 3), (2, 3), ('3', 3), ('3', 3)])
Out[3]: [0, 1, 2, 0, 1, 0, 1]

In [4]: W = nx.stochastic_graph(G)

In [5]: W
Out[5]: <networkx.classes.multidigraph.MultiDiGraph at 0x118428dd0>

In [6]: W.edges.data()
Out[6]: OutMultiEdgeDataView([(1, 2, {'weight': 0.3333333333333333}), (1, 2, {'weight': 0.3333333333333333}), (1, 2, {'weight': 0.3333333333333333}), (2, 3, {'weight': 0.5}), (2, 3, {'weight': 0.5}), ('3', 3, {'weight': 0.5}), ('3', 3, {'weight': 0.5})])
```
It just distributes the weight across all the multiple edges.

dschult:
Thank you for that!  I checked that it was dividing by the right value, but not that it was collapsing the edges (which it isn't).  So W is a MultiDiGraph afterall... Sorry to send you on a goose chase.

I guess the replacement code should be:
```python
    for _, nbr, wt in W.edges(n, data=weight):
        x[nbr] += alpha * xlast[n] * wt
```

MridulS:
This is ready for a review now.

MridulS:
Some rough benchmarking on my local machine with different graphs.
```
G_sparse : 0.05 density
G: 0.2 density
G_dense: 0.75 density
```

Screenshot 2020-11-16 at 13 32 18

dschult (approved):

MridulS:
Can someone restart the circleci build? It seems to be stuck on pending for some time now.
@rossbar @jarrodmillman

dschult:
I restarted the build -- and it completed now.
Ready for Review!
:}

jarrodmillman:
It looks like the scipy implementation is always the clear winner.  I think it makes sense to keep the different implementations for education purposes.  We could also test that we get the same output using different implementations, I suppose.

I am not sure why we should advertise the different implementations to users.  Would there be any disadvantage to making pagerank use the scipy implementation and then making the python and numpy implementations private?  That way folks looking at the source code can see the different implementations, but most casual users would never see (or accidentally use) the slow implementations.

jarrodmillman:
LGTM, but there are a couple of missing tests:

1. ``_pagerank_python`` needs a test for when ``nstart`` is not None.
2. ``google_matrix`` needs a test for when ``personalization`` is not None and ``p.sum() == 0``.

rossbar (approved):
Generally looks good! I think an instance of `impl` snuck through in the `_python_pagerank` signature, but that was it.

I had some other suggestions related to the warnings and tests that were too beefy for the GH suggestion feature, so I opened mriduls/networkx#3. They're just suggestions though, not blockers. This looks ready to go once the `_pagerank_python` signature is addressed.

jarrodmillman:
I forgot to ask you to document the deprecations in the release notes and our list of deprecation todo list, so I just made a PR:
https://github.com/MridulS/networkx/pull/4

jarrodmillman:
I think (assuming the PRs look good to you) that this is ready once you address Ross' comment and review/merge our two PRs. Thanks for doing this, I suspect this will be very helpful for many users.

jarrodmillman (approved):

PR #4320, opened 2020-11-05
Use GH actions for Linux CI

jarrodmillman:
@rossbar This is ready to go.  There is more to do in a follow-up PR:
- finish deploy doc
- maybe get rid of appveyor (add ``Windows`` to ``os: [Ubuntu, macOS]`` in ``test-default.txt``
- remove travis (and maybe appveyor)

But I would like to see how well this works first.  And this should give us more time with Travis, since we will only really use it to deploy docs on merge to master.


rossbar (approved):
LGTM - this adds quite a few jobs given the test matrices (ubuntu + macos) x (Python 3.6 - 3.9), but I'm +1 on starting with a large set and reducing later on. There are certainly areas where some jobs could probably be saved (e.g. running pre-releases with fewer Python versions).

A couple additional thoughts for follow on PRs:
 - Add jobs for other platforms/architectures (e.g. 32-bit). We've had tests fail on 32-bit architectures before that we've missed in CI.
 - See if we can't condense the "details" lines for the individual checks. This may become unnecessary if we reduce the number of jobs, but as it stands it's a lot of info and can be hard to find individual jobs (e.g. style/linting) in the mess of lines from the testing matrices.
 - Find a nicer way to display the coverage report in a nicer way (e.g. html)

jarrodmillman:
We are still using codecov.io for coverage reports:
https://codecov.io/gh/networkx/networkx

jarrodmillman:
We should definitely add 32-bit.  We used to do that.  But our tests were taking so long to run, at some point I started removing a bunch of tests.  GitHub actions are so fast that we can add them back.  (Part of the reason this has so many tests is I already added back some tests I had removed.  I also added a bunch more tests, though.)

jarrodmillman:
In terms of the detail, part of the change is that it used to be the checks displayed here were only links to longer lists somewhere else.  I prefer having them expanded here b/c it reduces the number of links I need to follow to see what is happening by one.  But if others prefer it the other way, it isn't a big deal for me.

MridulS (approved):

Issue #3823, opened 2020-02-19
connected_components should return frozensets


rsokl:
`connected_components` yields sets in a way that requires that they not be mutated during iteration:

https://github.com/networkx/networkx/blob/aeeeef92f2229b0cc8ebbbbfcfc87128f546d55f/networkx/algorithms/components/connected.py#L60-L65

This can lead to silent-but-deadly behavior. E.g. If I were to pop from these sets during iteration, which is a very natural way to access a node from each connected component,

```python
# getting a node from each component
[s.pop() for s in my_connected_components]
```
then the same connected component may be visited multiple times during the iteration.

This generator should yield `frozenset` objects instead so that the user cannot unwittingly produce such corrupt behavior.

rsokl:
(or you can just flip line 45 with line 46!)

dschult:
Yes -- it looks like we should swap lines 64 and 65 in ```connected.py```
Thanks!!

dschult:
@boothby has a slightly better solution in #3859 where in addition to switching lines 64 and 65, we change ```_plain_bfs``` to return a set instead of yielding each node. Look good?

rsokl:
Looks good! Thanks

PR #4333, opened 2020-11-10
Fix bug where changing a yielded component set breaks the component finding routine.

### LINKED ISSUE #4331: weakly_connected_components: Editing yielded sets during iteration has buggy side effects
When modifying the yielded component sets during iteration from `weakly_connected_components`, strange things happen:

```python
import networkx as nx
g = nx.MultiDiGraph()

g.add_edge(0, 1)                                                                      
g.add_edge(1, 0)                                                                      
g.add_edge(0, 3)                                                                      
g.add_edge(0, 3)                                                                      

print('Okay:')
print(list(nx.weakly_connected_components(g)))
# Everything alright, gives: [{0, 1, 3}]

print('Still okay:')
for n in nx.weakly_connected_components(g):
    print(n)
    n.remove(0)
# Will print once: {0, 1, 3}

print('NOT okay:')
# Deleting the wrong node during the iteration:
for n in nx.weakly_connected_components(g):
    print(n)
    n.remove(1)
# Will print the set twice:
# {0, 1, 3}
# {0, 1, 3}
```

In case this is actually desired (which I doubt), that should be at least documented that the connected component set is not to be edited.

dschult:
fix order of yield and seen.update in all connected components routines
Fixes #4331 
See also  #3859 & #3823 

A similar bug was fixed in March 2020 for connected components -- this fixes the directed versions of that bug.
We update the `seen` set with the component's nodes before yielding it so no longer has the problem reported in #4331 of changing the component set yielded makes the function report the same component again.

### DIFF
--- a/networkx/algorithms/components/tests/test_weakly_connected.py
+++ b/networkx/algorithms/components/tests/test_weakly_connected.py
@@ -76,3 +76,12 @@ class TestWeaklyConnected:
         pytest.raises(NetworkXNotImplemented, nx.weakly_connected_components, G)
         pytest.raises(NetworkXNotImplemented, nx.number_weakly_connected_components, G)
         pytest.raises(NetworkXNotImplemented, nx.is_weakly_connected, G)
+
+    def test_connected_mutability(self):
+        DG = nx.path_graph(5, create_using=nx.DiGraph)
+        G = nx.disjoint_union(DG, DG)
+        seen = set()
+        for component in nx.weakly_connected_components(G):
+            assert len(seen & component) == 0
+            seen.update(component)
+            component.clear()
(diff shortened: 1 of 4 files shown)


networkx/algorithms/components/tests/test_weakly_connected.py
```diff
@@ -76,3 +76,12 @@ def test_connected_raise(self):
         pytest.raises(NetworkXNotImplemented, nx.weakly_connected_components, G)
         pytest.raises(NetworkXNotImplemented, nx.number_weakly_connected_components, G)
         pytest.raises(NetworkXNotImplemented, nx.is_weakly_connected, G)
+
+    def test_connected_mutability(self):
+        DG = nx.path_graph(5, create_using=nx.DiGraph)
+        G = nx.disjoint_union(DG, DG)
+        seen = set()
+        for component in nx.weakly_connected_components(G):
+            assert len(seen & component) == 0
```
stefanv, line 85: What does this statement do?

stefanv (approved):
Looks good to me!  I have one tiny question, but it's not important.

PR #4355, opened 2020-11-15
Improve test coverage for coreviews.py

dschult:
Add tests to get test coverage up to 100% for coreviews.py


jarrodmillman:
Maybe we should deprecate copy before the 2.6 release.  We could then make sure to point out the deprecations in the release announcement and ask folks to let us know if they are using it.

dschult:
I've deprecated the `Filter*.copy()` methods.
I followed @rossbar deprecation style in `convert_matrix.py` and tests.
I left the FIXME comments in place until we resolve by removing that code (or not if people complain).

jarrodmillman (approved):

PR #4358, opened 2020-11-15
Update tutorial.rst - Fixes #4249

### LINKED ISSUE #4249: Add to tutorial info about edge order
With Python 3.6 or later the behavior of Graph/DiGraph is ordered because dict's became ordered with that version.

And... The order of adjacency reporting... g.adj, g.successors, g.predecessors is now the order of edge adding.
But the order of g.edges is not always the order of edge adding. The order of g.edges is the order of the adjacencies which includes both the order of the nodes and each node's adjacencies.

```python
g = nx.DiGraph()
g.add_edge(2, 1)   # adds the nodes in order 2, 1
g.add_edge(1, 3)
g.add_edge(2, 4)
g.add_edge(1, 2)
assert list(g.successors(2)) == [1, 4]
assert list(g.edges) == [(2, 1), (2, 4), (1, 3), (1, 2)]
```

We should add this kind of info to the tutorial.  (thought of because of question #4248 )

abdulelahsm:


MridulS (approved):

doc/tutorial.rst
```diff
@@ -131,6 +131,16 @@ At this stage the graph ``G`` consists of 8 nodes and 3 edges, as can be seen by
     >>> G.number_of_edges()
     3
 
+.. note:: With Python 3.6 or later the behavior of Graph/DiGraph is ordered. Since dictionaries became ordered with that version, the order of adjacency reporting: G.adj, G.successors, G.predecessors is now the order of edge adding. However, the order of G.edges is not always the order of edge adding. The order of G.edges is the order of the adjacencies which includes both the order of the nodes and each node's adjacencies. See example below:
```
jarrodmillman, line 134: I am not sure it is worth mentioning that this happened with Python 3.6, since we only work with 3.6 or later.  It is should also mention that the order is insertion order.  Maybe just:
```
The order of adjacency reporting (e.g., G.adj, G.successors, G.predecessors) is the order of edge addition. However, the order of G.edges is the order of the adjacencies which includes both the order of the nodes and each node's adjacencies. See example below:
```

jarrodmillman (changes requested):

doc/tutorial.rst
abdulelahsm: Thanks a lot for the review. I've reflected the changes.

dschult (approved):

PR #4359, opened 2020-11-15
Add unit tests for utils.misc.flatten

suvayu:
Also update flatten to always return a tuple. The previous behaviour of preserving the type was leading to test failure.  So we decided that while flattening there is no expectation to preserve the type.


MridulS:
@suvayu Can you look at the failing tests?

dschult (approved):

jarrodmillman (approved):

PR #2024, opened 2016-03-01
Add a way to draw self loops with Matplotlib

Sixshaman:
If you call networkx.draw(), you will face two issues:
1. If the draw() function puts three nodes on one line, you won't see the edge between two further nodes.
2. The draw() function can't draw self loops with Matplotlib because self loop edges have pos[0] == pos[1].

Solution: Instead of using simple line edges, use the Bezier curves.

Example of a Bezier-tweaked graph:
graph43

PR #4360, opened 2020-11-15
API: always use list of FancyArrowPatch rather than LineCollection

### LINKED ISSUE #3337: Error Attempting to Iterate Over LineCollection
This is less a bug with the code base and more a bug in the examples, but as a result a number of pieces of example code do not work any longer.

I have cobbled together code that uses the **Directed Graph** ([https://networkx.github.io/documentation/stable/auto_examples/drawing/plot_directed.html#](url)) example code to implement color maps, line alphas, and a colorbar.  Apparently matplotlib.collections.LineCollection objects are no longer iterable, and thus the code

```
for i, arc in enumerate(edges):
    arc.set_alpha(edge_alphas[i])
```

does not work (nor does using something like `for e in edges:` or `for i in range(len(edges)):`

The problem also arises in trying to create a colorbar legend by passing the `edges` object to `mpl.collections.PatchCollection()` as this too raise the error that LineCollection objects are not iterable.

Versions:

Python - 3.6
Networkx - 2.2
Matplotlib - 3.0.2

Here is the full code (assuming you have a graph `G` already):

```
pos = nx.circular_layout(G)

M = G.number_of_edges()
node_sizes = [3 + 10 * i for i in range(len(G))]
edge_colors = range(2, M + 2)
edge_alphas = [(5 + i) / (M + 4) for i in range(M)]

nodes = nx.draw_networkx_nodes(G, pos, node_color='red')
edges = nx.draw_networkx_edges(G, pos, edge_color=edge_colors,edge_cmap=plt.cm.coolwarm, width=2)

nx.draw_networkx_labels(G,pos,node_labels,font_size=16)

# set alpha value for each edge
for i, arc in enumerate(edges):
    arc.set_alpha(edge_alphas[i])

pc = mpl.collections.PatchCollection(edges, cmap=plt.cm.Blues)
pc.set_array(edge_colors)
plt.colorbar(pc)

plt.axis('off')
plt.show()
```

### LINKED ISSUE #3398: draw_network_edges allows curved edges or dashed edges, but not both.
Hi folks,

I would very much like to create curved, dashed  edges for either directed or undirected graphs.

However, it seems I can only have curved, **but not dashed** edges for directed graphs, even though I specify `style="dashed"`
```python
G = nx.DiGraph([('X','Y')])
pos = nx.spring_layout( G )
nx.draw_networkx_nodes(G, pos,
                       nodelist=list(G.nodes))
nx.draw_networkx_edges(G, pos,
                       edgelist=list(G.edges),
                       style='dashed',
                       connectionstyle="arc3,rad=0.3",
                      )
plt.savefig('solid-curved-directed-edge.png')
```

solid-curved-directed-edge

Furthermore, it seems I can only get dashed, **but not curved**, edges for undirected graphs. 
even though I specify `connectionstyle="arc3,rad=0.3"`

```python
G = nx.Graph([('X','Y')])
pos = nx.spring_layout( G )
nx.draw_networkx_nodes(G, pos,
                       nodelist=list(G.nodes))
nx.draw_networkx_edges(G, pos,
                       edgelist=list(G.edges),
                       style='dashed',
                       connectionstyle="arc3,rad=0.3",
                      )
plt.savefig('dashed-straight-undirected-edge.png')
```

dashed-straight-undirected-edge

Please advise.

Thanks so much,

Jeremy

tacaswell:
draw_networkx_edges used touse a LineCollection (presumably for
performance reasons) when drawing un-directed graphs and generate many
FancyArrowPatch objects when drawing directed graphs.  This resulted
in issues due to return type instability (#3337) and inconsistency
with handling the input (#3398).

closes #3398
closes #3337

Later commit adds ability to draw a self-loop edge.
closes #2024

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -36,7 +36,7 @@ API Changes
   In `to_*_array/matrix`, nodes in nodelist but not in G now raise an exception.
   Use G.add_nodes_from(nodelist) to add them to G before converting.
 
-- [ ]
+- [`#4360  <https://github.com/networkx/networkx/pull/4360>`_]
   Internally `.nx_pylab.draw_networkx_edges` now always generates a
   list of `matplotlib.patches.FancyArrowPatch` rather than using
   a `matplotlib.collections.LineCollection` for un-directed graphs.  This
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -36,6 +36,14 @@ API Changes
   In `to_*_array/matrix`, nodes in nodelist but not in G now raise an exception.
   Use G.add_nodes_from(nodelist) to add them to G before converting.
 
+- [ ]
+  Internally `.nx_pylab.draw_networkx_edges` now always generates a
+  list of `matplotlib.patches.FancyArrowPatch` rather than using
+  a `matplotlib.collections.LineCollection` for un-directed graphs.  This
+  unifies interface for all types of graphs.  In
+  addition to the API change this may cause a performance regression for
+  large graphs.
+
 Deprecations
 ------------
 
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -25,6 +25,7 @@ from networkx.drawing.layout import (
     random_layout,
     planar_layout,
 )
+import warnings
 
 __all__ = [
     "draw",
@@ -480,13 +481,13 @@ def draw_networkx_edges(
     edge_color="k",
     style="solid",
     alpha=None,
-    arrowstyle="-|>",
+    arrowstyle=None,
     arrowsize=10,
     edge_cmap=None,
     edge_vmin=None,
     edge_vmax=None,
     ax=None,
-    arrows=True,
+    arrows=None,
     label=None,
     node_size=300,
     nodelist=None,
@@ -536,11 +537,15 @@ def draw_networkx_edges(
        Draw the graph in the specified Matplotlib axes.
 
     arrows : bool, optional (default=True)
-       For directed graphs, if True draw arrowheads.
+       For directed graphs, if True draw arrowheads by default.  Ignored
+       if *arrowstyle* is passed.
+
        Note: Arrows will be the same color as edges.
 
-    arrowstyle : str, optional (default='-|>')
-       For directed graphs, choose the style of the arrow heads.
+    arrowstyle : str, optional (default=None)
+       For directed graphs and *arrows==True* defaults to ``'-|>'`` otherwise
+       defaults to ``'-'``.
+
        See :py:class: `matplotlib.patches.ArrowStyle` for more
        options.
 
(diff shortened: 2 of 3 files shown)


tacaswell:
The tricky thing about the self-loop is that when we compute the path that will be used to connect the two nodes we only (directly) know very local information (that the two points are the same point) which gives us no "natural" scale to use to pick the size of the loop (unlike where if the two points are distinct we have a natural scale of the distance between them and can then draw sensible arcs by going some fraction of the distance between them in the perpendicular direction).  

The first choice to make is fixed in screen size or fixed in data size.  This went with "fixed in data size" at ~10% of the total size of the visible graph.

image


```python
import networkx as nx

plt.ion()

G = nx.DiGraph([('X','Y'), ('X', 'X'), ('Y', 'X')])
pos = nx.spring_layout( G )
nx.draw_networkx_nodes(G, pos,
                       nodelist=list(G.nodes))
out = nx.draw_networkx_edges(G, pos,
                             edgelist=list(G.edges),
                             style='dashed',
                             connectionstyle="arc3,rad=0.3",
                             arrowstyle='-|>',
                             arrows=False
                             )


```

-----

One other possible issue with this PR is that because we now respect `arrowstyle` for all graphs, the user can add arrows to a `nx.Graph` and it will draw arrows.  That seems like a bit of a foot-cannon, but the foot cannon of backwards arrows, turning them off for directed graphs, etc is already there and the pay-off is fixing things like #3398.

dschult (approved):
Thanks Thomas. This is very helpful!

rossbar (approved):
Thanks @tacaswell , this is a very nice improvement.

Just to put some numbers on the performance regression for drawing edges of undirected graphs:

```python
>>> G  = nx.complete_graph(50)  # 1225 edges
>>>%timeit -r 1 -n 1 nx.draw_networkx_edges(G, nx.circular_layout(G))
```
gives `63.5 ms` on master and `1.4 s` for this PR. In my experience, layout/visualization get inherently difficult for >1K edges, so IMO this performance tradeoff is acceptable given the other advantages these changes bring. We can always add gallery examples for how to use LineCollection directly to draw graphs with very many edges.

The self-loop drawing is really nice. One thing to note is that it seems the nice self-loop paths are only visible when the user supplies a connection style:

```python
>>> G = nx.Graph([(0, 1), (0, 0)])
>>> pos = {0: (0, 0), 1: (1, 1)}
>>> nx.draw_networkx_nodes(G, pos)
>>> nx.draw_networkx_edges(G, pos)  # Two FancyArrowPatches in output, but only one visible
[<matplotlib.patches.FancyArrowPatch at 0x7f7f7bc90fa0>,
 <matplotlib.patches.FancyArrowPatch at 0x7f7f7bcd5ca0>]
```
fig0

```python
>>> plt.figure()
>>> nx.draw_networkx_nodes(G, pos)
>>> nx.draw_networkx_edges(G, pos, connectionstyle="arc3,rad=1.0")[<matplotlib.patches.FancyArrowPatch at 0x7f7f7af1ea30>,
 <matplotlib.patches.FancyArrowPatch at 0x7f7f7af1eca0>]
```
fig1

My vote would be to put this in as-is and see if we can't make self-loops visible with the default connectionstyle in a subsequent PR.

jarrodmillman:
It looks like there is still room for improvement, but I agree w/ Ross that we should merge this and improve in a separate PR.

jarrodmillman (approved):

PR #4364, opened 2020-11-17
Added implementation for MinHeap in heaps.py

jebogaert:
Implemented MinHeap class in heaps.py. The insert function is less efficient than the one in BinaryHeap, but it has the advantage of being the most efficient possible while maintaining no duplicate in the heap. Deleted get method in other heaps that extend MinHeap since the method is inherited.

### DIFF
--- a/networkx/utils/heaps.py
+++ b/networkx/utils/heaps.py
@@ -123,7 +123,7 @@ class MinHeap:
             heappush(self._heap, (key, value))
             return True
 
-    def _replaceValue(self,key, value, pos):
+    def _replaceValue(self, key, value, pos):
         """Replace the value of a given key with a new value.
         Recursivly calls itself in a DFS way until the value is found"""
         if pos < len(self._heap):
@@ -133,7 +133,9 @@ class MinHeap:
             elif key < self._heap[pos][0]:
                 return False
             else:
-                return self._replaceValue(key, value, pos*2+1) or self._replaceValue(key, value, pos*2+2)
+                return self._replaceValue(
+                    key, value, pos * 2 + 1
+                ) or self._replaceValue(key, value, pos * 2 + 2)
         return False
 
     def __nonzero__(self):
@@ -174,7 +176,6 @@ class PairingHeap(MinHeap):
 
     class _Node(MinHeap._Item):
         """A node in a pairing heap.
-
         A tree in a pairing heap is stored using the left-child, right-sibling
         representation.
         """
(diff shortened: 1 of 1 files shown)


dschult:
Hi!   

Can you give a description of your goal with this PR? 
Why are you implementing a MinHeap and what do you hope to use it for?

Your [comment elsewhere](https://github.com/networkx/networkx/issues/4224#issuecomment-729089918) suggests to me that you don't quite understand this Heap class yet. For example, the first doc_string in the module (for MinHeap) says that the heap is a collection of key-value pairs sorted by value.  That is not at all the same as sorting by keys. Indeed, in our use cases the keys are not guaranteed to be sortable (they are often nodes which don't have to be sortable).

Do you have other questions about BinaryHeap?  Are you interested in improving the tests for this module?

jebogaert:
Yes, sorry for this :') I just saw an unimplemented class and i thought i had the knowledge to implement it so i gave it a try. I don't have any particular idea of a use for it, i just thought it could be a usefull heap regardless of the usage i had in mind. Since you're talking about it, It could probably be usefull i one wants to sort unsortable objects as you said, it could be used by setting the unsortable object as values and giving them a key to order them in the heap. Maybe it is a better choice to create another class that extend MinHeap to implement this feature (so MinHeap would stay mostly unimplemented). Or maybe it is better to drop the idea, i don't know, what do you think about it ? I could also work on other features, but i don't know if i have the skills to do it. 

Just so you know, i started from scratch again, recloned the project and followed the suggested guidelines that i was not aware off the first time. If you agree with what i just suggested, the pull request is already ready (and went through all the tests this time).

dschult:
The current implementation does that in fact -- only it puts the unsortable object in the keys paired with a value that can be sorted.    :}

jarrodmillman:
I noticed that you are working on your master branch.  You should always work on a feature branch.  You may find reading our contributor guide helpful for learning how to set up your development environment and the basic workflow for contributing:

If anything is unclear in the contributor guide, please let us know.

Also, I would avoid implementing things in our utility subpackage unless you need it for a specific network analysis algorithm you are implementing.  The utility code is there to provide generic functionality needed by other parts of the library.  We are actively working to remove old utility code that we no longer need.

jebogaert:
Okay, i'll drop this idea, sorry for loosing your probably precious time :x I will try to work on something else, thanks a lot for the quick answers :) Also, i'm sorry, i didn't see i was pushing to the master branch, it is my first (unsuccessfull) contribution to an open source projet so i really appreciate the feedbacks :)

jarrodmillman:
No worries about difficulties with on the on ramp.  There is a fair bit of technical onboarding necessary to contribute to open source Python projects.  Fortunately, that work will apply to most open source Python projects, so it will get much easier after you've done a few PRs.  In the meantime, hopefully the contributor guide will provide enough guidance to help walk you through the process.

Please let us know if you have any questions.

PR #4370, opened 2020-11-19
MNT: make the self-loop appear in all cases

tacaswell:
Previously the logic to add the self-loop (which over-rides theConnectionStyle path!) would only be triggered if the user actually specified a ConnecitonStyle.  This pulls the default from FancyArrowPatch up to draw_networkx_edges.

Follow up to #4360 

```python
import networkx as nx

plt.ion()

G = nx.Graph([('X','Y'), ('X', 'X'), ('Y', 'X')])
pos = nx.spring_layout( G )
nx.draw_networkx_nodes(G, pos,
                       nodelist=list(G.nodes))
out = nx.draw_networkx_edges(G, pos,
                             edgelist=list(G.edges),
                             style='dashed',
                             #connectionstyle="arc3,rad=0.3",
                             )

```

image

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -492,7 +492,7 @@ def draw_networkx_edges(
     node_size=300,
     nodelist=None,
     node_shape="o",
-    connectionstyle=None,
+    connectionstyle="arc3",
     min_source_margin=0,
     min_target_margin=0,
 ):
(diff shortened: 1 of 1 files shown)


rossbar (approved):
Thanks for the quick follow-up. This fixes the visibility issue of the fancy arrow patch in #4360 and AFAICT the switch in the default value for the `connectionstyle` kwarg from `None` -> `arc3` doesn't affect the edge drawing for non-self-loops. LGTM!

jarrodmillman (approved):

PR #4380, opened 2020-11-23
Add an algorithm to produce maximum induced d-club cluster graphs.

aalekhpatel07:
### The theory that this PR is based on.

Given an undirected simple graph , let  denote an **induced subgraph** of  with the following property:
- Every connected component  of  has diameter at most .

Then  is called a -club cluster graph. Furthermore, if  has the largest number of edges amongst all -club cluster graphs of , then

denotes the minimum number of edges to be deleted from  to produce a -club cluster graph.

### What does this PR include?
In this PR, I have added an algorithm to compute  (along with a set of edges, i.e. a **minimum edge-cut**) if  is a forest (i.e. a collection of pair-wise disjoint trees), or if  and  is a general graph.

- For forests, the implemented algorithm runs in  where  denotes the largest degree of a vertex in .

- For general graphs  where , the algorithm relies on an efficient implementation of Maximum Matching in General Graphs. It uses the **Edmond's Blossom algorithm** implemented in [`networkx.algorithms.matching`](https://github.com/networkx/networkx/blob/master/networkx/algorithms/matching.py) and due to that, it runs in 

Multiple computational tests have been performed on random trees and random graphs to ensure that the algorithm produces a -club cluster graph from . Due to the exponential nature of an optimality test using the brute-force approach, we rely on the proof of correctness provided in a paper (currently WIP) by Oellermann, O., and Patel, A (2020).

### Some additional comments:

There is plenty of room for further advancement by considering other graph families  and implementing  for .

*FWIW, it is known that the problem of producing maximum induced -club cluster graph for arbitrary graphs  where  is **NP-Complete**.

A [Diploma Thesis of Alexander Shafer](https://pure.mpg.de/rest/items/item_1587743_4/component/file_1587742/content) has some interesting algorithms that one can implement to extend support for various other graph families like **Interval Graphs**, **Planar Graphs**, **Bipartite Graphs**, and graphs of **bounded Treewidth/Cliquewidth**.

### DIFF
--- a/networkx/algorithms/d_fragmentation.py
+++ b/networkx/algorithms/d_fragmentation.py
@@ -100,10 +100,7 @@ References
 """
 
 
-from typing import (
-    Dict,
-    Generator
-)
+from typing import Dict, Generator
 
 import networkx as nx
 from networkx.utils import not_implemented_for
--- a/networkx/algorithms/d_fragmentation.py
+++ b/networkx/algorithms/d_fragmentation.py
@@ -1,0 +1,378 @@
+"""
+Algorithm for computing a :math:s -club cluster graph (an undirected graph where
+all connected components have diameter at most :math:s) from a given graph :math:G.
+
+Given a positive integer :math:d, and an undirected simple graph :math:G,
+produce a d-club graph :math:G_d that is an induced subgraph
+of :math:G. Note that :math:G_d is not unique to :math:G.
+
+*d-fragmentation* is the process of removing a minimum number of edges from a
+graph :math:G such that the resultant graph is a *d-club* cluster graph. See [1]_ for a formal
+definition of a *d-club*.
+
+This implementation is for simple undirected forests and
+is based on the linear time algorithm presented in [2]_.
+
+
+Examples
+--------
+
+>>>
+>>> # A forest F with two trees.
+... F = nx.Graph()
+>>> F.add_edges_from(
+...     [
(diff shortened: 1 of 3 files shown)


networkx/algorithms/d_fragmentation.py
```diff
@@ -0,0 +1,375 @@
...
+
+    Parameters
+    ----------
+    G: A networkX graph.
+
+    Returns
+    -------
+    True if G is a forest, False otherwise.
+    """
+    return recognition.is_forest(G)
```
Erotemic, line 130: Why implement a custom function rather than just using `recognition.is_forest` anywhere where you would use this. There is only one place where it is used and it is decorated by `@not_implemented_for("directed", "multigraph")` anyway

aalekhpatel07: I agree with you!

Initially, I thought I might use it multiple times so I wanted a protected wrapper with a friendly name. But I ended up using it only once in the end. That has been corrected now.

Also, I have made some revisions.
**The algorithm suggested for d = 1 for arbitrary graphs is not correct.**

Note that the problem of producing a 1-club cluster graph from a graph G is the same as finding a family of pair-wise disjoint cliques such that the number of edges between the cliques is minimum.

The max weight matching approach is valid for weighted forests with d = 1. To resolve it, I have completely removed the support for graphs that are not forests.

In the future though, more algorithms can be added.

PR #4381, opened 2020-11-23
Add dissolve polygons example

brendan-ward:
This example builds on #4366 to add an example using networkx + geopandas to identify and dissolve spatially adjacent polygons, without using PySAL (already well demonstrated in #4366).

The goal here was to show the utility of networkx to handle a geospatial processing task that doesn't otherwise focus on networks as a primary outcome.  I'll leave it to the maintainers here to suggest if this belongs here or should be a linked example that lives in GeoPandas instead.

This may require adding rtree or pygeos (I recommend pygeos) to the example dependencies, depending on how geopandas is installed for those.

/cc @ljwolf @martinfleis


jarrodmillman:
Thanks!  I'll let @ljwolf @martinfleis provide feedback on the content.  For the failing style / format test, you just need to install black
```
pip install black
```
and run it on the files you are changing.  You can automate this using a pre-commit hook
```
pre-commit install
```

jarrodmillman:
Adding the edit comment from above (in case I edited the comment after you read it).

Also ``sphinx_gallery`` isn't running it b/c it only runs examples name ``plot_<something>.py``.  So just rename the example
``examples/geospatial/plot_dissolve_polygons.py``.

jarrodmillman:
@brendan-ward I suspect you are aware of this, so this comment is mainly to keep others from trying to investigate the failed test.

``nuts1.geojson`` is added in #4366.  So when that is merged and this is rebased on master, the test should no longer fail due to the missing file.

examples/geospatial/plot_dissolve_polygons.py
```diff
@@ -0,0 +1,107 @@
...
+### Read in example data.
+# Input file format could be any format supported by GeoPandas.
+filepath = "nuts1.geojson"
+european_regions = geopandas.read_file(filepath)
+
+### Spatially join regions to themselves using the spatial index.
+# This finds all regions that spatially intersect other regions.
+pairs = pandas.DataFrame(geopandas.sjoin(european_regions, european_regions))[
+    ["index_right"]
+]
```
martinfleis, line 49: I know this does the job, but seeing it used like this in documentation is not making me happy since it is not very straightforward and not the primary use of sjoin. That is the reason for something like `touches_matrix` proposed in https://github.com/geopandas/geopandas/pull/1674.

martinfleis (commented):
My feeling is that this example is overly (geo)pandas-heavy for networkx documentation (which is involved in two lines only). 

It is clearly a sub-optimal way of getting to the final result (I think it can be two lines with pysal) which may be too complex for examples anyway, no matter if here on in GeoPandas.

brendan-ward:
Added https://github.com/geopandas/geopandas/issues/1715 to take up discussion of this in geopandas instead.

This example certainly doesn't need to be here, but a crosslink from networkx to an example like this in geopandas might be helpful.  In terms of emphasis, @martinfleis is right, the spatial analysis side is the primary beneficiary of this approach rather than the network analysis side.

Closing as this doesn't seem like a good fit here, but feel free to reopen if this is something that would be helpful to keep here in `networkx` as well.

PR #4395, opened 2020-11-27
NXEP 3: Allow generators to yield from edgelists

dschult:
This NXEP presents ideas from #3036 as a reformulation of how graph generators produce output.
They can still return graphs as they currently do, but they can also yield from edgelists.

The generators are replaced with Graph Builders that allow create_using and edgelist options to be specified by the user.

The current edgelist reading tools like `G.update` will need to handle 1-tuples as potentially isolated nodes and also handle attribute dicts for graph, node, edge and multiedge objects as part of the edgelist stream.  The term edgelist may need to change to node_and_edgelist or something better.

We should create edgelist writing methods -- perhaps as part of `G.edges()` or perhaps a new method name. They need to basically be:  `chain( [(G.graph,)], map(tuple, G.nodes.data()), map(tuple, G.edges.data()))` (or with edge keys for multigraphs).

### DIFF
--- a/doc/developer/nxeps/index.rst
+++ b/doc/developer/nxeps/index.rst
@@ -11,6 +11,7 @@ NetworkX Enhancement Proposals (NXEPs) document major changes or proposals.
    nxep-0000
    nxep-0001
    nxep-0002
+   nxep-0003
 
 .. toctree::
    :hidden:
--- a/doc/developer/nxeps/nxep-0003.rst
+++ b/doc/developer/nxeps/nxep-0003.rst
@@ -43,6 +43,22 @@ that is desired. The latter is preferred stylistically because the main
 idea of the code phrase is to create a path graph and that comes first.
 Details such as what graph type to use are specified later in the phrase.
 
+Separating edge generation from graph data structure creation
+arguably makes a cleaner interface where independent tools can be put
+together in creative ways. To the extent that users need to generate
+edges rather than graphs, having an edge generator that doesn't store
+the graph is an advantage. It's not exactly clear how much demand there
+is for this feature. But e.g. `nx.utils.pairwise` would no longer be needed
+if we had `nx.path_edges(node_iterable)`.
+
+The `create_using` parameter is a mechanism to tell the function what
+class of graph data structure to start with. Separating edge generation
+from graph construction would mean the edge generator function would
+no longer need a type for the graph data structure since there isn't one.
+This NXEP proposes one way to provide an interface that separates edge
+generation from graph data structure creation when desired, while leaving
+a user-friendly mechanism for selection the graph type when desired.
+
 The changes proposed involve any nx function that creates a graph or an
 edgelist. The proposal is to make these functions return a graph of
 any type or an edgelist based on the choice indicated by the user.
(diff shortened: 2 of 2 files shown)


jarrodmillman:
@boothby We are planning to merge this soon (it is draft status so there is plenty of time to change or reject this later).  But I wanted to make sure you had a chance to review this first (if you have time at the moment).  If you are too busy at the moment, no pressure.

boothby:
Awesome!  Life has been getting in the way, but I'll set aside some time to read this tonight.

boothby (approved):
This looks good overall.  I'll try to get an implementation together in the next week or so.

rossbar (commented):
Thanks for writing this up! I'll admit that I had tried to follow the discussion in #3036 in the past and had trouble seeing the bigger picture, which this NXEP provides.

The most immediate thing that jumped out to me was: could this be structured in a way so that the end result is fully backward compatible? And if so, is there a reason not to? It seems one of the motivations is to *add* the ability to access only the edgelist generated by the "graph builder" rather than require a Graph object be created in the process. Would it be possible so that the builder supported both the `create_using=` and method-like syntaxes (e.g. `nx.path_graph(10, create_using=nx.DiGraph)` & `nx.path_graph.DiGraph(10)`)? 

The reason I think this is worth considering is that this would be a pretty significant API change - presumably every user of NetworkX relies on the graph creation functionality. Changing the API to drop the `create_using=` kwarg would affect a lot of downstream code. If there were a way to continue supporting `create_using` while simultaneously enabling the new syntax, that offers an avenue for incremental adoption without breaking users. OTOH, having two ways of doing the same thing is not exactly clean design, and there may be other reasons that `create_using=` is not desireable (which I either missed in my first reading, or could be added to the Motivatoin section).

dschult:
That approach to the API change where we leave `create_using` and also allow `nx.path_graph.DiGraph` is how we have usually done it in the past. As you point out, it reduces backward incompatibility problems -- but allows two ways to do something.  You can see echos of that in `G.edges()` vs `G.edges` and `G.degree(n)` vs `G.degree[n]` vs `G.degree`. 
Indeed there already is two ways to enact `create_using`.  `nx.DiGraph(nx.path_graph(9))` gives the same result as `nx.path_graph(9, create_using=nx.DiGraph)`.

As far as I know the argument for removing create_using is the idea that generators are usually, at their heart, simply edge generators. There is a feeling that a cleaner interface is possible if we separate the edge generation from storage in the data structure. But in my understanding, we never really came up with a better approach than `nx.DiGraph(nx.path_edges(9))`. I'm not sure this NXEP approach does that separation effectively either... :{

jarrodmillman:
I am, in principle, OK with breaking backward compatibility here.  But I don't know whether this NXEP is the best way forward (mostly for a lack of thinking about it on my part than any issue with the proposal).  I trust Dan's and Kelly's opinion about this more than my own (so I am almost certain to support anything the two of you agree on).  Since Dan said he isn't "sure this NXEP approach does that separation effectively," I wonder if a better solution will be found.

I suspect a working implementation/prototype (from @boothby ) may reveal more information.  I hate "to ask" Kelly to create a PR that may not ultimately be accepted, but it may be the best way forward if they are still willing to work on it knowing the NXEP is still a draft (and hasn't been accepted yet).  Of course, this NXEP is based on a longer conversation, so it isn't like this is just coming from nowhere and is likely to be rejected.  (I.e., it seems like Dan and Kelly have mostly agreed this is the right thing to do.)

Maybe Dan could add a short paragraph or two (I am imaging more or less cutting-and-pasting his comments from above) to the backward compatibility section explaining why we are breaking backward compatibility.

jarrodmillman:
If we remove ``create_using``, we should have a deprecation.  Make sure to add
1. a DeprecationWarning
2. a line to doc/release/release_dev.rst
3. a line to doc/developer/deprecations.rst
4. a line to  networkx/conftest.py

jarrodmillman (approved):
I like the new API!  I approve of this being merged in draft state (we can edit and change it to accepted in the implementation PR).

dschult:
This should be ready to merge. I added @boothby as author and added more motivation and alternatives.

boothby:
>  I hate "to ask" Kelly to create a PR that may not ultimately be accepted, but it may be the best way forward if they are still willing to work on it knowing the NXEP is still a draft (and hasn't been accepted yet).

No worries there.  It's mostly written in bits and pieces throughout the conversation, it's just a matter of patching it together and writing a few sample generators.

With regards to backwards-compatibility, I think that we can possibly get away without any breakage there. The proposal mentions moving generators into the `nx.builders` namespace.  Perhaps we can maintain the `nx.generators` namespace with an automated wrapper?  This is a little half-baked so disagreement is more than welcome.

PR #4398, opened 2020-11-28
fix small index iteration bug in kernighan_lin algorithm

jelgun:
The algorithm works incorrectly because of that bug as min_i in line 124 always gets 0.

### DIFF
--- a/networkx/algorithms/community/kernighan_lin.py
+++ b/networkx/algorithms/community/kernighan_lin.py
@@ -127,7 +127,7 @@ def kernighan_lin_bisection(G, partition=None, max_iter=10, weight="weight", see
         if min_cost >= 0:
             break
 
-        for _, _, (u, v) in costs[: min_i]:
+        for _, _, (u, v) in costs[:min_i]:
             side[u] = 1
             side[v] = 0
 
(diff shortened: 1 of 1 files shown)


dschult:
Thank You!!

dschult (approved):

rossbar (commented):
IMO the chained assignments made this difficult to read. I think it would make things more clear (and make errors like this less likely) to rewrite e.g. `i = totcost = 0` either as `i, totcost = 0, 0` or

```python
i = 0
totcost = 0
```
One other quick question just to double-check: the increment on `i` should happen after the yield, not before?

jarrodmillman:
It would be nice to have a test for this, if you have the time.

jelgun:
> IMO the chained assignments made this difficult to read. I think it would make things more clear (and make errors like this less likely) to rewrite e.g. `i = totcost = 0` either as `i, totcost = 0, 0` or
> 
> ```python
> i = 0
> totcost = 0
> ```
> 
> One other quick question just to double-check: the increment on `i` should happen after the yield, not before?

I agree. Did some refactoring. Incrementing 'i' before the yield actually was not a bug as it was handled differently on return.

rossbar (approved):
I agree that adding a test (or tests) for `_kernighan_lin_sweep` would be a big improvement, though this is a clear bug so +1 for fixing and not letting tests be a blocker.

PR #4400, opened 2020-11-29
[WIP] Add first pass at panther++

mrecachinas:
As discussed, follow on to #3886.

- Need to figure out how to convert `similarities` (i.e., `1 / || theta(v_i) - theta(v_j) ||`) into a true similarity metric. Perhaps just normalized to between 0 and 1?
- Probably should pull out the common core of `panther_similarity` and `panther_vector_similarity`
- Rename `panther_similarity` to `panther_path_similarity`?

### DIFF
--- a/networkx/algorithms/similarity.py
+++ b/networkx/algorithms/similarity.py
@@ -1596,6 +1596,129 @@ def panther_similarity(G, source, k=5, path_length=5, c=0.5, delta=0.1, eps=None
     return top_k_with_val
 
 
+def panther_vector_similarity(
+    G, source, D=50, k=5, path_length=5, c=0.5, delta=0.1, eps=None
+):
+    """Returns the Panther vector similarity of nodes in the graph ``G``
+    to node ``v``. Panther is a similarity metric that says
+    "two objects are considered to be similar if their structure
+    is " [1]_.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+        A NetworkX graph
+    source : node
+        Source node for whom to find the top ``k`` similar other nodes
+    D : int
+        The number of similarity scores to use (in descending order)
+        for each feature vector. Defaults to 50 per [1]_
+    k : int
+        The number of most similar nodes to return
+    path_length : int
+        How long the randomly generated paths should be (``T`` in [1]_)
+    c : float
+        A universal positive constant. Defaults to 0.5 per [1]_
+    delta : float
+        The probability that $S$ is not an epsilon-approximation to (R, phi)
+    eps : float
+        The error bound. Per [1]_, a good value is ``sqrt(1/|E|)``. Therefore,
+        if no value is provided, the recommended computed value will be used.
+
+    Returns
+    -------
+    similarity : dictionary
+        Dictionary of nodes to similarity scores (as floats). Note:
+        the self-similarity (i.e., ``v``) will not be included in
+        the returned dictionary.
+
+    Examples
+    --------
+
+        >>> import networkx as nx
+        >>> G = nx.star_graph(10)
+        >>> sim = nx.panther_vector_similarity(G, 0)
+
+    References
+    ----------
+    .. [1] Zhang, J., Tang, J., Ma, C., Tong, H., Jing, Y., & Li, J.
+           Panther: Fast top-k similarity search on large networks.
+           In Proceedings of the ACM SIGKDD International Conference
+           on Knowledge Discovery and Data Mining (Vol. 2015-August, pp. 1445–1454).
+           Association for Computing Machinery. https://doi.org/10.1145/2783258.2783267.
+    """
+    from scipy.spatial import cKDTree as KDTree
+    import numpy as np
+
+    num_nodes = G.number_of_nodes()
+    if num_nodes < k:
+        warnings.warn(
+            f"Number of nodes is {num_nodes}, but requested k is {k}. "
+            "Setting k to number of nodes."
+        )
+        k = num_nodes
+
+    # According to [1], they empirically determined
+    # a good value for ``eps`` to be sqrt( 1 / |E| )
+    if eps is None:
+        eps = np.sqrt(1.0 / G.number_of_edges())
+
+    inv_node_map = {name: index for index, name in enumerate(G.nodes)}
+    node_map = np.array(G)
+
+    # Calculate the sample size ``R`` for how many paths
+    # to randomly generate
+    t_choose_2 = _n_choose_k(path_length, 2)
+    sample_size = int((c / eps ** 2) * (np.log2(t_choose_2) + 1 + np.log(1 / delta)))
+    index_map = {}
+    _ = list(
+        generate_random_paths(
+            G, sample_size, path_length=path_length, index_map=index_map
+        )
+    )
+    S = np.zeros((num_nodes, D))
+
+    theta = np.zeros((num_nodes, D))
+    inv_sample_size = 1 / sample_size
+
+    # Calculate the path similarities
+    # for each node against each other node
+    for vi_idx, vi in enumerate(G.nodes):
+        # Calculate the path similarities
+        # between ``source`` (v) and ``node`` (v_j)
+        # using our inverted index mapping of
+        # vertices to paths
+        for node, paths in index_map.items():
+            # Only consider paths where both
+            # ``node`` and ``source`` are present
+            common_paths = set(index_map[vi]).intersection(paths)
+            S[vi_idx, inv_node_map[node]] = len(common_paths) * inv_sample_size
+
+        # Build up the feature vector using the
+        # largest ``D`` similarity scores
+        theta[vi_idx] = np.sort(np.partition(S[vi_idx], -D)[-D:])[::-1]
+
+    # Insert the feature vectors into a k-d tree
+    # for fast retrieval
+    kdtree = KDTree(theta)
+
+    # Retrieve top ``k`` similar vertices (i.e., vectors)
+    # (based on their Euclidean distance)
+    neighbor_distances, nearest_neighbors = kdtree.query(theta[source], k=k)
+
+    # The paper defines the similarity S(v_i, v_j) as
+    # 1 / || Theta(v_i) - Theta(v_j) ||
+    similarities = np.reciprocal(neighbor_distances)
+
+    # Add back the similarity scores (i.e., distances)
+    top_k_sorted_names = map(lambda n: node_map[n], nearest_neighbors)
+    top_k_with_val = dict(zip(top_k_sorted_names, similarities))
+
+    # Remove the self-similarity
+    top_k_with_val.pop(source, None)
+    return top_k_with_val
+
+
 def generate_random_paths(G, sample_size, path_length=5, index_map=None):
     """Randomly generate `sample_size` paths of length `path_length`.
 


rossbar:
@mrecachinas thanks for this PR and #3886!

There will be a NetworkX release in the near future (O[weeks]) which will have `panther_similarity` in it. One thing we would like to try to confirm is that the API for Panther and Panther++ is consistent, and that future development in this PR won't necessitate changes in the `panther_similarity` API. For example, in one of the bullets above, you mention changing the name from `panther_similarity` to `panther_path_similarity`. Changing the function name/signature for `panther_similarity` is painless now, but will require a deprecation cycle after release. I just wanted to check to see how happy you were with the current function names and signatures. Ideally we'd be able to get the function names and signatures settled prior to the NX 2.6 release. LMK what you think!

mrecachinas:
@rossbar I'm good with either -- definitely up to you all. The paper refers to the actual Panther path similarity algorithm simply as "Panther" and the vector similarity algorithm as "Panther++" -- 

image

I wasn't sure about naming it `panther_plus_plus` (since can't do `panther++`), but I'm happy with whichever you prefer.

storopoli:
Guys, since I was the one who brought the panther paper as a feature request, I will chip in and give my two cents.

Since the algorithm and code returns a dictionary of nodes, I prefer `panther_similarity` without any `paths`. We are not returning paths I guess only similar nodes.

Also `panther_similarity_plus_plus` is a little big. What about `panther_similarity_pp`?

dschult:
+1 for the `_pp` suffix. That's what VF2++ will use: either `vf2pp` or `vf2_pp`

mrecachinas:
Sounds good!

mrecachinas:
Also, wanted to revisit if @dschult @rossbar @jarrodmillman had anymore thoughts on a more permanent home for `generate_random_paths` (i.e., a Markovian random walk)?

dschult:
I think we should leave `generate_random_paths` where it is until NXEP3 is figured out. Then there will be many more "path constructors" to put this with.  (The idea is that many of our graph generators are really just constructing edges and/or paths so maybe we should make those available as edge constructors rather than -- or in addition to -- graph constructors.)

rossbar:
Re: naming, I agree with @dschult and @storopoli : **Panther** -> `panther_similarity` and **Panther++** -> `panther_similarity_pp`.

> I think we should leave generate_random_paths where it is until NXEP3 is figured out. 

I agree. If it is likely that `generate_random_paths` will be moved or wrapped up in some other object, it might be worth making it "private" for the 2.6 release (i.e. prepend `_` to the name). That way, it will still be available but the `_` will indicate to users that there may be a name change/API reshuffle on the horizon. OTOH this may be overly cautious, and the arguments for moving it somewhere else may never come to anything. I'm ambivalent, but if others think that there is a very high likelihood that we'd want to move `generate_random_paths` to a new location in the near future, it's worth thinking about.

jarrodmillman:
I would go with `_generate_random_paths` for the 2.6 release, and revisit it with NXEP3.  The rename could be a separate PR, since it should be merged quickly.  It probably makes sense to add a note to NXEP3 mentioning `_generate_random_paths` in the same rename PR, so that we don't forget to revisit this.

dschult:
The problem with making it a private function is that people won't see it or know they can use it. This is potentially a useful function for sampling paths. We shouldn't hide it.  If we use an underscore and still put it in the `__all__` variable and have it in the docs and in the main namespace, then I'm fine with making it private. But I'd prefer not to hide it from the documentation.

If it was up to me, I would leave it where it is without an underscore. If we get NXEP3 in place for v3.0 then I feel like we can still move the function to a new location within the package for v3.0. The API for the function will be the same. The function will still be the same -- we're only deprecating the location within the package. `nx.generate_random_paths` will work exactly as it does now.  Only users who used the full path to the function will have to change, and the error they get will be clear and easy to fix.

I'm open to adding the `_` if that's what should be done -- I'm just wanting verification that changing the location of a function within the package is a big enough change that the cost of keeping this function private is worth the benefit.

jarrodmillman:
I think we all understand the arguments for making it public or private.  So I will leave it to @dschult to decide.  It if is made public in 2.6 and deprecated in 2.7, we can still remove it in 3.0.  If it is deprecated after the 3.0 release, it can be removed in 3.2.

jarrodmillman:
@dschult Here is another suggestion (but you and Ross should make the call on this):

We could mark the function private in `networkx/algorithms/similarity.py`; but then import it in the top-level namespace as follows:
```
from networkx.algorithms.similarity import _generate_random_paths as generate_random_paths
```

dschult:
Interesting... Lots of Options...  
Maybe @rossbar and I can explore the possibilities and come up with a rationale for when to do which. :}

rossbar:
TL;DR - I agree with @dschult , I'm +1 on having `generate_random_paths` remain fully public (no `_`-prepended) for 2.6 and beyond.

I think the points in @dschult 's points are the most compelling. One thing that I hadn't considered was that `generate_random_paths` is available in the top-level namespace (i.e. `nx.generate_random_paths`), so which pkg/module it is defined in is basically an implementation detail that isn't interesting for most *users*. It can be moved to pretty much any module, including a new one - as long as it's still imported into the top-level namespace, the usage-impact will be minimal. Of course, moving it would cause minor changes, e.g. where it can be found in the reference docs etc, but the search bar alleviates that concern.

mrecachinas:
Huge apologies for the ...4 year delay @dschult @rossbar @storopoli @jarrodmillman 🙇 

Finally had some time this weekend to take a look at what was going on, finish the implementation, and get it passing CI. Let me know if you have any questions/comments after you review.

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+@nx._dispatchable(edge_attrs="weight")
+def panther_vector_similarity(
+    G, source, D=50, k=5, path_length=5, c=0.5, delta=0.1, eps=None, weight="weight"
+):
+    r"""Returns the Panther vector similarity of nodes in the graph ``G``
+    to node ``v``. This is known as "Panther++".
+
+    Panther is a similarity metric that says
+    "two objects are considered to be similar if their structures
+    are similar" [1]_.
```
dschult, line 1655:
```suggestion
    r"""Returns the Panther vector similarity of nodes in `G`.

     This approach to similarity is often called "Panther++" and the notion of
     similarity referred to here is low "graph edit distance". Panther is a
     similarity metric that says "two objects are considered to be similar
     if their structures are similar" [1]_. In this case, if their graph edit
     distances are small.
```

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+    weight : string or None, optional (default="weight")
+        The name of an edge attribute that holds the numerical value
+        used as a weight. If None then each edge has weight 1.
+
+    Returns
+    -------
+    similarity : dictionary
+        Dictionary of nodes to similarity scores (as floats). Note:
+        the self-similarity (i.e., ``v``) will not be included in
+        the returned dictionary.
```
dschult, line 1686:
```suggestion
    similarity : dict
        Dict of nodes to similarity scores (as floats).
        Note: the self-similarity (i.e., `node`) is not included in the dict.
```

dschult:
Welcome back! :)

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
     return top_k_with_val
 
 
+@nx._dispatchable(edge_attrs="weight")
+def panther_vector_similarity(
+    G, source, D=50, k=5, path_length=5, c=0.5, delta=0.1, eps=None, weight="weight"
```
dschult, line 1648: Let's make these parameters keyword only -- since this is a new function that will lead to easier to read code using this. 
```suggestion
    G, source, *, D=50, k=5, path_length=5, c=0.5, delta=0.1, eps=None, weight="weight"
```

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+        Source node for whom to find the top ``k`` similar other nodes
+    D : int
+        The number of similarity scores to use (in descending order)
+        for each feature vector. Defaults to 50 per [1]_
+    k : int
+        The number of most similar nodes to return
+    path_length : int
+        How long the randomly generated paths should be (``T`` in [1]_)
+    c : float
+        A universal positive constant. Defaults to 0.5 per [1]_
```
dschult, line 1672: Say what this constant does -- (the fact that it is constant can be said about any of these parameters).
Something like "Scale of the sample size when computing ___."

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+    delta : float
+        The probability that $S$ is not an epsilon-approximation to (R, phi)
+    eps : float
+        The error bound. Per [1]_, a good value is ``sqrt(1/|E|)``. Therefore,
+        if no value is provided, the recommended computed value will be used.
```
dschult, line 1677: Need to say which error you are talking about. And "the recommended computed value" doesn't mean anything without more information. Give the exact formula if you'd like. But be more specific in any case.

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+            f"Panther similarity is not defined for the isolated source node {source}."
+        )
+
+    num_nodes = G.number_of_nodes()
+    if num_nodes < k:
+        warnings.warn(
+            f"Number of nodes is {num_nodes}, but requested k is {k}. "
+            "Setting k to number of nodes."
+        )
+        k = num_nodes
```
dschult, line 1722: I'd prefer to raise when k > len(G) instead of trying to guess what the user wanted.
Same is true for D and maybe others.

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+    # Ensure D doesn't exceed the number of nodes
+    if num_nodes < D:
+        warnings.warn(
+            f"Number of nodes is {num_nodes}, but requested D is {D}. "
+            "Setting D to number of nodes."
+        )
+        D = num_nodes
+
+    # According to [1], they empirically determined
+    # a good value for ``eps`` to be sqrt( 1 / |E| )
```
dschult, line 1733:
```suggestion
    # In [1], an empirically good value for ``eps`` is sqrt( 1 / |E| )
```

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1643,170 @@ def panther_similarity(
...
+    # Calculate the sample size ``R`` for how many paths
+    # to randomly generate
+    t_choose_2 = math.comb(path_length, 2)
+    sample_size = int((c / eps**2) * (np.log2(t_choose_2) + 1 + np.log(1 / delta)))
+    index_map = {}
+    _ = list(
+        generate_random_paths(
+            G, sample_size, path_length=path_length, index_map=index_map, weight=weight
+        )
+    )
```
dschult, line 1749: If the output here is not used, you don't need to assign it. Just call `generate_random_paths`. Thought either way it might help to have a comment soemthing like: `# calling to fill index_map`

networkx/algorithms/tests/test_similarity.py
```diff
@@ -949,3 +949,136 @@ def user_match(x, y):
...
+        np.random.seed(42)
+
+        G = nx.Graph()
+        G.add_edge(0, 1)
+        G.add_edge(0, 2)
+        G.add_edge(0, 3)
+        G.add_edge(1, 2)
+        G.add_edge(2, 4)
+
+        # Just verify that the function works and returns results
```
dschult, line 964:
```suggestion
```

networkx/algorithms/tests/test_similarity.py
```diff
@@ -949,3 +949,136 @@ def user_match(x, y):
...
+        G = nx.Graph()
+        G.add_edge(0, 1)
+        G.add_edge(0, 2)
+        G.add_edge(0, 3)
+        G.add_edge(1, 2)
+        G.add_edge(2, 4)
+
+        sim = nx.panther_vector_similarity(G, 0, D=3, k=4, path_length=2)
+
+        # Check that we get the expected number of results
```
dschult, line 983: Don't need these comments -- the python code itself is almost english. :)

dschult (commented):
Here are some comments after a pass through this PR. I am still putting together big picture perspectives... but it looks good.

Can you have the docs explain why the function name is what it is? The docs don't talk about vector. Also how does it differ from panther_sim...? Why/when might you want to use this?

mrecachinas:
@dschult I think I added changes for each of your review comments. Let me know if that looks more in line with what you were thinking.

Also, let me know if I should make similar adjustments in `panther` to the docstring and change the warnings to exceptions. This feels like an API-breaking change, so I'm happy either way.

dschult:
Thanks for this question -- I hadn't opened up the file wider to see the similarity between panther and panther++.  MNost of that code is the same -- checking inputs and computing the panther path stats.

To help us avoid divergence in those code-paths does it make sense to have a helper function that computes as much as it can that is common?  Then if someone fixes an issue for e.g. panther, it will get fixed for panther++ too without having to remember to change it in both places. You can make that function "private" by starting the name with an underscore `_` and  dont put it in `__all__` at the top. Also don't worry about docs more than comments to help code-readers understand what is going on. 

I think that code change will also help me easily see what is different between similarities and vector similarities. The panther path generation will be the same, and the processing of the paths will hold the key to knowing the difference.

And -- yes -- let's update the panther code to match the panther++.  I'll bring up your questions about backward compatibility at the community meeting. The change will not lead to a hidden (surprise!) change -- it will raise an exception with a hopefully helpful message so it will be clear how to fix their code. That is, of course a change -- but better than a potentially hidden change.

dschult (approved):
This looks good to me!

networkx/algorithms/similarity.py
```diff
@@ -1505,6 +1507,124 @@ def _simrank_similarity_numpy(
...
+    t_choose_2 = math.comb(path_length, 2)
+    sample_size = int((c / eps**2) * (np.log2(t_choose_2) + 1 + np.log(1 / delta)))
+    index_map = {}
+
+    # Generate the random paths and populate the index_map
+    list(
+        generate_random_paths(
+            G, sample_size, path_length=path_length, index_map=index_map, weight=weight
+        )
+    )
```
amcandio, line 1614: Do you need to create the whole list here? Might be better to do something like
```suggestion
        for _ in generate_random_paths( 
             G, sample_size, path_length=path_length, index_map=index_map, weight=weight
        ):
            pass
```

mrecachinas: Good point.

amcandio (approved):
Very cool job! Left some minor comments / suggestions but this look good to me

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1751,168 @@ def panther_similarity(
...
+        # vertices to paths
+        for node, paths in index_map.items():
+            # Only consider paths where both
+            # ``node`` and ``source`` are present
+            common_paths = set(index_map[vi]).intersection(paths)
+            S[vi_idx, inv_node_map[node]] = len(common_paths) * inv_sample_size
+
+        # Build up the feature vector using the
+        # largest ``D`` similarity scores
+        theta[vi_idx] = np.sort(np.partition(S[vi_idx], -D)[-D:])[::-1]
```
amcandio, line 1876: Do you really need this matrix S? I see you use it to compute the top D scores. What about creating a new heap per node in the for loop and keeping the top D there?

networkx/algorithms/similarity.py
```diff
@@ -1505,6 +1507,124 @@ def _simrank_similarity_numpy(
...
+    )
+
+    # Return a named tuple with all the required data
+    return PantherPaths(
+        G=G,
+        inv_node_map=inv_node_map,
+        node_map=list(G),
+        sample_size=sample_size,
+        index_map=index_map,
+        inv_sample_size=1 / sample_size,
```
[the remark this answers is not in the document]
mrecachinas: So `PantherPaths` is more of a way to return the results from `_prepare_panther_paths` in a NamedTuple (i.e., without just returning 6-7 variables and worrying about any weirdness with unpacking/indexing). Also, `PantherPaths`, along with `_prepare_panther_paths` is scoped to this module and, more importantly not exposed outside of it.

To answer your question though: I don't think it makes sense for it to be 0 given that would blow up `inv_sample_size`, which is used later on, which would blow up the similarity metrics.

networkx/algorithms/similarity.py
```diff
@@ -1505,6 +1507,123 @@ def _simrank_similarity_numpy(
...
+# Define a named tuple for Panther paths data
+PantherPaths = namedtuple(
+    "PantherPaths",
+    [
+        "G",
+        "inv_node_map",
+        "node_map",
+        "sample_size",
+        "index_map",
+        "inv_sample_size",
```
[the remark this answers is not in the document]
mrecachinas: Good catch!

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1751,168 @@ def panther_similarity(
...
+    for vi_idx, vi in enumerate(G.nodes):
+        # Calculate the path similarities
+        # between ``source`` (v) and ``node`` (v_j)
+        # using our inverted index mapping of
+        # vertices to paths
+        for node, paths in index_map.items():
+            # Only consider paths where both
+            # ``node`` and ``source`` are present
+            common_paths = set(index_map[vi]).intersection(paths)
+            S[vi_idx, inv_node_map[node]] = len(common_paths) * inv_sample_size
```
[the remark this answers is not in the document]
mrecachinas: No it's fair feedback to avoid recomputing the set in each inner iteration. I'm not sure what I'd sum over since the intersection will return a result though.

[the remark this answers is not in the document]
amcandio: The sum would save you the intersection set creation and mem allocation. I meant sum(1 for x in set1 if x in set2) (sorry for bad formatting, reviewing from my phone 😅)

This is only significant if the intersection is big enough, otherwise it's just a micro-optimization. Intersection+len is nicer/clearer

networkx/algorithms/similarity.py
mrecachinas: Good idea. Take a look at https://github.com/networkx/networkx/pull/4400/commits/af878a2dc084e842b2e0e5f658ebee485f1b6107.

Idle thoughts:
- this should increase memory efficiency -- before I was calculating the NxN matrix (`O(n^2)` memory), but now it's just the top `D` scores (`O(n*D)`)
- time complexity will be the same, but we'll be avoiding sorting on each run

mrecachinas:
Not sure why test/extra is failing, but it looks unrelated?

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1749,187 @@ def panther_similarity(
...
+    # Pre-convert index_map values to sets for faster intersection
+    index_map_sets = {node: set(paths) for node, paths in index_map.items()}
+
+    # Calculate the path similarities for each node
+    for vi_idx, vi in enumerate(G.nodes):
+        # Get the set of paths for the current node
+        vi_paths = index_map_sets[vi]
+
+        # Use a min heap to track the top D scores
+        # Since heapq is a min heap, we negate scores to track largest values
```
amcandio, line 1869: I think you want to pop the minimum, right? So you shouldn't negate values

mrecachinas: No, higher similarity score == more similar, so we want a max heap here.

amcandio: When you do pop, do you want to pop the minimum score or the maximum score?

amcandio: Please bear with my confusion 😅 If you pop the maximum it means that you end up with the smallest scores in your heap. Is that what you are trying to do?

amcandio: So let's say your scores are 1,2,3,4,5 and you want the highest 3.

With a min heap (without negating) you will pop 1 and 2, ending up with 3, 4 and 5.

With a max heap (negating) you will pop -4 and -5, ending up with 1, 2 and 3.

So if you want the highest scores, you shouldn't negate. Does this make sense?

mrecachinas: Ah! I misunderstood and thought you meant when we go to pop the N largest off. Yep, I had the `elif` branch wrong where it was replacing the root node in the "max-heap".

Swapped it out to be a "min-heap" and am no longer negating.

amcandio:
Looks good to me, no more comments from my side!

dschult:
Can you merge main into this branch -- there is a fix @rossbar put in to get rid of some `extra` errors. And merging that into this will tell us if there is anything else affected there. I doubt there is anything after the found error, but it's good to make sure. :)

Thanks!!

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1749,182 @@ def panther_similarity(
...
+
+    # Pre-convert index_map values to sets for faster intersection
+    index_map_sets = {node: set(paths) for node, paths in index_map.items()}
+
+    # Calculate the path similarities for each node
+    for vi_idx, vi in enumerate(G.nodes):
+        # Get the set of paths for the current node
+        vi_paths = index_map_sets[vi]
+
+        # Use a min heap to track the top D scores
```
Peiffap, line 1868: I think this can be made faster using quickselect/introselect, but I don't know if the added code complexity is worth the reduction in time complexity (O(n + D log D) instead of O(n log D)). What are typical values of n and D?

Peiffap (commented):
Some nit-picky comments :)

amcandio:
@dschult are we good to merge or should we get more reviewers?

networkx/algorithms/tests/test_similarity.py
```diff
@@ -982,3 +982,120 @@ def user_match(x, y):
...
+    def test_panther_vector_similarity_unweighted(self):
+        """Test panther_vector_similarity with unweighted graph."""
+        G = nx.Graph()
+        G.add_edge(0, 1)
+        G.add_edge(0, 2)
+        G.add_edge(0, 3)
+        G.add_edge(1, 2)
+        G.add_edge(2, 4)
+
+        sim = nx.panther_vector_similarity(G, 0, D=3, k=4, path_length=2)
```
Peiffap, line 1013: This one doesn't use a fixed seed.

networkx/algorithms/similarity.py
```diff
@@ -1505,6 +1507,121 @@ def _simrank_similarity_numpy(
...
+    inv_node_map = {name: index for index, name in enumerate(G)}
+
+    # Calculate the sample size ``R`` for how many paths
+    # to randomly generate
+    t_choose_2 = math.comb(path_length, 2)
+    sample_size = int((c / eps**2) * (np.log2(t_choose_2) + 1 + np.log(1 / delta)))
+    index_map = {}
+
+    # Generate the random paths and populate the index_map
+    for _ in generate_random_paths(
```
Peiffap, line 1609: This call doesn't use a seed and makes the `panther*similarity` functions non-deterministic. Do we want them to be? Would it make sense to always explicitly use the same seed here, or to let the user specify a seed when calling `panther*similarity` (using an additional argument to the function)?
I'm asking because some of the tests occasionally fail if the random state is not set explicitly beforehand. I've indicated the ones I broke below, but there might be others.

If you think it's better to fix random seeds in our tests (as opposed to exposing them to users or setting them in the code), I think setting `np.random.seed(42)` works everywhere.

networkx/algorithms/tests/test_similarity.py
```diff
@@ -982,3 +982,120 @@ def user_match(x, y):
...
+
+        assert len(sim) == 3
+
+        assert 0 not in sim
+
+        for node in [1, 2, 3]:
+            assert node in sim
+
+        for node, score in sim.items():
+            assert 0 <= score <= 1
```
Peiffap, line 1023: I like how this looks in the function above
```suggestion
        assert len(sim) == 3
        assert 0 not in sim
        assert all(node in sim for node in [1, 2, 3])
        assert all(0 <= score <= 1 for score in sim.values())
```

networkx/algorithms/tests/test_similarity.py
```diff
@@ -982,3 +982,120 @@ def user_match(x, y):
...
+
+        assert len(sim) == 3
+
+        assert "v1" not in sim
+
+        for node, score in sim.items():
+            assert 0 <= score <= 1
+
+        for node in ["v2", "v3", "v4"]:
+            assert node in sim
```
Peiffap, line 1046: Same here
```suggestion
        assert len(sim) == 3
        assert "v1" not in sim
        assert all(0 <= score <= 1 for score in sim.values())
        assert all(node in sim for node in ["v2", "v3", "v4"])
```

networkx/algorithms/tests/test_similarity.py
```diff
@@ -982,3 +982,120 @@ def user_match(x, y):
...
+    def test_panther_vector_similarity_weighted(self):
+        """Test panther_vector_similarity with weighted graph."""
+        G = nx.Graph()
+        G.add_edge("v1", "v2", weight=5)
+        G.add_edge("v1", "v3", weight=1)
+        G.add_edge("v1", "v4", weight=2)
+        G.add_edge("v2", "v3", weight=0.1)
+        G.add_edge("v3", "v5", weight=1)
+
+        sim = nx.panther_vector_similarity(
```
Peiffap, line 1034: This one doesn't use a fixed seed.

networkx/algorithms/similarity.py
```diff
@@ -1505,6 +1507,121 @@ def _simrank_similarity_numpy(
...
+        k is not greater than the number of nodes in the graph.
+
+    Returns
+    -------
+    PantherPaths
+        A named tuple containing the prepared data:
+        - G: The graph (possibly with isolates removed)
+        - inv_node_map: Dictionary mapping node names to indices
+        - node_map: Array or list of node names
+        - sample_size: Number of paths to generate
```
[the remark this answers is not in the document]
Peiffap: My comment wasn't very clear. This got changed during code review and is no longer part of the named tuple.
```suggestion
```

networkx/algorithms/similarity.py
Peiffap: This is out of scope for this PR and might not even be faster in practice.

networkx/algorithms/similarity.py
```diff
@@ -1642,6 +1749,182 @@ def panther_similarity(
...
+
+            # Add to heap if we haven't reached D items yet
+            if len(top_d_heap) < D:
+                heapq.heappush(top_d_heap, similarity)
+            # Otherwise, replace the smallest value if new score is larger
+            elif similarity > top_d_heap[0]:
+                heapq.heapreplace(top_d_heap, similarity)
+
+        # Extract the top D scores in descending order
+        scores = heapq.nlargest(D, top_d_heap)
```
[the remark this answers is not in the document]
Peiffap: I checked this locally and just sorting works, no need for slices.

Peiffap (commented):
Some comments regarding randomness, mainly. Also some aesthetic suggestions that you're free to disregard. :)

networkx/algorithms/similarity.py
dschult: Yes -- let's make sure that any place that uses random number generators has a seed in the function signature (with the associated random decorator on the function) and that it passes that `seed` parameter to all internally called functions.  

Thanks @Peiffap !!

mrecachinas: Think I got everything in my two most recent commits https://github.com/networkx/networkx/pull/4400/commits/4a1ee39cca75cbcfbc030341ceb44f0ccc4b6d29 and https://github.com/networkx/networkx/pull/4400/commits/79e43c68677f6db2a0d5a2cb2f5e5be194bad6da.

networkx/algorithms/similarity.py
mrecachinas: My understanding is the typical number of nodes is 100k+ or millions, whereas k and D are usually 50-100.

networkx/algorithms/similarity.py
```diff
@@ -1642,7 +1771,198 @@ def panther_similarity(
...
+
+            # Add to heap if we haven't reached D items yet
+            if len(top_d_heap) < D:
+                heapq.heappush(top_d_heap, similarity)
+            # Otherwise, replace the smallest value if new score is larger
+            elif similarity > top_d_heap[0]:
+                heapq.heapreplace(top_d_heap, similarity)
+
+        # Extract the top D scores in descending order
+        scores = sorted(top_d_heap)
```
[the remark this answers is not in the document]
Peiffap: In an earlier state, this used [`nlargest()`](https://docs.python.org/3/library/heapq.html#heapq.nlargest), but one of the next paragraphs in that link says to use `sorted` instead for large values of `n`:
> The latter two functions perform best for smaller values of n. For larger values, it is more efficient to use the [sorted()](https://docs.python.org/3/library/functions.html#sorted) function. Also, when n==1, it is more efficient to use the built-in [min()](https://docs.python.org/3/library/functions.html#min) and [max()](https://docs.python.org/3/library/functions.html#max) functions. If repeated usage of these functions is required, consider turning the iterable into an actual heap.

As for why not pop from the heap—you could, but you end up with `lg n` complexity for each `heappop()` so it will very likely end up slower in practice than the optimized implementation in `sorted()`. And you have to reverse it...

Which does make me ask (and it's really my bad potentially since I suggested this change): shouldn't it be `sorted(top_d_heap, reverse=True)`, since we want it in decreasing order?

networkx/algorithms/similarity.py
```diff
@@ -1642,7 +1771,198 @@ def panther_similarity(
...
+        seed=seed,
+    )
+
+    G = paths.G
+    inv_node_map = paths.inv_node_map
+    node_map = paths.node_map
+    index_map = paths.index_map
+    inv_sample_size = paths.inv_sample_size
+    eps = paths.eps
+    num_nodes = G.number_of_nodes()
```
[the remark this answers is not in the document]
mrecachinas: No major reason, just a bit of an organizational thing to avoid messing up indexing (e.g., if anything was added or removed). Happy to change it to a non-named tuple.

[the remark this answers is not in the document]
mrecachinas: Addressed in [c06951b](https://github.com/networkx/networkx/pull/4400/commits/c06951b99323f55a95905329c499fc764b176fc6).

networkx/algorithms/similarity.py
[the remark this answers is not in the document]
mrecachinas: Thanks, good catch. Changing to the heap definitely took a bit of reasoning and tracing it out 😄

networkx/algorithms/similarity.py
```diff
@@ -1642,7 +1771,198 @@ def panther_similarity(
...
+    # Calculate reciprocals and normalize to [0, 1] range
+
+    # Handle the case where distances are very small or zero (common in small graphs)
+    # Use the passed in eps parameter instead of defining a new epsilon
+    neighbor_distances = np.maximum(neighbor_distances, eps)
+    similarities = np.reciprocal(neighbor_distances)
+
+    # Replace any NaN values with a reasonable value (1.0)
+    # This can happen in very small graphs where the source node is similar to itself
+    similarities = np.nan_to_num(similarities, nan=1.0)
```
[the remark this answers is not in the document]
mrecachinas: I remember running into `nan`s initially, specifically when I was testing very small graphs, so I added that to force it to a real number (as I note in the comment). As far as I can tell, it doesn't seem necessary anymore:
- KDTree distances are always real/finite (it'll throw an exception of it's nan/inf)
- `eps` is real (sort(1/|E|)
- `np.maximum(neighbor_distances, eps)` is real
- as you said, reciprocal would be `inf`, not `nan`

That said... since I added that (and re the comment), the source node isn't included in the returned results, but even if it were, identical nodes would have distance 0, which after `np.maximum(0, eps)` would be `eps`, and then the reciprocal would be real.

In any case, removed in [307050d](https://github.com/networkx/networkx/pull/4400/commits/307050d858c5b653307e0b2ea10993a67f5a5508).

dschult:
What size networks is this method aimed for? The heap approach is not faster in my quick tests until we have about 1 million nodes. I run `panther_similarity` on graphs with 200 nodes and it takes more than 1.5s. 400 nodes gets me to 7.5s. 800 nodes gets me to 40s. So 100,000 nodes could be about 2000 days.  

What do you think about removing the heapq stuff and going back to sorting and slice indexing?

mrecachinas:
@dschult I mentioned this in a comment buried somewhere (https://github.com/networkx/networkx/pull/4400#discussion_r2094301463):

> My understanding is the typical number of nodes is 100k+ or millions, whereas k and D are usually 50-100. At least, the paper benchmarks it on hundreds of millions of nodes IIRC.

(Note too that in the original paper's benchmarks, Panther was faster than Panther++, especially for large graphs, which makes sense since it's a more "naive" calculation than building feature vectors for each node.)

I can certainly revert it back to the numpy sorting/slicing. There was also mention of quickselect, which is an option. I suspect leaning on numpy might be faster here though.

dschult:
I suggest using numpy -- and it looks like it doesn't matter what we use at this part of the algorithm.

Peiffap:
I did some timing too, comparing the following:
1. `sort`, which sorts the entire array then slices the top `D` and reverses.
2. `heap`, which is the current implementation.
3. `quickselect`, which finds the `D` largest elements, sorts, and reverses.

This is what I got for `D = 100`—it agrees with Dan's comment, and shows that quickselect is even faster.
topd_timing
<details>
<summary>Timing code</summary>

```python
import numpy as np
import heapq
import timeit


def method_sort(arr, D):
    return np.sort(arr)[-D:][::-1]


def method_heap(arr, D):
    h = []
    for x in arr:
        if len(h) < D:
            heapq.heappush(h, x)
        else:
            heapq.heappushpop(h, x)
    return sorted(h, reverse=True)


def method_quickselect(arr, D):
    part = np.partition(arr, -D)[-D:]
    return np.sort(part)[::-1]


def benchmark(N=100_000, D=100, repeats=5):
    arr = np.random.random(N)
    res = []
    times = {}
    for name, method in [
        ("sort", method_sort),
        ("heap", method_heap),
        ("quickselect", method_quickselect),
    ]:
        res.append(method(arr, D))
        t = timeit.timeit(lambda: method(arr, D), number=repeats)
        times[name] = t / repeats
    assert np.allclose(res[0], res[1]) and np.allclose(res[0], res[2])
    return times


def run_plot():
    sizes = [10_000, 50_000, 100_000, 500_000, 1_000_000, 5_000_000, 10_000_000]
    D = 100
    results = {m: [] for m in ["sort", "heap", "quickselect"]}

    for N in sizes:
        times = benchmark(N, D)
        for method, t in times.items():
            results[method].append(t)

    for method, times in results.items():
        plt.plot(sizes, times, label=method)

    plt.xlabel("Array Size (N)")
    plt.ylabel("Time (s)")
    plt.title(f"Top {D} Largest Elements (Decreasing Order)")
    plt.legend()
    plt.grid(True)
    plt.tight_layout()
    plt.show()
```
if __name__ == "__main__":
    run_plot()

</details>

Since the quickselect implementation is reasonably significantly faster (especially at the multi-million node level they mention in the paper)— and I expect it to be faster for smaller values of `D` still—and not more complicated than sorting and slicing, my vote goes to doing that.

---

Then I set out to locally test all of that so I could just use the suggestion feature on here. (In the process, I discovered you can shift-select to get a much larger suggestion, which would've come in handy in previous PRs.) Only to find out that 5 years ago, you already came to this same conclusion for `panther_similarity`:
https://github.com/networkx/networkx/blob/d1c41f886add2ab86c1ce4c1d788d3c286721351/networkx/algorithms/similarity.py#L1629-L1633
That made me stop changing things locally, since I'd just be repeating what's already higher up. This does make me think that perhaps the similarity (pun intended) between the `panther_*similarity` functions is larger than just the preparation of paths—but that doesn't need to be done in this PR.

---

Tl;dr: let's use quickselect, like `panther_similarity` already does.

Some minor other notes: we can wait until we have a the top-`D` before multiplying by `inv_sample_size`, which I imagine would be marginally faster. And there's probably something to be said for taking care to avoid converting between lists/sets/numpy arrays more than strictly necessary.

dschult:
Can we stop optimizing the sort -- at least until we make sure that is the slow part of this algorithm?  The code we've got (with the quicksort) will never run for tens of thousands of nodes, let alone a million. 

Optimization should first identify where the bottlenecks are.

Peiffap:
Let me start off by saying I agree with the you that we shouldn't optimize things before we've identified them as bottlenecks. With that being said, we're in the innermost loop of the code, computing similarity between each node and _all other nodes_—and Michael claims that the algorithm is regularly used for hundreds of thousands or even millions of nodes. I'd be very surprised if that's not a significant part of the runtime... But it's not impossible of course. I'm not sure how much time the `KDTree` creation and querying will take.

Perhaps I should phrase it as a question instead: why do you say the code we have (with quicksort) will never run for tens of thousands of nodes?

It's my understanding that if we just sort and slice, we're sorting an array with `N` (the number of nodes, i.e. up to at least 1000000) entries, then slicing the top `D`, with `D` being ~100.
The heap implementation that's currently being reviewed goes through the `N` entries, keeping the top `D` and sorting only those. If that's what you're referring to, are you saying we should disregard your [original comment](https://github.com/networkx/networkx/pull/4400#issuecomment-2933025276) on the heap being slower than sorting and slicing?

Either way, the only satisfactory answer here is going to involve some more benchmarking, even if just to see whether collecting the top `D` entries is a bottleneck.

mrecachinas:
I just wrote a quick benchmarking script for the numpy implementation of Panther++ against a Barabasi-Albert graph. Below are the results:

(Quick note: I'm on a 2021 M1 Max MBP with 64 GB of memory.)

```
============================
Size     Nodes    Edges    Time (s)     Results 
--------------------------------------------------
10       10       21       0.2949       4       
100      100      291      0.0497       4       
1000     1000     2991     1.5536       4       
5000     5000     14991    29.7625      4       
10000    10000    29991    118.6974     4
```

Here's the core script logic with params:

```python
    for size in sizes:
        try:
            G = nx.barabasi_albert_graph(size, min(3, size-1), seed=42)
            degrees = dict(G.degree())
            source = max(degrees, key=degrees.get)
            D = min(50, G.number_of_nodes() - 1)
            panther_params = {
                'D': D,
                'k': min(5, G.number_of_nodes() - 1),
                'path_length': 5,
                'c': 0.5,
                'seed': 42
            }
```

Based on this, that probably means 100k nodes would be ~4 hours and 1M nodes would be ~10 days. It's still not as fast as the implementation in the paper. 🤔

Panther paper benchmarks

Peiffap:
What do you call the numpy implementation? What do the timings look like for `panther_similarity`? Do those match the timings in the paper?

mrecachinas:
@Peiffap The difference is I reverted the loop calculating the feature vectors from using a heap back to

```python
    similarities = np.zeros((num_nodes, num_nodes))
    theta = np.zeros((num_nodes, D))
    index_map_sets = {node: set(paths) for node, paths in index_map.items()}

    # Calculate the path similarities for each node
    for vi_idx, vi in enumerate(G.nodes):
        vi_paths = index_map_sets[vi]

        for node, node_paths in index_map_sets.items():
            # Calculate similarity score
            common_path_count = len(vi_paths.intersection(node_paths))
            similarities[vi_idx, inv_node_map[node]] = (
                common_path_count * inv_sample_size
            )

        # Build up the feature vector using the largest D similarity scores
        theta[vi_idx] = np.sort(np.partition(similarities[vi_idx], -D)[-D:])[::-1]
```

The timings for `panther_similarity` look like

```
Size     Nodes    Edges    Time (s)     Results 
--------------------------------------------------
10       10       21       0.0200       4       
100      100      291      0.0480       4       
1000     1000     2991     0.9605       4       
5000     5000     14991    14.4812      4       
10000    10000    29991    54.4178      4 
```

(Same general loop, just omitting `D` as a kwarg.)

mrecachinas:
Another note: From some quick profiling, the two most expensive operations were:
- `generate_random_paths` (~38%)
- calculating path similarities (the loop) (~62%)

On a 10k node/29k edge graph, this looks roughly like:
- 52 seconds for `generate_random_paths`
- 85 seconds for the path similarity loop

Trying something out to see if we can optimize generate_random_paths since it's the common denominator between both algorithms.

amcandio:
> Another note: From some quick profiling, the two most expensive operations were:
> - `generate_random_paths` (~38%)
> - calculating path similarities (the loop) (~62%)
> 
> On a 10k node graph, this looks roughly like:
> - 52 seconds for `generate_random_paths`
> - 85 seconds for the path similarity loop
> 
> Trying something out to see if we can optimize generate_random_paths since it's the common denominator between both algorithms.

Is your graph sparse? I see generate random path creates an adjacency matrix, which might be inefficient for sparse graphs.

mrecachinas:
Yeah -- I was still using a BA graph.

Anyway, I'm happy to pursue further optimizations in this PR or ship the PR and pursue them in a different PR. I'll defer to you all on that.

Peiffap:
Ah, I didn't realise the initial "sort" was already basically quickselect/introselect! I thought you were sorting the full list before selecting the top `D` entries. Disregard my previous comments!

mrecachinas:
@dschult What do you think about the current state?

mrecachinas:
Gentle nudge @dschult

dschult (approved):
This seems like a good approach -- too bad it isn't faster, but we can work on that. Presumably the publication wasn't using python. :}

I approve this PR!   Sorry it took so long.

networkx/algorithms/similarity.py
```diff
@@ -1505,9 +1507,126 @@ def _simrank_similarity_numpy(
...
+        pass
+
+    # Return a named tuple with all the required data
+    return (
+        G,  # The graph with isolated nodes removed
+        inv_node_map,
+        index_map,
+        1 / sample_size,
+        eps,
+    )
```
rossbar: This isn't actually a namedtuple, just a normal one. Do we want it to be named? If it's not necessary, then let's also remove the import at the top of the file.

networkx/algorithms/similarity.py
```diff
@@ -1573,47 +1700,29 @@ def panther_similarity(
...
+        path_length=path_length,
+        c=c,
+        delta=delta,
+        eps=eps,
+        weight=weight,
+        k=k,
+        seed=seed,
+    )
 
+    node_map = np.array(G.nodes)
```
rossbar: Is there an explicit reason why this must be a numpy array? AFAICT from looking at the code, no array-like features (vectorization, broadcasting, etc.) are every relied on. If not, my inclination would be to instead replace this with `node_map = list(G)`, which has the advantage of also simplifying the `top_k_with_val` construction in L1746.

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+
+    While basic Panther similarity measures how often two nodes appear on the same paths,
+    Panther vector similarity (Panther++) creates a ``D``-dimensional feature vector for each
+    node using its top similarity scores with other nodes, then computes similarity based
+    on the Euclidean distance between these feature vectors. This approach better captures
+    structural similarity and addresses the bias towards close neighbors present in
+    the original Panther algorithm.
+
+    This approach is preferred when:
+    1. You need better structural similarity than basic path co-occurrence
```
rossbar: I think a blank line is needed to get the rst enumerated list to render properly.

```suggestion
    This approach is preferred when:
    
    1. You need better structural similarity than basic path co-occurrence
```

networkx/algorithms/similarity.py
```diff
@@ -1505,9 +1507,126 @@ def _simrank_similarity_numpy(
...
+    # Generate the random paths and populate the index_map
+    for _ in generate_random_paths(
+        G,
+        sample_size,
+        path_length=path_length,
+        index_map=index_map,
+        weight=weight,
+        seed=seed,
+    ):
+        pass
```
rossbar: This is a pretty non-idiomatic pattern, so let's add a comment for readers!

```suggestion
    ):
        # NOTE: index_map is modified in-place by `generate_random_paths`
        pass
```

networkx/algorithms/similarity.py
```diff
@@ -1573,47 +1700,29 @@ def panther_similarity(
...
+        eps=eps,
+        weight=weight,
+        k=k,
+        seed=seed,
+    )
 
+    node_map = np.array(G.nodes)
     num_nodes = G.number_of_nodes()
+
+    # If there aren't enough nodes for the requested k, adjust k with a warning
```
rossbar:
```suggestion
    # Check number of nodes after any modifications by _prepare_panther_paths
```

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+    similarity : dict
+        Dict of nodes to similarity scores (as floats).
+        Note: the self-similarity (i.e., `node`) is not included in the dict.
+
+    Examples
+    --------
+
+        >>> import networkx as nx
+        >>> G = nx.star_graph(100)
+        >>> sim = nx.panther_vector_similarity(G, 0)
```
rossbar: Some suggestions for improvement:
 1. The explicit nx import isn't necessary (it's handled internally)
 2. Let's seed it and show the output (this also serves as a test for the seeding)
 3. My naive expectation would be that the "hub" node would have arbitrary simliarity to the "spoke" nodes, whereas the spoke nodes should (to the limit of stochasticity) be identical to each other. I think it's worth highlighting this in the example!
```suggestion
    >>> G = nx.star_graph(100)
    
    The "hub" node is distinct from the "spoke" nodes
    
    >>> nx.panther_vector_similarity(G, source=0, seed=42)
    {85: 0.10254413039936112, 61: 0.10187943587925867, 35: 0.10156480636337968, 65: 0.10155189938081624}
    
    But "spoke" nodes are similar to one another
    
    >>> nx.panther_vector_similarity(G, source=1, seed=42)
    {26: 1.0, 52: 1.0, 98: 1.0, 29: 1.0}
```

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+    References
+    ----------
+    .. [1] Zhang, J., Tang, J., Ma, C., Tong, H., Jing, Y., & Li, J.
+           Panther: Fast top-k similarity search on large networks.
+           In Proceedings of the ACM SIGKDD International Conference
+           on Knowledge Discovery and Data Mining (Vol. 2015-August, pp. 1445–1454).
+           Association for Computing Machinery. https://doi.org/10.1145/2783258.2783267.
+    """
+    import numpy as np
+    from scipy.spatial import cKDTree as KDTree
```
rossbar: The preferred import idiom for scipy in NetworkX is:

```suggestion
    import scipy as sp
```

With the corresponding change when creating the KDTree (see comment below)

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+            similarities[vi_idx, inv_node_map[node]] = (
+                common_path_count * inv_sample_size
+            )
+
+        # Build up the feature vector using the largest D similarity scores
+        theta[vi_idx] = np.sort(np.partition(similarities[vi_idx], -D)[-D:])[::-1]
+
+    # Insert the feature vectors into a k-d tree
+    # for fast retrieval
+    kdtree = KDTree(theta)
```
rossbar:
```suggestion
    kdtree = sp.spatial.KDTree(theta)
```

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+        path_length=path_length,
+        c=c,
+        delta=delta,
+        eps=eps,
+        weight=weight,
+        remove_isolates=False,
+        k=k,
+        seed=seed,
+    )
+    node_map = np.array(G.nodes)
```
rossbar: Same deal here - can we get away with a list?

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+    nearest_neighbors = np.atleast_1d(nearest_neighbors)
+
+    # The paper defines the similarity S(v_i, v_j) as
+    # 1 / || Theta(v_i) - Theta(v_j) ||
+    # Calculate reciprocals and normalize to [0, 1] range
+
+    # Handle the case where distances are very small or zero (common in small graphs)
+    # Use the passed in eps parameter instead of defining a new epsilon
+    neighbor_distances = np.maximum(neighbor_distances, eps)
+    similarities = np.reciprocal(neighbor_distances)
```
rossbar: I think there's a subtle reason to prefer `1 /` instead of `np.reciprocal` here - `reciprocal` will use same-kind casting by default, which means you would get wildly different answers if `neighbor_distances` were (somehow) e.g. integers vs. floats. In other words, `np.reciprocal` will give you `1 // neighbor_distances` or `1 / neighbor_distances` depending on the dtype `neighbor_distances` - far better to be explicit I think!

I suspect we *always* suspect `neighbor_distances` to be floating point (since we're dealing with distances), so I'd propose to switch this to be explicit:

```suggestion
    similarities = 1 / neighbor_distances
```

networkx/algorithms/similarity.py
```diff
@@ -1633,16 +1742,188 @@ def panther_similarity(
...
+
+    # Always normalize to ensure values are between 0 and 1
+    if similarities.size > 0:
+        max_sim = np.max(similarities)
+        if max_sim > 0:  # Avoid division by zero
+            similarities = similarities / max_sim  # Normalize to [0, 1]
```
rossbar: Minor suggestion, feel free to ignore:

```suggestion
    if len(similarities) > 0 and (max_sim := np.max(similarities)) > 0:
        similarities /= max_sim
```

Note this only works if `similarities` are guaranteed to be floating point (see my previous comment)

networkx/algorithms/similarity.py
```diff
@@ -1729,7 +2010,18 @@ def generate_random_paths(
     # Calculate transition probabilities between
     # every pair of vertices according to Eq. (3)
     adj_mat = nx.to_numpy_array(G, weight=weight)
-    inv_row_sums = np.reciprocal(adj_mat.sum(axis=1)).reshape(-1, 1)
+
+    # Handle isolated nodes by checking for zero row sums
```
rossbar, line 2054: I don't think this error handling is appropriate. As it stands (perhaps unfortunately), `generate_random_paths` is in the main NetworkX namespace, so it's not specific to the use in the panther algorithm. AFAIK random walks in graphs with isolated nodes is perfectly allowable (though would potentially be surprising for users who didn't know there were isolated nodes in their graph).

Either way, given the way things are currently structured, I think this checking needs to happen outside of this function so as not to break backward compatibility with non-panther usages.

rossbar (commented):
Thanks @mrecachinas , overall looks good but still some things to figure out I think.

Also when trying this out, I noticed there may be a small bug related to `k`. If I do:

```python
>>> G = nx.star_graph(100)
>>> nx.panther_vector_similarity(G, source=1, k=4, seed=42)
{26: 1.0, 52: 1.0, 98: 1.0}
```

Note that `k=4`, so I expect the result to contain the 4 most similar nodes, but there are only 3. This is true for the default to (`k=5`, but only 4 values returned). I suspect it's related to failing to account for the removal of the self-similar node from the result.

networkx/algorithms/similarity.py
mrecachinas: I think it was from an earlier version of my implementation where I was broadcasting.

networkx/algorithms/similarity.py
mrecachinas: Fixed!

mrecachinas:
@rossbar Thanks for the feedback:

> Let's remove the addition of the exception for isolated nodes inside generate_random_path. Off the top of my head there are two relatively simple alternatives that don't affect backward compatibility:

Yeah good call here. Should be addressed in cb5cf49c82ece25cff5670e92ec46336cc975a74.

> The doctest failure shows that the result is not deterministic. That may be okay, but let's double-check the source of the non-determinism and ensure it's not a result of failing to seed something.

Struggling with this one -- I can't reproduce the error we're seeing in CI. 🤔 

> The other test failure also seems real - I'd guess whatever fix was put in place for the k-off-by-one issue isn't working properly for k=1.

The issue was in querying the KD-tree. There was an implicit assumption that the source node would always be returned in the `k` requested nodes, hence why I updated it to query for `k+1`. Unfortunately, e.g., for the spokes of a star graph, there was no guarantee the source would be returned. Should be fixed in cccb49f3041ecc6369c546c5d164b66e2a195c8e. Basically, I added a check if we had `k` results, and if not, took the top `k` that were returned (sorted by similarity).

mrecachinas:
@rossbar I suspect the nondeterminism may be coming from "ties" when querying the KD-tree. I'm sorting _after_ the ties, but I don't know how the KD-tree internally would handle ties/stability.

At least for now, I've modified the doctest to illustrate the point (in d35320e).

rossbar:
> I suspect the nondeterminism may be coming from "ties" when querying the KD-tree. I'm sorting after the ties, but I don't know how the KD-tree internally would handle ties/stability.

I see - in that case it's probably worth a note in the docstring.

On a related note - while reading through the [kdtree docstring](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html) I noticed this:

> For large dimensions (20 is already large) do not expect this to run significantly faster than brute force. High-dimensional nearest-neighbor queries are a substantial open problem in computer science.

This seems directly relevant to our default choice of `D`, which is 50. IIUC, we're essentially constructing the tree out of an NxD array, which would be very inefficient. I see that the default `D=50` appears to have come from the paper, but perhaps we should consider a more reasonable (i.e. smaller) default given computational limitations. Note that KDTree switches over to brute force via the `leafsize` parameter, with a default value of 10. Again, IIUC, this means we're essentially getting no benefit from similarity vectors in this way.

mrecachinas:
@rossbar Think I covered your last two points. Lmk.

mrecachinas:
@dschult @rossbar Let me know if I've covered everything or if there's any remaining items.

dschult:
I think the remaining item for this is the default value of `D` (the number of dimensions to choose). 
The paper uses 50, but the SciPy code switches to the more inefficient handling for `D>= 10`. I see that the default here was originally at 50 and now has been set to 15. But that is still in the size where we don't get any benefit from similarity vectors (or am I understanding something incorrectly).

mrecachinas:
@dschult I've changed it to 10.

Thinking about this a bit more, the `leafsize` kwarg is "the number of points at which the algorithm switches over to brute-force", and it defaults to 10. Right now we just use the default. They also say going above 20 won't be _significantly faster_ than brute force. I'm wondering what the metrics are in practice and whether we should set leafsize based on `D` if it's still generally, although not significantly, faster than brute force.

In any case, I did change it from 15 -> 10.

dschult:
The sklearn version of KDtree has a default dimension of 40. 
And there is a pure python version of KDtree that claims to be about 60 lines of code.
But it has a CC0 license and I'm not sure if that plays nicely with us, but it means it probably wouldn't be too hard to code it up. 

I'm thinking we should probably merge this and figure out how to handle the KDtree issues in follow-up PRs. Thoughts?

mrecachinas:
SGTM

rossbar:
> I'm thinking we should probably merge this and figure out how to handle the KDtree issues in follow-up PRs. Thoughts?

Works for me, thanks for taking a look @dschult !

PR #4410, opened 2020-12-03
DOC: Remove repeated words

eumiro:
Remove repeated words from docs and comments.


dschult:
Thanks!  Hard to believe there were so many of these. :)

PR #4413, opened 2020-12-03
Safer repr format of variables

eumiro:
Replace `…'{x}'…` with `…{x!r}…` when formatting unknown string values. This will correctly escape quotes.


dschult (approved):
This looks good. Thanks!

rossbar (approved):
LGTM, thanks @eumiro !

PR #4421, opened 2020-12-04
Implement linear controllability algorithms

alexoshin:
Hello all,

As part of my course project for the Network Science class at Georgia Tech (https://www.cc.gatech.edu/~dovrolis/Courses/NetSci/), I have focused on implementing various controllability algorithms as defined by Lin et al. 2016 (https://doi.org/10.1103/RevModPhys.88.035006). I think these contributions will be valuable to the NetworkX codebase, as they allow for analysis of networks from a control theoretic perspective. There are many examples of such algorithms being used to analyze systems, ranging from the _C. Elegans_ connectome to protein-protein interactions networks and more (https://doi.org/10.1038/nature24056, https://doi.org/10.1073/pnas.1603992113, https://doi.org/10.1016/j.neuropsychologia.2018.01.001, etc.) 

These come largely from section II-B through II-F in the Lin et al. 2016 paper, and focus entirely on linear systems. In particular, I have implemented the following in a new directory under ```networkx/algorithms/control```. They are currently all located in a single file titled ```systems.py```. There are multiple unit tests in ```networkx/algorithms/control/tests/test_systems.py```.

1. A common interface for linear-time invariant systems
2. Constructing controllability matrices given A and B matricies
3. Kalman's rank test for controllability
4. Popov-Belevitch-Hautus test for controllability
5. Inaccessibility and dilation tests for testing structural controllability
6. Constructing a bipartite graph from a given directed graph (separates nodes into "in" and "out" partitions)
7. Finding matchings of a particular size t in a given graph
8. Finding the sets of maximum matchings in a given graph
9. Checking for constrained t-matchings
10. Testing if the system is strongly structurally controllable
11. Finding a minimum set of driver nodes (nodes that when driven can control the entire network)
12. Finding a minimum set of actuators (assuming each input can control one node, the minimum number of such inputs necessary to control the entire network)
13. Classifying the importance of an edge by checking the number of maximum matchings it participates in
14. Classifying the importance of a node by checking the number of minimum driver node sets it appears in
15. Classifying the importance of a node by checking whether deleting the node increases the minimum number of driver nodes necessary

This is my first contribution, so I think many of these implementations still need a lot of reworking, but this should be a good first step to a fully-functional control submodule in NetworkX. It might also be worth making each of these individual pull requests to make everything more organized, but this PR is largely to get the ball rolling.

Please let me know any suggestions you might have to improve the code here.

Thanks!


dschult:
Very nice!  Thank you!

You should at some point add a hook to the documentation builder in `doc/reference/algorithms` so the docstrings can be included in the reference section of the full package documentation.  But before that, a quick look made me wonder how robust this code is to non-string nodes. By that I mean, would it be a lot of work to make this code work with integer nodes? Maybe it already does, but I saw some adding "+" and "-" that might not always work.

We also have some matching algorithms. Does any of this fit with that, or duplicate/improve/etc the matching algorithms already here?

alexoshin:
Thanks for the tips! I will work on robustifying the code a bit, since as you mentioned it will not work with non-string nodes.

As for the matching algorithms, I could only find implementations of **maximal** matchings in the codebase, which is a matching that cannot add another edge and still be a matching, which I believe, in general, is not equal to a **maximum** matching, which is a matching of maximum possible size given a particular graph. I have also implemented a method for finding all matchings of a particular size. I think my additions are complementary to the methods already available in the existing matching algorithms. Would it be worth adding my contributions there?

dschult:
The other place in the library where we [add prefix/suffix to the nodes is in `union()`](https://github.com/networkx/networkx/blob/d4701d142ac8fc7b15d16100f589819d1d9d502a/networkx/algorithms/operators/binary.py#L65).  If you can think of a way to identify in and out without strings do that. But this could work too. It basically turns the node into a string if it isn't already -- and then adds the prefix. You may be able to do it more simply than that code.

I agree with you about matching that what we currently offer is maximal matchings, including maximal weight matchings. I couldn't know whether this overlapped much with what is there. I'm guessing the function to return all matchings of given size could live in `matching.py`. But let's hold off moving it there until that is more clear.  That function should probably yield the matchings rather than collect them in a list because the list could get *very* large and people might only want the first , or the first few.

This looks intriguing. I'm not going to have time to look at it in depth until close to the end of this month. Then I'll give a review with lots of comments all together instead of these piecemeal items.  :}   Thanks for this!

PR #4423, opened 2020-12-05
CNFG Link Prediction Algorithm Implementation

cgalo:
Hello!

As part of my course project for Social & Information Networks at the University of Toledo, me and @yashwalankar, focused on implementing the CNGF link prediction algorithm[1]. This algorithm was proposed due to the lack of link prediction algorithms utilizing network topology sufficiently to design the similarity function between entities.

### Pseudocode[1]
image

This is the first time I'm contributing to an open-source project, any feedback regarding the implementation, functionality, or documentation will be appreciated!

### Reference
[1] Liyan Dong, Yongli Li, Han Yin, Huang Le, Mao Rui, "The Algorithm of Link Prediction on Social Network", Mathematical Problems in Engineering, vol. 2013, Article ID 125123, 7 pages, 2013. https://doi.org/10.1155/2013/125123

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ networkx.egg-info/
 
 # PyTest Cache
 .pytest_cache
+
+# Development build
+networkx-dev/
--- a/doc/reference/algorithms/index.rst
+++ b/doc/reference/algorithms/index.rst
@@ -66,6 +66,7 @@ Algorithms
    smetric
    sparsifiers
    structuralholes
+   summarization
    swap
    threshold
    tournament
--- a/doc/reference/algorithms/summarization.rst
+++ b/doc/reference/algorithms/summarization.rst
@@ -1,0 +1,9 @@
+*************
+Summarization
+*************
+
+.. automodule:: networkx.algorithms.summarization
+.. autosummary::
+   :toctree: generated/
+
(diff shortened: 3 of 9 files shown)


dschult:
Very interesting -- thanks for this.

Could you make the doc_string better, especially as regards the paragraph after the intro line?
For example, what does CNGF stand for? and why is guidance named that? is it guiding something?
Also, the math can be typeset similar to the function two up from yours in the file.
The reference should have lines indented to match the first (see the function just above yours in the file).

A hook for this function should be added to the doc/reference/algorithms section to make this appear in the full docs reference section.

Setting up `subG` could avoid all those set commands via `subG = G.subgraph(cnbors | {u, v})` and I don't think you need the copy because you don't change `subG`.

For performance and readability you should avoid creating `x_sub_degree` and just put subG.degree(x) in the formula. 
Perhaps more importantly, since you only look up one node's degree at a time, it is faster to use `G.degree[n]` than `G.degree(n)`.  And I don't think `cnbors` can ever be `None` so the if statement is not needed.
Also, the summation for similarity can use Python's more simple `sum` function:
`sum(guidance(node) for node in cnbors)`
It even works when cnbors is empty.

Thanks!

PR #4427, opened 2020-12-07
Collect graphviz examples

jarrodmillman:
This moves all the graphviz examples to one section including the one that merged all the old pygraphviz examples.  It also sorts the subsections by filename.  Before it was by lines of code.


rossbar (approved):
Reorganizing the examples in this way makes sense to me. It might be worth adding a sentence or two about the fact that the current examples use `pygraphviz` for layout, but `nx_pylab` can still used for the actual drawing.

jarrodmillman:
I would like to keep the text in the README short and open for future changes.  I plan to add short sentences to the other sections as well.  It would also be good to see if we can get sphinx-gallery to display figures generated by graphviz (we should be able to display pngs and then remove some of the examples of using ``nx_pylab`` for drawing).  At that point we should also review the example docstrings and make things like the distinction between layout and drawing more explicit.

dschult (approved):

PR #4428, opened 2020-12-07
Deprecate jit

jarrodmillman:
How popular is the Javascript Infovis Toolkit?  It looks to me like it is a dead project and I propose deprecating this code.  

The last commit to https://github.com/philogb/jit was Sept. 2014.  The last release was July 1, 2013.  The newest question on stackoverflow was Feb. 2015.   The mailing list is getting questions, but no one is answering.

We did get a recent PR:  https://github.com/networkx/networkx/pull/3891

It looks like the only way to install it is to download the source (or clone the repo) and then type ``python make.py build``, but ``make.py`` is a Python 2 script.  Maybe if I fixed it, everything would work.  But it seems likely to me that there will be other things that need to be updated.

It looks like @MridulS contributed ``jit_data`` and ``jit_graph`` in #1945 based on #881.  Do you know more about if this widely used still?


MridulS:
It does look like the project is non maintained anymore, and I am not aware if this is still used.
+1 on deprecating it.

rossbar (approved):
Agreed - it doesn't seem that the project is active.

PR #4429, opened 2020-12-07
Remove advanced example section

jarrodmillman:
I don't think examples in the advanced section make sense.  (We could have an advanced section.  I just don't think the current examples are that great.)

So I moved
1. eigenvalues to drawing (It seems like a basic plot you might want to do.)
2. i deleted the umlaut one (unicode isn't special anymore)
3. the other two to algorithms


rossbar (approved):
I agree that "Advanced" is not a very descriptive category of examples, and there doesn't seem to be any criteria that make these examples a better fit in a stand-alone category. The new locations for the three moved examples make sense to me.

I also agree with removing the umlaut example as this seems to be very Python2-centric, unless I'm missing something.

dschult (approved):
I agree with the idea of eliminating the "advanced" section and the umlaut example
I approve this PR.

PR #4434, opened 2020-12-08
Fix broken links

dbxnr:
fix a few dead links to ActiveState


dschult (approved):
Thanks!

rossbar (approved):
I had been thinking we could add these to a `References` section and that hopefully that would allow nitpicky mode to detect broken links, but I don't think that'll work - thanks for finding/fixing these @dbxnr !

PR #4435, opened 2020-12-08
improve group betweenness centrality

guyroznb:
In this PR I improved the group betweenness centrality algorithm by using the algorithm of Puzis et al. "Fast algorithm for successive computation of group betweenness centrality".

In the previous implementation of the algorithm, for each pair of source - target non - group nodes permutation, all the shortest paths were found and if one of them go through one of the nodes in the group then the group betweenness centrality increases. This was done for each group.

Here I used Pusiz et al. implementation, which finds the GBC with running time of $O(K^3)$ for each group (k is the number of nodes in the group) and with a preprocessing time of $O(mn)$ for all the groups.

This unable us to find the GBC of many groups in one preprocessing step.
Furthermore, the article shows how to find the prominent group of size k (the group with the highest GBC among those with size k), which I would add in future PR.

### DIFF
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -225,7 +225,7 @@ def edge_betweenness_centrality(G, k=None, normalized=True, weight=None, seed=No
         else:  # use Dijkstra's algorithm
             S, P, sigma, D = _single_source_dijkstra_path_basic(G, s, weight)
         # accumulation
-        betweenness, delta = _accumulate_edges(betweenness, S, P, sigma, s)
+        betweenness = _accumulate_edges(betweenness, S, P, sigma, s)
     # rescaling
     for n in G:  # remove nodes to only return edges
         del betweenness[n]
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -124,14 +124,14 @@ def betweenness_centrality(
     for s in nodes:
         # single source shortest paths
         if weight is None:  # use BFS
-            S, P, sigma = _single_source_shortest_path_basic(G, s)
+            S, P, sigma, D = _single_source_shortest_path_basic(G, s)
         else:  # use Dijkstra's algorithm
-            S, P, sigma = _single_source_dijkstra_path_basic(G, s, weight)
+            S, P, sigma, D = _single_source_dijkstra_path_basic(G, s, weight)
         # accumulation
         if endpoints:
-            betweenness = _accumulate_endpoints(betweenness, S, P, sigma, s)
+            betweenness, delta = _accumulate_endpoints(betweenness, S, P, sigma, s)
         else:
-            betweenness = _accumulate_basic(betweenness, S, P, sigma, s)
+            betweenness, delta = _accumulate_basic(betweenness, S, P, sigma, s)
     # rescaling
     betweenness = _rescale(
         betweenness,
@@ -221,11 +221,11 @@ def edge_betweenness_centrality(G, k=None, normalized=True, weight=None, seed=No
     for s in nodes:
         # single source shortest paths
         if weight is None:  # use BFS
-            S, P, sigma = _single_source_shortest_path_basic(G, s)
+            S, P, sigma, D = _single_source_shortest_path_basic(G, s)
         else:  # use Dijkstra's algorithm
-            S, P, sigma = _single_source_dijkstra_path_basic(G, s, weight)
+            S, P, sigma, D = _single_source_dijkstra_path_basic(G, s, weight)
         # accumulation
-        betweenness = _accumulate_edges(betweenness, S, P, sigma, s)
+        betweenness, delta = _accumulate_edges(betweenness, S, P, sigma, s)
     # rescaling
     for n in G:  # remove nodes to only return edges
         del betweenness[n]
@@ -268,7 +268,7 @@ def _single_source_shortest_path_basic(G, s):
             if D[w] == Dv + 1:  # this is a shortest path, count paths
                 sigma[w] += sigmav
                 P[w].append(v)  # predecessors
-    return S, P, sigma
+    return S, P, sigma, D
 
 
 def _single_source_dijkstra_path_basic(G, s, weight):
@@ -303,7 +303,7 @@ def _single_source_dijkstra_path_basic(G, s, weight):
             elif vw_dist == seen[w]:  # handle equal paths
                 sigma[w] += sigma[v]
                 P[w].append(v)
-    return S, P, sigma
+    return S, P, sigma, D
 
 
 def _accumulate_basic(betweenness, S, P, sigma, s):
@@ -315,7 +315,7 @@ def _accumulate_basic(betweenness, S, P, sigma, s):
             delta[v] += sigma[v] * coeff
         if w != s:
             betweenness[w] += delta[w]
-    return betweenness
+    return betweenness, delta
 
 
 def _accumulate_endpoints(betweenness, S, P, sigma, s):
@@ -328,7 +328,7 @@ def _accumulate_endpoints(betweenness, S, P, sigma, s):
             delta[v] += sigma[v] * coeff
         if w != s:
             betweenness[w] += delta[w] + 1
-    return betweenness
+    return betweenness, delta
 
 
 def _accumulate_edges(betweenness, S, P, sigma, s):
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -124,9 +124,9 @@ def betweenness_centrality(
     for s in nodes:
         # single source shortest paths
         if weight is None:  # use BFS
-            S, P, sigma, D = _single_source_shortest_path_basic(G, s)
+            S, P, sigma, _ = _single_source_shortest_path_basic(G, s)
         else:  # use Dijkstra's algorithm
-            S, P, sigma, D = _single_source_dijkstra_path_basic(G, s, weight)
+            S, P, sigma, _ = _single_source_dijkstra_path_basic(G, s, weight)
         # accumulation
         if endpoints:
             betweenness, delta = _accumulate_endpoints(betweenness, S, P, sigma, s)
@@ -221,9 +221,9 @@ def edge_betweenness_centrality(G, k=None, normalized=True, weight=None, seed=No
     for s in nodes:
         # single source shortest paths
         if weight is None:  # use BFS
-            S, P, sigma, D = _single_source_shortest_path_basic(G, s)
+            S, P, sigma, _ = _single_source_shortest_path_basic(G, s)
         else:  # use Dijkstra's algorithm
-            S, P, sigma, D = _single_source_dijkstra_path_basic(G, s, weight)
+            S, P, sigma, _ = _single_source_dijkstra_path_basic(G, s, weight)
         # accumulation
         betweenness = _accumulate_edges(betweenness, S, P, sigma, s)
     # rescaling
--- a/networkx/algorithms/centrality/group.py
+++ b/networkx/algorithms/centrality/group.py
@@ -7,7 +7,7 @@ from networkx.utils.decorators import not_implemented_for
 from networkx.algorithms.centrality.betweenness import (
     _single_source_shortest_path_basic,
     _single_source_dijkstra_path_basic,
-    _accumulate_endpoints
+    _accumulate_endpoints,
 )
 
 
@@ -64,8 +64,9 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None, endpoints=F
 
     Returns
     -------
-    betweenness : list of floats
-       The list of group betweenness centrality of the groups C.
+    betweenness : list of floats or float
+       If C is a list with single group then return a float. If C is a list with
+       several groups then return a list of group betweenness centralities.
 
     See Also
     --------
@@ -217,14 +218,14 @@ def _group_preprocessing(G, set_v, weight):
                 # if node is connected to the two group nodes than continue
                 if group_node2 in D[node] and group_node1 in D[node]:
                     if (
-                            D[node][group_node2]
-                            == D[node][group_node1] + D[group_node1][group_node2]
+                        D[node][group_node2]
+                        == D[node][group_node1] + D[group_node1][group_node2]
                     ):
                         PB[group_node1][group_node2] += (
-                                delta[node][group_node2]
-                                * sigma[node][group_node1]
-                                * sigma[group_node1][group_node2]
-                                / sigma[node][group_node2]
+                            delta[node][group_node2]
+                            * sigma[node][group_node1]
+                            * sigma[group_node1][group_node2]
+                            / sigma[node][group_node2]
                         )
     return PB, sigma, D
 
--- a/networkx/algorithms/centrality/group.py
+++ b/networkx/algorithms/centrality/group.py
@@ -100,12 +100,9 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None, endpoints=F
 
     """
     GBC = []  # initialize betweenness
-    V = set(G)  # set of nodes in G
 
     #  check weather C contains one or many groups
-    if not any(isinstance(el, list) for el in C) and not any(
-        isinstance(el, set) for el in C
-    ):
+    if any(el in G for el in C):
         C = [C]
     set_v = {node for group in C for node in group}
     if set_v - G.nodes:  # element(s) of C not in G
--- a/networkx/algorithms/centrality/group.py
+++ b/networkx/algorithms/centrality/group.py
@@ -121,10 +121,19 @@ def group_betweenness_centrality(G, C, normalized=False, endpoints=True, weight=
         for group_node2 in set_v:
             for node in G:
                 # if node isn't connected to the two group nodes than continue
-                if D[node][group_node2] != float("inf") or D[node][group_node1] != float("inf"):
-                    if D[node][group_node2] == D[node][group_node1] + D[group_node1][group_node2]:
-                        PB[group_node1][group_node2] += delta[node][group_node2]*sigma[node][group_node1] * \
-                                                        sigma[group_node1][group_node2]/sigma[node][group_node2]
+                if D[node][group_node2] != float("inf") or D[node][
+                    group_node1
+                ] != float("inf"):
+                    if (
+                            D[node][group_node2]
+                            == D[node][group_node1] + D[group_node1][group_node2]
+                    ):
+                        PB[group_node1][group_node2] += (
+                            delta[node][group_node2]
+                            * sigma[node][group_node1]
+                            * sigma[group_node1][group_node2]
+                            / sigma[node][group_node2]
+                        )
 
     # the algorithm for each group
     for group in C:
@@ -143,7 +152,7 @@ def group_betweenness_centrality(G, C, normalized=False, endpoints=True, weight=
                     dxyv = 0
                     dvxy = 0
                     if not (
-                            sigma_m[x][y] == 0 or sigma_m[x][v] == 0 or sigma_m[y][v] == 0
+                        sigma_m[x][y] == 0 or sigma_m[x][v] == 0 or sigma_m[y][v] == 0
                     ):
                         if D[x][v] == D[x][y] + D[y][v]:
                             dxyv = 1.0 * sigma_m[x][y] * sigma_m[y][v] / sigma_m[x][v]
@@ -191,7 +200,6 @@ def group_betweenness_centrality(G, C, normalized=False, endpoints=True, weight=
             if not G.is_directed():
                 scale *= 2
             GBC_group *= scale
-
 
         GBC.append(GBC_group)
     return GBC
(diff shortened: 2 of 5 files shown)


rossbar (commented):
I haven't looked at this in detail but a few things jumped out to me:
 1. This PR proposes to change the signature for `group_betweenness_centrality` in a way that is not backwards-compatible. Doing so requires a [more lengthy avenue to adoption](https://networkx.org/documentation/stable/developer/deprecations.html#deprecations). 
 2. Can the private implementations of shortest path/djikstra be replaced with the implementations already available in `shortest_path`? Introducing multiple copies of functions that do the same thing is bad practice and significantly increases maintenance burden and the probability of bugs. Even if it requires a bit of refactoring, it would be much preferable if this could be recast to use the functions that are already included in NX.

Additionally, it would be great if some preliminary benchmarking could be provided to help motivate the change. I'm not familiar with this algorithm, even simple benchmarks demonstrating performance improvement for the new implementation would certainly bolster the case for the new implementation.

guyroznb:
hi,
1) I changed the signature of the function so it is now backwards-compatible. 
2) I took the implementation from betweenness centrality implementation (in `betweenness.py`). However I can't use the function in `betweenness.py` because I need that the functions `_accumulate_endpoints` would return the `delta` variable, as well as the `betweenness`. Should I change the implementation in `betweenness.py`?

There is a reduction in running time between the two implementation in the case of finding GBC for single group - `n^{2}*O(nlogn+ m)` for the current algorithm vs.  `O(m*n) + O(nk^2) + O(k^3)`, where k is the number of vertices in the group.
But the real difference is in the case of finding GBC for multiple groups. In the current implementation we need to execute the algorithm for each group. In the new implementation the added running time for each group is `O(k^3) + O(nk^2)`.

thanks

networkx/algorithms/centrality/group.py
```diff
@@ -84,49 +84,196 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+            scale = 1 / ((v - c) * (v - c - 1))
+            if not G.is_directed():
+                scale *= 2
+            GBC_group *= scale
+
+        GBC.append(GBC_group)
+    return GBC
+
+
+def _single_source_shortest_path_basic(G, s):
```
dschult, line 207: I think you can use the helper functions from betweenness.py as suggested by @rossbar becuse that function returns `S, P, sigma` and then you can build delta using the code after your comment: `accumulate`.

networkx/algorithms/centrality/tests/test_group.py
```diff
@@ -10,58 +10,97 @@
...
-            1 node and normalized
+        Group betweenness centrality for group with more than
+        1 node and normalized
         """
         G = nx.path_graph(5)
         C = [1, 3]
-        b = nx.group_betweenness_centrality(G, C, weight=None, normalized=True)
-        b_answer = 1.0
+        b = nx.group_betweenness_centrality(
+            G, C, weight=None, normalized=True, endpoints=False
```
dschult, line 31: Do you need to put `endpoints=False` on these tests? It is the default.
And can you why the answer is not the same value (2.0 instead of 1.0) as the old function?

dschult (commented):
Thanks for this. I've put a number of comments in the code. None are deal brekers, but I think some simplify things somewhat.

I think you should use the helper functions from betweenness.py and the compute delta using the output of those functions, rather than recreating the functions here.

Also, can you try a simple timing to see the difference in time. Asymptotic speed is good to know, but implementation is often about the constants that hide in the Big-O. 
Something like this from inside ipython:

```python
G=nx.binomial_graph(1000, 0.3)
%timeit  old_gbc(G, range(90))
%timeit new_gbc(G, range(90))
```

networkx/algorithms/centrality/group.py
```diff
@@ -66,10 +69,7 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
     Notes
     -----
     The measure is described in [1]_.
-    The algorithm is an extension of the one proposed by Ulrik Brandes for
-    betweenness centrality of nodes. Group betweenness is also mentioned in
-    his paper [2]_ along with the algorithm. The importance of the measure is
-    discussed in [3]_.
+    The algorithm is presented in [2].
```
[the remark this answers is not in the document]
guyroznb: I changed to:

> Group betweenness centrality is described in [1]_ and its importance discussed in [3]_. initial implementation of the algorithm is mentioned in [2]_. This function uses an improved algorithm presented in [4]_.

networkx/algorithms/centrality/group.py
```diff
@@ -84,49 +84,196 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
-        if not G.is_directed():
-            scale *= 2
-    else:
-        scale = None
-    if scale is not None:
-        betweenness *= scale
-    return betweenness
+
+    #  check weather C contains one or many groups
+    if not any(isinstance(el, list) for el in C):
```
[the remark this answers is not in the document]
guyroznb: this if statment doesn't work if C is a list of lists. I use this instead. Do you another idea?
```suggestion
    if not any(isinstance(el, list) for el in C) and not any(isinstance(el, set) for el in C):
```

networkx/algorithms/centrality/group.py
```diff
@@ -84,49 +84,196 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+        C = [C]
+    set_v = set()
+    for i in range(len(C)):
+        set_v.update(C[i])
+        if len(set_v - V) != 0:  # element(s) of C not in V
+            C[i] = list(set_v.intersection(C[i]))
+            raise nx.NodeNotFound(
+                "The node(s) " + str(list(set_v - V)) + " are not " "in the graph."
+            )
```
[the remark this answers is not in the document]
guyroznb: chaged

networkx/algorithms/centrality/group.py
```diff
@@ -84,49 +84,196 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+
+    # pre-processing
+    sigma = dict.fromkeys(G)
+    delta = dict.fromkeys(G)
+    D = dict.fromkeys(G)
```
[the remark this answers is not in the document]
guyroznb: changed

networkx/algorithms/centrality/group.py
guyroznb: I also need to get the dictionary  `D` (the distance from s to all other nodes) from `_single_source_shortest_path_basic` function. Should I change the function in betweenness.py, keep the helper functions in group.py, or calculate D from `S, P, sigma` (which would increase the running time of the algorithm)?

networkx/algorithms/centrality/tests/test_group.py
guyroznb: The reason is that in the previous implementation the shortest path from u to v in undirected graph is count as 1 (like betweenness.py). In my implementation (and in most of the papers that I know) the number of shortest path counts as 2. Should I change this to be like betweenness.py? should I write this in the docs?

guyroznb:
Hi,
thanks for the review @dschult @rossbar.
For the old implementation the %timeit is:
2.5 s ± 186 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
I used a graph of size n=100 and group of size 9 because for larger group the old implementation took too long.
For the new implementation the runtime is:
244 ms ± 25 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

networkx/algorithms/centrality/group.py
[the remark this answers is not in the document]
dschult: I thought this one does work for a list of lists... But maybe I am misunderstanding. 

If C is a list of lists or a set of sets (or a list of sets or a set of lists)  -> container of containers... Then C is good to go and we don't do anything.  If C is a collection of nodes, then we put that inside a list to make it a list of a single collection.  Do I have this part right?

So, what we want to check for is whether we have a collection of nodes. 
`if all(el in V for el in C):` does that. But again, maybe I am not understanding... What doesn't work here?

networkx/algorithms/centrality/group.py
```diff
@@ -49,27 +49,28 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
     Raises
     ------
     NodeNotFound
        If node(s) in C are not present in G.
 
     Returns
     -------
-    betweenness : float
-       Group betweenness centrality of the group C.
+    betweenness : list of floats
```
dschult, line 62: If there is only one value to return should we return a float (backward compatible) and for more than one we return a list of values? @rossbar, does that violate some principle?  What's the best way to transition to returning a collection of values instead of a single value?

networkx/algorithms/centrality/group.py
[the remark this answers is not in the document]
guyroznb: If C is a list of lists  (or a set of lists) then we get `TypeError: unhashable type: 'list'`.
On second though I change el to be always a set: `if all(set(el) in V for el in C):`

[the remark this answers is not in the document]
guyroznb: Hi,
two problems with this approach:
1) C may have nodes not in the graph (like the test test_group_betweenness_node_not_in_graph).
2) C may be list of lists or set of lists and then we receive  `TypeError: unhashable type: 'list'`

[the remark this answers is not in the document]
dschult: Use `if all(el in G for el in C):` to avoid that `TypeError:` in case 2). Using `in G` instead of `in V` doesn't raise the TypeError. There isn't really any reason to have `V` defined with the new code. Let's just not define V. We can use `in G` to check for a node, and set operations can use `G.nodes`.

What do you think of checking whether the first element of C is in the Graph. If it is then we treat C as a group.  If the first element isn't in the Graph, it is presumably a container of nodes and we can assume the rest of the elements of C are containers of nodes as well. There is probably some way that this messes up too.  But I can't see it.

```python
thing = nx.utils.arbitrary_element(C)
if thing in g:
    C = [C]
nodes = {node for node in group for group in C}
if nodes - G.nodes:  # elements(s) of C are not in G
    raise nx.NodeNotFound(f"The node(s) {nodes - G.nodes} are in C but not in G")
```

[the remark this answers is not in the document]
guyroznb: I suggest something like this: `if any(el in G for el in C):`. If C has one element which is a node in the graph then we treat C as a group. The one way to mess it up would be to have every element in C not in G. In that case we would get a `TypeError: 'int' object is not iterable`. @dschult what do you think about that?

networkx/algorithms/centrality/group.py
guyroznb: should I return the values in a dictionary in case of many groups, and as a float in case of single group in order to maintain backward compatible?

networkx/algorithms/centrality/group.py
dschult: I think `dist` is already calculated in the helper functions in `betweenness.py` so we should change those two helper functions to return `S, pred, sigma, dist` ... which will require changes in every function that uses these helper functions -- which I believe is once(for each function) in `betweenness.py`, once in `percolation.py` and once in `betweenness_subset.py`.

Then in `group.py` you can call it and compute `delta`.  Does this make sense?

networkx/algorithms/centrality/group.py
dschult: I'm leery of using a dictionary because the keys could need to be a list or set of nodes. Can we return the Group Betweenness values in the same order as they appear in the original incoming list `C`?  Is that too confusing? 

Let's make it a single value if there is a single group and a list of values if `C` is a list of nodes... That is, create a flag variable in the `if thing in G:` clause to indicate scalar or list and then choose the output type based on that.

networkx/algorithms/centrality/group.py
guyroznb: I change the helper functions in betweenness.py so they return delta and D. I also change each function which uses those helper functions.
Now the group betweenness centrality function calls the helper functions from betweenness.

networkx/algorithms/centrality/group.py
guyroznb: I made the change. If C is a single group then return a single betweenness float value. If C is a list (even one containing a single group) the function returns a list of floats.

networkx/algorithms/centrality/tests/test_group.py
dschult: Looks like this is the last thing to discuss...

What do the references use? Is there a conflict in the literature? 
That would be indicated by any paper mentioning this discrepancy...

guyroznb: The reference counts the number of shortest path in undirected graph from u to v as 2 (from u to v and from v to u). That is also the case in Brandes's article. I am not familiar with a paper stating that the number of shortest path between two nodes in undirected graph is 1, but I assume that there is one if betweenness.py did that. Should I state that in the description of the algorithm?

dschult: Remember that if you use all starting nodes, you need to count each path as one because you will find the path from u to v and then again from v to u. So you count each undirected path twice even if you increment the counter by 1 for each direction.

Also, the mathematical formula given in the docstring sums over `s<t` so that indicates to me that you only count one path for v to u and u to v even though there are paths both directions.  This is an important difference (a factor of 2). So we certainly can't just change the algorithm so it returns a different value. What is the right value? Check the papers again for simple cases. I'll do that too.

networkx/algorithms/centrality/group.py
```diff
@@ -93,40 +100,134 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
        Group Centrality Maximization via Network Design.
        SIAM International Conference on Data Mining, SDM 2018, 126–134.
        https://sites.cs.ucsb.edu/~arlei/pubs/sdm18.pdf
+    .. [4] Rami Puzis, Yuval Elovici, and Shlomi Dolev.
+       "Fast algorithm for successive computation of group betweenness centrality."
+        https://journals.aps.org/pre/pdf/10.1103/PhysRevE.76.056709?casa_token=YvH6xTIc6AUAAAAA%3A90BkxMRiQyOg9y7PEU_lrcf1FXPF_sqTCI_Y_2lD0QodSeyXAQtz-s0_e1Z0CtkZ1bWdVCkK9S5fXw
```
dschult, line 105:
```suggestion
        https://journals.aps.org/pre/pdf/10.1103/PhysRevE.76.056709
```
The `casa_token` part of the URL is not needed.

networkx/algorithms/centrality/group.py
```diff
@@ -93,40 +100,134 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+                    if y != v:
+                        PB_m_v[x][y] -= PB_m[x][v] * dxyv
+                    if x != v:
+                        PB_m_v[x][y] -= PB_m[v][y] * dvxy
+            tmp = PB_m
+            PB_m = PB_m_v
+            PB_m_v = tmp
+            tmp = sigma_m
+            sigma_m = sigma_m_v
+            sigma_m_v = tmp
```
dschult, line 157:
```suggestion
            PB_m, PB_m_v = PB_m_v, PB_m
            sigma_m, sigma_m_v = sigma_m_v, sigma_m
```

networkx/algorithms/centrality/group.py
```diff
@@ -93,40 +100,134 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+            for x in group:
+                for y in group:
+                    dxvy = 0
+                    dxyv = 0
+                    dvxy = 0
+                    if not (
+                        sigma_m[x][y] == 0 or sigma_m[x][v] == 0 or sigma_m[v][y] == 0
+                    ):
+                        if D[x][v] == D[x][y] + D[y][v]:
+                            dxyv = 1.0 * sigma_m[x][y] * sigma_m[y][v] / sigma_m[x][v]
```
dschult, line 141: Why do we have the `1.0 * ` in these expressions?

networkx/algorithms/centrality/group.py
```diff
@@ -93,40 +100,134 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+            tmp = sigma_m
+            sigma_m = sigma_m_v
+            sigma_m_v = tmp
+
+        # endpoints
+        v, c = len(G), len(group)
+        if not endpoints:
+            scale = 0
+            # if the graph is connected than deduce the endpoints of all the nodes in the graph. else find which nodes
+            # are connected to the group's nodes and deduce them
```
dschult, line 164:
```suggestion
            # if the graph is connected then subtract the endpoints from 
            # the count for all the nodes in the graph. else count how many
            # nodes are connected to the group's nodes and subtract that.
```

networkx/algorithms/centrality/group.py
guyroznb: changed

networkx/algorithms/centrality/group.py
guyroznb: changed

networkx/algorithms/centrality/group.py
guyroznb: I don't need that. Deleted

networkx/algorithms/centrality/group.py
guyroznb: changed

networkx/algorithms/centrality/tests/test_group.py
guyroznb: O.K. I see that I made two mistakes:
1) **undireced vs. directed path counting**: The definition of group betweenness centrality at the references is:
image
This formula is true for both directed and undirected graphs.  So, we need to calculate the shortest path from s to t and from t to s. I change the formula in the docstring.
2) **normalize**: I made a mistake in the normalization. If the algorithm for directed and undirected graphs are the same, then the normalization factor should be the same.

dschult: OK!   Good work finding those...
I think the answers now agree with the old version except for the case of unnormalized, undirected graphs.   The test `test_group_betweenness_single_node` shows that it is different (6.0 instead of 3.0). 

Can we just add at line 180 (just after the normalization):
```
        elif not G.is_directed():
            GBC_group /= 2
```
There might be a better way... Change the algorithm so you only do one direction of paths when G is undirected. That would presumably save computation time. But I'm fine just scaling the final answer for this PR. The speedup is significant alreay.

guyroznb: The reason that `test_group_betweenness_single_node` shows different result is because G in this test is undirected. Do you want to change the result of the algorithm for the undirected unnormalized case, so it would agree with the old version (by dividing the result by 2)?

dschult: Yes...  I'd like the answers to agree with the old version of the code -- just be faster. :)
That's the only case that differs now. Let's divide the result by 2 so it represents the number of undirected shortest paths through the group.

dschult (approved):
This is ready to merge as far as I can see.

rossbar (approved):
The implementation is complicated and I admit that I didn't thoroughly compare it to the reference you provided (thanks for that!). Aside from some very minor nits though, the code/docs/testing changes LGTM! Thanks @guyroznb !

networkx/algorithms/centrality/betweenness.py
```diff
@@ -124,14 +124,14 @@ def betweenness_centrality(
     for s in nodes:
         # single source shortest paths
         if weight is None:  # use BFS
-            S, P, sigma = _single_source_shortest_path_basic(G, s)
+            S, P, sigma, D = _single_source_shortest_path_basic(G, s)
```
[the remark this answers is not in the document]
rossbar: Feel free to ignore this BTW!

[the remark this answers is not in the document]
guyroznb: yes I change it to `_` instead of `D`

networkx/algorithms/centrality/group.py
```diff
@@ -36,40 +41,41 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
 
     Notes
     -----
-    The measure is described in [1]_.
-    The algorithm is an extension of the one proposed by Ulrik Brandes for
-    betweenness centrality of nodes. Group betweenness is also mentioned in
-    his paper [2]_ along with the algorithm. The importance of the measure is
-    discussed in [3]_.
+    Group betweenness centrality is described in [1]_ and its importance discussed in [3]_.
+    initial implementation of the algorithm is mentioned in [2]_. This function uses
```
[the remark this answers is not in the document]
guyroznb: changed

networkx/algorithms/centrality/group.py
```diff
@@ -93,40 +99,133 @@ def group_betweenness_centrality(G, C, normalized=True, weight=None):
...
+                            else:
+                                scale += 2
+            GBC_group -= scale
+
+        # normalized
+        if normalized:
+            scale = 1 / ((v - c) * (v - c - 1))
+            GBC_group *= scale
+
+        # If undirected than count only the undirected edged
```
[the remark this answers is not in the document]
guyroznb: changed

dschult:
Thanks @guyroznb !

guyroznb:
Thanks for the review @dschult  @rossbar

PR #4438, opened 2020-12-09
Improves description bfs_predecessors and bfs_successors.

matnagel:
Corrects the return type of bfs_predecessors.
Adds more details to the return type of bfs_successors.

Adds doctests for bfs_predecessors and bfs_successors that
are indicative for the behaviour of these functions on DiGraphs.
Previous tests only used these functions on undirect Graphs.

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -250,8 +250,8 @@ def bfs_predecessors(G, source, depth_limit=None, sort_neighbors=None):
     Returns
     -------
     pred: iterator
-        (node, predecessors) iterator where predecessors is the list of
-        predecessors of the node.
+        (node, predecessor) iterator where `predecessor` is a predecessor of
+        `node` in a shortest path from `source` to `node`.
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -250,8 +250,8 @@ def bfs_predecessors(G, source, depth_limit=None, sort_neighbors=None):
     Returns
     -------
     pred: iterator
-        (node, predecessors) iterator where predecessors is the list of
-        predecessors of the node.
+        (node, predecessor) iterator where `predecessor` is a predecessor of
+        `node` in a shortest path from `source` to `node`.
```
dschult, line 254:
```suggestion
        (node, predecessor) iterator where `predecessor` is the predecessor of
        `node` in a breadth first search starting from `source`.
```

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -310,8 +314,9 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
     Returns
     -------
     succ: iterator
-       (node, successors) iterator where successors is the list of
-       successors of the node.
+       (node, successors) iterator where `successors` is the list of
+       successors of `node`, `node` is a transitive successor of `source`
+       and `successors` is non-empty.
```
dschult, line 319: Would it be better to define transitive successor here? Also, the first time I read this I couldn't tell if it is a list of 3 traits that successors has or if the latter phrases are simply describing other true statements. When I got to `successors is non-empty`, I decided it must be 3 ideas about successors.  I suggest splitting the sentence:

```suggestion
       (node, successors) iterator where `successors` is the non-empty list of
       successors of `node` in a breadth first search from `source`.
       To appear in the iterator,  `node` must have successors.
```

dschult (commented):
Nice -- Thanks for this!

Feel to correct/change the wording in my suggestions. They aren't deal breakers.

networkx/algorithms/traversal/breadth_first_search.py
matnagel: Thanks! Splitting the sentence makes it more readable.

I think, we should not go into (direct) successors vs. transitive successors here. If (at all), I feel the natural place would be in the doctest of DiGraph.successors (maybe that could be another pull request).

matnagel:
Thank you for suggestions! I have incorporated all of them.

dschult (approved):
Looks good to me!  Thanks!

jarrodmillman (approved):

PR #4461, opened 2020-12-14
binomial_tree() with "create_using parameter

### LINKED ISSUE #4460: Allow binomial_tree() to create directed graphs (via create_using parameter)
`binomial_tree() ` doesn't seem to allow the `create_using` parameter to get passed and as a result, one can't create a directed binomial graph. 

This seem like a inadvertent omission and easy to fix. I will put in a PR to address this soon.

rocky:
Fixes #4460

### DIFF
--- a/networkx/generators/classic.py
+++ b/networkx/generators/classic.py
@@ -200,6 +200,9 @@ def binomial_tree(n, create_using=None):
     n : int
         Order of the binomial tree.
 
+    create_using : NetworkX graph constructor, optional (default=nx.Graph)
+       Graph type to create. If graph instance, then cleared before populated.
+
     Returns
     -------
     G : NetworkX graph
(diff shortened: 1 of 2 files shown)


dschult:
It looks like the [Wikipedia entry on binomial tree](https://en.wikipedia.org/wiki/Binary_tree)indicates it could be directed or not. And other sources I found agree that the direction should be outward from the root -- just as you are implementing it.

The other case your code allows is MultiGraph/MultiDiGraph and I think this works well for that -- no multiedes are created, but I don't think people will expect them to be and this doesn't try to add the same edge twice (relying on it not being MultiGraph).  Still, it would be good to include MultiGraph and MultiDiGraph in your tests if that is easy.

Thanks for this!

rocky:
> The other case your code allows is MultiGraph/MultiDiGraph and I think this works well for that

Ok - I'll add the other two kinds of graphs to the tests when I get a chance.

rocky:
@dschult I looked at the multitable generators and the simple thing of adding them to the `create_using` iterator is not enough.

I am sorry to report that I don't understand well enough multitables for binomial trees and the motivation behind it. For the directed and undirected case, binimomial trees makes sense.

I've added checks to disallow mulititable creation. 

If you or other understand multitables and feel it is important to add here, feel free just to keep comitting on top. I think I checked the box that allows maintainers to edit the PR.

jarrodmillman (approved):

dschult (approved):

PR #4463, opened 2020-12-15
adds implementation of SNAP summarization algorithm

dpfens:
An implementation of the SNAP graph summarization algorithm from [Efficient Aggregation for Graph Summarization](http://pages.cs.wisc.edu/~jignesh/publ/summarization.pdf).

This algorithm creates a summary graph of supernodes, where supernodes represent sets of nodes that have the same attributes and edges with nodes in the same supernodes in the summary graph.

### DIFF
--- a/examples/algorithms/plot_snap.py
+++ b/examples/algorithms/plot_snap.py
@@ -57,7 +57,7 @@ plt.title(
     "Original (%s nodes, %s edges)"
     % (original_graph.number_of_nodes(), original_graph.number_of_edges())
 )
-pos = nx.spring_layout(original_graph)
+pos = nx.spring_layout(original_graph, seed=7482934)
 node_colors = [d["color"] for _, d in original_graph.nodes(data=True)]
 
 edge_type_visual_weight_lookup = {"Weak": 1.0, "Strong": 3.0}
@@ -82,7 +82,7 @@ plt.title(
     "SNAP Aggregation (%s nodes, %s edges)"
     % (summary_graph.number_of_nodes(), summary_graph.number_of_edges())
 )
-summary_pos = nx.spring_layout(summary_graph)
+summary_pos = nx.spring_layout(summary_graph, seed=8375428)
 node_colors = []
 for node in summary_graph:
     color = summary_graph.nodes[node]["color"]
--- a/examples/algorithms/plot_snap.py
+++ b/examples/algorithms/plot_snap.py
@@ -1,0 +1,108 @@
+"""
+==================
+SNAP Graph Summary
+==================
+An example of summarizing a graph based on node attributes and edge attributes
+using the Summarization by Grouping Nodes on Attributes and Pairwise
+edges (SNAP) algorithm (not to be confused with the Stanford Network
+Analysis Project).  The algorithm groups nodes by their unique
+combinations of node attribute values and edge types with other groups
+of nodes to produce a summary graph.  The summary graph can then be used to
+infer how nodes with different attributes values relate to other nodes in the
+graph.
+"""
+import networkx as nx
+import matplotlib.pyplot as plt
+
+
+nodes = {
+    "A": dict(color="Red"),
+    "B": dict(color="Red"),
+    "C": dict(color="Red"),
+    "D": dict(color="Red"),
+    "E": dict(color="Blue"),
+    "F": dict(color="Blue"),
+    "G": dict(color="Blue"),
+    "H": dict(color="Blue"),
+    "I": dict(color="Yellow"),
+    "J": dict(color="Yellow"),
+    "K": dict(color="Yellow"),
+    "L": dict(color="Yellow"),
+}
+edges = [
+    ("A", "B", "Strong"),
+    ("A", "C", "Weak"),
+    ("A", "E", "Strong"),
+    ("A", "I", "Weak"),
+    ("B", "D", "Weak"),
+    ("B", "J", "Weak"),
+    ("B", "F", "Strong"),
+    ("C", "G", "Weak"),
+    ("D", "H", "Weak"),
+    ("I", "J", "Strong"),
+    ("J", "K", "Strong"),
+    ("I", "L", "Strong"),
+]
+original_graph = nx.Graph()
+original_graph.add_nodes_from(n for n in nodes.items())
+original_graph.add_edges_from((u, v, {"type": label}) for u, v, label in edges)
+
+
+plt.suptitle("SNAP Summarization")
+
+base_options = dict(with_labels=True, edgecolors="black", node_size=500)
+
+ax1 = plt.subplot(1, 2, 1)
+plt.title(
+    "Original (%s nodes, %s edges)"
+    % (original_graph.number_of_nodes(), original_graph.number_of_edges())
+)
+pos = nx.spring_layout(original_graph)
+node_colors = [d["color"] for _, d in original_graph.nodes(data=True)]
+
+edge_type_visual_weight_lookup = {"Weak": 1.0, "Strong": 3.0}
+edge_weights = [
+    edge_type_visual_weight_lookup[d["type"]]
+    for _, _, d in original_graph.edges(data=True)
+]
+
+nx.draw_networkx(
+    original_graph, pos=pos, node_color=node_colors, width=edge_weights, **base_options
+)
+
+node_attributes = ("color",)
+edge_attributes = ("type",)
+summary_graph = nx.snap_aggregation(
+    original_graph, node_attributes, edge_attributes, prefix="S-"
+)
+
+plt.subplot(1, 2, 2)
+
+plt.title(
+    "SNAP Aggregation (%s nodes, %s edges)"
+    % (summary_graph.number_of_nodes(), summary_graph.number_of_edges())
+)
+summary_pos = nx.spring_layout(summary_graph)
+node_colors = []
+for node in summary_graph:
+    color = summary_graph.nodes[node]["color"]
+    node_colors.append(color)
+
+edge_weights = []
+for edge in summary_graph.edges():
+    edge_types = summary_graph.get_edge_data(*edge)["types"]
+    edge_weight = 0.0
+    for edge_type in edge_types:
+        edge_weight += edge_type_visual_weight_lookup[edge_type["type"]]
+    edge_weights.append(edge_weight)
+
+nx.draw_networkx(
+    summary_graph,
+    pos=summary_pos,
+    node_color=node_colors,
+    width=edge_weights,
+    **base_options
+)
+
+plt.tight_layout()
+plt.show()
--- a/networkx/algorithms/summarization.py
+++ b/networkx/algorithms/summarization.py
@@ -59,10 +59,10 @@ For more information on graph summarization, see `Graph Summarization Methods
 and Applications: A Survey <https://dl.acm.org/doi/abs/10.1145/3186727>`_
 """
 import networkx as nx
+from collections import Counter, defaultdict
 
-__all__ = [
-    "dedensify",
-]
+
+__all__ = ["dedensify", "snap_aggregation"]
 
 
 def dedensify(G, threshold, prefix=None, copy=True):
@@ -210,3 +210,352 @@ def dedensify(G, threshold, prefix=None, copy=True):
             G.add_edge(compression_node, node)
         compressor_nodes.add(compression_node)
     return G, compressor_nodes
+
+
+def _snap_build_graph(
+    G,
+    groups,
+    node_attributes,
+    edge_attributes,
+    neighbor_info,
+    edge_types,
+    prefix,
+    supernode_attribute,
+    superedge_attribute,
+):
+    """
+    Build the summary graph from the data structures produced in the SNAP aggregation algorithm
+
+    Used in the SNAP aggregation algorithm to build the output summary graph and supernode
+    lookup dictionary.  This process uses the original graph and the data structures to
+    create the supernodes with the correct node attributes, and the superedges with the correct
+    edge attributes
+
+    Parameters
+    ----------
+    G: networkx.Graph
+        the original graph to be summarized
+    groups: dict
+        A dictionary of unique group IDs and their corresponding node groups
+    node_attributes: iterable
+        An iterable of the node attributes considered in the summarization process
+    edge_attributes: iterable
+        An iterable of the edge attributes considered in the summarization process
+    neighbor_info: dict
+        A data structure indicating the number of edges a node has with the
+        groups in the current summarization of each edge type
+    edge_types: dict
+        dictionary of edges in the graph and their corresponding attributes recognized
+        in the summarization
+    prefix: string
+        The prefix to be added to all supernodes
+    supernode_attribute: str
+        The node attribute for recording the supernode groupings of nodes
+    superedge_attribute: str
+        The edge attribute for recording the edge types represented by superedges
+
+    Returns
+    -------
+    summary graph: Networkx graph
+    """
+    output = G.__class__()
+    node_label_lookup = dict()
+    for index, group_id in enumerate(groups):
+        group_set = groups[group_id]
+        supernode = "%s%s" % (prefix, index)
+        node_label_lookup[group_id] = supernode
+        supernode_attributes = {
+            attr: G.nodes[next(iter(group_set))][attr] for attr in node_attributes
+        }
+        supernode_attributes[supernode_attribute] = group_set
+        output.add_node(supernode, **supernode_attributes)
+
+    for group_id in groups:
+        group_set = groups[group_id]
+        source_supernode = node_label_lookup[group_id]
+        for other_group, group_edge_types in neighbor_info[
+            next(iter(group_set))
+        ].items():
+            if group_edge_types:
+                target_supernode = node_label_lookup[other_group]
+                summary_graph_edge = (source_supernode, target_supernode)
+
+                edge_types = [
+                    dict(zip(edge_attributes, edge_type))
+                    for edge_type in group_edge_types
+                ]
+
+                has_edge = output.has_edge(*summary_graph_edge)
+                if output.is_multigraph():
+                    if not has_edge:
+                        for edge_type in edge_types:
+                            output.add_edge(*summary_graph_edge, **edge_type)
+                    elif not output.is_directed():
+                        existing_edge_data = output.get_edge_data(*summary_graph_edge)
+                        for edge_type in edge_types:
+                            if edge_type not in existing_edge_data.values():
+                                output.add_edge(*summary_graph_edge, **edge_type)
+                else:
+                    superedge_attributes = {superedge_attribute: edge_types}
+                    output.add_edge(*summary_graph_edge, **superedge_attributes)
+
+    return output
+
+
+def _snap_eligible_group(G, groups, group_lookup, edge_types):
+    """
+    Determines if a group is eligible to be split.
+
+    A group is eligible to be split if all nodes in the group have edges of the same type(s)
+    with the same other groups.
+
+    Parameters
+    ----------
+    G: graph
+        graph to be summarized
+    groups: dict
+        A dictionary of unique group IDs and their corresponding node groups
+    group_lookup: dict
+        dictionary of nodes and their current corresponding group ID
+    edge_types: dict
+        dictionary of edges in the graph and their corresponding attributes recognized
+        in the summarization
+
+    Returns
+    -------
+    tuple: group ID to split, and neighbor-groups participation_counts data structure
+    """
+    neighbor_info = {node: {gid: Counter() for gid in groups} for node in group_lookup}
+    for group_id in groups:
+        current_group = groups[group_id]
+
+        # build neighbor_info for nodes in group
+        for node in current_group:
+            neighbor_info[node] = {group_id: Counter() for group_id in groups}
+            edges = G.edges(node, keys=True) if G.is_multigraph() else G.edges(node)
+            for edge in edges:
+                neighbor = edge[1]
+                edge_type = edge_types[edge]
+                neighbor_group_id = group_lookup[neighbor]
+                neighbor_info[node][neighbor_group_id][edge_type] += 1
+
+        # check if group_id is eligible to be split
+        group_size = len(current_group)
+        for other_group_id in groups:
+            edge_counts = Counter()
+            for node in current_group:
+                edge_counts.update(neighbor_info[node][other_group_id].keys())
+
+            if not all(count == group_size for count in edge_counts.values()):
+                # only the neighbor_info of the returned group_id is required for handling group splits
+                return group_id, neighbor_info
+
+    # if no eligible groups, complete neighbor_info is calculated
+    return None, neighbor_info
+
+
+def _snap_split(
+    groups,
+    neighbor_info,
+    group_lookup,
+    group_id,
+):
+    """
+    Splits a group based on edge types and updates the groups accordingly
+
+    Splits the group with the given group_id based on the edge types
+    of the nodes so that each new grouping will all have the same
+    edges with other nodes.
+
+    Parameters
+    ----------
+    groups: dict
+        A dictionary of unique group IDs and their corresponding node groups
+    neighbor_info: dict
+        A data structure indicating the number of edges a node has with the
+        groups in the current summarization of each edge type
+    edge_types: dict
+        dictionary of edges in the graph and their corresponding attributes recognized
+        in the summarization
+    group_lookup: dict
+        dictionary of nodes and their current corresponding group ID
+    group_id: object
+        ID of group to be split
+
+    Returns
+    -------
+    dict
+        The updated groups based on the split
+    """
+    new_group_mappings = defaultdict(set)
+    for node in groups[group_id]:
+        signature = tuple(
+            frozenset(edge_types) for edge_types in neighbor_info[node].values()
+        )
+        new_group_mappings[signature].add(node)
+
+    # leave the biggest new_group as the original group
+    new_groups = sorted(new_group_mappings.values(), key=len)
+    for new_group in new_groups[:-1]:
+        # Assign unused integer as the new_group_id
+        # ids are tuples, so will not interact with the original group_ids
+        new_group_id = len(groups)
+        groups[new_group_id] = new_group
+        groups[group_id] -= new_group
+        for node in new_group:
+            group_lookup[node] = new_group_id
+
+    return groups
+
+
+def snap_aggregation(
+    G,
+    node_attributes,
+    edge_attributes=(),
+    prefix="Supernode-",
+    supernode_attribute="group",
+    superedge_attribute="types",
+):
+    """Creates a summary graph based on attributes and connectivity.
+
+    This function uses the Summarization by Grouping Nodes on Attributes
+    and Pairwise edges (SNAP) algorithm for summarizing a given
+    graph by grouping nodes by node attributes and their edge attributes
+    into supernodes in a summary graph.  This name SNAP should not be
+    confused with the Stanford Network Analysis Project (SNAP).
+
+    Here is a high-level view of how this algorithm works:
+
+    1) Group nodes by node attribute values.
+
+    2) Iteratively split groups until all nodes in each group have edges
+    to nodes in the same groups. That is, until all the groups are homogeneous
+    in their member nodes' edges to other groups.  For example,
+    if all the nodes in group A only have edge to nodes in group B, then the
+    group is homogeneous and does not need to be split. If all nodes in group B
+    have edges with nodes in groups {A, C}, but some also have edges with other
+    nodes in B, then group B is not homogeneous and needs to be split into
+    groups have edges with {A, C} and a group of nodes having
+    edges with {A, B, C}.  This way, viewers of the summary graph can
+    assume that all nodes in the group have the exact same node attributes and
+    the exact same edges.
+
+    3) Build the output summary graph, where the groups are represented by
+    super-nodes. Edges represent the edges shared between all the nodes in each
+    respective groups.
+
+    A SNAP summary graph can be used to visualize graphs that are too large to display
+    or visually analyze, or to efficiently identify sets of similar nodes with similar connectivity
+    patterns to other sets of similar nodes based on specified node and/or edge attributes in a graph.
+
+    Parameters
+    ----------
+    G: graph
+        Networkx Graph to be summarized
+    edge_attributes: iterable, optional
+        An iterable of the edge attributes considered in the summarization process.  If provided, unique
+        combinations of the attribute values found in the graph are used to
+        determine the edge types in the graph.  If not provided, all edges
+        are considered to be of the same type.
+    prefix: str
+        The prefix used to denote supernodes in the summary graph. Defaults to 'Supernode-'.
+    supernode_attribute: str
+        The node attribute for recording the supernode groupings of nodes. Defaults to 'group'.
+    superedge_attribute: str
+        The edge attribute for recording the edge types of multiple edges. Defaults to 'types'.
+
+    Returns
+    -------
+    networkx.Graph: summary graph
+
+    Examples
+    --------
+    SNAP aggregation takes a graph and summarizes it in the context of user-provided
+    node and edge attributes such that a viewer can more easily extract and
+    analyze the information represented by the graph::
+
+    >>> nodes = {
+    ...     "A": dict(color="Red"),
+    ...     "B": dict(color="Red"),
+    ...     "C": dict(color="Red"),
+    ...     "D": dict(color="Red"),
+    ...     "E": dict(color="Blue"),
+    ...     "F": dict(color="Blue"),
+    ... }
+    >>> edges = [
+    ...     ("A", "E", "Strong"),
+    ...     ("B", "F", "Strong"),
+    ...     ("C", "E", "Weak"),
+    ...     ("D", "F", "Weak"),
+    ... ]
+    >>> G = nx.Graph()
+    >>> for node in nodes:
+    ...     attributes = nodes[node]
+    ...     G.add_node(node, **attributes)
+    ...
+    >>> for source, target, type in edges:
+    ...     G.add_edge(source, target, type=type)
+    ...
+    >>> node_attributes = ('color', )
+    >>> edge_attributes = ('type', )
+    >>> summary_graph = nx.snap_aggregation(G, node_attributes=node_attributes, edge_attributes=edge_attributes)
+
+    Notes
+    -----
+    The summary graph produced is called a maximum Attribute-edge
+    compatible (AR-compatible) grouping.  According to [1]_, an
+    AR-compatible grouping means that all nodes in each group have the same
+    exact node attribute values and the same exact edges and
+    edge types to one or more nodes in the same groups.  The maximal
+    AR-compatible grouping is the grouping with the minimal cardinality.
+
+    The AR-compatible grouping is the most detailed grouping provided by
+    any of the SNAP algorithms.
+
+    References
+    ----------
+    .. [1] Y. Tian, R. A. Hankins, and J. M. Patel. Efficient aggregation
+       for graph summarization. In Proc. 2008 ACM-SIGMOD Int. Conf.
+       Management of Data (SIGMOD’08), pages 567–580, Vancouver, Canada,
+       June 2008.
+    """
+    edge_types = {
+        edge: tuple(attrs.get(attr) for attr in edge_attributes)
+        for edge, attrs in G.edges.items()
+    }
+    if not G.is_directed():
+        if G.is_multigraph():
+            # list is needed to avoid mutating while iterating
+            edges = [((v, u, k), etype) for (u, v, k), etype in edge_types.items()]
+        else:
+            # list is needed to avoid mutating while iterating
+            edges = [((v, u), etype) for (u, v), etype in edge_types.items()]
+        edge_types.update(edges)
+
+    group_lookup = {
+        node: tuple(attrs[attr] for attr in node_attributes)
+        for node, attrs in G.nodes.items()
+    }
+    groups = defaultdict(set)
+    for node, node_type in group_lookup.items():
+        groups[node_type].add(node)
+
+    eligible_group_id, neighbor_info = _snap_eligible_group(
+        G, groups, group_lookup, edge_types
+    )
+    while eligible_group_id:
+        groups = _snap_split(groups, neighbor_info, group_lookup, eligible_group_id)
+        eligible_group_id, neighbor_info = _snap_eligible_group(
+            G, groups, group_lookup, edge_types
+        )
+    return _snap_build_graph(
+        G,
+        groups,
+        node_attributes,
+        edge_attributes,
+        neighbor_info,
+        edge_types,
+        prefix,
+        supernode_attribute,
+        superedge_attribute,
+    )
--- a/networkx/algorithms/tests/test_summarization.py
+++ b/networkx/algorithms/tests/test_summarization.py
@@ -1,9 +1,8 @@
 """
-Unit tests for dedensification/densification
+Unit tests for dedensification and graph summarization
 """
-
+import pytest
 import networkx as nx
-from networkx.algorithms.summarization import dedensify
 
 
 class TestDirectedDedensification:
@@ -45,7 +44,7 @@ class TestDirectedDedensification:
         Verify that an empty directed graph results in no compressor nodes
         """
         G = nx.DiGraph()
-        compressed_graph, c_nodes = dedensify(G, threshold=2)
+        compressed_graph, c_nodes = nx.dedensify(G, threshold=2)
         assert c_nodes == set()
 
     @staticmethod
@@ -92,7 +91,7 @@ class TestDirectedDedensification:
         """
         G = self.build_original_graph()
         compressed_G = self.build_compressed_graph()
-        compressed_graph, c_nodes = dedensify(G, threshold=2)
+        compressed_graph, c_nodes = nx.dedensify(G, threshold=2)
         for s, t in compressed_graph.edges():
             o_s = "".join(sorted(s))
             o_t = "".join(sorted(t))
@@ -108,7 +107,7 @@ class TestDirectedDedensification:
         """
         G = self.build_original_graph()
         original_edge_count = len(G.edges())
-        c_G, c_nodes = dedensify(G, threshold=2)
+        c_G, c_nodes = nx.dedensify(G, threshold=2)
         compressed_edge_count = len(c_G.edges())
         assert compressed_edge_count <= original_edge_count
         compressed_G = self.build_compressed_graph()
@@ -164,7 +163,7 @@ class TestUnDirectedDedensification:
         Verify that an empty undirected graph results in no compressor nodes
         """
         G = nx.Graph()
-        compressed_G, c_nodes = dedensify(G, threshold=2)
+        compressed_G, c_nodes = nx.dedensify(G, threshold=2)
         assert c_nodes == set()
 
     def setup_method(self):
@@ -197,7 +196,7 @@ class TestUnDirectedDedensification:
         correct edges to/from the compressor nodes in an undirected graph
         """
         G = self.build_original_graph()
-        c_G, c_nodes = dedensify(G, threshold=2)
+        c_G, c_nodes = nx.dedensify(G, threshold=2)
         v_compressed_G = self.build_compressed_graph()
         for s, t in c_G.edges():
             o_s = "".join(sorted(s))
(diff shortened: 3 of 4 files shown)


rossbar:
I'm not familiar with the SNAP algorithm, but one thing that struck me from just a quick look over was that the implementation doesn't seem to conform that well to most of the tools in NetworkX, which have a more functional flavor. Is it feasible to implement the algorithm as a function, e.g. something like

```python
def snap(G, ...):
    # Perform SNAP summarization
    return summary_graph, supernodes
```

From the included gallery example (thanks for that!) it seems that the algorithm lends itself to such an implementation.

dpfens:
@rossbar sorry for the delay, I just updated the implementation to be purely functional

dschult:
I haven't looked at the code much yet. I can make a comment on the name though...
SNAP is a big project with many aspects to it. It refers to much more than summarization. 
Can we find a better name for what this algorithm does? Even `snap_summarization` is 
more communicative than `snap`. Even better would be a word that describes what it does.
Maybe `snap_aggregation`?

dpfens:
@dschult I just updated the function name to be `snap_aggregation`.  If it helps, here is a high-level view of how this algorithm works:

1.  Group nodes by node attribute values.
2.  Iteratively split groups until all nodes in each group have edges to nodes in the same groups
3.  Build the output summary graph, where the groups are represented by super-nodes.  Edges represent the edges shared between all the nodes in each respective groups.

dschult:
Great!  Thanks for that overview.  
For step 2) does "have edges to nodes in the same groups" mean *all* nodes in the same groups? 
Or is it OK an edge to *at least one* node of the other group? 

Groups {a,b} and {c,d} with edges a-c and b-d. Would the splitting stop at this point?

**How to start?**
One Group {a, b, c, d}  any edges you like between those 4 nodes...  This looks like it satisfies stopping step 2).

Thanks!

dpfens:
@dschult Yes, that's correct, all nodes in a group must have 1 or more edges to nodes in the same groups, they are not required to have edges to all the nodes in the same groups (also, I realized that I worded step 2 poorly, so I updated it to hopefully clarify it).

In the case of groups {a,b} and {c,d} with edges a-c and b-d, the algorithm would stop splitting the groups, as all the nodes in each group have 1 or more edges with the same groups.

If all the node attributes between all nodes in the original graph are the same, then the algorithm would only split if there were some nodes that did not have any edges with any nodes.

Does that make sense?

Loosely related, there are two closely related algorithms called KSNAP (Top-Down) and KSNAP (Bottom-Up) from the same paper that implement logic where relationship heterogeneity is allowed within groups (I have [ an implementation](https://github.com/dpfens/networkx/blob/ksnap/networkx/algorithms/summarization.py#L871-L1535) for those as well, but I am still wrapping up testing).  I plan on eventually submitting a pull request for those as well.

examples/algorithms/plot_snap.py
```diff
@@ -0,0 +1,43 @@
+"""
+==================
+SNAP Graph Summary
```
dschult, line 3: Somewhere you should say that SNAP stands for:
(**S**ummarization by Grouping **N**odes on **A**ttributes and **P**airwise Relationships)

Also that it does not (in this case) stand for the Stanford Network Analysis Project (SNAP).

I was getting those confused.

dpfens: Makes sense.  Sorry for not clearing that up earlier.

I went ahead and added the distinction to both.

dpfens:
I believe the test job failed with the release of [pyyaml==5.4](https://pypi.org/project/PyYAML/#history), just submitted #4548 to address the issue.

dschult:
OK!   Thanks for that!   We've merged your fix. 
I guess that means you need to pull/rebase from networkx:master to avoid the trouble in this branch.

dpfens:
No problem.  I've merged the changes into this branch.

networkx/algorithms/summarization.py
```diff
@@ -210,3 +213,536 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+                ] += 1
+
+    return bitmap, participation_counts
+
+
+def snap_attribute_compatible_groups(G, node_attributes):
+    """
+    Produces a list of Attribute compatible (A-compatible) groupings.
+        An A-compatible grouping is where each group of nodes each have all of the
+    same values for each node attribute.
```
dschult, line 352:
```suggestion
    """Produces a list of Attribute compatible (A-compatible) groupings.

    An A-compatible grouping is where each group of nodes each have all of the
    same values for each node attribute.
```

networkx/algorithms/summarization.py
```diff
@@ -210,3 +213,536 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+                    output.add_edge(*summary_graph_edge, **edge_attributes)
+
+    return output, supernode_lookup
+
+
+def snap_build_data_structures(G, node_attributes, relationship_attributes):
+    """
+    Finds the AR-compatible grouping of graph G and returns data structures
+    to efficiently process those groupings.
+
```
dschult, line 518:
```suggestion
    """Returns the AR-compatible grouping of `G`
    
    Finds the AR-compatible grouping of graph G and returns data structures
    to efficiently process those groupings.

```

networkx/algorithms/summarization.py
```diff
@@ -210,3 +213,536 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+def snap_aggregation(
+    G, node_attributes, relationship_attributes=None, prefix="Supernode-"
+):
+    """
+    The Summarization by Grouping Nodes on Attributes and Pairwise
+    Relationships (SNAP) algorithm for summarizing a given graph by grouping
+    nodes by node attributes and their edge attributes into supernodes in a
+    summary graph.  Not to be confused with the Stanford Network Analysis
+    Project (SNAP).
```
dschult, line 687:
```suggestion
    """Creates a summary graph based on attributes and connectivity.
    
    This function uses the Summarization by Grouping Nodes on Attributes
    and Pairwise Relationships (SNAP) algorithm for summarizing a given
    graph by grouping nodes by node attributes and their edge attributes
    into supernodes in a summary graph.  This name SNAP should not be 
    confused with the Stanford Network Analysis Project (SNAP).
```

dschult:
This looks pretty good.  Thanks for the addition!!

You've got a test to look at the supernodes.
Can you add some tests to make sure the entire code is working as you'd expect?
Maybe a full (but small) example where you know what it should be and can compare to that.
(Or maybe the examples file already does this.... I didn't look at that.)

networkx/algorithms/summarization.py
dpfens: Good idea to add the steps.  I just added them.

dpfens:
@dschult I just added replaced the test with one that all aspects of the constructed summary graph match (node attributes, edge attributes, supernodes, and connectivity).  The smallest thorough example I could come up with ended up using 12 nodes and 12 edges.  It should catch any incorrect behavior in the implementation.

I also updated the example to make a clearer distinction between the original nodes and the supernodes in the summary graph.   It uses the same graph I used in the test.

dpfens:
Just updated the wording on the example.

Also, for some reason the [continuous-integration/appveyor/pr](https://ci.appveyor.com/project/dschult/networkx-pqott/build/job/n23scwnhgx6h6dtd) job randomly failed twice for Python3.9 before succeeding on the third push attempt.  I'm not sure why it failed/succeeded.

dpfens:
@dschult these changes are ready for another look

dschult:
Thanks for the nudge. :}  

I don't understand the term "relationship type" or "relationship key".  Is there a more revealing name? Does relationship have something to do with edges? And "key" is confusing especially since the returned variable is named "value". The terms "key" and "value" have specific python meanings for dicts. Is there a better term for either of these?  Miscellaneous comments follow:

- It looks like `snap_relationship_keys` is a one liner:  `{snap_relationship_key(G, edge, attrs) for edge in G.edges}`
Is it worth making a function for that?

- Similarly, what would `snap_relationship_key` be used for? Is it: 
`tuple(v for k, v in G.edges[edge] if k in relationship_attributes)`
And is there a reason for changing the empty result to `None` as is done in the code at present?  There may be reasons for these choices that I don't understand 

- The documentation is not getting linked to the reference section of the docs. Can you add to `/doc/reference/algorithms/summarization.rst`?  

- Can you go through the doc_strings and check that each starts with a single line short description followed by a blank line followed by a paragraph saying what the function does and how it does it. 

- Can you remove the `logger` stuff -- or explain why we need to add that as a requirement for the library?

- In your examples you are calling the function using the whole path to the function instead of `nx.snap_aggregation`.  The shorter form is better for examples and allows us to move code more easily.

- The karate_club example in the doc_string doesn't actually show anything about what is output. It is useful to show possible input values, but is there an example where you look at the supernodes or the output more generally that would help explain to a new user what the function does and/or why it might be useful?

networkx/algorithms/summarization.py
```diff
@@ -210,3 +209,627 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+    )
+
+    group_sets = snap_attribute_compatible_groups(G, node_attributes)
+    group_ids = list(range(len(group_sets)))
+
+    group_lookup = dict()
+    for group_id, group in zip(group_ids, group_sets):
+        for node in group:
+            group_lookup[node] = group_id
+
```
dschult, line 565: Use `enumerate` to finesse the double-list creation + zip idiom.  A dict-comprehension also helps.
```suggestion
group_lookup = {node: gid for gid, group in enumerate(group_sets) for node in group}
```

networkx/algorithms/summarization.py
```diff
@@ -210,3 +209,627 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+    if edge_attributes:
+        unique_edge_types = {
+            snap_edge_type_id(G, edge, edge_attributes) for edge in G.edges
+        }
+    else:
+        unique_edge_types = [None]
+    unique_edge_types = list(unique_edge_types)
+    edge_index_lookup = dict(
+        (value, index) for index, value in enumerate(unique_edge_types)
+    )
```
dschult, line 556: You can leave `unique_edge_types` as a set to use `enumerate`. And a dict-comprehension is helpful again.
Also, I don't think you need to use and `if` to set `[None]`. You can just use an empty set. Does that mess up something elsewhere?  
```suggestion
    unique_edge_types = {snap_edge_type_id(G, edge, edge_attributes) for edge in G.edges}
    edge_index_lookup = {v: i for i, v in enumerate(unique_edge_types)}
```

dschult (commented):
What do you think about staring the supernodes membership info as a node attribute on the `summary_graph`?  The attribute could be named 'groups' or there is probably a better name.

Then to get the supernames dict you would do: 
```supernodes = nx.get_node_attributes(summary_graph, 'group')```

Then the summary_graph is the only output of the `snap_aggregation` function.

File: networkx/algorithms/summarization.py (hunk 5 above)
Comment (dpfens) [reply networkx/algorithms/summarization.py]: Good idea.  I changed the `supernodes` to be stored in an attribute node called `group`.  I couldn't come up with a better name for it.  To account for potential conflicts with developer attributes, I added a `supernode_attribute` for changing the attribute name.

File: networkx/algorithms/summarization.py (hunk 6 above)
Comment (dpfens) [reply networkx/algorithms/summarization.py]: No, this definitely works.  The if/else was leftover from an earlier, more crude implementation.

File: networkx/algorithms/summarization.py (hunk 5 above)
Comment (dschult) [reply networkx/algorithms/summarization.py]: +1 for flexible name of supernode attribute

**Review by dschult — COMMENTED** [review]
I'm concerned with the large number of index -> value conversions and value-> index conversions. One of the nice things about python loops and dict-storage is that you don't need indexes. It is possible that I am missing subtle things here. But I think this code could be simplified quite a bit by eliminating the storage of indexes and IDs.

Similarly, see the `bitmap` comment below where I think you can combine the bitmap info and the participation_counts info into a single dict-of-dict structure that would simplify both creation and use of that information.

This comment sort of cuts to the heart of what you've put together here, so feel free to push back. But before you do, see if it makes the code easier to read and faster to eliminate one of the index/ID data structures and also to use dict-comprehensions to construct the data structures.

File: networkx/algorithms/summarization.py [hunk 7]
```diff
@@ -210,3 +209,622 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+):
+    """
+    Produces the neighbor-groups bitmap and group-ground participation_counts data structures
+
+    Produces a bitmap indicating if a given node in the original graph has one
+    or more edges of a given edge type to nodes in another group.  Also produces a
+    group-group participation_counts indicating how many nodes in a group have edges with
+    node(s) with another group with a given edge type.  These structures are produced from the
+    original graph G and the group ids and group sets from the previous split operation in the
+    SNAP aggregation algorithm.
```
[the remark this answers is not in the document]
dpfens: @dschult Good idea.  I was able to remove the `bitmap` and `participation_counts` data structures in favor of the `neighbor_groups`.

I found a couple place to use dict-comprehensions, but there very well may be more.  Just let me know if I missed them, and I'll update them.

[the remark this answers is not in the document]
dpfens: I have not been able to find any alternatives to `group_lookup` without introducing more overhead for finding what group an node is a part of.  Same with the `edge_index_lookup`.  If you happen to know of way to mitigate those 

As you suggested, I was able to merge `group_ids` and `group_sets` into `groups` using dict` storage.

dschult:
These errors are due to an error in a new scipy release. I'll get that fixed.
I'll make time to go through what you improved too. I hope sometime this week,
but maybe on the weekend.  Thanks!

networkx/algorithms/summarization.py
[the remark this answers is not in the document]
dschult: I looked at this some -- mostly at `edge_index_lookup` to get rid of the index lookups.

I think that `snap_edge_type_id` is called too often. Can we start by constructing a dict keyed by edge to the `snap_edge_type`?  Then we replace function calls with dict lookups -- and we don't have to reconstruct the tuple each time we examine the edge.  Something like:  
```python
snap_edge_type = {edge: tuple(attrs.get(attr) for attr in edge_attributes) for edge, attrs in G.edges.items()}
```

Next, in `snap_calculate_neighbor_groups()`, where you construct a list of edge types for each neighbor group of a node, you could switch from a list of integers to a single `collections.Counter` object which makes counting easy while avoiding indexes. Counter is a dict that is set up for counting. You don't have to populate it with the possible values to be counted -- it just returns 0 if it hasn't counted that key before.  To use it, do something like:
```python
neighbor_groups[node] = {group_id: Counter() for group_id in groups}
```
Then where you use `edge_index_lookup[edge_type]`  you can use something like:
```python
neighbor_groups[node][neighbor_group_id][snap_edge_type[(node, neighbor)]] +=1
```
That saves the edge type construction and the edge index lookup (from an inner-inner-inner loop!). 
You use  `snap_edge_type[edge]` to give a tuple instead of computing the edge type and then looking up in `edge_index_lookup` to get an integer.

Also, in that function you can use `G.neighbors(node)` for both directed and undirected cases since `G.out_neighbors` is the same as `G.neighbors` for a `DiGraph`.

It looks like another inner loop would be simplified by using the dict `snap_edge_type` in the function that builds the graph, but I haven't dug into that one much yet.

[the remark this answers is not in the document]
dpfens: I just made these changes.  I also simplified the `snap_build_graph` as you suggested as well.

dschult:
Looks good -- can you remove `edge_attributes` from the inputs to the functions that don't use it?

It looks like I forgot about having `edge_types` store both directions of the edge. That makes the creation step slightly uglier, but saves the failed lookups in the inner loop. To correct that we have to replace `for edge, attrs in G.edges.items()` which only returns one direction of an undirected edge with `for node, nbrdict in G.adjacency() for nbr, attrs in nbrdict.items()` and then in place of `edge` put `(node, nbr)`.

Come to think of it, is this algorithm suppose to run with DiGraphs?  What about MultiGraphs?
To make it work with those, we could try a different approach -- after creating `edge_types` we can add the other direction like this:
```python
edge_types.update(((v, u), edge_type) for (u, v), edge_type in edge_types.items())
```
Then of course remove the `if` code from the inner loops that switch the node order in the edge.

I think you can remove the function that constructs the groups also by creating `groups` and `group_lookup` using the nice `collections.defaultdict` structure:
```python
group_lookup = {node: tuple(attrs[attr] for attr in node_attributes) for node, attrs in G.nodes.items()}
groups = collections.defaultdict(set)
for node, node_type in node_types.items():
    groups[node_type].add(node)
```

`snap_is_eligible` also seems like it has to compute the same thing each time it is called and it might be made faster and easier to read. I think with the recent changes that you've got `neighbor_groups[node][other_group_id]` to be a Counter. If so, you could avoid the sum of the values by just checking the length of the Counter.  Counts aren't stored unless they are more than 0 so if any of them exist the sum will be positive and `group_edge_count` can be incremented.
```python
# Are there any connections to the other group?
if neighbor_groups[node][other_group_id]:
    group_edge_count += 1
```

I have a question about a loop in that function too... It loops over all `other_group_id`s. But doesn't `current_group` come up one of the times?  How do we keep it from returning True in that case? Or maybe we want it to.... I get confused.

dpfens:
Thank you for going through this in such detail.

> can you remove edge_attributes from the inputs to the functions that don't use it?

Done

This algorithm can be ran on Digraphs and MultiGraphs, so I added the code you provided to account for the other direction in `edge_types`.  

I also updated `snap_is_eligible` to be useful on a MultiGraph, where groups may be connected by multiple edges with different types.  I did have to use an inner-inner-inner loop to do it, though.

`snap_is_eligible` does check if the nodes of `current_group` all have edges of the same type(s) with  itself.  `snap_is_eligible` returns `True` for `current_group` when the group of nodes is not homogeneous in their edges (or lack thereof) with any group based on edge types.  As long as all the nodes in `current_group` are homogeneous in terms of edges (or lack thereof) and edge types with both itself and all other groups, it will return `False`.  That is why we check if `edge_counts[edge_type] in (0, group_size)`:

1.  `edge_counts[edge_type]  == 0` means `current_group` has no nodes with an edge of `edge_type` with `other_group`
2.  `edge_counts[edge_type]  == group_size` means all nodes in `current_group` have one or more edges of `edge_type` with `other_group`

dschult:
OK.... now looking at `snap_split` and `snap_update_groups`.  The fancier data structures allow us to simplify the updates to the point where we should just in-line that code and remove one function from the API. 

`snap_update_groups` can avoid using len/sum/range/zip by assigning `new_group_id` to be `len(groups)`...  it works out the same because `groups` grows at just the right rate to make it what the current code makes it.

```python
# keep the first group in it's current group and group_id
for new_group in new_groups[1:]:
    new_group_id = len(groups)
    groups[new_group_id] = new_group
    groups[old_group_id] -= new_group
    for node in new_group:
        group_lookup[node] = new_group_id
```

That's short enough to just include in the function that calls it (`snap_split`).
The `snap_split` can save a fair amount because of the `defaultdict` and `Counter` data structures. The dict `new_group_mappings` should be a defaultdict(set) so you can avoid the `setdefault`.  The next change I'm suggesting removes the `tuple(sorted(...))` portion. Instead of sorting, it'd be nice to use `set`s which compare irrespective of the order. That allows the `edge_types` to be non-sortable objects, and should speed things up too. But we can't use sets as key's in the `new_group_mappings` dict. So, use `frozenset` instead.
Finally, we can take advantage of the Counter in `neighbors_groups[node][grouping]` to avoid storing all edge_types with a boolean for `count>0`.  With a Counter, the only keys that iterate are the keys with count>0.  So the `grouping_signature` is not longer needed:
```python
# Construct new groups based on edge_types to the other groups
new_group_mappings = defaultdict(set)
for node in groups[old_group_id]:
    signature = tuple(frozenset(edge_types) for edge_types in neighbor_groups[node].values())
    new_group_mappings[signature].add(node)
new_groups = list(new_group_mappings.values())
```

When combining the two functions you could think about whether to sort `new_groups` by length to use the largest group as the "old" group_id. If sorting is likely to take longer than saving moving a few nodes, don't bother changing. If you think it might help, use code like: 
```python
new_groups = sorted(new_group_mappings.values(), key=len)
for new_group in new_groups[:-1]:
```

I also noticed that these functions return `groups` but they don't actually need to because the original `groups` was input and was updated in place.  But the API may be nicer to avoid thinking about inputs getting updated. Up to you....

Thanks!   And if you prefer I can save these kind of things up until I've looked at all the code and just make one comment...

dschult:
When you construct the graph, it looks like you add the same edge multiple times with different edge attributes.
```python
for edge_type in neighbor_groups[node][other_group]:
    edge_attributes = dict(zip(edge_attributes, edge_type))
    output.add_edge(*summary_graph_edge, **edge_attributes)
```
Two problems with how this is set up... 
1) edge_attributes is turned into a dict instead of a list on the first time through the loop. Luckily this doesn't impact the rest of the loop because iterating over a dict that has the same keys as elements in a list gives the same result.  But still, did you intend to name this something different?
2) Adding the same edge with different values for the same attributes overwrites the previous edge.  So only the last set of edge attributes will actually be present in the `output` Graph.
```python
G.add_edge(1, 2, color='red')
G.add_edge(1, 2, color='blue')
print(G.edges[1, 2])
```
gives output `{"color": "blue"}`.   What do you want here?  Should there only be one edge_type? That's what I am figuring. If so, maybe remove the loop and use:
```python
(edge_type,) = neighbor_groups[node][other_group]
edge_attr = dict(zip(edge_attributes, edge_type))
output.add_edge(summary_source_node, summary_target_node, **edge_attr)
```
But I might be missing something.

dschult:
One more question on graph construction:
The `deterministic_labels` option requires that the original nodes be sortable...
More precisely, it requires that the first elements of `supernode_lookup[k]` are sortable, which if I parse this right are the first node in each group_set. Those are original nodes, so the original nodes must be sortable. NetworkX doesn't require sortable nodes (think of a Graph whose nodes are Graphs).

Furthermore this doesn't enable the supernodes to be the same if the same graph is summarized twice. Python sets are not ordered, so the first element chosen from a `group_set` may not be the same between two summarizations. So the representative node from a group might be different and they would sort into a different order.

Does this make sense?

dschult:
Here's a speed up for `snap_is_eligible`. We can push the inner loop into CPython for speed by using Counter.update:
Also, we don't need to check for size 0 because the Counter will simply not report those edge types.
```python
    current_group = groups[group_id]
    group_size = len(current_group)
    for other_group_id in groups:
        edge_counts = Counter()
        for node in current_group:
            edge_counts.update(neighbor_groups[node][other_group_id])
        if not all(count == group_size for count in edge_counts.values()):
            return True
    return False
```

dpfens:
I really appreciate your thorough explanations.  I've merged `snap_update_groups` into `snap_splits` as you said, and updated `snap_is_eligible` to use your more efficient version.

Regarding adding the same edge multiple times with different attributes:  All of your points make sense.  I was attempting to account for cases where all nodes in a group have multiple edges (of 2 or more of the same edge types) to nodes in another group.  In the summary graph, the supernode would need to have multiple edges (representing each respective edge edge type) to the other supernode to represent the multiple edge types.

The best solution I can think is to modify the `snap_build_graph` function to store edge types as a list on each edge in the summary graph.  Something like this (truncated for brevity):

```python
summary_graph_edge = (summary_source_node, summary_target_node)
edge_types = []
for edge_type in neighbor_groups[node][other_group]:
    edge_type_attributes = dict(zip(edge_attributes, edge_type))
    edge_types.append(edge_type_attributes)
output.add_edge(*summary_graph_edge, types=edge_types)
```
where the `types` attribute on the `output` edge can be overridden via `kwargs`.  As always, I am open to other ideas if you have them.

You're right about the `deterministic_labels` option as well.  I originally added that intending to use it to write robust unit tests for this.  The current unit tests use `nx.is_isomorphic` which relies on the supernodes on the summary graph having the same labels on each execution.  I am open to removing the `deterministic_labels` option, but I am not sure how to re-write a robust unit test for this at the moment.

dschult:
Excellent -- thanks for this helpful description. 

I didn't take into account that the original graph being a MultiGraph would make the summary graph a MultiGraph as well!  So adding the same edge more than once with different edge_type will work just fine.  No need to store a list of edge types because we have the MultiGraph structure to do just what we want anyway. :}
I'm not sure if others will fall into the same trap I did, but could you leave the loop that adds edges and add a comment above it reminding the reader that this handles multiedges.  

I made a suggestion in an Edit [above](https://github.com/networkx/networkx/pull/4463#issuecomment-803510418) that also assumes we aren't dealing with multiedges. So we can safely ignore that suggestion as well. I'll add an edit that readers should ignore that edit. :}

For the `deterministic_labels`, can we include that code in the test suite? For the tests we control that there are sortable nodes in our tests so relabeling to check the answer should be OK.  I haven't looked into whether this is a reasonable suggestion or not...

Also, we should get some simple multigraph (and digraph) examples in the tests.

dschult:
We could also work on getting this merged with a `not_implemented_for` restriction to only use Graph classes and then make another PR to update it for DiGraph and MultiGraph.  I'm fine with either approach.

dpfens:
I moved the `deterministic_labels` functionality to the test suite, like you said.  It makes much more sense for that to be there. than in the actual implementation.

I also added a test for DiGraphs.  I was able to create a scenario where a summary graph for a DiGraph would have multiple edge types with another group.  Since the current implementation cannot account for that scenario, and MultiGraphs are not directed, I went ahead and changed the `snap_build_graph` function to store edge types as a list on each edge in the summary graph.  I can change it to store them another way if you have an idea.

I am willing to go ahead and make the implementation compatible with MultiGraphs, if you think that is feasible.  I was not able to create a test for a MultiGraph because I get the following error:

```
>   edge_types.update(((v, u), edge_types[(u, v)]) for (u, v) in list(edge_types))
E   ValueError: too many values to unpack (expected 2)
```
on line 365 of the `summarization module.  What is the best way to convert this loop to run on MultiGraphs as well as Graph/DiGraph instances?

dschult:
Multiedges are 3-tuples: (u, v, edge_key) where edge_key is usually an integer, but is allowed to be any hashable. You can get just the node 2-tuple using G.edges() when G is a MultiGraph.  You can get the 3-tuples by using G.edges

To make the line you ask about work don't split apart the edge:
`edge_type[e] for e in list(edge_types)`

dpfens:
That line is where we swap the source/target nodes of the edge to account for both directions of undirected edges in `edge_types`.  So far, this is the best I can come up with:

```python
edge_types.update(((edge[1], edge[0], edge[2]), edge_types[edge]) if isinstance(G, nx.MultiGraph) else ((edge[1], edge[0]), edge_types[edge]) for edge in list(edge_types))
```

In snap_calculate groups`, I am attempting to convert :
```python
 neighbors = set(G.neighbors(node))
for neighbor in neighbors:
    edge = (node, neighbor)
```
to:
```python
for edge in G.edges(node):
    neighbor = edge[1]
    edge = (node, neighbor)
```
so the logic will work as expected on MultiGraphs. But `G.edges(node)` is only returning a `2-tuple` for each edge in the MultiGraph.  I am having trouble finding an appropriate method to return 3-tuples as expected.  What is the best way to handle this scenario?

dschult:
Sorry -- I missed the context and the edge order swap...

You can use `G.is_multigraph()` to check if it is a multigraph. It is often best to do this outside the loop.  
```python
if not G.is_directed():
    # swap node order to get both directions of the undirected edge
    if G.is_multigraph():
        edge_types.update(((e[1], e[0], e[2]), edge_types[e]) for e in list(edge_types))
    else:
        edge_types.update((e[1], e[0]), edge_types[e]) for e in list(edge_types))
```

For the code in `snap_calculate_groups`:  `G.edges(keys=True)` gives 3-tuples for multigraphs. 
```python
edges = G.edges(node, keys=True) if G.is_multigraph() else G.edges(node)
for edge in edges:
    edge_type = edge_types_lookup[edge]
    ....
```

dpfens:
I believe I almost have this ready for use on MultiGraphs by modifying `snap_build_data_structures`, `snap_is_eligible` and `snap_build_graph`.

In the case of a summary MultiGraph, would it make more sense to represent each edge type as its own edge instead of having a `types` edge attribute that stores all the types represented by the edge?

dschult:
Yes, I think it makes more sense that if you start with a MultiGraph (or MultiDiGraph) you get back a MultiGraph with one edge for each edge type between the two supernodes.
:)

networkx/algorithms/summarization.py
```diff
@@ -210,3 +210,456 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+    }
+    if not G.is_directed():
+        if G.is_multigraph():
+            edge_types.update(
+                ((e[1], e[0], e[2]), edge_types[e]) for e in list(edge_types)
+            )
+        else:
+            edge_types.update(
+                ((v, u), edge_types[(u, v)]) for (u, v) in list(edge_types)
+            )
```
dschult, line 404:
```suggestion
        if G.is_multigraph():
            # list is needed to avoid mutating while iterating
            edges = [((v, u, k), etype) for (u, v, k), etype in edge_types.items()]
        else:
            # list is needed to avoid mutating while iterating
            edges = [((v, u), etype) for (u, v), etype in edge_types.items()]
        edge_types.update(edges)
```

networkx/algorithms/summarization.py
```diff
@@ -210,3 +210,456 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+
+    neighbor_groups = snap_calculate_neighbor_groups(
+        G,
+        groups,
+        edge_types,
+        group_lookup,
+    )
+    eligible_groups = [
+        g_id for g_id in groups if snap_is_eligible(groups, neighbor_groups, g_id)
+    ]
```
dschult, line 422: Does it save anything to do these two steps in a single function call?
We always construct `eligible_groups` after we call `neighbor_groups`.
Could we check for eligible groups while we construct neighbor_groups?
Maybe I just need to look at it more to understand what is going on...

networkx/algorithms/summarization.py
```diff
@@ -210,3 +210,456 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+        groups = snap_split(
+            G, groups, neighbor_groups, edge_types, group_lookup, group_id
+        )
+        neighbor_groups = snap_calculate_neighbor_groups(
+            G,
+            groups,
+            edge_types,
+            group_lookup,
+        )
+        eligible_groups = [
```
dschult, line 434: Do we need to check all the `snap_is_eligible` when we are just going to use the first one (with `pop`)?
Perhaps we could reduce the computation considerably here...

networkx/algorithms/summarization.py
```diff
@@ -210,3 +210,456 @@ def dedensify(G, threshold, prefix=None, copy=True):
...
+    Returns
+    -------
+    boolean: Indicates if the group can/should be split into multiple groups
+    """
+    current_group = groups[group_id]
+    group_size = len(current_group)
+    for other_group_id in groups:
+        edge_counts = Counter()
+        for node in current_group:
+            edge_counts.update(neighbor_groups[node][other_group_id].keys())
```
dschult, line 466:
```suggestion
    node_neighbor_groups = neighbor_groups[node]
    current_group = groups[group_id]
    group_size = len(current_group)
    for other_group_id in groups:
        edge_counts = Counter()
        for node in current_group:
            edge_counts.update(node_neighbor_groups[other_group_id])
```
Feel free to rename as you like... this just moves the dict lookup outside the loop and removes the superfluous `keys`.

networkx/algorithms/tests/test_summarization.py
```diff
@@ -1,9 +1,9 @@
 """
-Unit tests for dedensification/densification
+Unit tests for dedensification and graph summarization
 """
 
 import networkx as nx
-from networkx.algorithms.summarization import dedensify
+from networkx.algorithms import summarization
```
dschult, line 6: Doesn't `nx.dedensify` work?  Do we need to import the summarization module explicitly?
`nx.snap_aggregation` should work too (any function to put into `__all__`).

dschult (commented):
Here are some more list/dict comprehensions and other "simplifications".
The 2 big questions in these comments are:
1) would it save time/storage if we combined the `snap_calculate_neighbor_groups` and the formation of `eligible_groups`?
2) could we just get by with computing the first eligible group and avoid computing the rest?

dschult (commented):
`snap_split` doesn't need/use `edge_types` so we should remove that input parameter.
Part of this confusion is (I think) because `neighbor_groups` actually stores the edge types too -- not just the groups.  What do you think about renaming `neighbor_groups` as something more generic like `neighbor_info`. That occurs in many places so it will be tricky to get right.  What is a good, short name that captures what we store there?
Sorry for this because it is a pain to change variables names, but a better name will help the reader and a shorter name will help readability too...

While we're changing names, could `snap_build_data_structures` be called `snap_find_groups`. It is more accurate and shorter. At least this one is in few places.

Also inside `snap_split`, we now sort the groups by len, but we need to change the index on `for new_group in new_groups[1:]` to `[:-1]` to leave the biggest new_group as the old group. Right now we are leaving the smallest group as the old group.

And... Could you add some comments near that loop to help readers? Something like:
```
# leave the biggest new_group as the original group
new_groups = ...
for ...
    # Assign unused integer as the new_group_id
    # Won't interact with the original group_ids. They are tuples.
    new_group_id = ...
```

dpfens:
I made almost all of the changes you suggested.  They were good and necessary changes.  Regarding the big questions:

1.  I tried merging `snap_is_eligible` into `snap_calculate_info` (I was not able to think of a more appropriate name) and did not see any time/storage savings.  This may be due to having already changed the logic to only identify a single eligible group, instead of calculating all of them (the second question).
 2. You're right, calculating them all was completely unnecessary when only one eligible group needs to be identified at a time.  I made that change.

I had to add the `keys` to `snap_is_eligible` to account for when a node has edges (of the same type) to multiple nodes in the same group in a MultiGraph.  Without `keys`, this caused the `edge_counts` to be greater than `group_size` so the group was falsely be considered eligible for splitting.  `keys` fixes this by ensuring each edge type the `node` has with any node in the `other_group_id` is only counted once.  This ensures the upper bound for any edge type in `edge_counts` is `group_size`,s is only `group_size` when all the nodes have edges of the same type with `other_group_id`.  I used that scenario as a test in `test_summarization.TestSNAPUndirectedMulti` using the  `I` to `G` and `I` to `H` edges in the original graph.  As always,  if you have an idea of how to update that code to do the same logic better, I'm willing to change it.

dschult:
ooo.. nice work figuring out that `.keys()` is needed. That's a feature of `Counter` I hadn't thought about. I knew that `Counter.update` worked on an iterator and counted how many times each element occurs. But if that iterator is a `Counter` then it doesn't just treat it like an iterator. It adds the counts (which are like a dict's values) for the two Counter objects.  So, yes you need the `.keys()`.  That's good to realize... :}

dschult:
```python
    neighbor_info = {node: {gid: Counter() for gid in groups} for node in group_lookup}
    for edge, edge_type in edge_types.items():
        node, nbr = edge[:2]
        neighbor_info[node][group_lookup[nbr]][edge_type] += 1
```
I tested that this does the same thing as `snap_calculate_neighbor_info`.  
Does it make sense that it should? Am I missing any special cases? It handles multigraph OK.

I'm still trying to see if `snap_is_eligible` could be combined with this to shortcut the process once it finds an eligible group for splitting. The code here (if correct) shows that we are looping over the edges once. I think we have to do this before we can start counting edges between groups of each edgetype.

dpfens:
Yes, that makes sense.  The logic seems identical, so I went ahead and replaced `snap_calculate_neighbor_info` with that logic.

I have not yet been able to come up with a way to combine `snap_is_eligible` with this shortcut.  To split a group, we only need complete `neighbor_info` on `node`s in the `group_id` to be split.  Based on that, the best I have come up with so far uses the previous approach to calculating the `neighbor_info`:

```python
def snap_eligible_group(G, groups, group_lookup, edge_types):
    """
    Determines if a group is eligible to be split.

    A group is eligible to be split if all nodes in the group have edges of the same type(s)
    with the same other groups.

    Parameters
    ----------
    G: graph
        graph to be summarized
    groups: dict
        A dictionary of unique group IDs and their corresponding node groups
    group_lookup: dict
        dictionary of nodes and their current corresponding group ID
    edge_types: dict
        dictionary of edges in the graph and their corresponding attributes recognized
        in the summarization

    Returns
    -------
    tuple: group ID to split, and neighbor-groups participation_counts data structure
    """
    neighbor_info = {node: {gid: Counter() for gid in groups} for node in group_lookup}
    for group_id in groups:
        current_group = groups[group_id]

        # build neighbor_info for nodes in group
        for node in current_group:
            neighbor_info[node] = {group_id: Counter() for group_id in groups}
            edges = G.edges(node, keys=True) if G.is_multigraph() else G.edges(node)
            for edge in edges:
                neighbor = edge[1]
                edge_type = edge_types[edge]
                neighbor_group_id = group_lookup[neighbor]
                neighbor_info[node][neighbor_group_id][edge_type] += 1

        # check if all nodes have edges with the same groups with same edge types
        group_size = len(current_group)
        for other_group_id in groups:
            edge_counts = Counter()
            for node in current_group:
                edge_counts.update(neighbor_info[node][other_group_id].keys())

            if not all(count == group_size for count in edge_counts.values()):
                # only the neighbor_info of the group_id is required for handling group splits
                return group_id, neighbor_info

    # if no eligible groups, complete neighbor_info is calculated
    return None, neighbor_info
```

which would reduce the `while` loop in `snap_find_groups` to:
```python
eligible_group_id = snap_eligible_group(G, groups, group_lookup, edge_types)
while eligible_group_id:
    groups = snap_split(groups, neighbor_info, group_lookup, eligible_group_id)
    eligible_group_id = snap_eligible_group(G, groups, group_lookup, edge_types)
```

This approach gets ris of `snap_calculate_neighbor_info`.  It does not have any significant impact of the speed of the tests, but would have an impact when used on larger graphs.  It still doesn't use the shortcut you came up with though.

dschult:
Nice!!  
This definitely cuts the computation -- it does one group's nodes at a time instead of all nodes into neighbor_info and then shifting to checking if eligible.  I especially like the part that the neighbor_info for just the one group is needed for snap_split.  I think it is easier (at least for me) to read when the eligible check is right there with the neighbor_info collection code. :}

dpfens:
I just added that logic and removed `snap_calculate_neighbor_info`.

dschult:
I had a thought today (while doing something else)...  
Do we need to recompute the first groups in neighbor_info? If all nodes in them are pointing with a single edge type to each other group, then even if that other group gets split, the first groups won't need to be split. If we skip them, then neighbor_info for those nodes wouldn't have the right group_ids because of splitting of later groups. So I guess we'd have to go through once at the end and rebuild neighbor_info. Would it save time to only look at groups that haven't been examined before?

dpfens:
There are scenarios where the `neighbor_info` for a group may need to be examined multiple times.

Let's say one of the first groups (Group A) does not need to be split, but the next group (Group B) does need to be split and all of the Group B nodes share an edge with Group A. After Group B is then split, all the nodes in Group A may no longer have edges with all nodes in the newly split Group B **and** newly created Groups C etc and so Group A would need to be re-calculated to determine that it needs to be split.  To keep it simple as possible, this scenario assumes that there is only one edge type in the graph.

A visual example may be easier to quickly grasp, so here is a quick & dirty one:

illustration

dschult:
Nice description!  That means we will definitely have to recompute the `neighbor_info` after splitting.
There might even be a way to keep track of which groups must be recomputed. That would reduce the time even more, but I think that could be done in a separate PR after this is merged. 

We're getting close...  I think we just have the build_graph code basically.
Question:

- It's a little confusing with all the many different kinds of nodes here. Can we find names to make it more clear which does what?  We have `node`, `supernode`, `original_node`, `summary_source_node` and `summary_target_node`. My question is really about what these terms refer to.  `supernode` sounds like the node that will represent the group in the summary.  But then `node` is the word used to hold that node object.  Am I understanding correctly?
- Then `node_label_lookup` seems like a dict to lookup the supernode labels.  But so does `supernode_lookup`, which holds the groups for each supernode.  Generically, dicts are described by the `values` sometimes with a phrase like "keyed by ...".   So, for example, `groups` is a dict of groups keyed by group ID. I suppose it would be more complete to call that dict `groups_by_id`. and `group_lookup` as `node_type_by_node`.  But that can get long and is often not needed. Are there better names for `node_label_lookup` and `supernode_lookup` that would help me keep straight which type of keys and values these dicts hold?
- In the second loop it seems that `node` refers to what `original_node` used to be. And `summary_source_node` and `target_summary_node` are supernodes.  

- The `if` logical construct for multigraph and for non-multigraph are confusingly opposite. One has a long construction leading to adding an edge and the other seems like the opposite logical expansion leading to a continue that doesn't add the edge. It's hard to tell  if the logic is the same or different for the two cases: multigraph and graph.
- Near the end, a call to `nx.set_node_attributes` seems to set the supernode attributes, but I think that is done in the first loop. But I might be missing something.

- Finally, `snap_aggregation` consists of calling two functions that don't ever get called in any other way by any other functions. Would it be better to transfer the code of each of those functions into `snap_aggregation` so a reader doesn't have to track it down?  I guess some parts of the doc_strings should be merged as well. What do you think?  Let's not do it yet -- until the build_graph code is pretty much done. But think about it.
:)

dpfens:
>  It's a little confusing with all the many different kinds of nodes here. Can we find names to make it more clear which does what? We have node, supernode, original_node, summary_source_node and summary_target_node. My question is really about what these terms refer to. supernode sounds like the node that will represent the group in the summary. But then node is the word used to hold that node object. Am I understanding correctly?
> 
> Then node_label_lookup seems like a dict to lookup the supernode labels. But so does supernode_lookup, which holds the groups for each supernode. Generically, dicts are described by the values sometimes with a phrase like "keyed by ...". So, for example, groups is a dict of groups keyed by group ID. I suppose it would be more complete to call that dict groups_by_id. and group_lookup as node_type_by_node. But that can get long and is often not needed. Are there better names for node_label_lookup and supernode_lookup that would help me keep straight which type of keys and values these dicts hold?

To make this simpler all-around, I found a way to get rid of `node`, `original_node`, `supernode`, and `supernode_lookup`.  And I also renamed `summary_*node` to `source_supernode` and `target_supernode`.

> The if logical construct for multigraph and for non-multigraph are confusingly opposite. One has a long construction leading to adding an edge and the other seems like the opposite logical expansion leading to a continue that doesn't add the edge. It's hard to tell if the logic is the same or different for the two cases: multigraph and graph.

I figured out that I no longer need the `if...continue` logic, after all the changes we have made.  In short, the remaining `if output.is_multigraph()` logic determines if an edge with a given edge doesn't exist before adding it between too supernodes, to make sure there are no duplicate edges.

> Finally, snap_aggregation consists of calling two functions that don't ever get called in any other way by any other functions. Would it be better to transfer the code of each of those functions into snap_aggregation so a reader doesn't have to track it down? I guess some parts of the doc_strings should be merged as well. What do you think? Let's not do it yet -- until the build_graph code is pretty much done. But think about it.
> :)


This makes sense to me.  When I implement the `KSNAP` variants, I may split out `snap_build_graph` into its own function, but for now it makes complete sense to bundle everything into one function.

dschult:
If you are thinking that you will use `build_snap_graph` from another PR then leave it separate. But you could "inline" the `snap_find_groups`.

The logic in `build_snap_graph` for the multigraph case is still hard to read. And a couple of the if clauses can be done outside the loop. What do you think of something like:
```python
                has_edge = output.has_edge(*summary_graph_edge)                   
                if output.is_multigraph():                                        
                    if not has_edge:                                              
                        for edge_type in edge_types:                              
                            output.add_edge(*summary_graph_edge, **edge_type)     
                    elif not output.is_directed():                                
                        existing_edge_data = output.get_edge_data(*summary_graph_edge)                     
                        for edge_type in edge_types:
                            if edge_type not in existing_edge_data.values():                                            
                                output.add_edge(*summary_graph_edge, **edge_type) 
                    # This should not be reached... right?   If so, you could make the previous `elif:` into an `else:`
                    else:
                        raise NetworkXError("We should not find a duplicate edge for directed graph in summary graph")
```

Then -- on to the tests!  The current test class structure organizes the graph type cases nicely. But it is hard to see if the text code itself is the same, or if that differs. Certainly the graphs being tests change from one class to another.  To make that more clear and also to avoid duplicating identical code, you can put the duplicate code into the preliminary class `TestSNAP` that you already use for `node_attributes` and `deterministic_labels`. But if you add the method `test_summary_graph` to that class you'll need to change it's name so that pytest doesn't look for tests in it until it is subclassed.  Perhaps change the name from `TestSNAP` to `CheckSNAP`. Then add the `test_summary_graph` to it and remove from all the subclasses. 

Finally, all the `test_empty` tests don't use the class structure so it'd be better to use `@pytest.mark.parametrize` to loop over the 4 graph types, testing each in turn. You'll need to `import pytest` at the top to make this work.  Something like:

```python
graph_types = [nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph]

@pytest.mark.parametrize("graph_type", graph_types)
def test_summarization_empty(graph_type):
    G = graph_type()
    summary_graph = nx.snap_aggregation(G,  node_attributes=("color",))
    assert nx.is_isomorphic(summary_graph, G) is True
```

Lookin good! :}

dschult (approved):
This looks good to me!   We need another set of eyes and we're ready to merge.
:)

dpfens:
Sounds good.  Thank you so much for all your help on this, it's immensely better/faster now than in my initial implementation.  I really appreciate it.

dpfens:
@dschult the latest push changes the default value of relationship_attributes from `None` to `()` to make `relationship_attributes` an iterable when `relationship_attributes` is not specified.  It also added a test to verify that `snap_aggregation` will work when `relationship_attributes` is not specified.

dschult:
Can someone take a look at this PR?  I approved it, but it needs another and it'd be good to get in before 2.6 is released.

networkx/algorithms/tests/test_summarization.py
```diff
@@ -213,10 +212,429 @@ def test_dedensify_edge_count(self):
...
+
+
+class TestSNAPEmpty:
+    graph_types = [nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph]
+
+    @pytest.mark.parametrize("graph_type", graph_types)
+    def test_summarization_empty(self, graph_type):
+        G = graph_type()
+        summary_graph = nx.snap_aggregation(G, node_attributes=("color",))
+        assert nx.is_isomorphic(summary_graph, G) is True
```
rossbar, line 231:
```suggestion
@pytest.mark.parametrize("graph_type", [nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph])
def test_summarization_empty(self, graph_type):
    G = graph_type()
    summary_graph = nx.snap_aggregation(G, node_attributes=("color",))
    assert nx.is_isomorphic(summary_graph, G)
```

networkx/algorithms/tests/test_summarization.py
```diff
@@ -213,10 +212,429 @@ def test_dedensify_edge_count(self):
...
+
+        return nx.relabel_nodes(G, label_mapping)
+
+
+class TestSNAPNoEdgeTypes(AbstractSNAP):
+    relationship_attributes = ()
+
+    def test_summary_graph(self):
+        original_graph = self.build_original_graph()
+        summary_graph = self.build_summary_graph()
```
rossbar, line 268: This pattern is generally unsafe - calling methods that are only defined in child classes but not in the `AbstractSNAP`.

rossbar (commented):
Generally LGTM - I've left a couple suggestions and questions mostly related to minor cleanups/clarification.

There are several new functions defined in `summarization`, but it seems like only `snap_aggregation` is intended to be exposed. It might be worth prepending the other functions `snap_build_graph`, `snap_eligible_group`, and `snap_split` with underscores to make it completely clear that those functions are not intended to be used directly by the user.

The only real concern I have is with the test suite - the organization relies on an "abstract" base class, but the pattern isn't implemented properly (not a huge deal, I don't think). Also, I would characterize the tests that were added as integration tests in that the test case - it would be a nice improvement to have unit tests for the individual functions as well. Let's not worry about it here, but I'll open an issue to keep track of some suggestions for improving the tests after merge.

I'll add the approval as soon as the individual comments are addressed - the suggestions don't need to be accepted of course, I just want to make sure that you see them @dpfens ; feel free to mark as resolved as you see fit!

dpfens:
@rossbar These suggestions make sense.

> is the intent here to be able to provide a sub-set of all the node attributes? For example, if you have nodes with attributes "weight", "color", and "affinity", but you only want to perform summarization over weight and color, could you use node_attributes=("weight", "color")?

Yes, exactly.  It gives users the option of specifying which attributes they want to be consider when performing the summarization.  Your suggested wording was much clearer, so I went ahead and applied the change.

> It would be nice to have a shorter, more illustrative example, especially since this example is already included in the gallery. Just an area for improvement in the future though - not a blocker for this PR

I was able to condense the example using fewer nodes/edges:
```python
nodes = {
    "A": dict(color="Red"),
    "B": dict(color="Red"),
    "C": dict(color="Red"),
    "D": dict(color="Red"),
    "E": dict(color="Blue"),
    "F": dict(color="Blue"),
}
edges = [
    ("A", "E", "Strong"),
    ("B", "F", "Strong"),
    ("C", "E", "Weak"),
    ("D", "F", "Weak"),
]
```
I think this simpler example would readers a quicker understanding of what is happening, than in the `plot_snap` example.

> This pattern is generally unsafe - calling methods that are only defined in child classes but not in the `AbstractSNAP`.

I added a placeholder method in `AbstractSNAP` for `build_original_graph` and `build_summary_graph`:

```python
class AbstractSNAP:
    node_attributes = ("color",)

    def build_original_graph(self):
        pass

    def build_summary_graph(self):
        pass
```
would that a safer implementation?

rossbar:
> Yes, exactly. It gives users the option of specifying which attributes they want to be consider when performing the summarization. Your suggested wording was much clearer, so I went ahead and applied the change.

:+1: 

> I think this simpler example would readers a quicker understanding of what is happening, than in the plot_snap example.

Agreed! Thanks for condensing it down.

> I added a placeholder method in AbstractSNAP for build_original_graph and build_summary_graph:

Generally I think you'd want to use `@abstractmethod` to enforce this, but honestly that's all beside the point :). I have a strong preference to try to use pytest features like fixtures rather than OO patterns anyways as I think they generally end up being both faster and more readable, but let's not worry about it here. We can always tinker with the tests later - for now, the tests due what they're supposed: assert that the function behaves as you expect, so this LGTM!

rossbar (approved):
One final little tweak - I went ahead and added seeds to the layouts in the `plot_snap` example so that the graphs look the same each time the example is run.

Thanks for all the hard work @dpfens !

farellf:
Hello all,
is the snap_aggregation algorithm already available in netwokx newest version? I have tried the "plot_snap.py" example but got an error: module 'networkx' has no attribute 'snap_aggregation'

Thanks for your help.

farellf:
Dear All, 

@dpfens 
is there any version of Networkx or any implementation code that already allows to test the snap_aggregation algorithm? I've tried the example provided but got an error: _AttributeError: module 'networkx' has no attribute 'snap_aggregation_'

I would like to test the summarisation algorithm on some datasets.

Thanks.

Regards.

dpfens:
@farellf `snap_aggregation` has been implemented for v2.6, which you can install the latest version of using the following command:
```bash
pip install networkx==2.6.1
```
[This post](https://groups.google.com/g/networkx-discuss/c/cf6F3VLtW4k) (and the issue mentioned) explains that v2.6 and v2.6.1 were yanked from pypi due as some default dependencies need to removed.

farellf:
Hello @dpfens, 

it worked perfectly. I had 2.5.1 installed and using python3 (so I've used pip3)

Were you one of the author that wrote the paper on SNAP in 2008? I have read this paper for my phd research. I've tried to write an implementation of the algorithm but wasn't that successful.

Many thanks.

dpfens:
@farellf  I'm glad that worked.

No, I am not an author of the 2008 paper.  Is the modified logic you want to implement similar to the KSNAP algorithms, where only `k` (user-provided) groups exist in the final summary graph?

farellf:
@dpfens,

Yes definitely, it is is definitely close to that. Speficy a k number of groups to obtain, but not necessarily basing the stop condition on the same original attributes, ie start building and splitting according to some attribute but the condition to stop is based on something else.

farellf:
I have read the papaer again and this merely describes what I would like to have:

"_Users control how many groups are present in the summary by specifying
the required number of groups, denoted as k. There are many different groupings of
size k, thus there is a need to measure the qualities of the different groupings. The
Δ-measure is proposed to assess the quality of a k-SNAP summary by examining
how different it is to a hypothetical “ideal summary.”_"

farellf:
Hello @dpfens ,

It is me again, is there a method to get the nodes (members) of each of the Supernodes of the summary_graph, output by the Python implementation of the snap aggregation algorithm? i.e all the groups built according to the edge and node attributes provided, from a method or an extra property of summary_graph.

Thanks once more for your help.

dschult:
The original graph nodes are stored in a node attribute of the summary graph.  You can see them with
```list(summary_graph.nodes(data="group"))```, or create a dict from summary nodes to original nodes using ```dict(summary_graph.nodes(data="group"))```

farellf:
Morning @dschult,

Thank you very much.

PR #4474, opened 2020-12-20
Fixed docs + added decorator for k_components

atomassi:
Just re-added the decorator and updated the docs for the `k_components` function in the `networkx.algorithms.approximation` module.


dschult (approved):
Thanks!

rossbar (approved):
Thanks @atomassi !

PR #4476, opened 2020-12-20
Approximated Diameter 

### LINKED ISSUE #4475: Approximated Metrics - Diameter
Hi all,
In some cases, especially for large graphs, an exact result for a metric might not be necessary and a bound could be good enough.  
This might be the case for the *diameter*, which currently on networkx is present only as an [exact metric](https://networkx.org/documentation/stable//reference/algorithms/generated/networkx.algorithms.distance_measures.diameter.html).  

I think it might be worth to add functionalities to computed an approximated diameter (i.e., a lower bound).  
Good candidates might be the *2-sweep* [[1]](#1) and *2d-sweep* [[2]](#2)  for the undirected and directed case, respectively. 

In addition to the results on the paper, here some comparisons made on my laptop (2.9 GHz Dual-Core Intel Core i5) with a draft implementation of the *2-sweep* on undirected graphs considering the [Twitch Snap Dataset](http://snap.stanford.edu/data/twitch-social-networks.html).  
Time is expressed in seconds. The code used can be found [here](https://gist.github.com/atomassi/3ee4d827afc6f7f20e3a70bba97ed697).

| network    	| \|V\| 	|  \|E\|  	| Time (exact) 	| Value (exact) 	| Time (LB) 	| Value (LB) 	|
|------------	|:-----:	|:-------:	|:------------:	|:-------------:	|:---------:	|:----------:	|
| musae_ENGB 	| 7,126 	|  35,324 	|     166.7    	|       10      	|    0.06   	|     10     	|
| musae_RU   	| 4,385 	|  37,304 	|     63.2     	|       9       	|    0.03   	|      9     	|
| musae_FR   	| 6,549 	| 112,666 	|     255.5    	|       7       	|    0.11   	|      7     	|
| musae_DE   	| 9,498 	| 153,138 	|     602.7    	|       7       	|    0.18   	|      7     	|
| musae_ES   	| 4,648 	|  59,382 	|     101.9    	|       9       	|    0.07   	|      8     	|
| musae_PTBR 	| 1,912 	|  31,299 	|     17.5     	|       7       	|    0.02   	|      7     	|

If you think it might make sense to add this metric for the 2 cases, I can add a module (e.g., `networkx/algorithms/approximation/distance_measures.py`) with a proposed implementation and the tests for it.

## References
<a id="1">[1]</a> 
Magnien, C., Latapy, M. and Habib, M., 2009. *Fast computation of empirically tight bounds for the diameter of massive graphs*. Journal of Experimental Algorithmics (JEA), 13, pp.1-10.
<a id="1">[2]</a>
Crescenzi, P., Grossi, R., Lanzi, L. and Marino, A., 2012, June. *On computing the diameter of real-world directed (weighted) graphs*. In International Symposium on Experimental Algorithms (pp. 99-110). Springer, Berlin, Heidelberg.

atomassi:
Closes #4475 

The PR aims at implementing a module with the functionalities to compute a `lower bound` on the diameter for both undirected and directed graphs.

### DIFF
--- a/networkx/algorithms/approximation/distance_measures.py
+++ b/networkx/algorithms/approximation/distance_measures.py
@@ -1,0 +1,95 @@
+"""Distance measures approximated metrics."""
+
+import networkx as nx
+from networkx.utils.decorators import py_random_state
+
+__all__ = ["diameter"]
+
+
+@py_random_state(1)
+def diameter(G, seed=None):
+    """Returns a lower bound on the diameter of the graph G.
+
+    The function computes a lower bound on the diameter
+    (i.e., the maximum eccentricity) of G.
+
+    @TODO to be added the detailed description and the references.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+
+    seed : integer, random_state, or None (default)
+        Indicator of random number generation state.
+        See :ref:`Randomness<randomness>`.
+
+    Returns
+    -------
+    d : integer
+       Lower Bound on the Diameter of G
+
+    Raises
+    ------
+    NetworkXError
+        If the graph is empty or consists of multiple components
+
+    Notes
+    -----
+    The algorithm was proposed in the following papers:
+
+    See Also
+    --------
+    diameter
+    eccentricity
+    """
+    if not G:
+        raise ValueError("Expected non-empty NetworkX graph!")
+    # if G is a directed graph
+    if G.is_directed():
+        if not nx.is_strongly_connected(G):
+            raise nx.NetworkXError("DiGraph not strongly connected.")
+        return _two_sweep_directed(G, seed)
+    # else if G is an undirected graph
+    else:
+        if not nx.is_connected(G):
+            raise nx.NetworkXError("Graph not connected.")
+        return _two_sweep_undirected(G, seed)
+
+
+def _two_sweep_undirected(G, seed=None):
+    """Helper function for finding a lower bound on the diameter
+        for undirected Graphs.
+
+        The idea is to pick the farthest node from a random node
+        and return its eccentricity.
+
+        ``G`` is a NetworkX undirected graph.
+
+    .. note::
+
+        ``seed`` is a random.Random or numpy.random.RandomState instance
+    """
+    # pick a random node u
+    u = seed.sample(G.nodes(), 1)[0]
+    # compute the distances from node u
+    distances = nx.single_source_shortest_path_length(G, u)
+    max_distance = max(distances.values())
+    farthest_nodes = [node for node, dist in distances.items() if dist == max_distance]
+    # get a node v that is (one of) the farthest nodes from u
+    v = seed.choice(farthest_nodes)
+    # return the eccentricity of j
+    distances = nx.single_source_shortest_path_length(G, v)
+    return max(distances.values())
+
+
+def _two_sweep_directed(G, seed=None):
+    """Helper function for finding a lower bound on the diameter
+        for directed Graphs.
+
+        ``G`` is a NetworkX directed graph.
+
+    .. note::
+
+        ``seed`` is a random.Random or numpy.random.RandomState instance
+    """
+    raise NotImplementedError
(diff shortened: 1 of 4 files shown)


networkx/algorithms/approximation/distance_measures.py
```diff
@@ -0,0 +1,138 @@
...
+        ``seed`` is a random.Random or numpy.random.RandomState instance
+    """
+    # select a random source node
+    source = seed.sample(G.nodes(), 1)[0]
+    # take a node v that is (one of) the farthest nodes from the source
+    *_, (_, v) = nx.bfs_edges(G, source)
+    # compute the distances from v to the other nodes
+    distances = nx.single_source_shortest_path_length(G, v)
+    # return the eccentricity of v
+    return max(distances.values())
```
dschult, line 101: There is already an eccentricity function in networkx. Can we use that?
And `seed.choice` seems more appropriate than `sample` when selecting one node.
Finally, `shortest_path_length` has some advantages over `bfs_edges` in terms of unpacking the result to get a node. It is also faster for all graphs I tried (from lattice to barabasi-albert to binomial_graph) but about 10 times faster for the binomial.

```suggestion
    # select a random source node
    source = seed.choice(G)
    # take a node that is (one of) the farthest nodes from the source
    *_, node = nx.shortest_path_length(G, source) 
    return nx.eccentricity(G, node)
```

Similarly for directed only do it twice, once for G and once for G_reversed and take max of the two.

networkx/algorithms/approximation/distance_measures.py
```diff
@@ -0,0 +1,138 @@
...
+        if not nx.is_strongly_connected(G):
+            raise nx.NetworkXError("DiGraph not strongly connected.")
+        return _two_sweep_directed(G, seed)
+    # else if G is undirected
+    if not nx.is_connected(G):
+        raise nx.NetworkXError("Graph not connected.")
+    return _two_sweep_undirected(G, seed)
+
+
+def _two_sweep_undirected(G, seed=None):
```
dschult, line 81: For the private functions, could we take out the default value for seed? It won't work as None, and we shouldn't call it without the seed argument anyway.

dschult:
This looks nice -- take a look at my comments. They may be off-base, but see what you think.

atomassi:
Thank you for the review! I totally agree with your suggestions and I’ve updated the pull request accordingly.

dschult:
Great!   Now I have one more thought -- When we check for a connected or strongly_connected graph do we turn the two_sweep method into 3 sweeps? 

If so, maybe we can reduce it back to 2-sweeps by including the check for connected as a step in the two step process.  Is it correct that shortest_path_length with a source specified will return all the nodes in G if-and-only-if the graph is connected (I think so...)?  What about strongly connected for a DiGraph (I think not...)? But we go both directions with the DiGraph, so maybe there is some similar relationship. 

A 3-sweep method is still linear in number of nodes. But it'd be good to use 2-sweep if it saves time.

atomassi:
> Great! Now I have one more thought -- When we check for a connected or strongly_connected graph do we turn the two_sweep method into 3 sweeps? If so, maybe we can reduce it back to 2-sweeps by including the check for connected as a step in the two step process.

You're absolutely right and that's a very good suggestion, I didn't consider that.

> it correct that shortest_path_length with a source specified will return all the nodes in G if-and-only-if the graph is connected (I think so...)?

True, so for the undirected case we can just check the number of nodes in the dictionary returned by the `shortest_path_length` function for a chosen node.

> What about strongly connected for a DiGraph (I think not...)? But we go both directions with the DiGraph, so maybe there is some similar relationship.

For the directed case, in order to be strongly connected, we can pick a source node and check if the source can reach all the other nodes and if all the other nodes can reach the source to determine if the digraph is strongly connected.   
In practice, given a chosen node, both the dictionaries returned by `shortest_path_length` in the forward and backward case need to contain all the nodes.


I'll update the PR accordingly.

dschult:
Thanks for the gentle nudge. :} 

This looks good to me! Thanks very much for the submission.
I approve this pull request.

dschult (approved):

PR #4477, opened 2020-12-20
Update docs for clustering Fixes #4348

### LINKED ISSUE #4348: Clustering coefficient of weighted graph
Hi! I'm trying to use networkx.algorithms.cluster.clustering to calculate the clustering coefficient of a weighted network. I noticed that the formula listed in the networkx reference is slightly different from the formula in the paper being cited [1]. Could I ask why?

In the [networkx reference](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.clustering.html#networkx.algorithms.cluster.clustering), the constant number in the numerator is 1. In the paper[1], that constant number is 2. 

I believe it should be 2, as it should be adapted from the calculation of all possible triangles, namely deg(u)(deg(u-1))/2. 

Thanks!

[1] Intensity and coherence of motifs in weighted complex networks by J. P. Onnela, J. Saramäki, J. Kertész, and K. Kaski, Physical Review E, 71(6), 065103 (2005).

dschult:
Tries to make more clear that number of # of triangles double counts in the undirected functions.
Doesn't change that behavior (which also saves dividing by 2 and then multiplying by 2 later).
Just tries to make it more clear.

Fixes #4348

### DIFF
--- a/networkx/algorithms/cluster.py
+++ b/networkx/algorithms/cluster.py
@@ -85,6 +85,9 @@ def _weighted_triangles_and_degree_iter(G, nodes=None, weight="weight"):
     """Return an iterator of (node, degree, weighted_triangles).
 
     Used for weighted clustering.
+    Note: this returns the geometric average weight of edges in the triangle.
+    Also, each triangle is counted twice (each direction).
+    So you may want to divide by 2.
 
     """
     if weight is None or G.number_of_edges() == 0:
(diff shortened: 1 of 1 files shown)


rossbar (approved):
This seems to clear up #4348 nicely

jarrodmillman (approved):

PR #4486, opened 2020-12-23
Fix bfs_predecessors to reverse edges for directed graphs

Chris7:
This may be a misunderstanding on my part, but I believe bfs_predecessors is broken for directed graphs. Here is a sample code showing a simple example where `G.predecessors` correctly returns the predecessors, but the `nx.bfs_predecessors` fails to:

```
import networkx as nx
G = nx.DiGraph()
G.add_edges_from([(1, 0), (2, 0)])
nx.draw_networkx(G, with_labels=True)
print(list(G.predecessors(0)), list(nx.bfs_predecessors(G, 0)))
[1, 2] []
```
image

I attached a patch .with a simple fix for this.

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -289,7 +289,11 @@ def bfs_predecessors(G, source, depth_limit=None, sort_neighbors=None):
     edge_bfs
     """
     for s, t in bfs_edges(
-        G, source, depth_limit=depth_limit, sort_neighbors=sort_neighbors
+        G,
+        source,
+        reverse=G.is_directed(),
+        depth_limit=depth_limit,
+        sort_neighbors=sort_neighbors,
     ):
         yield (t, s)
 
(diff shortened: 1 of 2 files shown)


dschult:
Read the docs for `nx.bfs_predecessors` and `nx.Graph.predecessors`.  They do not intend to return the same thing.  In the example you show the BFS doesn't take any steps and so visits only the source node and there is no predecessor for the source in a BFS.

Chris7:
Thanks for your quick response. I do not expect they should return the same thing, but that was a check for my understanding of what a predecessor should be (my intuition is the node(s) that point at a given node, which nx.G.predecessors matches).

I'm confused though how a BFS would make sense to find predecessors if it can't travel against the defined edge direction. It seems like it would only be a successor search then.

dschult:
Yes -- that's correct. The BFS only searches outward along the directed edges of a graph. So you might think of the traversal as being successors only. But the search itself follows a directed tree across the network with each node having a parent node from which it was first "discovered" in the search. So the word "bfs_predecessor" is used to refer to a predecessor/parent node during the BFS process rather than a predecessor in the original graph/network.  We apparently expect the adjective BFS in "BFS predecessor" to signify the context (either original graph or the search tree) within which the predecessor is determined.

Chris7:
Thanks a ton for your detailed explanation! And also, for maintaining/updating/expanding this package long after the initial publication. Its been a bedrock of graph analysis for me in grad school (for plotting out reaction networks and contextualizing with metabolomic/transcriptomic/proteomic/etc. data) and now in industry (I use it often to provide visuals of sample transfer histories for our liquid handlers at Ginkgo).

Enjoy the holidays/new year.

PR #4491, opened 2020-12-25
Simplify version information

jarrodmillman:


stefanv (commented):
This looks good to me!

The only part I don't understand is "We no longer require that prepared tarballs include a version.py."

jarrodmillman:
This still doesn't work.  I am changing it in stages, so I can check that I am not breaking it more...

jarrodmillman:
See #4492.

PR #4492, opened 2020-12-26
Simplify version information

jarrodmillman:
As part of the pygraphviz cleanup, I noticed that it had a unmaintained version number system.  I looked into the nx version number system and noticed it was also largely unmaintained (although more up to date than the one in pygraphviz).

## What

I planned to update ``networkx/release.py`` and started in #4491.  But decided we didn't need to have something complicated.  So this PR removes ``networkx/release.py`` and there will no longer be an autogenerated ``networkx/version.py``.  Instead, we would now have something like
```
__version__ = "2.6rc1.dev0"
```
in ``networkx/__init__.py``.  I moved the other useful information in ``networkx/release.py`` to ``setup.py``.

Moving forward everytime we make a release, the release maintainer will update the version number.  For example,
```
__version__ = "2.6rc1.dev0"  # now
__version__ = "2.6rc2.dev0"  # release 2.6rc1
__version__ = "2.6rc2"  # if needed, otherwise
__version__ = "2.6"  # release 2.6
__version__ = "2.7rc1.dev0"  # immediately after 2.6 release
__version__ = "2.7rc2.dev0"  # release 2.7rc1
__version__ = "2.7rc2"  # if needed, otherwise
__version__ = "3.0"  # release 3.0
__version__ = "3.1rc1.dev0"  # immediately after 3.0 release
```

## Why

The old code didn't really work and we don't use the information (it mostly didn't generate) anyway.   Now that we ask developers to test development releases using ``pip install -e .``, we always create ``networkx/version.py`` when they run that command.  Thus, generating a ``networkx/version.py``, which is never modified until they update or reinstall.  I suspect few of us ever hit the "dynamic" version number stuff and, if we did, I suspect we never look at it.  For example, the dynamic version number stuff has never been updated to work with git.  The only update to handle git was to add this line
```
    if os.path.isdir(gitdir):
        vcs = "git"
        # For now, we are not bothering with revision and tag.
```
to ``networkx.release.get_revision``.  That was put there 10 years ago and hasn't changed since.

While numpy and scipy still do something complicated, scikit-image uses the simpler ``__version__`` information (https://github.com/scikit-image/scikit-image/blob/master/skimage/__init__.py#L74).

If desired, we could put more information on https://networkx.org/documentation/latest/.  For example, if we wanted to list the git commit hash, we could do that in ``doc/Makefile`` or ``doc/conf.py`` with something like
```
git rev-parse --short HEAD
```


dschult (approved):
This looks good to me. Simplifying this code is long overdue.
Thansk!

PR #4496, opened 2020-12-27
[DRAFT] Autogenerating lazy initialization everywhere

Erotemic:
Forked from the conversation in #4349 this is a test to see how easy it is to switch to mkinit for `__init__.py` file management. This tests to see if lazy imports work. So far I have pytest running with 134 failed, 4115 passed.


stefanv:
It'd be great if mkinit could support the new stubs loader as well.

See also https://github.com/networkx/networkx/pull/6048

Erotemic:
@stefanv Initial support for using `lazy_loader` inside mkinit is added in the 1.0.0 branch: https://github.com/Erotemic/mkinit/pull/32

Using `mkinit --lazy_loader` will now use the lazy_loader module instead of using its own boilerplate. Not quite sure how you were thinking of integrating stubs support with mkinit. Please make an issue on the mkinit repo if you can provide more details on your thoughts there.

stefanv:
Thanks, Jon! I'm out of office for a bit, but will check when I get back.

PR #4499, opened 2020-12-29
Clean up list comprehensions in bipartite/generators `configuration_model`

alexpsimone:
In algorithms/bipartite/generators.py, `configuration_model` generates several unnecessary intermediate lists. This PR removes those and changes one list comprehension a to generator.

This is my first PR for NetworkX; any guidance/feedback is welcome and appreciated.


dschult (approved):
Thanks...  
These clean-ups are definitely better -- and few enough of them for an easy review. 👍

PR #4513, opened 2021-01-03
Delete old test

jarrodmillman:
@dschult @rossbar I tried reenabling this in case Travis was the issue.  After looking into it more, I think we should just delete this test.  I noticed it wasn't consistent even on my laptop:
time_strongly_connected_components
The code to produce the figure is here:  https://gist.github.com/jarrodmillman/a170d27fc8c5fcb5e03ff450e260ce42

### DIFF
--- a/networkx/algorithms/components/tests/test_strongly_connected.py
+++ b/networkx/algorithms/components/tests/test_strongly_connected.py
@@ -1,3 +1,4 @@
+import time
 import pytest
 import networkx as nx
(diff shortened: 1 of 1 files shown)


dschult:
I have not had good success using timing to calibrate order of magnitude dependence unless I use the same machine under similar conditions. So, it may be that this test is overly optimistic about the testing configuration. 

I'm +1 for removing it.

dschult (approved):

PR #4519, opened 2021-01-04
Clustering for signed weighted graphs

Huang-W:
The current code `(wt1 * wt2 * wt3) ** (1 / 3)` returns a complex number when the product of edge weights is negative. 

Negative edge weights can cause the `average_clustering` function to raise a `TypeError` when the parameter `count_zeroes = False`.

<details>
<summary>show testcases</summary>
Here are the graphs tested in `test_cluster.py`
`TestDirectedWeightedClustering::test_triangle_and_negative_edge`
graph1
`TestWeightedClustering::test_triangle_and_negative_edge`
graph2
`TestClustering::test_k5_signed`
graph3
`TestAverageClustering::test_average_clustering_signed`
graph4
</details>

Please let me know if I am misunderstanding "clustering coefficients", if the current code is as intended, or if there is anything that should be changed.

### DIFF
--- a/networkx/algorithms/cluster.py
+++ b/networkx/algorithms/cluster.py
@@ -3,7 +3,7 @@
 from itertools import chain
 from itertools import combinations
 from collections import Counter
-from math import pow
+from numpy import cbrt
 
 from networkx.utils import not_implemented_for
 
@@ -85,7 +85,7 @@ def _triangles_and_degree_iter(G, nodes=None):
 def _weighted_triangles_and_degree_iter(G, nodes=None, weight="weight"):
     """Return an iterator of (node, degree, weighted_triangles).
 
-    Used for weighted (clustering).
+    Used for weighted clustering.
     Note: this returns the geometric average weight of edges in the triangle.
     Also, each triangle is counted twice (each direction).
     So you may want to divide by 2.
@@ -103,14 +103,6 @@ def _weighted_triangles_and_degree_iter(G, nodes=None, weight="weight"):
     def wt(u, v):
         return G[u][v].get(weight, 1) / max_weight
 
-    def cubed_root(num):
-        if num > 0:
-            return pow(num, 1.0 / 3.0)
-        elif num < 0:
-            return -pow(abs(num), 1.0 / 3.0)
-        else:
-            return 0
-
     for i, nbrs in nodes_nbrs:
         inbrs = set(nbrs) - {i}
         weighted_triangles = 0
@@ -123,7 +115,7 @@ def _weighted_triangles_and_degree_iter(G, nodes=None, weight="weight"):
             # loop.
             wij = wt(i, j)
             weighted_triangles += sum(
-                cubed_root(wij * wt(j, k) * wt(k, i)) for k in inbrs & jnbrs
+                cbrt([(wij * wt(j, k) * wt(k, i)) for k in inbrs & jnbrs])
             )
         yield (i, len(inbrs), 2 * weighted_triangles)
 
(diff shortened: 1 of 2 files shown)


networkx/algorithms/cluster.py
```diff
@@ -102,6 +103,14 @@ def _weighted_triangles_and_degree_iter(G, nodes=None, weight="weight"):
     def wt(u, v):
         return G[u][v].get(weight, 1) / max_weight
 
+    def cubed_root(num):
+        if num > 0:
+            return pow(num, 1.0 / 3.0)
+        elif num < 0:
+            return -pow(abs(num), 1.0 / 3.0)
+        else:
+            return 0
```
dschult, line 112:
```suggestion
        if num >= 0:
            return pow(num, 1 / 3)
        return -pow(-num, 1 / 3)
```
I suppose you could use `numpy.cbrt` also...

dschult (commented):
Thanks for this.  Please beef up the documentation to explain the handling of negative edge weights. 

Do you have evidence that this is the best way to handle negative weights? There are other ways to extend the original definition to the negative weight case. 
Here's a discussion: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0088669

networkx/algorithms/cluster.py
Huang-W: From what I could gather, sign-adapted Zhang performs better than sign-adapted Onnela.

They made a few observations:
1. The coefficient that adapted Zhang produces is more **linear** than the other 2 algorithms ( when produced as a function of the proportion of negative triangles ), which makes it more human-understandable.
2. Adapted Zhang and adapted Onnela are more **noise-resistant** than adapted Watts, with similar performance when the threshold for noise is raised.

I will revert the changes for now, and work on directed graphs some other time.

Do you have any suggestions for how to proceed in implementing adapted Zhang? 
Should a signed, weighted graph be explicitly specified by the user?

Thank you for the help and direction!

Huang-W: From what I could understand, adapted Zhang is less impacted than adapted Onnela by random noise.

I will revert the changes and leave clustering for directed graphs for some other time.

Do you have any suggestions on how to proceed from here?
Thank you for the help and direction.

dschult:
Is it true that your PR implements the Onnela form?  
That seems like the one most commonly referred to in other papers. There aren't very many papers that look into this corner case (possibly negatively weighted graph clustering) in a systematic way.

Perhaps we should implement the adapted Onnela and make sure the docs state that this formula works for negative weights too.  If you (or someone else) wants to pursue the Zhang alternative we could add a separate function for that approach.

What do you think?

Huang-W:
Yes, this PR implements the adapted Onnela form mentioned in the [referenced discussion](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0088669).

If there is not much information, maybe letting the user decide which algorithm to use is a good idea.

For now, we can update the docs / code for signed Onnela and add Zhang / adapted Zhang alternative in another PR.

dschult:
That sounds like a good plan.   Thanks!

dschult:
The numpy import has to be handled specially so the package can work if someone doesn't have numpy installed.

1) put the import into each function that uses it (at the start of the code right after the doc_string).

2) In the tests you can skip some tests that will use the numpy import. Use `pytest.importorskip("numpy")`.  I think the easiest way for you to do that here is to create a `setup` method that just calls importorskip in each of the two "weighted"  classes.  I don't think any other tests will run into `np.cbrt`. But if there are others, they would need an `importorskip` too.

I hope this makes sense.

Huang-W:
Thank you.

I wasn't sure whether to remove some test cases, stop using numpy, or wait until 2.6

dschult (approved):
This looks good to me.
Thanks!

rossbar (approved):
I was initially thrown by the fact that `np.cbrt` and `** (1/3)` don't give the same result, but this makes sense. +1 for `np.cbrt`. Thanks @Huang-W !

PR #4522, opened 2021-01-05
`arrows` should be True by default for directed graphs

stefanv:


dschult:
Perhaps better to put the correct default value on [line 448](https://github.com/networkx/networkx/blob/3e12d7683676481fcaca2f4f1166789d46189c06/networkx/drawing/nx_pylab.py#L488)?  I'm fine either way, but the default should definitely be True.  

This code was changed recently -- good catch!  This is also another example of how image comparison in tests might be helpful.

stefanv:
The problem with setting it up there is that that check will then trigger this warning:

```
    if arrowstyle is not None and arrows is not None:
        warnings.warn(
            f"You passed both arrowstyle={arrowstyle} and "
            f"arrows={arrows}.  Because you set a non-default "
            "*arrowstyle*, arrows will be ignored."
        )
```

Looks like `None` is used to detect whether the user specified it intentionally or not.

dschult:
That warning was also added in the recent change to using FancyArrowPatch for both directed and undirected edges. The parameter description for `arrows` already says that the parameter is ignored if `arrowstyle` is specified. So I don't think that warning is needed. Also, it is a little misleading... If `arrows is True`, the warning is given but the arrows still show--which might be confusing. Anyway, I think we should remove that warning.

Also, the parameter descriptions in docstrings for the more general functions `draw` and `draw_networkx` should be updated. We need to make it clear that `arrowstyle` can be used for either directed or undirected graphs (if we want that -- it is a change) and we need to be clear what happens if `arrows==False` and `arrowstyle` is set to some value.

dschult (approved):

rossbar (approved):
LGTM, thanks @stefanv & @dschult !

PR #4536, opened 2021-01-12
Improve community metrics complexity and memory  (coverage and perfor…

attilamester:
The original implementation of `performance` (described in https://arxiv.org/abs/0906.0612) creates the complement graph, `coverage` builds several subgraphs to calculate the intra-community-edges and the inter-community-non-edges.
This implementation does not build any additional graphs, the calculation is based solely on the edges of the graph, and its partition. Complexity: O(L + C^2) L - edges, C - number of communities

### DIFF
--- a/networkx/algorithms/community/quality.py
+++ b/networkx/algorithms/community/quality.py
@@ -178,6 +178,7 @@ def performance(G, partition):
            <https://arxiv.org/abs/0906.0612>
 
     .. deprecated:: 2.6
+        See `nx.community.partition_quality`.
     """
     # Compute the number of intra-community edges and inter-community
     # edges.
@@ -234,6 +235,7 @@ def coverage(G, partition):
            <https://arxiv.org/abs/0906.0612>
 
     .. deprecated:: 2.6
+        See `nx.community.partition_quality`.
     """
     intra_edges = intra_community_edges(G, partition)
     total_edges = G.number_of_edges()
@@ -351,7 +353,7 @@ def partition_quality(G, partition):
 
     partition : sequence
         Partition of the nodes of `G`, represented as a sequence of
-        sets of nodes. Each block of the partition represents a
+        sets of nodes (blocks). Each block of the partition represents a
         community.
 
     Complexity
--- a/networkx/algorithms/community/quality.py
+++ b/networkx/algorithms/community/quality.py
@@ -146,6 +146,9 @@ def inter_community_non_edges(G, partition):
 def performance(G, partition):
     """Returns the performance of a partition.
 
+    .. deprecated:: 2.6
+       Use `partition_quality` instead.
+
     The *performance* of a partition is the ratio of the number of
     intra-community edges plus inter-community non-edges with the total
     number of potential edges.
@@ -177,8 +180,6 @@ def performance(G, partition):
            *Physical Reports*, Volume 486, Issue 3--5 pp. 75--174
            <https://arxiv.org/abs/0906.0612>
 
-    .. deprecated:: 2.6
-        See `nx.community.partition_quality`.
     """
     # Compute the number of intra-community edges and inter-community
     # edges.
@@ -200,6 +201,9 @@ def performance(G, partition):
 @require_partition
 def coverage(G, partition):
     """Returns the coverage of a partition.
+
+    .. deprecated:: 2.6
+       Use `partition_quality` instead.
 
     The *coverage* of a partition is the ratio of the number of
     intra-community edges to the total number of edges in the graph.
@@ -234,8 +238,6 @@ def coverage(G, partition):
            *Physical Reports*, Volume 486, Issue 3--5 pp. 75--174
            <https://arxiv.org/abs/0906.0612>
 
-    .. deprecated:: 2.6
-        See `nx.community.partition_quality`.
     """
     intra_edges = intra_community_edges(G, partition)
     total_edges = G.number_of_edges()
--- a/networkx/algorithms/community/quality.py
+++ b/networkx/algorithms/community/quality.py
@@ -4,7 +4,7 @@ communities).
 """
 
 from functools import wraps
-from itertools import product
+from itertools import product, combinations
 
 import networkx as nx
 from networkx import NetworkXError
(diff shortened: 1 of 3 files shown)


attilamester:
Comparison on Louvain communities (https://python-louvain.readthedocs.io/en/latest/), with Python's `timeit` module:

Algorithm | Time
---|---
**34 nodes, 78 edges** | http://networkrepository.com/soc-karate.php 
existing _coverage_ | `204 µs ± 1.12 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)`
existing _performance_ | `3.29 ms ± 107 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`
_fast_coverage_performance_ | `33.8 µs ± 404 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)`
**620 nodes, 2.1k edges** | http://networkrepository.com/fb-pages-food.php
existing _coverage_ | `3.68 ms ± 20.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`
existing _performance_ | `1.22 s ± 34.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
_fast_coverage_performance_ | `776 µs ± 33.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)`
**5.9k nodes, 41.7k edges** | http://networkrepository.com/fb-pages-politician.php
existing _coverage_ | `56.2 ms ± 584 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`
existing _performance_ | `2min 33s ± 6.58 s per loop (mean ± std. dev. of 7 runs, 1 loop each)`
_fast_coverage_performance_ | `17.4 ms ± 167 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`

In the existing implementation, building the complement graph, quadratic in the nodes, has a huge memory and time impact.

rossbar:
+1 for the performance improvements. My initial reaction is that it might be better to replace the existing `coverage` and `performance` implementations rather than adding a new function. Any thoughts on whether that's a reasonable idea, or is it better to have a separate "fast" function?

attilamester:
Agree. Shall I remove the existing `coverage` and `performance` functions, and introduce the new function `coverage_performance` (with unit tests, too)? Since for both metrics nearly the same steps are executed, for a large graph it is useful to calculate them together.

Or, would you like to keep `coverage` + `performance`, separately?

dschult:
I agree that they should probably be computed together.  How about the naming scheme:

```python
def partition_quality(G, partition):
    """Returns the coverage and performance of a partition of G.
```

That way deprecation is also straightfoward. The new function can be put in place with deprecation notices for the others that suggest using the new code.  As for the name, I'm open to other options, but I chose "quality" because it is a general term for what coverage and performance measure -- and it is the name of the module and not used elsewhere in the module/api.

networkx/algorithms/community/quality.py
```diff
@@ -177,6 +177,7 @@ def performance(G, partition):
            *Physical Reports*, Volume 486, Issue 3--5 pp. 75--174
            <https://arxiv.org/abs/0906.0612>
 
+    .. deprecated:: 2.6
```
dschult, line 180:
```suggestion
    .. deprecated:: 2.6
        See `nx.community.partition_quality`.
```

networkx/algorithms/community/quality.py
```diff
@@ -232,6 +233,7 @@ def coverage(G, partition):
...
```
dschult, line 236:
```suggestion
    .. deprecated:: 2.6
        See `nx.community.partition_quality`.
```

networkx/algorithms/community/quality.py
```diff
@@ -330,3 +332,98 @@ def community_contribution(community):
...
+    intra-community edges plus inter-community non-edges with the total
+    number of potential edges.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+
+    partition : sequence
+        Partition of the nodes of `G`, represented as a sequence of
+        sets of nodes. Each block of the partition represents a
```
dschult, line 354:
```suggestion
        sets of nodes (blocks). Each block of the partition represents a
```

networkx/algorithms/community/quality.py
```diff
@@ -330,3 +332,98 @@ def community_contribution(community):
...
+        for node in community:
+            node_community[node] = i
+
+    # `performance` is not defined for multigraphs
+    if not G.is_multigraph():
+        # Iterate over the communities, quadratic, to calculate `possible_inter_community_edges`
+        possible_inter_community_edges = 0
+        for i in range(0, len(partition) - 1):
+            for j in range(i + 1, len(partition)):
+                possible_inter_community_edges += len(partition[i]) * len(partition[j])
```
dschult, line 400: Probably more performant to push the loops into a library like `itertools.combinations`
```suggestion
        possible_inter_community_edges = sum(len(p1) * len(p2) for p1, p2 in itertools.combinations(partitions, 2))
```

networkx/algorithms/community/quality.py
```diff
@@ -330,3 +332,98 @@ def community_contribution(community):
...
+    inter_community_non_edges = possible_inter_community_edges
+
+    # Iterate over the links to count `intra_community_edges` and `inter_community_non_edges`
+    for e in G.edges():
+        if node_community[e[0]] == node_community[e[1]]:
+            intra_community_edges += 1
+        else:
+            inter_community_non_edges -= 1
+
+    coverage = intra_community_edges / len(G.edges())
```
dschult, line 424:
```suggestion
    coverage = intra_community_edges / G.size()
```

dschult (commented):
This looks great!  Thank you -- this is definitely the way this should be computed.

dschult:
Given my mistakes that require edits of my comments, you should probably rerun at least a subset of the performance tests. I think my suggestions should be faster, but better to check...

dschult:
The performance results still hold by my checks. 
This is ready to merge.

dschult (approved):

rossbar (commented):
I took the liberty of making a few doc changes:
 - Adding the `partition_quality` function to the reference guide autosummary
 - Moving the deprecation warnings to the top of the docstring (and modifying the links)

There's one other documentation issue - the docstring standard doesn't permit arbitrary heading names, so the Complexity section is not being parsed through to the HTML document. I left a suggestion on how to fix that, but leave it up to you how/where you'd like to add the information.

Other than that, this LGTM. Thanks @attilamester !

networkx/algorithms/community/quality.py
```diff
@@ -330,3 +334,97 @@ def community_contribution(community):
...
+    partition : sequence
+        Partition of the nodes of `G`, represented as a sequence of
+        sets of nodes (blocks). Each block of the partition represents a
+        community.
+
+    Complexity
+    ----------
+    O(C^2 + L)
+        C = number of communities
+        L = links
```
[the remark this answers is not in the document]
attilamester: Moved the complexity description in the summary, before the `Parameters` header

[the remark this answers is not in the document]
rossbar: Great thanks, it's [rendering properly](https://1501-890377-gh.circle-artifacts.com/0/doc/build/html/reference/algorithms/generated/networkx.algorithms.community.quality.partition_quality.html#networkx.algorithms.community.quality.partition_quality) now!

rossbar (approved):
Thanks @attilamester !

PR #4537, opened 2021-01-12
Replace generate_unique_node internally where not needed

dschult:
Relates to #4224
Relates to #4454

All internal uses of generate_unique_node are better served using
more specific methods to choose a node. Two uses end up finding
an unused node by counting down from -1 until the integer is not
a node. I'm not sure that justifies a utility function to do that
so I haven't implemented one, but we could make one (or a graph method)
with api: `generated_unused_node(G)` or maybe `G.find_unused_node()`.

I like the idiom:
```python
# find a node that is not in G
node = -1
while node in G:
    node -= 1
```

### DIFF
--- a/networkx/algorithms/flow/capacityscaling.py
+++ b/networkx/algorithms/flow/capacityscaling.py
@@ -8,14 +8,12 @@ from itertools import chain
 from math import log
 import networkx as nx
 from ...utils import BinaryHeap
-from ...utils import generate_unique_node
 from ...utils import not_implemented_for
 from ...utils import arbitrary_element
 
 
 def _detect_unboundedness(R):
     """Detect infinite-capacity negative cycles."""
-    s = generate_unique_node()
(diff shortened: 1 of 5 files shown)


rossbar (approved):
I definitely think this approach is preferable to the current `generate_unique_node` utility function. Do you want to deprecate `generate_unique_node` here or maybe it'd be best in a separate PR? With these changes in place, I'd be in favor of closing #4454 and deprecating instead. Does that sound like the right approach?

dschult:
Yes, now that you state it like that, it does seem better to deprecate now. I guess a user who needs it can just replace it with a direct call to uuid4.  I'm going to put the deprecation into a separate PR along with putting a description of the steps needed to deprecate into the contributors guide. I always forget at least one of them. :)  The separate PR also means it will be easier for you to check what I did wrong.    :)

PR #4546, opened 2021-01-19
updated cutoff def in weighted.py

AbhayGoyal:
The changes were made to the definition of cutoff param in the weighted.py file

### DIFF
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -249,7 +249,7 @@ def single_source_dijkstra_path(G, source, cutoff=None, weight="weight"):
        Starting node for path.
 
     cutoff : integer or float, optional
-       Length (sum of edge weights) at which the search is stopped.
+        Length (sum of edge weights) at which the search is stopped.
         If cutoff is provided, only return paths with summed weight <= cutoff.
 
     weight : string or function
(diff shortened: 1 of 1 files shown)


dschult:
For the failing style / format test, you just need to install black
```
pip install black
```
and run it on the files you are changing. You can automate this using a pre-commit hook
```
pre-commit install
```
More details about contributing to NX are available here:

AbhayGoyal:
> and run it on the files you are changing. You can automate this using a pre-commit hook
> 
> ```
> pip install -r requirements/developer.txt
> pre-commit install
> ```
> 
Could you please tell me how to use the hook? I am trying to understand where should I use it so that it can automatically be taken care of.
Thanks

dschult:
After installing `pre-commit`, each time you use `git commit` it will run black before accepting the changes. If black finds the style acceptable the commit completes as usual. If black finds style issues, it automatically makes the changes and stops the commit process. You can check (or not) the changes and then commit again. 

Personally I like to use `black --diff filename` to see what is being suggested before I rely on the pre-commit hook. But that's my personality more than any specific reason for doing it. I just don't like black changing things and not telling me what it changed. :}

AbhayGoyal:
I installed black and then tried to commit again. It still shows the styling error

dschult:
Hmmm...  Something is strange

It looks like you commit still has two lines with spaces and no letters. `black` replaces those with blank lines.  But it should do this as you commit... 

It's hard to know what is going on without more info.  So try running black locally:  `black weighted.py`
The commit that change (if there is one).

MridulS:
`black` seems happy now, failed tests are due to the recent `yaml` issue. Should be fixed after https://github.com/networkx/networkx/pull/4548 is merged.

PR #4548, opened 2021-01-20
fixes Github Actions failures

dpfens:
Github Actions failing due to new release of [pyyaml==5.4](https://pypi.org/project/PyYAML/#history) this change will ensure that version 5.3 is used.  There may be a better/simpler way to account for this issue, just let me know.

### DIFF
--- a/networkx/readwrite/nx_yaml.py
+++ b/networkx/readwrite/nx_yaml.py
@@ -102,5 +102,5 @@ def read_yaml(path):
     except ImportError as e:
         raise ImportError("read_yaml() requires PyYAML: http://pyyaml.org/") from e
 
-    G = yaml.load(path, Loader=yaml.FullLoader)
+    G = yaml.load(path, Loader=yaml.Loader)
     return G


stefanv:
:wave: @dpfens, thanks for the fix!  I don't think we can pin to such a specific range of pyyaml, so could we rather address the error introduced?

dpfens:
@stefanv Changing [this line](https://github.com/networkx/networkx/blob/master/networkx/readwrite/nx_yaml.py#L105) to use the `yaml.Loader` instead of `yaml.FullLoader`. `yaml.BaseLoader` should also work, but it could [introduce new issues](https://github.com/yaml/pyyaml/issues/266#issuecomment-600736794).

stefanv:
This seems to be an ongoing issue for PyYAML; [the issue](https://github.com/yaml/pyyaml/issues/266) @dpfens mentioned above is also the one that tracks this problem.  They sorted it out in 5.2 apparently, and now it is appearing again.

@dpfens Thanks for addressing, this seems like a fine workaround (strictly speaking unnecessary, but that's for PyYAML to address).

MridulS (approved):

dschult (approved):

jarrodmillman:
LGTM.  We are removing ``networkx/readwrite/nx_yaml.py``, so in a a couple months we won't have to worry about this anymore.

PR #4565, opened 2021-01-24
Add 'compressed' parameter to numeric_mixing_matrix and others

### LINKED ISSUE #4368: Extremely slow calculation of numeric_assortativity_coefficient
This simple example takes a minute

```python
import networkx as nx
G = nx.path_graph(2)
G.nodes[0]['attr'] = 19999
G.nodes[1]['attr'] = 20000
nx.numeric_assortativity_coefficient(G, 'attr')
```

I did not check source code, but it looks like the algorithm calculates all possible combinations of integers between 0 and 20000. An equation in corresponding Newman's article is

image

and it is easy to see that only non-zero terms make sense.

DKozl50:
As for now, this is a pull-request to fix #4368 — the problem here was that `numeric_assortativity_coefficient` was generating a 20001x20001 matrix and doing operations on it. 
These calculations happen in `numeric_mixing_matrix`: we find the greatest value of the attributes and create a matrix of this size.

The solution is to omit empty lines: we already have a set of all used attribute values, and we already pass some kind of mapping to `dict_to_numpy_array` — why not make a matrix have only 'important' values? 

My approach does it by iterating through set elements and giving them a smallest possible index. In an extreme case such as #4368 matrix gets from 20001x20001 to just 2x2. In general this change should help with graphs with 'sparse' attribute values and have almost no impact on time in the worst case.

Not only that, this solution should now work well if attribute values can be negative or even not integers — I believe the only remaining requirement is that you should be able to make a set of them.

As for 'compressed' keyword: the default values is False, so everything should behave as it did before the commit. At the moment, `numeric_assortativity_coefficient` is the only place where `numeric_mixing_matrix` is used, and here it is ok to use compressed matrix.

I assume other assortativity algorithms may have a similar problem and solution, so maybe we should add 'compressed' option to something else. 
Yet, I didn't look into it as for now.

### DIFF
--- a/networkx/algorithms/assortativity/mixing.py
+++ b/networkx/algorithms/assortativity/mixing.py
@@ -160,7 +160,7 @@ def degree_mixing_matrix(G, x="out", y="in", weight=None, nodes=None, normalized
     return a
 
 
-def numeric_mixing_matrix(G, attribute, nodes=None, normalized=True):
+def numeric_mixing_matrix(G, attribute, nodes=None, normalized=True, compressed=False):
     """Returns numeric mixing matrix for attribute.
 
     The attribute must be an integer.
@@ -179,6 +179,9 @@ def numeric_mixing_matrix(G, attribute, nodes=None, normalized=True):
     normalized : bool (default=True)
        Return counts if False or probabilities if True.
 
+    compressed: bool (default=False)
+        Build the matrix without empty lines and columns
+
     Returns
     -------
     m: numpy array
@@ -188,8 +191,11 @@ def numeric_mixing_matrix(G, attribute, nodes=None, normalized=True):
     s = set(d.keys())
     for k, v in d.items():
         s.update(v.keys())
-    m = max(s)
-    mapping = {x: x for x in range(m + 1)}
+    if compressed:
+        mapping = {x: i for x, i in zip(s, range(len(s)))}
+    else:
+        m = max(s)
+        mapping = {x: x for x in range(m + 1)}
     a = dict_to_numpy_array(d, mapping=mapping)
     if normalized:
         a = a / a.sum()
--- a/networkx/algorithms/assortativity/correlation.py
+++ b/networkx/algorithms/assortativity/correlation.py
@@ -226,7 +226,7 @@ def numeric_assortativity_coefficient(G, attribute, nodes=None):
     .. [1] M. E. J. Newman, Mixing patterns in networks
            Physical Review E, 67 026126, 2003
     """
-    a = numeric_mixing_matrix(G, attribute, nodes)
+    a = numeric_mixing_matrix(G, attribute, nodes, compressed=True)
     return numeric_ac(a)
 
 


DKozl50:
All that said I assume I made a wrong assumption that such compressed algorithm is correct, as I get different results for the same random graph in Release version and mine. 
Code to replicate:

```python
import random
import networkx as nx
random.seed(42)
N_NODES = 100
G = nx.generators.random_graphs.gnp_random_graph(N_NODES, 0.5, seed=42)
for n in G.nodes():
    G.nodes[n]['attr'] = random.randrange(1, N_NODES)
print(nx.numeric_assortativity_coefficient(G, 'attr'))
```
On release version it returns -0.022885731786025598, on my version it returns -0.024618554125752406

networkx/algorithms/assortativity/mixing.py
```diff
@@ -188,8 +191,11 @@ def numeric_mixing_matrix(G, attribute, nodes=None, normalized=True):
     s = set(d.keys())
     for k, v in d.items():
         s.update(v.keys())
-    m = max(s)
-    mapping = {x: x for x in range(m + 1)}
+    if compressed:
+        mapping = {x: i for x, i in zip(s, range(len(s)))}
+    else:
+        m = max(s)
+        mapping = {x: x for x in range(m + 1)}
```
dschult, line 198: Would the following code do what this PR is proposing? 
Can you describe why someone would want the compression option and other people not want is compressed?
I don't understand why someone would want the empty rows and columns.
```
    # collect all possible values of the attribute
    s = set(d)
    s.update(attr_value for key, nbrs in d.items() for att_value in nbrs)
    if compressed:
        # include only attribute values that appear in G for some node
        mapping = {x: i for i, x in enumerate(s)}
    else:
        # include each integer even if unused in G 
        mapping = {x: x for x in range(max(s) + 1)}
```

DKozl50: Compression parameter is added for compatibility: if I want to calculate a mixing matrix, then I would need these empty rows and columns.
In order to calculate assortativity coefficient we need a mixing matrix, and _I assumed_ that deleting empty rows and columns would significantly help with calculation time and still give correct answers. 
It really helped with time, but answers were wrong, and as for now I haven't yet found the cause of the error

dschult: Reading the papers by Newman, it seems that the difference these "empty rows" make is that they aren't empty. They are filled with zeros. So, if you average or find standard deviation those values could be quite different if you remove the zeros. 

There may be a way to do the computation without storing all the zeros. I haven't worked through it, but if we could use scipy.sparse matrices instead of arrays -- or maybe just work with the dict-of-dict which is also a sparse matrix representation.  We'd have to compute the sigmas and averages using formulas instead of the traditional functions. But it seems reasonable that we could make it fast and also correct. Certainly we don't need to store all those zeros to get the answer.

DKozl50: Yeah, my approach was wrong.
I believe it is better for me to close this pull-request and make another one when my solution is ready — the description and even name for this PR are kind of wrong

Issue #4415, opened 2020-12-03
Add a test for kernighan-lin sweep


dschult:
There was a bug in this code fixed by #4398   But there are still no tests for it. 
Let's add one that makes sure this is doing something reasonable.

ldelille:
I made a test using max_iter argument to reveal the bug fixed by #4398

PR #4575, opened 2021-01-28
Add test Kernighan Lin Algorithm

ldelille:
This is a proposal to fix issue #4415 . I added a test which calls kernighan_lin_bisection() with max_iter argument set to 1. 

Since the test case is a simple situation, the algorithm finds the minimal cut at first iteration. Before issue #4398 was fixed, the test I wrote would fail because loop on line 130 on kernighan_lin_bisection only maxe one swap ( the _kernighan_lin_sweep function was always returning 0 in second position). 

The test case I used can be found on page 7 here : http://users.ece.northwestern.edu/~haizhou/357/lec2.pdf

NB: This is my first time contributing so I would happy to have any feedbacks.

### DIFF
--- a/networkx/algorithms/community/tests/test_kernighan_lin.py
+++ b/networkx/algorithms/community/tests/test_kernighan_lin.py
@@ -62,3 +62,14 @@ def test_multigraph():
         assert_partition_equal(
             [A, B], [{mapping[0], mapping[1]}, {mapping[2], mapping[3]}]
         )
+
+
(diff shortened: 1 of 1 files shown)


dschult:
This looks good.  I've verified that the old code fails this test and that the new code works.
Thank you very much!

dschult (approved):

stefanv:
Thank you, @ldelille!

PR #4583, opened 2021-01-31
Pygraphviz choco

jarrodmillman:
Bash and PowerShell have different syntax for line continuations and for ``for`` loops.  So you can not do something like
```
    - name: Before install
      run: |
        if [ "$RUNNER_OS" == "Linux" ]; then
          sudo apt-get update
          sudo apt-get install graphviz graphviz-dev
        elif [ "$RUNNER_OS" == "macOS" ]; then
          brew install graphviz
        elif [ "$RUNNER_OS" == "Windows" ]; then
          choco install graphviz
        fi
```
without also doing something like
```
  extra:
    runs-on: ${{ matrix.os }}
    defaults:
       run:
         shell: bash
```

That uses Bash for all OSes; but then Windows fails to build extension code because the development environment for Windows is not setup/configured for Git Bash.  It may be possible to get this working, but I decided to use the default shell for each OS.

This requires moving the logic that detects what OS you are running on outside of all the ``run`` steps.  For example,
```
    - name: Before install (macOS)
      if: runner.os == 'macOS'
      run: brew install graphviz
```
You can see more about context and expression syntax here:
- https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

In particular, you can use these operators to do more complex checks:
- https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#operators

One thing that tripped me up a bit was the fact that double quotes are invalid.  That is, this prevents the whole ``test.yml`` workflow from even loading:
```
    - name: Before install (macOS)
      if: runner.os == "macOS"
      run: brew install graphviz
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Before install (non-windows)
+    - name: Before install (Linux)
       if: runner.os == 'Linux'
       run: sudo apt-get update && sudo apt-get install graphviz graphviz-dev
 
@@ -77,7 +77,7 @@ jobs:
       if: runner.os == 'macOS'
       run: brew install graphviz
 
-    - name: Before install (windows)
+    - name: Before install (Windows)
       if: runner.os == 'Windows'
       run: choco install graphviz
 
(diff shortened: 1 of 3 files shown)


rossbar (approved):
Thanks for figuring this out @jarrodmillman . Though it's a bit inconvenient to have to have OS-specific steps, it's all very clear and easy to follow. This looks good given the constraints, +1 for going with what works rather than trying to figure out how to configure a windows build environment with a different shell.

dschult (approved):
Hooray for testing graphviz on Windows!
:}   I approve.

stefanv:
Nice work, thanks!

PR #4588, opened 2021-02-01
Enhancement intersection function 3265

ldelille:
This is a proposition to fix issue #3265 - Enhancement for the intersection of two graphs. 

The code does not raise a Value Error anymore when we try to intersect two graphs that don't have the same node sets. It now returns a graph which contains all the nodes that are the intersection of the two node sets and edges that are the intersection of the two edge sets.

I added a few tests and modified those which where expecting an error when running intersection function for two different node sets. I updated documentation accordingly.

### DIFF
--- a/networkx/algorithms/operators/binary.py
+++ b/networkx/algorithms/operators/binary.py
@@ -137,7 +137,7 @@ def disjoint_union(G, H):
 
 
 def intersection(G, H):
-    """Returns a new graph that contains only the edges and the nodes that exist in
+    """Returns a new graph that contains only the nodes and the edges that exist in
     both G and H.
 
     Parameters
@@ -162,7 +162,6 @@ def intersection(G, H):
     >>> R.remove_nodes_from(n for n in G if n not in H)
     >>> R.remove_edges_from(e for e in G.edges if e not in H.egdes)
     """
-
     if not G.is_multigraph() == H.is_multigraph():
         raise nx.NetworkXError("G and H must both be graphs or multigraphs.")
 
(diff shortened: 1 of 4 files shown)


networkx/algorithms/operators/binary.py
```diff
@@ -137,15 +137,13 @@ def disjoint_union(G, H):
...
+    """Returns a new graph that contains only the nodes and the edges that exist in
     both G and H.
 
-    The node sets of H and G must be the same.
-
     Parameters
     ----------
     G,H : graph
-       A NetworkX graph.  G and H must have the same node sets.
+       A NetworkX graph. G and H can have different node sets but must be both graphs or both multigraphs.
```
dschult, line 146: Could one be a MultiGraph and the other be a Graph?
What happens in that case?

ldelille: It raises a NetworkX error, I didn't find any straightforward definition for the intersection of graphs of different types.

dschult (approved):
Looks good to me!   Thanks!

jarrodmillman (approved):

stefanv:
What happens to edge attributes?  Should whatever happens be explicitly mentioned?

ldelille:
> What happens to edge attributes? Should whatever happens be explicitly mentioned?

Like node and graph attributes, they are not copied to the new graph. It is mentioned in the Notes of the function (with a few lines of code to describe how to get the intersection graph with all attributes).

stefanv:
Great, thanks for clarifying!

PR #4589, opened 2021-02-02
Check nodelist input to floyd_warshall

### LINKED ISSUE #4587: Documentation and code in `floyd_warshall_numpy` are inconsistent
### Current Behavior

Using `floyd_warshall_numpy` with a specified set of nodes will only find paths that are confined to that subset of nodes.  I'm not sure I agree with that choice, and certainly the documentation does not make it clear.

### Expected Behavior

Based on the documentation, I would expect it to find a path that starts at one node and ends at another, even if that path must go through additional nodes not in the provided list.

### Steps to Reproduce

https://stackoverflow.com/q/65771537/2966723

### Environment

Python version:  3.9
NetworkX version:  2.5

dschult:
Add tests for NetworkXError when nodelist is not a full order.

Fixes #4587

### DIFF
--- a/networkx/algorithms/shortest_paths/dense.py
+++ b/networkx/algorithms/shortest_paths/dense.py
@@ -15,42 +15,41 @@ def floyd_warshall_numpy(G, nodelist=None, weight="weight"):
 
     This algorithm for finding shortest paths takes advantage of
     matrix representations of a graph and works well for dense
-    graphs where all-pairs shortest path is desired. The results
-    are returns in a numpy array with each column and row
-    representing a node and entries providing the distance along
-    the shortest path between that row's node and column's node.
-    If no path exists the distance is Inf.
+    graphs where all-pairs shortest path lengths are desired.
+    The results are returned as a NumPy array, distance[i, j],
+    where i and j are the indexes of two nodes in nodelist.
+    The entry distance[i, j] is the distance along a shortest
+    path from i to j. If no path exists the distance is Inf.
 
     Parameters
     ----------
     G : NetworkX graph
 
-    nodelist : list, optional (default= the order of G.nodes)
+    nodelist : list, optional (default=G.nodes)
        The rows and columns are ordered by the nodes in nodelist.
        If nodelist is None then the ordering is produced by G.nodes.
        Nodelist should include all nodes in G.
 
-    weight: string, optional (default= 'weight')
+    weight: string, optional (default='weight')
        Edge data key corresponding to the edge weight.
 
     Returns
     -------
     distance : NumPy matrix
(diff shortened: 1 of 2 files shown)


networkx/algorithms/shortest_paths/dense.py
```diff
@@ -13,13 +13,22 @@
 def floyd_warshall_numpy(G, nodelist=None, weight="weight"):
     """Find all-pairs shortest path lengths using Floyd's algorithm.
 
+    This algorithm for finding shortest paths takes advantage of
+    matrix representations of a graph and works well for dense
+    graphs where all-pairs shortest path is desired. The results
```
stefanv, line 18: Maybe a tiny typo in here?  "an all-pairs shortest path" or "all-pairs shortest paths"?

networkx/algorithms/shortest_paths/dense.py
```diff
@@ -13,13 +13,22 @@
...
+    are returns in a numpy array with each column and row
```
stefanv, line 19:
```suggestion
    are returned as a NumPy array with each column and row
```

networkx/algorithms/shortest_paths/dense.py
```diff
@@ -13,13 +13,22 @@
...
+    representing a node and entries providing the distance along
+    the shortest path between that row's node and column's node.
```
stefanv, line 21: Some commas or colons would help read this more easily.

I could also see a slightly different phrasing working, but the above should be fine with some punctuation / splitting into multiple sentences.

> The results are returned as a NumPy array, ``distance[i, j]``, where ``i`` and ``j`` represent nodes.  The entry ``distance[i, j]`` is the distance along the shortest path between ``i`` and ``j``.

networkx/algorithms/shortest_paths/dense.py
```diff
@@ -13,13 +13,22 @@
...
+    If no path exists the distance is Inf.
+
     Parameters
     ----------
     G : NetworkX graph
 
-    nodelist : list, optional
+    nodelist : list, optional (default= the order of G.nodes)
```
stefanv, line 28:
```suggestion
    nodelist : list, optional (default=G.nodes)
```

stefanv (approved):
Minor suggestions for docstring, but looks good to me.  Thanks @dschult!

networkx/algorithms/shortest_paths/dense.py
dschult: Thanks for these suggestions! It reads much better.

jarrodmillman (approved):

stefanv:
Looks great, thanks @dschult!

PR #4607, opened 2021-02-08
Add approximation algorithms for traveling salesman problem

dschult:
Combines #4083 and #3585 and brings up to master to avoid conflicts.


rossbar:
I did a first pass through the docstrings to try to get familiar with some of the new features added in this PR. These are some very nice additions! I took the liberty of making minor changes related to rst-formatting and grammar as I was reading through. I aim to do a more complete readthrough ASAP - thanks @dschult for combining & organizing the TSP contributions into one place!

networkx/algorithms/approximation/tests/test_traveling_salesman.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+        pytest.raises(
+            nx.NetworkXError,
+            nx_app.simulated_annealing_tsp,
+            self.unweightedDG,
+            source=0,
+            cycle=[],
+        )
+
+    def test_simulated_annealing_fails(self):
```
rossbar, line 220: It might be good to add a docstring to this test explaining what the expected failure mode is (too few iterations, bad tolerance, etc). If it's a concise reason, giving the test a more descriptive name would work too, e.g. `def test_simulated_annealing_fails_too_few_iterations(self)`

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,716 @@
...
+        Indicator of random number generation state.
+        See :ref:`Randomness<randomness>`.
+
+    Returns
+    -------
+    The solution after move is applied. (A neighbor solution.)
+    """
+    a = seed.randint(1, len(soln) - 2)
+    listb = list(range(1, a)) + list(range(a + 1, len(soln) - 1))
+    b = seed.choice(listb)
```
rossbar, line 711: If I parsed this correctly it looks like the intent here is to sample two indices in the range `[1, len(soln) - 1]` without replacement. If so, then I think the following should work:

```suggestion
    a, b = seed.sample(range(1, len(soln) - 1), k=2)
```

If sampling two indices without replacement wasn't the intent - then maybe add a comment here to explain the intended behavior (it took me a second to parse).

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,716 @@
...
+    if move == "1-1":
+        soln[a], soln[b] = soln[b], soln[a]
+    elif move == "1-0":
+        soln.insert(b, soln.pop(a))
```
rossbar, line 715: AFAICT there is no check for an invalid move - either add one here:

```suggestion
        soln.insert(b, soln.pop(a))
    else:
        raise ValueError(
            f"Invalid move: {move}. Expected one of ('1-1', '1-0')"
        )
```

Or add a check in the top-level SA and TA functions.

networkx/algorithms/matching.py
```diff
@@ -92,13 +98,21 @@ def is_matching(G, matching):
...
+            msg = f"Matching includes edge {e} not in G"
+            raise nx.NetworkXNotImplemented(msg)
     # TODO This is parallelizable.
     return all(len(set(e1) & set(e2)) == 0 for e1, e2 in combinations(matching, 2))
 
 
 def is_maximal_matching(G, matching):
-    """Decides whether the given set or dictionary represents a valid
-    maximal matching in ``G``.
+    """Return True if ``matching`` is a mximal matching of ``G``
```
rossbar, line 115:
```suggestion
    """Return True if ``matching`` is a maximal matching of ``G``
```

networkx/algorithms/matching.py
```diff
@@ -141,8 +155,7 @@ def is_maximal_matching(G, matching):
 
 
 def is_perfect_matching(G, matching):
-    """Decides whether the given set represents a valid perfect matching in
-    ``G``.
+    """Return True is ``matching`` is a perfect matching for ``G``
```
rossbar, line 158:
```suggestion
    """Return True if ``matching`` is a perfect matching for ``G``
```

networkx/algorithms/matching.py
```diff
@@ -176,6 +189,45 @@ def is_perfect_matching(G, matching):
     return all(counts[v] == 1 for v in G)
 
 
+@not_implemented_for("multigraph")
+@not_implemented_for("directed")
+def min_weight_matching(G, maxcardinality=False, weight="weight"):
+    """Use reciprocal edge weights to find max reciprocal weight matching.
+
+    This method replaces the weights with their reciprocal and
+    then runs :func:``max_weight_matching``.
```
rossbar, line 198:
```suggestion
    then runs :func:`max_weight_matching`.
```

rossbar (commented):
Just a few comments and questions - this looks good overall though!

dschult:
I'm collecting things to change on the TSP files as I go through the comments.
- Better names for kwargs `iterations` and `tolerance`.
- Make `cycle` a required input.
- I think the idea of taking out the highest cost edge does not give the lowest cost path through all nodes. If cutting that edge makes it possible to stop 3 edges earlier because you already traversed those nodes, then removing one edge is not sufficient.  We'd have to remove the edge from the smaller, complete graph problem -- that is, remove the edge before adding nodes to the path due to the incomplete graph.
-

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,716 @@
...
+    for e in nx.selfloop_edges(G):
+        raise ValueError("Christofides algorithm does not allow selfloops")
+    n = len(G)
+    m = G.size()
+    if (n * (n - 1) // 2) != m:
+        raise ValueError("Christofides algorithm needs a complete graph")
+
+    if tree is None:
+        tree = nx.minimum_spanning_tree(G, weight=weight)
+    L = nx.Graph(G)
```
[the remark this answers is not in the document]
dschult: The subtle difference between `nx.Graph(G)` and `G.copy()` is that if `G` is a multigraph, the first version converts it to a graph.  Here that isn't an issue because G is declared to be not directed and not multigraph.

So, I'll change it to G.copy() as you suggest. :}

networkx/algorithms/approximation/traveling_salesman.py
dschult: Yes, this took me much more than a minute to parse...  
It seems like the chance of the value `len(soln) - 1` is not the same as the chance of the other numbers.

Also, in both the code and your suggestion, 0 is not chosen.  
Why isn't 0 an allowed choice for either a or b?
Maybe it is because the list has the same node at beginning and end of the cycle.
But it could also be that we want to keep the same starting node, or we'll end up chasing our tail around the cycle.  Certainly a comment is needed here.

networkx/algorithms/approximation/traveling_salesman.py
dschult: Yes!  This is important because a mis-spelling would proceed without ever making a move!
It's probably good to put in SA and TA functions. But it's cheap to put it here even if it is used in an inner loop. The `if` already exists... it's just the else that's added.

networkx/algorithms/matching.py
dschult: I think I'm getting confused between double ` and single `.  
Aren't we supposed to use `G` instead of ``G``?
What's the difference?

networkx/algorithms/approximation/traveling_salesman.py
dschult: OK... I missed that both 0 and len(soln)-1 are values that this code (either version) does not choose. It's making more sense to me.

It is assuming the list is a cycle and we don't want to switch the first node which must be the same as the last node.  I'll add a comment to this effect.

dschult:
I guess I disagree with using text to designate which move to make. I think the user should supply a function -- not the text. We should supply a couple of nice functions and then they are free to make their own functions.

I'll try to implement that here unless there is a reason to prefer the text '1-1' or '1-0' interface.

rossbar:
> I guess I disagree with using text to designate which move to make. I think the user should supply a function -- not the text.

I didn't think of this on the first pass, but I agree that would be a better approach. It makes it much more extensible and removes the ugly case-matching code mapping strings to functions :+1:

networkx/algorithms/approximation/traveling_salesman.py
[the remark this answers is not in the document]
rossbar: Got it, thanks. I wasn't sure what the differences were so thanks for the explanation :)

networkx/algorithms/approximation/traveling_salesman.py
rossbar: Yeah I wasn't entirely sure on this either - my suggestion based on trying to replicate what the behavior of the original was without truly understanding *why* that was the desired behavior. A comment would definitely be helpful!

networkx/algorithms/matching.py
rossbar: In principle, parameters are supposed to be in single backticks (so this *should* be `` `G` ``, not ``` ``G`` ```) but in practice it's not a huge difference, so I try not to be *too* picky about it.

It's actually tricky in networkx's case because the parameters-in-single-backticks rule comes from the numpydoc standard, but the `default_role=obj` in the networkx docs configuration actually clashes with this suggestion and results in many more "broken links" when sphinx is run in nitpicky mode. The end result in the rendered html looks good though and I'm not sure the best way to tackle this. So in the mean time my feeling is not to worry about it too much.

dschult:
OK... I think this is pretty much done. I have two API questions left.
1) The `traveling_salesman_problem` function uses a user identified input function to improve the solution. Is it better to have users pass both a function and a `**kwargs`?  Or is it better make the user "curry" the function to have only 2 inputs -- explicitly putting in the `kwargs` before giving it to us?

2) Should we require the user to explicitly input an initial guess of the solution, usually calculated via `nx_app.greedy_tsp(G, weight, source)`? Or should there be a default way to handle it?

It looks ugly to always have this boilerplate code to create the initial solution. I added a shortcut word "greedy" which if input does the greedy algorithm for the user. But it's still ugly to have `nx_app.simulated_annealing_tsp(G, "greedy", weight="weight")` rather than 'nx_app.simulated_annealing_tsp(G)`.

Making the user construct the initial cycle could allow us to get rid of the `source` input as that is simply provided to `greedy_tsp`. But we have to find a good default way to pick it for the shortcut word "greedy".

networkx/algorithms/matching.py
dschult: We changed the default role before. I think it was math and we made it python object. We might have changed it once before that too.  It would be good to have that figured out, but I agree that we shouldn't probably worry about it too much. :}

rossbar:
Sounds good, I will try to take another pass today. Also I think there are a still a few comments left from the first round - you may have to click the "Load More..."` button in the middle of the comment listing to be able to see them. Perhaps you saw them already in which case feel to close them - I just didn't want to do it in case you hadn't seen them yet!

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,716 @@
...
+      A function that returns a cycle on all nodes and approximates
+      the solution to the traveling salesman problem on a complete
+      graph. The returned cycle is then used to find a corresponding
+      solution on `G`. `method` should be callable; take inputs
+      `G`, and `weight`; and return a list of nodes along the cycle.
+
+    Returns
+    -------
+    list
+        List of nodes in `G` along a path with a 3/2-approximation of the minimal
```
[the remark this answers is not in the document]
dschult: only for chistofides...   more general language would be better here.

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,716 @@
...
+    For more information and how algorithm is inspired see:
+    http://en.wikipedia.org/wiki/Simulated_annealing
+    """
+    if cycle is None:
+        # Construct an initial solution using a greedy algorithm.
+        cycle = greedy_tsp(G, weight=weight, source=source)
+        if G.number_of_nodes() == 2:
+            return cycle
+
+        cost = sum(G[u][v][weight] for u, v in zip(cycle, cycle[1:]))
```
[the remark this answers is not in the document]
dschult: I have gone back and forth on this -- mostly when I was trying to use e.g. `SA_tsp` in `traveling_salesman_problem` which required currying to provide the cycle. But I've re-come-around.  I think it is best to require the initial guess and I think it should not be called "cycle" so people don't confuse it with "cycle" from the other function. I'm thinking it should be called `initial_soln` or similar -- and really the name isn't as important if it isn't a kwarg anyway.

+1

dschult:
You guessed it!   I had not seen the comments in the middle -- my eyes glossed over the "load more..." between the other comments. I was thinking that it was like the breaks in code when github shows a diff.   :{  
Thanks for the comments and ideas!

rossbar (commented):
It seems to me that the major points have been addressed and we're soon entering opinion/bikeshedding territory :)

I have a slight preference for `move` and `init_cycle` *requiring* functions instead of functions or shortcut strings (e.g. "1-1" for the former and "greedy" for the later) as I think that leads to a clearer, less over-loaded API , but I don't think it should be a blocker. 

A few minor comments from another readthrough, but this looks good to me!

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,846 @@
...
+    # Remove selfloops if necessary
+    loop_nodes = nx.nodes_with_selfloops(G)
+    try:
+        node = next(loop_nodes)
+    except StopIteration:
+        pass
+    else:
+        G = G.copy()
+        G.remove_edge(node, node)
+        G.remove_edges_from((n, n) for n in loop_nodes)
```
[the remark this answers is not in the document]
rossbar: Ah, I think I see now: turns out `number_of_selfloops` calls `selfloop_edges` under the hood, so though a bit shorter in code, my suggestion is likely less efficient. +1 for performance!

rossbar (approved):
The API makes a lot of sense and I think the documentation does a nice job of pointing out recommended uses while the defaults are very convenient. Having looked at this a couple times, nothing stands out to me as a blocker and this will obviously be a very nice addition. Thanks @dschult!

dschult:
OK... I think this is ready to get merged!

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,849 @@
...
+    See Also
+    --------
+        swap_two_nodes
+    """
+    a, b = seed.sample(range(1, len(soln) - 1), k=2)
+    soln.insert(b, soln.pop(a))
+    return soln
+
+
+@not_implemented_for("directed")
```
boothby, line 122: From the looks of it, this isn't implemented for multigraphs (see call to `min_weight_matching` which is so decorated).  Should it be? We could homogenize to a non-multigraph by taking the minimum weight edge for each node pair?

networkx/algorithms/approximation/traveling_salesman.py
```diff
@@ -0,0 +1,849 @@
...
+    nodes = []
+    for u, v in circuit:
+        if v in nodes:
+            continue
+        if not nodes:
+            nodes.append(u)
+        nodes.append(v)
+    nodes.append(nodes[0])
+    return nodes
+
```
boothby, line 192: stopping here for now

networkx/algorithms/approximation/traveling_salesman.py
dschult: Even better might be to get `max_weight_matching` updated to work with multigraph.
But that's another PR.  I put the homogenization into the other checks of the graph.

networkx/algorithms/matching.py
```diff
@@ -92,13 +98,21 @@ def is_matching(G, matching):
...
         matching = matching_dict_to_set(matching)
+    for e in matching:
+        if len(e) == 1:  # self-loop
+            if next(iter(e)) not in G:
+                msg = f"Matching includes edge {e} not in G"
+                raise nx.NetworkXNotImplemented(msg)
+            continue
+        if not G.has_edge(*e):
+            msg = f"Matching includes edge {e} not in G"
+            raise nx.NetworkXNotImplemented(msg)
```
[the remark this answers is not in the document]
dschult: I changed this to the handling put in place by #4638 and merged to current `main`

dschult:
I'm not on what level you are asking this question. So I'll answer on many levels.

The purpose of this pull request (PR) in particular is to bring together various submissions of approximate solutions to the traveling wave solution, give them a unified interface and get them included into the NetworkX package. 

You may be coming to this PR from the "[projects page](https://github.com/networkx/networkx/blob/main/doc/developer/projects.rst)" which lists projects that might be of interest. The intent of that project is to extend this PR with further functions which create approximate solutions.  In particular it mentions some algorithms that are published but not implemented here.

To find out more about the traveling salesman problem more generally, take a look at wikipedia.

dschult:
I think this TSP PR should be merged before Monday, so let me know if you want to do something with it by Saturday night. If I don't hear from you I will merge it Sunday morning.

eskountis:
Hey Dan,
Yes we were planning to make the Pull Request tomorrow, but there is a
possibility that the tests won't be ready, almost everything is ready. One
more thing, we should make the pull request on networkx main branch, right?
Can we make the PR and before you merge it we add the additional commits
referred to the tests?

Στις Παρ, 14 Μαΐ 2021 στις 10:57 μ.μ., ο/η Dan Schult <
***@***.***> έγραψε:

dschult:
You should wait until this gets merged with the main branch and then make your PR to the main networkx branch.
I think that is the simplest way to proceed. Your current branch doesn't need to be made into a PR on this PR. It can just be a PR on the main networkx repo  (at least that's the way it should be... :}  if it isn't easy, then this branch will still exist and we can figure out how to get it working)

Issue #4623, opened 2021-02-16
Add gallery example of using mpl to annotate a graph drawn with nx.draw


jarrodmillman:
NetworkX can draw graphs that can be annotated using the matplotlib drawing tools because the result of `nx.draw` is a matplotlib figure on which you can use matplotlib to draw anything else you'd like on top of the networkx drawing of the graph.

This idea is motivated by this SO post:
- https://stackoverflow.com/questions/66221054/3-dimensional-graph-lightgraphs-graphplot-julia-or-networkx-python

Someone could start with the example from the SO post; but it would be nice to have an example with more annotations and with nicer looking output.  (I suspect someone with more matplotlib skill than me would be able to make something very nice).   It also might be worth explaining a bit more about potential use cases in the example (perhaps by using real data where this type of annotation provides useful information about the underlying system).  If we add a nice example to the gallery, we should also add a link to the example to the SO post.

sharwinbobde:
Id love to contribute to it with @SA-networkx

jarrodmillman:
Excellent!  We are working to improve the quality of the gallery examples.  So it would be a good idea to briefly discuss what might be a good example.

Do you already have some specific ideas about what this example should be?  If so, could you quickly describe your plan?  If not, maybe it would be best to first identify a useful graph drawing from the literature.  For example, I am not sure representing nodes by vertical bars is the best thing.  (It may be.)

I am not certain what to do, so this is a list of possible ideas:

- Maybe we could use an example where there are several numerical attributes for each node.  Then the drawing could attempt to display those multiple attributes at the same time.  For example, one attribute could be represented with shades of gray; one attribute could be represented by the node diameter, and another attribute could be represented with the height of a vertical bar or column.  Maybe there is a better way to represent the magnitudes.  Ideas are welcome.
- Maybe the nodes could be icons.   For example, it could represent a network with computers and people (or something).  Then some nodes could be displayed with a person icon and other nodes could be represented by a computer icon.
- Maybe subsets of nodes could be grouped.  Often this is done with color grouped node with the same color.  However, you could also draw boxes, circles, or more custom shapes around nodes to group them.  Maybe these shapes could also include text labels.

Here are some example networks that may provide you with some ideas:
- https://www.frontiersin.org/files/Articles/445334/fgene-10-00294-HTML-r1/image_m/fgene-10-00294-g001.jpg
[...]

PR #4631, opened 2021-02-22
Gallery example: composite plot

sharwinbobde:
The following code is intended to address issue number #4623 

Draw a composite plot on matplotlib showing the distribution of nodes' features and the network topology at the same time.

This can be useful when you have a dataset with feature information for different items and some links between different
items.


sharwinbobde:
image
Here is the resulting plot so that you don't have to run it to see the output :)

rossbar:
Note that if you place the example in one of the existing directories, e.g. `examples/drawing/plot_composite_feature_network.py`, it will be built automatically by CI and can be previewed via the build artifact link in CI summary (you may have to click "Show all checks" to see it).

sharwinbobde:
thnx @rossbar :)
moved it to the `drawing` directory.

MridulS:
@sharwinbobde is this ready to be reviewed? It's still in "Draft" mode :)

Issue #4261, opened 2020-10-15
Improve gallery examples


jarrodmillman:
### Restructure

#4422 #4427 #4428 #4429 

### Existing examples
- [x] examples/drawing/plot_house_with_colors.py #4263 
- [x] examples/drawing/plot_four_grids.py #4264 
- [x] examples/drawing/plot_degree_rank.py #4259 (see also: #4265)
- [x] examples/drawing/plot_labels_and_colors.py #4266 
- [x] examples/drawing/plot_giant_component.py #4267 
- [x] examples/drawing/plot_knuth_miles.py #4251 
- [x] examples/drawing/plot_chess_masters.py #4252 
- [x] examples/basic/plot_read_write.py #4260 
- [ ] examples/graph/plot_napoleon_russian_campaign.py
- [x] examples/algorithms/plot_decomposition.py #4278 
- [ ] examples/algorithms/plot_krackhardt_centrality.py
- [x] examples/graph/plot_words.py #4409 
- [x] examples/applications/plot_circuits.py #4408 
- [x] examples/algorithms/plot_rcm.py #4411 

### New examples
- [x] gene-gene network #4269

### Geospatial examples
 - [x] pysal, geopandas, momepy, OSMnx, etc. #4366, #4381, #4383, #4407 

### External libraries examples

We should move the javascript, jit, and pygraphviz libraries here.  I think it makes sense to keep geospatial separate as it looks like it will have a number of specialized examples.

- [x] igraph #4404
- [ ] graphtools (maybe via igraph?)
- [ ] scikit-learn (maybe a spectral clustering example?)
- [ ] RAPIDS cuGraph
    - https://medium.com/rapids-ai/rapids-cugraph-networkx-compatibility-d119e417557c
    - https://github.com/ptheywood/cuda-cmake-github-actions
    - FLAMEGPU/FLAMEGPU2_dev#290
- [ ] graph plotting libraries ?

Are there other examples we should include?  I know a lot of libraries build on networkx.  It would be nice to have list of ones that provide functions to convert to/from NetworkX.

eikNamo:
We would like to contribute a scikit-learn example to the gallery @SA-networkx
So far we are thinking of coming up with some clustering example or feature extraction from a graph.

YaoM1886:
Hi, I and @eikNamo are working together on this issue. We have got this initial result by plotting spectral clustering in scikit-learn with networkx graph object. Hope this helps!
[...]

PR #4643, opened 2021-02-27
Example: spectral clustering based graph

sharwinbobde:
**[Updated]**
This example is meant to address Issue #4261
More specifically, using scikit-learn to cluster data using Spectral Clustering. Then using the adjacency matrix from the same to generate a graph for the data points to understand the clustering.

image

Follow the below thread and the issue to view the iterations of the demo.


sharwinbobde:
@dschult we changed the colour schemes for plots of Step 1 and 2 to indicate that because clustering is an unsupervised method, the grouping does not necessarily indicate class membership.

dschult:
This is much clearer--thank you.

But it still is not clear how you are assigning the edges in a network.  It looks like you start with a dataset with three variables for each observation ==> two are continuous variables and the third is a binary classification. Then you use an AI method (which one?) to classify the observations into two groups. The clustering output shows these groups and they are close but not the same as the original groups. Is this correct?

Then you show the network -- but there is no explanation for what the nodes represent, how the edges are chosen, etc.

sharwinbobde:
I'll write a detailed flow here first thing in the morning :)

sharwinbobde:
So here is what we decided to do
## Task overview
- We do everything on a toy dataset
- We know data comes from 2 distributions only to compare results.
- The data is noisy (to show actual problems with clustering).
- We will use Spectral Clustering to (I) understand the data and (ii) cluster it
- We can also build a network using the affinity matrix which results in the following:
     - nodes are data points from the feature space
     - edges' weight is the value in the affinity matrix
     - nodes have cluster membership information
     
#### Usage
- This can be used to understand unlabelled data in a network form. For example we can represent a 10-dimensional dataset using networks while preserving local information and making it obvious.
- The resulting graph is a condensed form of the feature space information
- We can indirectly use graph algorithms on feature space information
- Edge weights do not have small noisy information from feature measurements, only the neighbourhood relation information.
 

## Step 1: Generating sample dataset
- Generate 2-class data which is not linearly separable 
```
n_samples = 100
X, y = datasets.make_moons(n_samples=n_samples)
```
- Add gaussian noise to make the problem difficult: `X = X + np.random.normal(1, 0.1, X.shape)`
- The problem can look much cleaner with really clean clustering if we reduce the noise, but the readers will need to know that clustering has its limitations and not to use it blindly.

## Step 2: Clustering
- Lets conveniently select **2 as the number of clusters**. (this is just an example/demonstration :) )
- We will use *nearest-neighbour* based affinity matrix for [SpectralClustering](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html). This is what the method does under the hood:
     - Given only the feature space information, find 10 nearest neighbours for each data point using euclidean distance.
     - create the affinity matrix. Here is the code snippet from the scikit-learn source code. 
     ```
     connectivity = kneighbors_graph(X, n_neighbors=self.n_neighbors,
                                            include_self=True,
                                            n_jobs=self.n_jobs)
     self.affinity_matrix_ = 0.5 * (connectivity + connectivity.T)
     ```

     - this is what a sample affinity sparse matrix looks like:  
```
    (0, 76) 0.5
    (0, 70) 0.5
    (0, 61) 0.5
    (0, 36) 0.5
    (0, 66) 1.0
    (0, 92) 1.0
    (0, 15) 1.0
    ...
    (98, 83) 1.0
    (98, 99) 1.0
    (98, 98) 1.0
    (99, 25) 1.5
    (99, 20) 1.0
    (99, 48) 1.0
```
- label the data points according to the cluster membership
     
   
## Step 3: constructing the network
- get the affinity matrix
- get the labels
- remove self-loops
```
    clusters = sc.fit(X)
    cluster_affinity_matrix = clusters.affinity_matrix_
    print(cluster_affinity_matrix)

    pred_labels = clusters.labels_.astype(int)
    G = nx.from_scipy_sparse_matrix(cluster_affinity_matrix)
    # remove self edges
    G.remove_edges_from(nx.selfloop_edges(G))
```

*Then just plot the different steps*

sharwinbobde:
_Oops! posted the messages on the issue thread instead of this PR before.
Deleted the content on the Issue to keep things clean._

@dschult hope this is clearer :)
If there are any doubts we will keep updating the above explanation.
We do believe from our side that this is a strong direction for the Spectral Clustering example because it covers many possibilities of usage while still showing the limitations.

dschult:
Much more complete information here...  Thank you!

The original description had the sentence: "Then using the data from the spectral clustering to generate a graph for the data points."   If I understand correctly, the "graph" is constructed using the affinity matrix by computing k-nearest-neighbors.  But maybe here you mean "graph" (as in draw) a scatter plot of the feature space showing the clusters. In general, stay away from the word "graph" to mean a scatter plot when you've got graph theory in the mix. Graph Theory graphs can be called "networks", while graphs of datasets can be called scatter plots (or whatever other plot they are).

What is the right visualization of the network that will make this algorithm more clear?  I think you have formed the network using distance in the feature space between data points. When you plot it you use a spring_layout for the nodes (which are the data points).  Maybe it would be better to use the feature values as the coordinates of the nodes.  Then you will show the essentially the scatter plot with edges between neighbors. 

Another approach would be to plot the nodes using laplacian eigenvector values for two main eigenvalues. That gives a spectral layout which may (or may not depending on the meaning of "spectral clustering") show how the clustering is being determined. 

Perhaps both pictures of the network would be helpful in getting the reader to understand 1) where the matrix/network is coming from, and 2) where the clustering is coming from once the network is formed.

What do you think?  Or maybe you can explain why you chose to show the axes and positions you chose -- what is your pedagogical goal and how are you working to attain it?

rossbar (commented):
> representing items in a network and creating avenues to using graph algorithms on feature-space information.

This example could be made much more compelling by actually applying a graph algorithm on feature-space information. The one high-level comment I have is that it's not clear why networkx is even necessary, since the data generation and analysis (clustering) are both done w/ scikit-learn. It's nice to be able to explicitly see the edges after converting to a networkx graph, and the demonstration of the layout is also nice, but in its current form this seems more like a pure scikit-learn example than a scikit-learn + networkx example.

That's not to say it isn't valuable though! I'm not an expert so it's quite possible I'm missing interesting details - in that case adding some narrative text in the form of rst comments might help.

sharwinbobde:
I see that some choices we made can be a bit confusing and deliver the wrong message. We will try to address these with a modified example.
image

Here you can clearly see that.
- Step 1. Some data exists. we know there are 3 clusters.
- Step 2. We do Spectral clustering (based on the nearest neighbours method) with the number of clusters as 3.
- Step 3. We plot the graph obtained from the affinity matrix.

Modification considerations:
- clustering is more apparent with consistent colour schemes.
- it is more apparent that edges have neighbour information
- what is a scatterplot and what is a simple network is clearer.
- use case is comparatively more apparent.


This example can be used to observe cluster quality for high-dimensional data using a 2d network plot.
@rossbar I agree with you that there should be a Step 4. where we use some sort of graph technique.

Clarifications for comments:
- @dschult This modification makes the distinction between scatterplots and network plots clearer.
- @dschult @rossbar the affinity matrix will bot store real distance information at the clustering phase.
     - edges exist only if two nodes are neighbours
     - the edge weight is either **0.5** (if one node  is neighbour of the other) or **1.0** (if both nodes are neighbours of each other)
     - Thus, the network representation has information that depends on feature-space distances but is NOT equal to the euclidean distance. _The graph is a condensed form of information_.
- @rossbar In Step 3. We represent neighbours for different cluster members and thus draw the condensed high dimensional feature information using networkx, that is separated properly.
- @rossbar for step 4. we might compute some heuristics to show clustering quallity.

sharwinbobde:
I'll update the above "Demonstration Explanation"

sharwinbobde:
image

One more Iteration!
In Step 4 we record the cut-size for different cluster-pair and plot the confusion matrix. The above figure shows that the cut size of clusters 2 and 1 is 0 so they are cleanly separated, whereas clusters 2 and 0 have a lot of similar neighbours. This cut-size is the heuristic Data Scientist can use to actually see how clean the clustering algorithm's output is and quantifiably measure the how good the clustering is.

rossbar:
One thing to keep in mind is that the example should highlight something about graph/network analysis. There is a fine line to walk between making an example compelling and making it understandable. For me anyways, making the example *more* complex is heading in the wrong direction :). That's not to say that it isn't valuable - in fact, as the scope of the example here continues to increase, it may be a better fit for the [networkx-notebooks](https://github.com/networkx/notebooks) repo, which is a more natural fit for larger examples and examples that would benefit from some narrative/explanatory text.

Also - there's no need to continue attaching images to the GH comments - all examples are run by the CI and can be previewed by clicking on the "ci/circleci: build artifact" tab in the CI check summary. This is actually better for review because it also guarantees that the image is being generated as expected (which is [not currently the case](https://1592-890377-gh.circle-artifacts.com/0/doc/build/html/auto_examples/index.html#external-libraries)).

lobpcg:
The PR appears abandoned, and the code no longer runs on the current release.

examples/external/spectral_clustering_graph.py
```diff
@@ -0,0 +1,121 @@
...
+import networkx as nx
+import seaborn as sns
+
+np.random.seed(0)
+
+
+def spectral_clustering_to_graph(X, n_clusters: int, sc: SpectralClustering):
+    # get affinity matrix from spectral clustering
+    clusters = sc.fit(X)
+    cluster_affinity_matrix = clusters.affinity_matrix_
```
lobpcg, line 24: cluster_affinity_matrix = clusters.affinity_matrix_.H

examples/external/spectral_clustering_graph.py
```diff
@@ -0,0 +1,121 @@
...
+
+    pred_labels = clusters.labels_.astype(int)
+    G = nx.from_scipy_sparse_matrix(cluster_affinity_matrix)
```
lobpcg, line 27: G = nx.from_scipy_sparse_array(cluster_affinity_matrix)

rossbar:
Thanks for the ping @lobpcg - I still have the sense that this is a better fit for a scikit-learn gallery rather than networkx. Since it's been 18 months since last activity and is still in draft, I'm inclined to close. I'll do so soon if there are no objections!

lobpcg:
@rossbar Yes, I agree that this PR could be closed as stalled. I like the idea, but the proposed implementation is still far from being acceptable unfortunately.

lobpcg:
@rossbar I have made https://github.com/networkx/networkx/pull/7040 replacing this PR - could you please have a look?

PR #4647, opened 2021-03-03
Adding automatic reformatting from initial commit

MaxBromberg:
Simply committed hierarchy coordinates as a new project in _algorithms_, though without testing as of yet. New to such projects, I'd appreciate some guidance regarding testing and whatever else's needed. An issue with graph entropy remains to be resolved.

[Paper the algorithms are based on][2], [Earlier git with readme][1], [initial discussion][3]

[1]: https://github.com/MaxBromberg/Hierarchy_Coordinates
[2]: https://arxiv.org/abs/1303.2503
[3]: https://groups.google.com/g/networkx-discuss/c/Kltl3aQAWzU/m/SyQhne4FBgAJ

### DIFF
--- a/networkx/algorithms/hierarchy_coordinates/hierarchy_coordinates.py
+++ b/networkx/algorithms/hierarchy_coordinates/hierarchy_coordinates.py
@@ -1,0 +1,892 @@
+import numpy as np
+import networkx as nx
+import copy
+from utils import distribute, matrix_normalize
+
+"""
+Implementation of the hierarchy coordinates via networkX from 
+Hierarchy in Complex Networks: The Possible and the Actual [B Corominas-Murtra - 2013]  [*] - Supporting Information
+
+Though implemented for unweighted networkX graphs, in the context of its original application, 
+these are applied to weighted graphs by averaging over the unweighted graphs resulting from 
+applying thresholds to normalized weighted graphs. 
+"""
+
+
+# Hierarchy Coordinate Functions
+def weakly_connected_component_subgraphs(G, copy=True):
+    """Generate weakly connected components as subgraphs. Re-imported to ensure later NetworkX compatibility
+
+    Parameters
+    ----------
+    G : NetworkX Graph
+       A directed graph.
+
+    copy : bool
+        If copy is True, graph, node, and edge attributes are copied to the
+        subgraphs.
+
+    Notes
+    -----
+    Simply brought in from earlier version of NetworkX to ensure compatibility with later versions.
+    Not sure why it was dropped...
+    """
+    for comp in nx.weakly_connected_components(G):
+        if copy:
+            yield G.subgraph(comp).copy()
+        else:
+            yield G.subgraph(comp)
+
+
+def node_weighted_condense(A, num_thresholds=8, threshold_distribution=None):
+    """Returns a series of node_weighted condensed graphs (DAGs) [1]_ and their original nx_graphs.
+
+    Parameters
+    ----------
+    A: numpy array
+        Adjacency matrix, as square 2d numpy array
+    num_thresholds: int, default: 8
+        Number of thresholds and resultant sets of node-weighted Directed Acyclic Graphs
+    threshold_distribution: float, optional
+        If true or float, distributes the thresholds exponentially, with an exponent equal to the float input.
+
+    Returns
+    -------
+    largest_condensed_graphs: list of networkX Graphs
+        list of node weighted condensed networkx graphs reduced from unweighted digraphs determined by thresholds. (See note)
+    nx_graphs: list of networkX Graphs
+        list of unweighted graphs produced from applying thresholds to the original weighted network
+
+    Examples
+    --------
+    Graphing the resultant network is recommended, as otherwise this is difficult to visualize...
+
+    a = np.array([
+        [0, 0.2, 0, 0, 0],
+        [0, 0, 0, 0.7, 0],
+        [0, 0.4, 0, 0, 0],
+        [0, 0, 0.1, 0, 1.0],
+        [0, 0, 0, 0, 0],
+    ])
+    condensed_networks, base_binary_networks = hc.node_weighted_condense(a)
+    for network in condensed_networks:
+        print(nx.to_numpy_array(network))
+
+    Notes
+    ------
+    TODO: As multiple independent graphs may form from applying threshold cutoffs to a weighted graph,
+    only the largest is considered. This might be worth considering in re-evaluating the meaning of
+    weighted network hierarchy coordinate evaluations. (See pages 7, 8 of [1]_, supplementary material)
+
+    An threshold_distribution of None results in a linear distribution, otherwise
+     the exponential distribution is sampled from exp(x) \in (0, 1)
+
+    .. [1] "On the origins of hierarchy in complex networks."
+     Corominas-Murtra, Bernat, Joaquín Goñi, Ricard V. Solé, and Carlos Rodríguez-Caso,
+     Proceedings of the National Academy of Sciences 110, no. 33 (2013)
+    """
+
+    # Establishing Thresholds
+    if num_thresholds == 1 or np.isclose(np.max(A) - np.min(A), 0, 1e-2):
+        nx_graphs = [nx.from_numpy_matrix(A, create_using=nx.DiGraph)]
+    else:
+        if threshold_distribution is None:
+            try:
+                thresholds = list(
+                    np.round(
+                        np.arange(
+                            np.min(A),
+                            np.max(A),
+                            (np.max(A - np.min(A))) / num_thresholds,
+                        ),
+                        4,
+                    )
+                )  # linear distribution
+            except:
+                thresholds = [np.max(A)] * num_thresholds
+        else:
+            thresholds = distribute(
+                dist=threshold_distribution,
+                end_value_range=(np.min(A), np.max(A)),
+                n=num_thresholds,
+            )
+        # Converting to binary nx_graphs according to thresholds:
+        nx_graphs = [
+            nx.from_numpy_matrix(np.where(A > threshold, 1, 0), create_using=nx.DiGraph)
+            for threshold in thresholds
+        ]
+    nx_graphs = [
+        graph for graph in nx_graphs if not nx.is_empty(graph)
+    ]  # eliminates empty graphs
+    condensed_graphs = [
+        nx.condensation(nx_graphs[index]) for index in range(len(nx_graphs))
+    ]
+    largest_condensed_graphs = []
+    for condensed_graph in condensed_graphs:
+        largest_condensed_graphs.append(
+            nx.convert_node_labels_to_integers(
+                max(
+                    weakly_connected_component_subgraphs(condensed_graph, copy=True),
+                    key=len,
+                )
+            )
+        )
+        # networkx.weakly_connected_component_subgraphs comes from networkx 1.10 documentation, and has sense been discontinued.
+        # For ease of access and future networkx compatibility, it was copied directly to this file before the class declaration.
+        members = nx.get_node_attributes(largest_condensed_graphs[-1], "members")
+        node_weights = [len(w) for w in members.values()]
+        for node_index in range(len(node_weights)):
+            largest_condensed_graphs[-1].nodes[node_index]["weight"] = node_weights[
+                node_index
+            ]
+
+    return largest_condensed_graphs, nx_graphs
+
+
+def weight_nodes_by_condensation(condensed_graph):
+    """Weights nodes according to the integer number of other nodes they condensed. Proposed in _[1]
+    e.g. if a cycle contained 3 nodes (and became one in condensation)
+    the resulting node of the condensed graph would then gain weight = 3.
+
+    Parameters
+    ----------
+    condensed_graph: NetworkX Graph
+        result of a networkx.condensation call, such that the 'members'
+        attribute is populated with constituent cyclical nodes
+
+    Return
+    ------
+    condensed_graph: NetworkX Graph
+        node weighted condensed graph
+
+    Examples
+    --------
+    Visualization also recommended here (with node size prop. weighting)
+    b = np.array([
+        [0, 0.2, 0, 0, 0],
+        [0, 0, 0, 0.7, 0],
+        [0, 0.4, 0, 0, 0],
+        [0, 0, 0.1, 0, 1.0],
+        [0, 0, 0, 0, 0],
+    ])
+
+    num_thresholds = 2
+    condensed_networks, _ = hc.node_weighted_condense(b, num_thresholds=num_thresholds)
+    for network_index in range(num_thresholds):
+        print(f"Network {network_index}:")
+        for node_index in range(len(condensed_networks[network_index].nodes)):
+            print(f"Node {node_index}, new weight:", condensed_networks[network_index].nodes[node_index]["weight"])
+        print()
+
+    Note:
+    ------
+    TODO: Might wish to eliminate return, or enable copying?
+    .. [1] "On the origins of hierarchy in complex networks."
+     Corominas-Murtra, Bernat, Joaquín Goñi, Ricard V. Solé, and Carlos Rodríguez-Caso,
+     Proceedings of the National Academy of Sciences 110, no. 33 (2013)
+    """
+    node_weights = [
+        len(w) for w in nx.get_node_attributes(condensed_graph, "members").values()
+    ]
+    for node_index in range(len(node_weights)):
+        condensed_graph.nodes[node_index]["weight"] = node_weights[node_index]
+    return (
+        condensed_graph  # Not be necessary, as the graph itself is updated (not copied)
+    )
+
+
+def max_min_layers(G, max_layer=True):
+    """
+    Returns the maximal (k_in = 0, highest in hierarchy) layer (those nodes with in degree = 0) or the minimal layer (k_out = 0)
+
+    Parameters
+    ----------
+    G: NetworkX Graph
+        A directed graph.
+    max_layer: bool, default: True
+        if True, returns maximal layer (k_in = 0), else returns nodes for which k_out = 0, minimal layer
+
+    Return
+    ------
+    min/max_layer: list
+        list of node indices as ints
+
+    Examples:
+    a = np.array([
+        [0, 0, 1, 1, 0, 0, 0],
+        [0, 0, 1, 1, 0, 0, 0],
+        [0, 0, 0, 0, 1, 1, 0],
+        [0, 0, 0, 0, 0, 1, 1],
+        [0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0],
+    ])
+
+    G = nx.from_numpy_matrix(a, create_using=nx.DiGraph)
+    print(hc.max_min_layers(G))
+    print(hc.max_min_layers(G, max_layer=False))
+
+    Notes:
+    -------
+    TODO: Should be two functions?
+    """
+    if max_layer:
+        return [node for node in G.nodes() if G.in_degree(node) == 0]
+    else:
+        return [node for node in G.nodes() if G.out_degree(node) == 0]
+
+
+def leaf_removal(G, forward=True):
+    """Returns a pruned network, with either maximal (k_in=0)
+    or minimal (k_out = 0) nodes removed upon call.
+
+    Parameters
+    -----------
+    G: NetworkX Graph
+        A directed graph.
+
+    forward: bool, default: True
+        if True, prunes from k_in=0 nodes
+
+    Return
+    -------
+    NetworkX Graph:
+        copy of the original graph G without either maximal or minimal nodes
+
+    Examples:
+    ---------
+    a = np.array([
+        [0, 0, 1, 1, 0, 0, 0],
+        [0, 0, 1, 1, 0, 0, 0],
+        [0, 0, 0, 0, 1, 1, 0],
+        [0, 0, 0, 0, 0, 1, 1],
+        [0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0],
+    ])
+
+    G = nx.from_numpy_matrix(a, create_using=nx.DiGraph)
+    print(nx.to_numpy_array(hc.leaf_removal(G)))
+    print(nx.to_numpy_array(hc.leaf_removal(G, forward=False)))
+
+    Raises
+    ------
+    TODO: NonDAG: will loop infinitely for cyclic input
+
+    Notes
+    ------
+    Crafted as a component of the leaf-removal algorithm given in [1]_;
+
+    .. [1] "On the origins of hierarchy in complex networks."
+     Corominas-Murtra, Bernat, Joaquín Goñi, Ricard V. Solé, and Carlos Rodríguez-Caso,
+     Proceedings of the National Academy of Sciences 110, no. 33 (2013)
+
+    """
+    layer = max_min_layers(G, max_layer=forward)
+    peeled_graph = copy.deepcopy(G)
+    for node in layer:
+        peeled_graph.remove_node(node)
+    return peeled_graph
+
+
+def recursive_leaf_removal(G, from_top=True, keep_linkless_layer=False):
+    """Prunes nodes from top (maximal) or bottom (minimal) recursively. Original DAG is given as first element
+    Parameters
+    ----------
+    G: NetworkX graph
+        A directed graph.
+    from_top: bool, default: True
+        Determines whether nodes are removed from the top or bottom of the hierarchy
+    keep_linkless_layer: bool, default: False
+        if True, empty, single node and all zero layers are preserved
+
+    Return
+    -------
+    list of NetworkX Graphs:
+        Starting with the initial DAG and with successively pruned hierarchy layers
+
+    Examples
+    ---------
+    a = np.array([
+        [0, 0, 1, 1, 0, 0, 0],
+        [0, 0, 1, 1, 0, 0, 0],
+        [0, 0, 0, 0, 1, 1, 0],
+        [0, 0, 0, 0, 0, 1, 1],
+        [0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0],
+    ])
+    G = nx.from_numpy_matrix(a, create_using=nx.DiGraph)
+    pruned_to_ground = hc.recursive_leaf_removal(G, from_top=True)
+    pruned_from_ground = hc.recursive_leaf_removal(G, from_top=False)
+    print("Pruned from top:")
+    for pruned_tree in pruned_to_ground:
+        print(nx.to_numpy_array(pruned_tree))
+    print("Pruned from bottom:")
+    for pruned_tree in pruned_from_ground:
+        print(nx.to_numpy_array(pruned_tree))
+
+    """
+    dissected_graphs = [copy.deepcopy(G)]
+    while len(dissected_graphs[-1].nodes()) > 1:
+        dissected_graphs.append(leaf_removal(dissected_graphs[-1], forward=from_top))
+    if not keep_linkless_layer:
+        while (
+            nx.is_empty(dissected_graphs[-1]) and len(dissected_graphs) > 1
+        ):  # catches empty graphs, which are eliminated in node condense
+            dissected_graphs = dissected_graphs[
+                :-1
+            ]  # removes empty or single node layer
+    return dissected_graphs
+
+
+def orderability(
+    G, condensed_nx_graph=None, num_thresholds=8, threshold_distribution=None
+):
+    """Evaluates orderability, or number of nodes which were not condensed
+     over total number of nodes in original uncondensed graph. Evaluated as in [1]_
+
+    Parameters
+    ----------
+    G: NetworkX Graph
+        A directed graph.
+    condensed_nx_graph: optional
+        Directed acyclic networkX graph if already evaluated to save computational time
+    num_thresholds: int, optional
+        only applicable as node_weighted_condense parameter if G is weighted
+    threshold_distribution: float, optional
+        only applicable as node_weighted_condense parameter if G is weighted
+
+    Return
+    -------
+    float: orderability
+        Hierarchy coordinate
+
+
+    Examples
+    ---------
+    a = np.array([
+        [0, 0.2, 0, 0, 0],
+        [0, 0, 0, 0.7, 0],
+        [0, 0.4, 0, 0, 0],
+        [0, 0, 0.1, 0, 1.0],
+        [0, 0, 0, 0, 0],
+    ])
+
+    G = nx.from_numpy_matrix(a, create_using=nx.DiGraph)
+    print(hc.orderability(G))
+
+
+    Notes
+    ______
+    TODO: Not sure what's proper re: optional num_thresholds and threshold_distribution parameter
+    .. [1] "On the origins of hierarchy in complex networks."
+     Corominas-Murtra, Bernat, Joaquín Goñi, Ricard V. Solé, and Carlos Rodríguez-Caso,
+     Proceedings of the National Academy of Sciences 110, no. 33 (2013)
+    """
+
+    if not np.array_equal(
+        np.unique(nx.to_numpy_array(G)), [0, 1]
+    ):  # unweighted (non-binary) check
+        o = 0
+        condensed_graphs, original_graphs = node_weighted_condense(
+            nx.to_numpy_array(G),  # creates binary graphs
+            num_thresholds=num_thresholds,
+            threshold_distribution=threshold_distribution,
+        )
+
+        for index in range(len(condensed_graphs)):
+            o += orderability(original_graphs[index], condensed_graphs[index])
+        return o / len(condensed_graphs)
+
+    if condensed_nx_graph is None:
+        condensed_nx_graph = weight_nodes_by_condensation(nx.condensation(G))
+    non_cyclic_nodes = [
+        node[0]
+        for node in nx.get_node_attributes(condensed_nx_graph, "weight").items()
+        if node[1] == 1
+    ]
+    total_acyclic_node_weight = sum(
+        [
+            weight
+            for weight in nx.get_node_attributes(condensed_nx_graph, "weight").values()
+        ]
+    )
+    return len(non_cyclic_nodes) / total_acyclic_node_weight
+
+
+def feedforwardness_iteration(G):
+    """Performs a single iteration of the feedforward algorithm described in [1]_
+
+    Parameters
+    ----------
+    G: NetworkX Graph
+        a directed, acyclic graph [1]_
+
+    Return
+    --------
+    g: float
+        sum of feedforwardness for a single Directed Acyclic Graph
+    num_paths: int
+        sum total paths considered
+
+    Examples
+    ---------
+    a = np.array([
+        [0, 0.2, 0, 0, 0],
+        [0, 0, 0, 0.7, 0],
+        [0, 0.4, 0, 0, 0],
+        [0, 0, 0.1, 0, 1.0],
+        [0, 0, 0, 0, 0],
+    ])
+    condensed_networks, base_binary_networks = hc.node_weighted_condense(a)
+    for network in condensed_networks:
+        print(nx.to_numpy_array(network))
+        g, paths = hc.feedforwardness_iteration(network)
+        print('g: {0}, # paths: {1}'.format(g, paths))
+
+    Notes
+    ______
+    g(G) = sum of F over all paths from maximal to minimal nodes
+
+    .. [1] "On the origins of hierarchy in complex networks."
+     Corominas-Murtra, Bernat, Joaquín Goñi, Ricard V. Solé, and Carlos Rodríguez-Caso,
+     Proceedings of the National Academy of Sciences 110, no. 33 (2013)
+
+    """
+    max_layer = max_min_layers(G, max_layer=True)
+    min_layer = max_min_layers(G, max_layer=False)
+    weights = nx.get_node_attributes(G, "weight")
+    g = 0
+    num_paths = 0
+    for max_node in max_layer:
+        for min_node in min_layer:
+            for path in nx.all_simple_paths(G, source=max_node, target=min_node):
+                g += len(path) / sum(
+                    [weights[node] for node in path]
+                )  # where each path calculation is F(path)
+                num_paths += 1
+    return g, num_paths
+
+
+def feedforwardness(DAG):
+    """
+    Parameters
+    ----------
+    DAG: NetworkX Graph
+        A Directed, Acyclic Graph.
+
+    Return
+    -------
+    float:
+        feedforwardness, hierarchy coordinate
+
+    Examples
+    ---------
+    a = np.array([
+        [0, 0.2, 0, 0, 0],
+        [0.4, 0, 0, 0.7, 0],
+        [0, 0.4, 0, 0, 0],
+        [0, 0, 0.1, 0, 1.0],
+        [0, 0, 0, 0, 0],
+    ])
+    condensed_networks, base_binary_networks = hc.node_weighted_condense(a)
+    for network in condensed_networks:
+        print("\nDAG: \n", nx.to_numpy_array(network))
+        print("Feedforwardness: {0}".format(hc.feedforwardness(network)))
+
+    Notes
+    ______
+    feedforwardness is calculated by pruning layers of the original graph and averaging
+    over subsequent feedforwardness calculations.
+    """
+    # Must be fed the set of graphs with nodes lower on the hierarchy eliminated first
+    successively_peeled_nx_graphs = recursive_leaf_removal(DAG, from_top=False)
+    if (
+        len(successively_peeled_nx_graphs) == 1
+        and len(successively_peeled_nx_graphs[0].nodes()) == 1
+    ):
+        return 0
+
+    f = 0
+    total_num_paths = 0
+    for nx_graph in successively_peeled_nx_graphs:
+        g, paths = feedforwardness_iteration(nx_graph)
+        f += g
+        total_num_paths += paths
+    return f / total_num_paths
+
+
+def graph_entropy(DAG, forward_entropy=False):
+    """
+    Parameters
+    ----------
+    DAG: NetworkX Graph
+        Directed Acyclic Graph
+    forward_entropy: Bool, default: False
+        if True, calculates entropy from maximal nodes (k_in = 0) nodes to others.
+        Otherwise calculates using paths from the bottom (minimal nodes) of the network.
+
+    Return
+    -------
+    entropy: float
+        graph entropy, in [1]_, eq.14/16 ( H_b/f (G_C) )
+
+    Examples
+    ---------
+    b = np.array([
+        [0, 0, 1, 0, 0, 0, 1],
+        [0, 0, 1, 0, 0, 0, 0],
+        [0, 0, 0, 0, 1, 0, 0],
+        [0, 0, 1, 0, 1, 0, 0],
+        [0, 0, 0, 0, 0, 1, 0],
+        [0, 0, 0, 1, 0, 0, 1],
+        [0, 0, 0, 0, 0, 0, 0],
+    ])
+
+    condensed_network_layers = hc.recursive_leaf_removal(nx.condensation(nx.from_numpy_matrix(b, create_using=nx.DiGraph)))
+    fwd_graph_entropy = [round(hc.graph_entropy(net, forward_entropy=True), 3) for net in condensed_network_layers]
+    bkwd_graph_entropy = [round(hc.graph_entropy(net), 3) for net in condensed_network_layers]
+    print("fwd graph entropy (from top | bottom): {0} | {1}".format(fwd_graph_entropy, bkwd_graph_entropy))
+
+    Notes
+    ______
+    As described in the text of _[1]:
+    Graph Entropy, B(G)_ij, is measure of probability of crossing a node n_j when following a path from n_i.
+    Found in _[1], proposed and developed in _[2], _[3].
+
+    .. [1] "On the origins of hierarchy in complex networks."
+     Corominas-Murtra, Bernat, Joaquín Goñi, Ricard V. Solé, and Carlos Rodríguez-Caso,
+     Proceedings of the National Academy of Sciences 110, no. 33 (2013)
+
+    .. [2] "Topological reversibility and causality in feed-forward networks."
+      Corominas-Murtra, Bernat, Carlos Rodríguez-Caso, Joaquín Goñi, and Ricard Solé.
+      New Journal of Physics 12, no. 11 (2010): pg 113051
+
+    .. [3] "Measuring the hierarchy of feedforward networks."
+    Corominas-Murtra, Bernat, Carlos Rodríguez-Caso, Joaquin Goni, and Ricard Solé.
+    Chaos: An Interdisciplinary Journal of Nonlinear Science 21, no. 1 (2011): pg 016108.
+
+
+    """
+    dag = nx.convert_node_labels_to_integers(DAG)
+    L_GC = len(
+        recursive_leaf_removal(DAG)
+    )  # Could be passed in most contexts to reduce redundant computation
+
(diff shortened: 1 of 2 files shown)


dschult:
Wow -- this is a lot of code and docs and tests..   Thank you!
It might take a while to go through in its entirety but I've got a few comments to start.

- the weakly_connected_components_subgraphs function was removed because it is a simple extension of `weakly_connected_components`.  And we didn't feel that the subgraphs were usually needed, and when they were it isn't clear if they should be subgraph views or subgraph views that are copied to make networkx graphs.  So  we think it is better to let the user build that rather than provide all possible options in a convoluted api.  Something like this -- though I don't think you should actually make a list of all of them. Most likely you want to process one at a time to save memory:
```python
[G.subgraph(cc).copy() for cc in nx.weakly_connected_components(G)]
```
- It seems like the functions in `utils.py` should be available as numpy commands. Can we avoid this whole module?  Perhaps these functions could be "private" functions (start with an underscore and not included in `__all__`) within the main module, or simply replaced by their numpy equivalent if there is one.
- The doc_strings in the main `hierarchy_coordinates.py` module should start with an explanation of what hierarchy coordinates are and why someone might want to use them. Then for the functions themselves the doc_strings should start with a single line that explains what it does in very terse lamguage. This is the line the doc-builder sphinx puts near links in the docs to this function. It should describe the function enough that people know whether to click to find out more.  Then the details can be put in the rest of the doc_string. After the first line, there should be a blank line and then paragraph(s) describing what the function does in more detail, the algorithm used, etc. Then parameters, return values and examples, "See Also" links if reasonable and References.  The first line and paragraphs are all I noticed that could use changes.
- Your subpackage and its main module are named the same thing so namespace collisions are possible. But I'd prefer to make this a single module addition rather than a subpackage if possible (a single hierarchy.py file within algorithms and test file in `algorithms/tests`). But if you expect this to expand I'm open to reconsidering.
- You should connect your new files to the documentation by adding appropriate rst files to `/doc/reference/algorithms`. Similarly, adjust `algorithms/__init__.py` to include your functions in the main networkx namespace.  Take a look at what is in both these places to try to mimic it for your files. 

If you have any questions about any of this, reply here and we'll help out. 

I look forward to learning more about hierarchy coordinates. :}  Thanks!

MaxBromberg:
Glad hierarchy coordinates can feature in networkx! 

I restructured as you suggested, though in my latest push, it appears I've made a mistake in formatting examples, and all the checks have thus failed. Whoops; what did I do wrong? 
I'll be happy to fix it (an the below issues) with a bit of guidance; also, let me know if the documentation is a touch too terse. Hierarchy coordinates a a bit tricky to explain in concisely...

Notes:
 - I am awaiting a reply from Prof. Solé to resolve the present graph entropy degeneracy, though both definitions _work_, albeit one is far less computationally efficient, though potentially more as intended.
 - There are various TODO/WIP notes, most of which I just think are general best-practice stuff which I don't know the proper conventions for.
 - Weighted networks will transition from considering only the _largest_ (though that's generally the only) DAG to all resultant DAGs after threshold reduction, though I've yet to implement this. This is where the 
 `[G.subgraph(cc).copy() for cc in nx.weakly_connected_components(G)]`
came in, where the list is made because I'm looking for the max of this set. 
- Computational optimization wasn't exactly a priority, so I'm sure it could be improved all around in time...

dschult:
You are right -- there are many errors.  You can see the pytest description of those error by looking at the "details" link for one of the Github Action test sessions that has a red "x".  Scroll down to just below the list of all tests and it shows the errors and what the error is.  You can also get this locally on your system by installing pytest and running `pytest --doctest-modules filename.py`

It looks like 1 error is calling a private function (starts with an underscore) from the `nx.` namespace. You'll have to import those functions from the module's namespace specifically.  

Most of the others are spacing issues.  The examples to a very simple character by character comparison of what you put on the line after the python command and what that python command prints. So extra spaces (or not enough) will make it fail.

:)

MaxBromberg:
I see! I corrected the example formatting, but I still get this numpy import error. How do I resolve this? (Should I import functions/classes from numpy rather than numpy as np?)

MaxBromberg:
I'm afraid I don't quite understand what I should do regarding the numpy imports; I now import only the needed modules within each function, but the general imports in the tests raise an error. (Though I did not get this via `pytest --doctest-modules test_hierarchy_coordinates.py` or `pytest --doctest-modules hierarchy_coordinates.py`)

dschult:
The import within a test function would need the `np = pytest.importorskip("numpy")` syntax to know to skip that test.  But I think you will want to skip **all** tests in this module. Right?  Don't all those tests call functions that use numpy? If so, just do the import at the top of the test module using `pytest.importorskip`.  

I'm trying to say that it is not sufficient to check whether the test code itself uses numpy. If it calls a function that imports numpy you will want to skip that test.

dschult:
Other comments:
- You should put `algorithms/hierarchy_coordinates.py` into `networkx/conftest.py` in the `needs_numpy` list near line 200. Sorry I didn't remember to say that before.

- it's our "standard" to leave `np.log` etc using the `np` notation. That makes it clearer to know that the `log` function is coming from numpy and easier to find where numpy is used.  So, for your in-the-function imports, can you switch to `import numpy as np` and revert to the `np.` syntax?

- Would it be "good form" to allow input `A` to be an array-like object by starting the functions with something like `A = np.asarray(A)`?  I'm not sure what the usual way to treat this is. Maybe @rossbar knows.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1228 @@
...
+                n=num_thresholds,
+            )
+        # Converting to binary nx_graphs according to thresholds:
+        nx_graphs = [
+            nx.from_numpy_matrix(np.where(A > threshold, 1, 0), create_using=nx.DiGraph)
+            for threshold in thresholds
+        ]
+    # removes isolated nodes (0 in & out degree) from binary nodes. (not needed for consolidation)
+    for index in range(len(nx_graphs)):
+        nx_graphs[index].remove_nodes_from(list(nx.isolates(nx_graphs[index])))
```
dschult, line 149:
```suggestion
    for G in nx_graphs:
        G.remove_nodes_from(nx.isolates(G))
```
You should remove any idioms that use `for index in range(len(...)):`. It can/should almost always be done by iterating over the items themselves. Python makes this faster then the indexes and it is more readable.

Also, can you move the comments from the end of the line to the line previous?  I think that is preferred style for python and it will allow the `black` formatter to put many of these commands on one line too.

MaxBromberg: In-code comments were moved to new lines, which was what I understood to be intended. I fixed the superfluous indices; those that remain are iterating over multiple items. This is not to say that the code could not be optimized considerably; however I have not yet heard from [Professor Solé](http://complex.upf.es/~ricard/Main/RicardSole.html) regarding the implementation of the graph entropy, who I understand to be the intended correspondent for the original paper.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+        for non_extremal_node in non_extremal_nodes:
+            if forward_entropy:
+                entropy += P[layer_node][non_extremal_node] * np.log(
+                    dag.out_degree(layer_node)
+                )  # nan for 0 outdegree
+            else:
+                entropy += P[non_extremal_node][layer_node] * np.log(
+                    dag.in_degree(layer_node)
+                )
+    entropy /= len(boundary_layer)
```
dschult, line 780:
```substitution
    deg = dag.out_degree if forward_entropy else dag.in_degree
    e = sum(P[u][v] * np.log(deg(u) for u in boundary_layer for v in non_extremal_nodes)
    entropy = e / len(boundary_layer)
```

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+        for node in nx.get_node_attributes(condensed_nx_graph, "weight").items()
+        if node[1] == 1
+    ]
+    total_acyclic_node_weight = sum(
+        [
+            weight
+            for weight in nx.get_node_attributes(condensed_nx_graph, "weight").values()
+        ]
+    )
+    return len(non_cyclic_nodes) / total_acyclic_node_weight
```
dschult, line 530:
```suggestion
    weights = nx.get_node_attributes(condensed_nx_graph, "weight").values()
    total_acyclic_node_weight = sum(weights)
    number_non_cyclic_nodes = sum(1 for wt in weights if wt == 1)
    return number_non_cyclic_nodes / total_acyclic_node_weight
```

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+        o = 0
+        condensed_graphs, original_graphs = node_weighted_condense(
+            nx.to_numpy_array(G),
+            num_thresholds=num_thresholds,
+            threshold_distribution=threshold_distribution,
+        )
+
+        for index in range(len(condensed_graphs)):
+            o += orderability(original_graphs[index], condensed_graphs[index])
+        return o / len(condensed_graphs)
```
dschult, line 515:
```suggestion
        condensed_graphs, original_graphs = node_weighted_condense(
            nx.to_numpy_array(G),
            num_thresholds,
            threshold_distribution,
        )

        o = sum(orderability(OG, CG) for CG, OG in zip(condensed_graphs, original_graphs))
        return o / len(condensed_graphs)
```

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+    Notes:
+    -------
+
+    """
+    if not G.is_directed():
+        raise nx.NetworkXError("G must be a DiGraph for min/max layer evaluation")
+    if max_layer:
+        return [node for node in G.nodes() if G.in_degree(node) == 0]
+    else:
+        return [node for node in G.nodes() if G.out_degree(node) == 0]
```
dschult, line 303:
```suggestion
    if max_layer:
        return [node for node, deg in G.in_degree if deg == 0]
    else:
        return [node for node, deg in G.out_degree if deg == 0]
```

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+    if not nx.is_directed_acyclic_graph(condensed_graph):
+        raise nx.NetworkXError(
+            "G must be a directed acyclic graph for node weighted condensation"
+        )
+
+    node_weights = [
+        len(w) for w in nx.get_node_attributes(condensed_graph, "members").values()
+    ]
+    for node_index in range(len(node_weights)):
+        condensed_graph.nodes[node_index]["weight"] = node_weights[node_index]
```
dschult, line 247:
```suggestion
    for node, attrs in condensed_graph.nodes.data():
        attrs["weight"] = len(attrs["members"])
```

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+                    np.round(
+                        np.arange(
+                            np.min(A),
+                            np.max(A),
+                            (np.max(A - np.min(A))) / num_thresholds,
+                        ),
+                        4,
+                    )
+                )  # linear distribution
+            except:
```
dschult, line 135: What kind of exception might occur here? Can we provide the exception type, i.e. `except ValueError:`?

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+                    key=len,
+                )
+            )
+        )
+        members = nx.get_node_attributes(largest_condensed_graphs[-1], "members")
+        node_weights = [len(w) for w in members.values()]
+        for node_index in range(len(node_weights)):
+            largest_condensed_graphs[-1].nodes[node_index]["weight"] = node_weights[
+                node_index
+            ]
```
dschult, line 173:
```suggestion
        for node, attrs in largest_condensed_graphs[-1].nodes.data():
            attrs["weight"] = len(attrs["members"])
```

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+    for condensed_graph in condensed_graphs:
+        largest_condensed_graphs.append(
+            nx.convert_node_labels_to_integers(
+                max(
+                    [
+                        condensed_graph.subgraph(cc).copy()
+                        for cc in nx.weakly_connected_components(condensed_graph)
+                    ],
+                    key=len,
+                )
```
dschult, line 165: Can the `subgraph(cc).copy()` be pulled out of the `max`? I think `max(nx.weakly_...), key=len)` will give the largest `cc`.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+        threshold_distribution=threshold_distribution,
+    )
+    for index in range(len(condensed_graphs)):
+        t += treeness(condensed_graphs[index])
+        f += feedforwardness(condensed_graphs[index])
+        o += orderability(original_graphs[index], condensed_graphs[index])
+    t /= len(condensed_graphs)
+    f /= len(condensed_graphs)
+    o /= len(condensed_graphs)
+    return t, f, o
```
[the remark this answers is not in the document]
MaxBromberg: I left out the list comprehension here so that I could feed orderability the original graphs directly, as I think that's often faster than re-deriving the original (unweighted) graphs, seen in:
`o += orderability(original_graphs[index], condensed_graphs[index])`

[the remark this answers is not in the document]
MaxBromberg: I left out the list comprehension here so that I could feed orderability the original graphs directly, as I think that's often faster than re-deriving the original (unweighted) graphs, seen in:
`o += orderability(original_graphs[index], condensed_graphs[index])`

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1220 @@
...
+                    nx.all_simple_paths(dag, source=start_node, target=end_node)
+                )
+            else:
+                paths = list(
+                    nx.all_simple_paths(
+                        dag.reverse(copy=True), source=start_node, target=end_node
+                    )
+                )
+            for path in paths:
+                if forward_entropy:
```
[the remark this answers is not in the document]
MaxBromberg: yes! Good call, I'm hope this whole algorithm shouldn't be needed, as the analytic version simply called `graph_entropy` would be far faster, and is what is described in their appendix. However, as it disagrees with the apparent method seen in their infographic, I implemented the infographic version as well in hopes of easing explanation for author review. (To agree perfectly with the infographic, change np.log to np.log2, as I have done in latest commit)

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: That's better! I accounted for the transposed indices with `P = P if forward_entropy else P.T` and implemented your change.

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: This was rendered obsolete by the earlier `np.isclose(np.max(A) - np.min(A), 0, 1e-15):` check, so I removed it.

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: Just comparing the two, it appears `max(nx.weakly_...), key=len) ` yields only the set of nodes in the maxima. (Class `Sized`)
```
print(max([condensed_graph.subgraph(cc).copy() for cc in nx.weakly_connected_components(condensed_graph)], key=len))
print(max(nx.weakly_connected_components(condensed_graph), key=len))
```
So in short, it's likely possible, though I know not how.

MaxBromberg:
Thanks for the feedback; I implemented what I could, modified if needed. I abstained from list (sum) comprehension in `average_hierarchy_coordinates.py`, (see comment) but otherwise revised as suggested. 
...But apparently I did something that led to all the checks failing? I tested beforehand a lá `pytest --doctest-modules ...`, so not sure what I did :/

networkx/algorithms/hierarchy_coordinates.py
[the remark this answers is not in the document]
dschult: Thanks -- I missed that...   My suggestion should read:

```python
o = sum(orderability(OG, CG) for OG, CG in zip(original_graphs, condensed_graphs)) / len(condensed_graph)
```

networkx/algorithms/hierarchy_coordinates.py
dschult:
```python
max_cc = max(nx.weakly_connected_components(condensed_graph), key=len))
condensed_graph.subgraph(max_cc).copy()
```
I guess it could all be on one line too if you prefer.

dschult:
The test failures are definitely not something you did. It looks like a widespread issue -- probably due to a new version of one of the packages used on our testing system  maybe `decorator`.  We'll look into that. If your branch passes tests locally you are probably good. We'll just rerun the tests once the other problem is fixed.

I think the `np.log` vs `np.log2` issue is a miscommunication about their infographic.  But it's hard to know for sure. I'm not bothered by using either but it would be good to track it down as you say.  I'll take a closer look at all this stuff soon. :}

dschult:
I pushed the recent changes from `main` onto this branch. That includes the fix needed for tests to work correctly.
You will need to pull them to your local computer from your github repository using something like 
`git pull origin adding-hierarchy-coords`.  
Hopefully that is smooth.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1166 @@
...
+            "G must be a directed acyclic graph for graph entropy evaluation"
+        )
+
+    dag = nx.convert_node_labels_to_integers(DAG)
+    # Could be passed in most contexts to reduce redundant computation
+    L_GC = len(recursive_leaf_removal(DAG))
+
+    if forward_entropy:
+        # as w sklearn: B_prime = normalize(nx.to_numpy_array(dag), axis=1, norm='max')  # Row normalization
+        B_prime = _matrix_normalize(nx.to_numpy_array(dag), row_normalize=True)
```
dschult, line 740: It looks like the comments normalizes using the max, but this code normalizes the sum.
Also, you might consider replacing `row_normalize=True` with `axis=1` which matches numpy

networkx/algorithms/tests/test_hierarchy_coordinates.py
```diff
@@ -0,0 +1,225 @@
...
+    G = nx.cycle_graph(5, create_using=nx.DiGraph)  # No cycles
+    pytest.raises(nx.NetworkXError, nx.recursive_leaf_removal, G)
+
+
+def test_weight_nodes_by_condensation():
+    center_cycle = [(0, 1), (1, 2), (2, 3), (3, 1), (3, 4)]
+    G = nx.DiGraph(center_cycle)  # unit weights by default
+    pytest.raises(
+        nx.NetworkXError, nx.weight_nodes_by_condensation, G
+    )  # Must be condensed (acyclic)
```
dschult, line 89: Can you move this comment to the line prior to the pytest?
And other comments in this file that end a line...   Thanks!

networkx/algorithms/tests/test_hierarchy_coordinates.py
```diff
@@ -0,0 +1,225 @@
...
+        A=weighted_center_cycle_A, num_thresholds=5, threshold_distribution=None
+    )
+    assert len(condensed_Gs) == len(
+        binary_Gs
+    )  # Not necessarily equal to num thresholds, as empty graphs are dropped
+    for index in range(len(binary_Gs)):
+        assert np.array_equal(
+            np.unique(nx.to_numpy_array(binary_Gs[index])), [0, 1]
+        )  # binary checks
+        assert np.array_equal(np.unique(nx.to_numpy_array(condensed_Gs[index])), [0, 1])
```
dschult, line 147:
```suggestion
    for binaryG, condensedG in zip(binary_Gs, condensed_Gs):
        # binary checks
        assert np.array_equal(np.unique(nx.to_numpy_array(binaryG)), [0, 1])
        assert np.array_equal(np.unique(nx.to_numpy_array(condensedG)), [0, 1])
```
You could easily split this into two loops also.   Other places with `index` can use `zip` as well.
More readable, should be faster because done in C, more "pythonic" (whatever that means).

MaxBromberg: I implemented the suggested changes, though I didn't find other places which could better use `zip`; I think you might've caught them all already.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1166 @@
...
+    else:
+        column_sums = matrix.sum(axis=0)
+        return np.array(
+            [
+                matrix[:, index] / column_sums[index]
+                if column_sums[index] != 0
+                else [0] * column_sums.size
+                for index in range(column_sums.size)
+            ]
+        ).T
```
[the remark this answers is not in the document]
MaxBromberg: Clever! I fear I haped upon this convoluted mean myself, because I ran into some error originally, likely solved by your `keepdims=True`. Implemented as suggested, with axis replacing row_normalize and sklearn comment deleted.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1139 @@
...
+    if forward_entropy:
+        B_prime = _matrix_normalize(nx.to_numpy_array(dag), axis=1)
+        # +1 as k \in ( 1, L(G_C) )
+        P = sum([np.power(B_prime, k) for k in range(1, L_GC + 1)])
+        # TODO: Not so sure about this unless k coincides with the number of steps already taken (and the sum is odd)
+        # (Presently awaiting response from original authors for clarification)
+    else:
+        # as w sklearn: B = normalize(nx.to_numpy_array(dag), axis=0, norm='max')  # Column normalization
+        B = _matrix_normalize(nx.to_numpy_array(dag), axis=0)
+        P = sum([np.power(B, k) for k in range(1, L_GC + 1)])
```
dschult, line 743: I think this whole section collapses to a couple of lines now that axis is the only thing that changes.
The comments indicate that we maybe should be using `max` instead of `sum` to normalize. 
(i.e. use the infinity norm instead of the L_1 norm).   Thoughts?

If this does collapse, we should probably inline the _matrix_normalize

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1139 @@
...
+    ...     [0, 0.4, 0, 0, 0],
+    ...     [0, 0, 0.1, 0, 1.0],
+    ...     [0, 0, 0, 0, 0],
+    ... ])
+
+    >>> num_thresholds = 2
+    >>> condensed_networks, _ = nx.node_weighted_condense(b, num_thresholds=num_thresholds)
+    >>> for network_index in range(num_thresholds):
+    ...     for node_index in range(len(condensed_networks[network_index].nodes)):
+    ...         print(f"Node {node_index}, new weight:", condensed_networks[network_index].nodes[node_index]["weight"])
```
dschult, line 204:
```suggestion
    >>> for cG in condensed_networks:
    ...     for node_index, (node, weight) in enumerate(cG.nodes.data("weight")):
    ...         print(f"Node {node}, new weight: {weight}")
```
I couldn't easily see whether the node_index is needed here. If condensed_networks has integer nodes then it isn't.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1139 @@
...
+            "Ensure feedforwardness is fed nodes weighted by consdensation"
+            ", e.g. G = nx.weight_nodes_by_condensation(nx.condensation(G))"
+        )
+
+    successively_peeled_nx_graphs = recursive_leaf_removal(DAG, from_top=False)
+    if (
+        len(successively_peeled_nx_graphs) == 1
+        and len(successively_peeled_nx_graphs[0].nodes()) == 1
+    ):
+        return 0
```
dschult, line 646: Does this special case treatment avoid an error later? If not, this could be removed...
If so, then don't we need to check all peeled graphs for a single node?

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1139 @@
...
+                # where each path calculation is F(path)
+                g += len(path) / sum([weights[node] for node in path])
+                num_paths += 1
+    return g, num_paths
+
+
+@not_implemented_for("undirected")
+def feedforwardness(DAG):
+    """Returns feedforwardness hierarchy coordinates, ~extent information passes vertically
+    (vs horizontally) within a hierarchy.
```
dschult, line 562: The one-line description is used to labels documentation links to this function.
So it really should be short and sweet (one line:)

Here, you should split this an start a next paragraph with the description of what feedforwardness is.
Then  the paragrph you've got about how it is computed -- which should be clarified... it is too vague
as currently written.  What are the weights, how is the "repeated calculation" combined to get a single value (or is it combined?)

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1139 @@
...
+    ----------
+    DAG: NetworkX Graph
+        A Directed, Acyclic Graph.
+
+    Return
+    -------
+    float:
+        feedforwardness, hierarchy coordinate
+
+    Examples
```
dschult, line 581: Should this example have some text describing what is going on?
- The matrix is hard to understand as a graph -- perhaps create a network graph by adding weighted edges?
- What are the 8 graphs that are produced and how are they related to the original?

Maybe there should be a big picture paragraph somewhere in this docstring to explain what this function does to a graph.

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: Yes! I've collapsed this as suggested. The sklearn `max` norm would've been an improper implementation, as **B** is defined as the adjacency matrix normalized by total _in_ (backward entropy) or _out_ (forward entropy) degree. 
It appears `hierarchy_coordinates.py` does not use `_matrix_normalize` more than the once, though I'm not quite sure how to inline it. (Would seem to have too much state for a lambda function)

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: You're right; no need for the `node_index`; implemented as suggested.

MaxBromberg: You're right; no need for the `node_index`; implemented as suggested.

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: So long as `recursive_leaf_removal` is passed DAGs (with `keep_linkless_layer=False` implicitly), the only time a singleton graph will be passed is when the original DAG itself was a single node, and thus the trivial 0 feedforwardness. I think this occurs not infrequently for well connected graphs whose giant component condenses to a single node, and so this 0 feedforwardness weighting is needed in the final average.

MaxBromberg: So long as `recursive_leaf_removal` is passed DAGs (with `keep_linkless_layer=False` implicitly), the only time a singleton graph will be passed is when the original DAG itself was a single node, and thus the trivial 0 feedforwardness. I think this occurs not infrequently for well connected graphs whose giant component condenses to a single node, and so this 0 feedforwardness weighting is needed in the final average.

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: I wrote up something more of a description, though perhaps the added math is a bit much. Let me know if the description could be improved.

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: I've simplified the example to just the infographic, without weighed edges; perhaps that better conveys the essence of it. 
The 8 graphs from the old example result from the default 8 (linear) thresholds produced from `node_weighted_condense`, a lá
```
nx_graphs = [nx.from_numpy_matrix(np.where(A > threshold, 1, 0), create_using=nx.DiGraph) for threshold in thresholds]
```
 I'd add some more big picture explanation, though I'm not sure where's appropriate.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1133 @@
...
+        raise nx.NetworkXError(
+            "G must be a directed acyclic graph for graph entropy evaluation"
+        )
+
+    dag = nx.convert_node_labels_to_integers(DAG)
+    # Could be passed in most contexts to reduce redundant computation
+    L_GC = len(recursive_leaf_removal(DAG))
+
+    axis = 1 if forward_entropy else 0
+    B = _matrix_normalize(nx.to_numpy_array(dag), axis=axis)
```
dschult, line 734:
```suggestion
    B = nx.to_numpy_array(dag)
    # normalize so rows (forward_entropy) or columns (backward_entropy) sum to 1. 
    sums = B.sum(axis=axis, keepdims=True)
    sums[sums==0] = 1
    B = B / sums
```
This is all I was thinking of for "inlining"...

networkx/algorithms/hierarchy_coordinates.py
dschult: Then maybe reduce this (and maybe put before the recursive_leaf_removal) to:   ???
```python
if len(DAG) == 1:
    return 0
```

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: done!

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: done!

MaxBromberg:
At present, neither [Prof. Sole](http://complex.upf.es/~ricard/Main/RicardSole.html) nor [Dr. Rodriguez](https://www.linkedin.com/in/carlos-rodriguez-caso-836a3514) have responded to my email asking for clarification on the entropy calculation for this original algorithm, though otherwise I imagine we're roughly ready for release. Is there something I ought to do to better prepare the code at present?

Relevant Emails I found:
ricard.sole@upf.edu
carlos.rodriguez@upf.edu

dschult:
I don't think we should rely on a response from the authors if we haven't got one yet.
What is the issue/question a bout entropy that you are asking them about?

MaxBromberg:
This was the body of the question I sent them:

> The LaTeX(t) suggests a entropy path is a successive sum for every non-maximal node in a given path over P_ij, the probability of a walker on node i going to node j, multiplied by log(degree of non-maximal node). As I understand the text however, this would necessitate P to be calculated independently for each layer, and not as a sum, i.e: eq[13] P(v_j <-- v_i) = \sum_{1 \leq k \leq L(G_C)} ( [B^T]^k (G) )_ij would then instead become P(v_j <-- v_i) = ( [B^T]^k (G) )_ij,  k = num steps already taken. 
> The infographic's above sums appear to suggest a path's entropy is 1/(num divergences) * log (num divergences), but the sums above do not add to the values showed below unless log = log base 2, not ln; is this an arbitrary choice?

I imagine that the choice of log base is relatively arbitrary, but the difference between the infographic entropy, as implemented, and the different versions of the probability of transition (i.e. with or without the sum over the exponentiated edge normalized adjacency matrix, B) is a difference between a computationally easy analytic calculation (matrix exponentiation) or a sum over all (simple) paths.

dschult:
That is helpful, but I still have questions. First of all, does the choice of logarithmic base cause the difference in formulas?  My intuition is that the formula should be the same either way, just with a different base.  The change is base is simply a difference of multiplying by a constant: log_2(x) = ln(x)/ln(2). and my intuition of entropy definitions is that you can multiply by a constant and the important features don't change.  Is that correct?

Can you describe again what drives the difference in formulas?  I don't understand why P would need to be calculated independently for each layer.  Or rather it seems to me that P_ij is computed for each edge. Then the entropy path is computed by walking the path and getting a weighted sum of the edges weighted by log(degree of outgoing node).

MaxBromberg:
> That is helpful, but I still have questions. First of all, does the choice of logarithmic base cause the difference in formulas? My intuition is that the formula should be the same either way, just with a different base. The change is base is simply a difference of multiplying by a constant: log_2(x) = ln(x)/ln(2). and my intuition of entropy definitions is that you can multiply by a constant and the important features don't change. Is that correct?

Yes! The choice of log is naught to worry about, methinks. Fundamentally, the character of entropy remains intact.

> Can you describe again what drives the difference in formulas? I don't understand why P would need to be calculated independently for each layer. Or rather it seems to me that P_ij is computed for each edge. Then the entropy path is computed by walking the path and getting a weighted sum of the edges weighted [multiplied] by log(degree of outgoing node).

This is the idea in both cases; entropy is calculated by walking every path between extremal nodes, while maintaining a running sum of the probability of stepping to a given (j th) node multiplied by its degree. 
In `infographic_graph_entropy`, this is done directly, considering every path explicitly thus the `nx.all_simple_paths` call, where at every branch the (integer) number of possible directions is added to a running sum, let's call it **F** (for forks), which then factors into the sum over each path as log(**F**) / **F**.  This sum is then divided by the starting number of extremal nodes  to yield the entropy. Notice that the probability of transition is here only _implicitly_ considered, as entropy is a literal weighted sum of possible forks.

The `graph_entropy` call is hopefully a bit more clever, and computationally concise. It uses the fact that the n-th power of a degree-normalized adjacency matrix yields a matrix whose i,j-th element is the probability of transition from node i to node j after n steps. This means the sum of the powers of the degree normalized matrix (` P = sum([np.power(B, k) for k in range(1, L_GC + 1)]) `)  yield the probability of all possible transitions, accounting for more or less direct paths, with proportionately varying number of steps, between any two given nodes. 
Therefore, the sum over this probability matrix **P** need only consider each intermediary node once per maximal, minimal node pair, and all possible paths need not be calculated explicitly, as via:
```
e = sum(P[u][v] * np.log(deg(u)) for u in boundary_layer for v in non_extremal_nodes)
```
The difference is that `infographic_graph_entropy` (with present implementation) calculates single layer entropy, equivalent to the `graph_entropy` algorithm using only extremal nodes and powers of **B** independently (not summed). By using the summed powers of **B**, (**P**) in `graph_entropy` , and evaluating over all nodes (not just extremal nodes, but all paired with starting nodes) we consider all pruned layers as well. That said, I'm not sure which is intended; treeness does consider all layers independently in the infographic, but seemingly collectively in computation. (specifically in the [appendix][1], eq. 15 takes into account all possible transition paths, which in eq. 16, appears to be evaluated per node)

This said, I just saw I need to change a couple things in the `graph_entropy` function, specifically replacing u with v in `np.log(deg(u))` above, and ensuring that `B.T` is exponentiated for backwards entropy. I'll push these changes soon, but it doesn't change the main point of confusion. Despite all this, both forms of graph entropy make sense to me, and it seems a relatively minor semantic difference in the grand scheme of getting a clear notion of hierarchy coordinates.

[1]: www.pnas.org/lookup/suppl/doi:10.1073/pnas.1300832110/-/DCSupplemental

dschult:
Glad that this is getting closer to matching the mathematical description in the paper.

Not that you need more to think about here, but I'm concerned with network algorithms that rely on formulas that use matrix exponentials.  The formulas themselves are usually compact and easy to read (yay for linear algebra!). But in a network setting the matrix is usually sparse -- and the equivalent of exponentiation can be done effectively by traversing the graph rather than forming a (dense) matrix and exponentiating it.  

Are these methods intended to be used on fairly dense networks? What is a typical use case for this algorithm?

Also, am I missing something that makes the linear algebra approach stronger here even when we don't have dense networks/matrices?

MaxBromberg:
> Not that you need more to think about here, but I'm concerned with network algorithms that rely on formulas that use matrix exponentials. The formulas themselves are usually compact and easy to read (yay for linear algebra!). But in a network setting the matrix is usually sparse -- and the equivalent of exponentiation can be done effectively by traversing the graph rather than forming a (dense) matrix and exponentiating it.

Good point! This is something which I addressed in an earlier use of matrix exponentiation for Random Walker Effective Distances using scipy's sparse optimization (and could be likewise adapted into this implementation when matrix exponentiation is called for) but I was under the mistaken impression that matrix exponentiation is easier computationally than repeated `all_simple_paths` calls. 

> Are these methods intended to be used on fairly dense networks? What is a typical use case for this algorithm?

There is no particular reason why hierarchy coordinates would be used more in dense networks.

> Also, am I missing something that makes the linear algebra approach stronger here even when we don't have dense networks/matrices?

...Likely not, and I'm mistaken as to which is more efficient. However, just rephrasing my earlier note, my reasoning was that the (sum of) matrix exponentiation(s) occurs but once, and then for each pair of extremal nodes, each node is considered but once. In the `nx.all_simple_paths` dependent case (`infographic_graph_entropy`) , for each pair of extremal nodes there's a new `nx.all_simple_paths` call, and all paths are evaluated individually, such that paths which are identical save for a single deviation nonetheless are computed wholly independently. 
From what a quick reference to [Wikipedia][1] suggests, square matrix multiplication of dims (n x n)^k would take at best ~O(k * n^2.4) time, whereas the `nx.all_simple_paths` algorithm (constant, O(V+E) time) would be called |M| * |mu| times, where |M| and |mu| are the number of maximal, minimal nodes respectively. ...Which just goes to show why I should've considered this earlier; it appears it would be an extraordinary circumstance indeed which leads to a faster matrix computation than simply using the existing `infographic_graph_entropy` call. 
My bad then! Apart from the possible change of log base, increasing tests, relabeling `infographic_graph_entropy` to `graph_entropy`,  `graph_entropy` to `analytic_graph_entropy`, and dealing with whatever made my last push fail to pass all checks, we might be good. Let me know if that seems a reasonable conclusion, or what else it might need.

[1]: https://www.wikiwand.com/en/Computational_complexity_of_mathematical_operations

MaxBromberg:
Just

MaxBromberg:
Just re-iterativing this, I might've slipped by in my edit rather than repost:
> Apart from the possible change of log base, increasing tests, relabeling infographic_graph_entropy to graph_entropy, graph_entropy to analytic_graph_entropy, and dealing with whatever made my last push fail to pass all checks, we might be good. Let me know if that seems a reasonable conclusion, or what else it might need.

dschult:
My head has not been able to reload this thread completely yet...
But your last comment does indeed sound reasonable.  

I think the last two tests are not passing due to a fix that has been merged with the main branch. To get them to pass probably just requires merging with the main branch again.

MaxBromberg:
Thanks for the step by step explanation; I hope you have a copy pasta for the confused like myself. 
`git remote -v` shows the upstream already as `git@github.com:networkx/networkx.git`, and `git pull upstream main` results in an access error:
```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
Though I thought read access was allowed...

dschult:
That is indeed confusing...  Can you check that connecting to other similar repositories still works correctly? I know [github is making changes to its authentication rules,](https://github.blog/2020-07-30-token-authentication-requirements-for-api-and-git-operations/) but this doesn't seem like it is related. And this worked before, so that's confusing.

I guess you could try making a new remote using a different interface (like https://github.com/networkx/networkx.git)-- or even deleting and re-adding the remote git@github.com.  The repository definitely should give read access for everyone.

MaxBromberg:
Alright! Hopefully this works, I took your advice and used  `https://github.com/networkx/networkx.git` rather than `git@github.com...` and it seemed to then allow the `git pull upstream main`.

networkx/algorithms/tests/test_hierarchy_coordinates.py
```diff
@@ -0,0 +1,273 @@
...
+        (5, 6),
+    ]
+    G = nx.weight_nodes_by_condensation(
+        nx.condensation(nx.DiGraph(infographic_network))
+    )
+    assert np.round(nx.feedforwardness(G), 2) == 0.56
+
+
+def test_hierarchy_coordinate():
+    import numpy as np
```
dschult, line 245: Shouldn't need import numpy since imported at top.

networkx/algorithms/tests/test_hierarchy_coordinates.py
```diff
@@ -0,0 +1,273 @@
...
+            [0, 0, 0, 0, 0, 0, 0],
+        ]
+    )
+    a_hc = [round(hc, 2) for hc in nx.hierarchy_coordinates(a)]
+    b_hc = [round(hc, 2) for hc in nx.hierarchy_coordinates(b)]
+    assert a_hc == [-0.08, 0.87, 0.81]
+    assert b_hc == [-0.56, 0.56, 0.43]
+
+
+test_hierarchy_coordinate()
```
dschult, line 273: Let pytest run the test (it also loads the local repository).

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1107 @@
+"""
+========================
+Hierarchy Coordinates
+========================
+Functions to calculate hierarchy coordinates for directed networks.
+
+Hierarchy coordinates consist of treeness, feedforwardness, and orderability, which collectively create a
```
dschult, line 7: Can you pull the docs into an 80-char width (or so) for better readability?

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1107 @@
...
+    "graph_entropy",
+    "infographic_graph_entropy",
+    "treeness",
+    "hierarchy_coordinates",
+]
+
+
+# Hierarchy Coordinate Functions
+def node_weighted_condense(A, num_thresholds=8, threshold_distribution=None):
+    """Creates a series of networkx graphs based on generated edge weight thresholds, and returns their node weighted
```
dschult, line 46: Can you make the doc_strings start with a single-line description. This is used as text to go with links to this function in the documentation.  Keep it short.

Then follow up with a paragraph that actually says more completely what the function does.  Try to keep all doc lines around 80-chars or less.

You need to define what "node weighted condense and original graphs" means.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1107 @@
...
+    Node 2, new weight: 1
+
+    Raises
+    ------
+    NetworkXError
+        If input not a directed acyclic graph
+
+    Note:
+    ------
+    TODO: Might wish to eliminate return, or enable copying?
```
dschult, line 219: If it is changed "in-place" then don't return the graph. Also state in the documentation that the graph is updated in place.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1107 @@
...
+    # WIP TODO: May not be necessary, as the graph itself is updated (not copied)?
+    return condensed_graph
+
+
+@not_implemented_for("undirected")
+def max_min_layers(G, max_layer=True):
+    """Returns the maximal (k_in = 0) layer or the minimal layer (k_out = 0)
+
+    Returns the maximal (k_in = 0, highest in hierarchy) layer (those nodes with in degree = 0)
+     or the minimal layer (k_out = 0) of a directed network.
```
dschult, line 245: Maybe say what the words hierarchy and layer mean?  Just give some context.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1107 @@
...
+    [4, 5, 6]
+
+    Raises
+    ------
+    NetworkXError
+        If input not a directed graph
+
+    Notes:
+    -------
+
```
dschult, line 284: If you don't have notes, you can leave that section out.

networkx/algorithms/hierarchy_coordinates.py
```diff
@@ -0,0 +1,1107 @@
...
+        raise nx.NetworkXError("G must be a DiGraph for min/max layer evaluation")
+    if max_layer:
+        return [node for node, deg in G.in_degree if deg == 0]
+    else:
+        return [node for node, deg in G.out_degree if deg == 0]
+
+
+@not_implemented_for("undirected")
+def leaf_removal(G, top=True):
+    """Returns a pruned network, with either maximal (k_in=0)
```
dschult, line 296: Tighten up this language.  And add a paragraph to describe it more fully. What do you mean by pruned? Say something about why this function is useful, or how it fits wit the other code around it -- when should users use it?

networkx/algorithms/hierarchy_coordinates.py
dschult: Add big picture description everywhere you can. It's really needed in this module. Users come to the docs via google searches so you can't tell where they will start. You need to make every doc_string be good enough to be the first doc_string people see.  If you end up repeating ideas in many functions, use the same language in each so the user recognizes that it is the same ideas as they saw before -- but if its the first time they understand what's going on.
:}

networkx/algorithms/hierarchy_coordinates.py
MaxBromberg: I wasn't sure how to best implement this; presumably as a member of the graph class, such that a G.weight_nodes_by_condensation would make sense, but I think it might be better to either do without the function (it's only used once), or make it general enough to be generally useful. Or it could be best as an optional argument in the condense function?

MaxBromberg: I wasn't sure how to best implement this; presumably as a member of the graph class, such that a G.weight_nodes_by_condensation would make sense, but I think it might be better to either do without the function (it's only used once), or make it general enough to be generally useful. Or it could be best as an optional argument in the condense function?

MaxBromberg:
Thanks for the suggestions, which I hope I've made appropriate edits for, save re: weight_nodes_by_condensation, as commented above. I hope this makes hierarchy coordinates nearly deployable, at least as the original author's used them. 

There remains a significant edit to accommodate the edge cases addressed in a paragraph of their appendix (p. 7) when they consider how condensation can break the graph into more than one single connected component. In their numerical investigations, they only consider the largest connect components, however theoretically outline taking a weighted average of all components, which we should perhaps likewise do. In code, this would translate to accommodating the consequences of changing
```
    for condensed_graph in condensed_graphs:
        largest_condensed_graphs.append(nx.convert_node_labels_to_integers(condensed_graph.subgraph(
                    max(nx.weakly_connected_components(condensed_graph), key=len))).copy())
```
to 
```
    for condensed_graph in condensed_graphs:
        condensed_graphs_per_threshold.append([nx.convert_node_labels_to_integers(condensed_graph.subgraph(c)) 
for c in sorted(
        nx.weakly_connected_components(condensed_graph), key=len, reverse=True).copy()])
                                              )
```

MaxBromberg:
Thanks for the suggestions, which I hope I've made appropriate edits for, save re: weight_nodes_by_condensation, as commented above. I hope this makes hierarchy coordinates nearly deployable, at least as the original author's used them. 

There remains a significant edit to accommodate the edge cases addressed in a paragraph of their appendix (p. 7) when they consider how condensation can break the graph into more than one single connected component. In their numerical investigations, they only consider the largest connect components, however theoretically outline taking a weighted average of all components, which we should perhaps likewise do. In code, this would translate to accommodating the consequences of changing
```
    for condensed_graph in condensed_graphs:
        largest_condensed_graphs.append(nx.convert_node_labels_to_integers(condensed_graph.subgraph(
                    max(nx.weakly_connected_components(condensed_graph), key=len))).copy())
```
to 
```
    for condensed_graph in condensed_graphs:
        condensed_graphs_per_threshold.append([nx.convert_node_labels_to_integers(condensed_graph.subgraph(c)) 
for c in sorted(
        nx.weakly_connected_components(condensed_graph), key=len, reverse=True).copy()])
                                              )
```

dschult:
I think you are right that the proper treatment is averaging over components. 
I think your code example has a `copy()` misplaced -- it should copy the subgraph, not the sorted list. Otherwise it looks good to me.

I think there are a couple of places in the newly revised docstrings where the indent level changes. Can you correct those? And also a place with 3 blank lines before the "See Also" section.  Perhaps reduce to 1 or 2?

Thanks for this!

MaxBromberg:
I'm sorry I haven't returned to this in very long! I've been busy with a new job, and have some difficulty reloading the project into mental RAM, but I promise I'm alive and aware I ought to!

dschult:
No problem Maqz!  (Darn that real life stuff -- it keeps getting in the way!)

Quick question in case you know the answer easily.  The docs say that hierarchy coordinates can classify all networks structures. "all" is a strong word. Does this mean that the three hierarchy sets of values uniquely identify the network?
I'm just try to find the right words there...
Thanks!

PR #4653, opened 2021-03-05
Optimize multipartite_layout

### LINKED ISSUE #4061: Tighten partite_layout if/else code.
Tighten up the bipartite_layout and multipartite_layout  if/else as describes in #3815

z3y50n:
Fixes #4061 

Since the logic for drawing the graph in both `vertical` and `horizontal` alignment is the same we can use one loop and just flip the positions.

Results are the same as before.

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -310,6 +310,10 @@ def bipartite_layout(
 
     import numpy as np
 
+    if align not in ("vertical", "horizontal"):
+        msg = "align must be either vertical or horizontal."
+        raise ValueError(msg)
+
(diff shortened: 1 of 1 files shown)


dschult (approved):
Nice!   That certainly reduces the duplicated code. :)

rossbar (commented):
Looks good - there is a similar pattern in `bipartite_layout` that would be good to fix in this PR as well if you have the time. If not, we can merge this one but leave the issue open.

z3y50n:
Yes I was thinking the same thing, I will work on fixing `bipartite_layout` as well

rossbar (approved):
Awesome, thanks @z3y50n , LGTM!

PR #4658, opened 2021-03-06
[fix] for negative_edge_cycle incorrect call of bellman_ford

shichuzhu:
Use keyword argument instead of positional argument to avoid argument mismatch.
Affected function: `negative_edge_cycle`

### DIFF
--- a/networkx/algorithms/shortest_paths/tests/test_weighted.py
+++ b/networkx/algorithms/shortest_paths/tests/test_weighted.py
@@ -338,6 +338,14 @@ class TestWeightedPath(WeightedTestBase):
         G.add_edge(9, 10)
         pytest.raises(ValueError, nx.bidirectional_dijkstra, G, 8, 10)
 
+    def test_negative_edge_cycle_custom_weight_key(self):
+        d = nx.DiGraph()
+        d.add_node("a")
+        d.add_node("b")
+        d.add_edge("a", "b", w=-2)
+        d.add_edge("b", "a", w=-1)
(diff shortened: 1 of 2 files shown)


dschult:
Great!
Can you add your example code to create the error as a test so we don't fall into that again?
Thanks!

shichuzhu:
Sure I can do that.
Shall I commit to the same branch and request a PR again?

networkx/algorithms/shortest_paths/tests/test_weighted.py
```diff
@@ -338,6 +338,14 @@ def test_negative_edge_cycle(self):
         G.add_edge(9, 10)
         pytest.raises(ValueError, nx.bidirectional_dijkstra, G, 8, 10)
 
+    def test_negative_edge_cycle_custom_weight_key(self):
+        d = nx.DiGraph()
+        d.add_node("a")
+        d.add_node("b")
```
dschult, line 344: These two lines aren't needed because the `add_edge` automatically adds the nodes if needed.

dschult:
Good -- pushing to the same branch automatically adds to the PR so you don't need to request a PR again (as I think you found out). 
Thanks!

networkx/algorithms/shortest_paths/tests/test_weighted.py
shichuzhu: Done.

dschult (approved):
Looks good to me!
Thanks!

rossbar (approved):
LGTM, thanks @shichuzhu

PR #4659, opened 2021-03-06
Add an `initial` parameter to simple and dual Barábasi-Albert random graphs

### LINKED ISSUE #3281: barabasi_albert_graph does not start off with a connected network
`barabasi_albert_graph` starts with `G=empty_graph(m)` instead of `G = complete_graph(m)`.
As a result, the early new nodes are almost certain to become hubs, since the distribution is heavily skewed in their favour. The initial `m` nodes have almost zero chance to be a hub.

To reproduce:
```
hubs = []
for i in range(100):
    G = nx.barabasi_albert_graph(1000,500)
    hubs.append(sorted(G, key=lambda n: len(G[n]), reverse=True)[0])
count_dict={}
for hub in hubs:
    count_dict[hub]=hubs.count(hub)
print ('\nfrequency:\t',count_dict)

# prints
# frequency:	 {500: 34, 501: 27, 502: 13, 503: 10, 504: 11, 505: 2, 506: 3}
```

The first new node (500) has the highest chance of becoming the hub. This is not random enough.

Quote from Wikipedia ( https://en.wikipedia.org/wiki/Barab%C3%A1si%E2%80%93Albert_model ):
> The network begins with an initial connected network of m_0 nodes. 

And logically it has to start from a connected network for the initial `m` nodes to have a chance to be a hub.

The effect would not be noticeable in "small world" kind of networks (medium node degree << number of nodes) but is profound in networks with n and m close enough to each other.

### LINKED ISSUE #3459: Allow generators to start with a seed network
For example, instead of always starting Barabasi-Albert networks with the same base network, add a new parameter initial that takes a graph, and the preferential attachment proceeds from there.  This would allow the study of the growth of the Barabasi-Albert network, capturing its behavior after every few steps, instead of always starting from scratch.

Here is some sample code for B-A:
```
def barabasi_albert_graph(n, m, seed=None, initial=None):
    """Returns a random graph according to the Barabási–Albert preferential
    attachment model.
    A graph of $n$ nodes is grown by attaching new nodes each with $m$
    edges that are preferentially attached to existing nodes with high degree.
    Parameters
    ----------
    n : int
        Number of nodes
    m : int
        Number of edges to attach from a new node to existing nodes
    initial : Graph or None
        If Graph, starts with that Graph as the initial graph
        If None, starts with m nodes and zero edges
    Returns
    -------
    G : Graph
    Raises
    ------
    NetworkXError
        If `m` does not satisfy ``1 <= m < n``.
    References
    ----------
    .. [1] A. L. Barabási and R. Albert "Emergence of scaling in
       random networks", Science 286, pp 509-512, 1999.
    """

    if m < 1 or m >= n:
        raise nx.NetworkXError("Barabási–Albert network must have m >= 1"
                               " and m < n, m = %d, n = %d" % (m, n))
    
    if initial == None:
        # Add m initial nodes (m0 in barabasi-speak)
        G = empty_graph(m)
        # Target nodes for new edges
        targets = list(range(m))
        # List of existing nodes, with nodes repeated once for each adjacent edge
        repeated_nodes = []
        # Start adding the other n-m nodes. The first node is m.
        source = m
    else:
        G = initial
        targets = list(G.nodes())
        repeated_nodes = [i for n,d in G.degree() for i in [n]*d]
        source = len(G.nodes())
    while source < n:
        # Add edges to m nodes from the source.
        G.add_edges_from(zip([source] * m, targets))
        # Add one node to the list for each new edge just created.
        repeated_nodes.extend(targets)
        # And the new node "source" has m edges to add to the list.
        repeated_nodes.extend([source] * m)
        # Now choose m unique nodes from the existing nodes
        # Pick uniformly from repeated_nodes (preferential attachment)
        targets = _random_subset(repeated_nodes, m, seed)
        source += 1
    return G
```
(I don't have the time to create a fork and pull request right this moment.)

Aufinal:
As asked in #3459, adds a new `initial` parameter to Barábasi-Albert random graph models that allows the user to specify a base graph.

Notes:
- my IDE formatted some imports automatically; I left them in the PR but I can remove those changes if needed,
- as noted in [this paper](http://www.mathematik.uni-muenchen.de/%7Ekpanagio/ModelsLitSS12/DegreeSequenceBA.pdf) (page 3), and somewhat in #3281, the BA model is ill-defined whenever some degrees are zero. For this reason, the initial graphs for `barabasi_albert_graph` and `dual_barabasi_albert_graph` have been changed to stars on (m+1) and max(m1, m2) + 1 vertices, respectively. For the simple model, this does not change the end result, since the first added vertex was always connected to the m others.
- The tests have been improved: previously, the `repeat` parameter was useless since the seed (and hence the randomness) was fixed. The new behavior tests the functions for several distinct seeds.

Fixes #3459, fixes #3281


networkx/generators/tests/test_random_graphs.py
```diff
@@ -137,28 +142,49 @@ def test_dual_barabasi_albert(self, m1=1, m2=4, p=0.5):
...
         pytest.raises(NetworkXError, dbag, 100, m1, m2, 1.5)
+        pytest.raises(
+            NetworkXError,
+            dbag,
+            100,
+            m1,
+            m2,
+            p,
+            initial=nx.complete_graph(max(m1, m2) - 1),
+        )
```
dschult, line 187: This might make it easier to see that p has changed as well as initial relative to previous lines. 
```suggestion
        initial=nx.complete_graph(max(m1, m2) - 1)
        pytest.raises(NetworkXError, dbag, 100, m1, m2, p, initial=initial)
```

dschult (commented):
This looks great -- and is overdue -- so Thank You!
@hagberg might want to look too.

My pass through the code didn't take long -- well done -- and I only came up with a couple of questions -- some of which aren't really related to the heart of the PR at all. I don't want them to bog down getting this merged but I will put them here anyway and if we get off track we can move that discussion to another PR.

- [x] Should the keyword argument be more descriptive at the expense of length? `initial` to `initial_graph`?  I'm not happy with either.  Can anyone find a better name? `initial` is actually pretty good.
- [x] Why do we import all these functions at the top of the test module? They are all at the top level of the `nx` namespace. The list at the top does tell us which functions are tested in the module. And putting `nx.` in front of each function will take some work. But the current treatment is not like our other test modules.  What is the current best practice?

networkx/generators/tests/test_random_graphs.py
Aufinal: Agreed, fixed it !

Aufinal:
- Regarding the keyword argument, `initial` does seem like the most natural to me, although I'm completely open to changing it.
- I admit that introducing the toplevel `networkx` import introduced some redundancy, but I didn't want to clutter this PR by making those changes. As an aside, being a new contributor, I couldn't find any best practices on how to write tests... Maybe this causes some of the clutter/confusion in the tests ?

networkx/generators/tests/test_random_graphs.py
```diff
@@ -137,28 +142,42 @@ def test_dual_barabasi_albert(self, m1=1, m2=4, p=0.5):
...
+
+            BA3 = barabasi_albert_graph(100, m1, seed)
+            DBA3 = dual_barabasi_albert_graph(100, m1, m1, p, seed)
+            # We can't compare edges here since randomness is "consumed" when drawing
+            # between m1 and m2
+            assert BA3.size() == DBA3.size()
+
+            DBA = dual_barabasi_albert_graph(100, m1, m2, p, seed, initial=initial)
+            BA1 = barabasi_albert_graph(100, m1, seed, initial=initial)
+            BA2 = barabasi_albert_graph(100, m1, seed, initial=initial)
```
dschult, line 168: Isn't `initial` mutated and returned?  So DBA and BA1 and BA2 will end up all the same object?
If true, we will need to make the mutating nature very clear -- or change to `G = initial.copy()`

Aufinal: Good point ! I changed it to `initial.copy()`.

networkx/generators/random_graphs.py
```diff
@@ -704,7 +718,8 @@ def dual_barabasi_albert_graph(n, m1, m2, p, seed=None):
     Raises
     ------
     NetworkXError
-        If `m1` and `m2` do not satisfy ``1 <= m1,m2 < n`` or `p` does not satisfy ``0 <= p <= 1``.
+        If `m1` and `m2` do not satisfy ``1 <= m1,m2 < n``, `p` does not satisfy ``0 <= p <= 1``,
+        or the initial graph size m0 does not satisfy m1, m2 <= m0 <= n.
```
dschult, line 722: Is `m0` the size of G or the order? Maybe "initial graph number of nodes m0 does not satisfy"...

networkx/generators/random_graphs.py
```diff
@@ -730,34 +745,39 @@ def dual_barabasi_albert_graph(n, m1, m2, p, seed=None):
...
+        if len(initial) < max(m1, m2) or len(initial) > n:
+            raise nx.NetworkXError(
+                f"Barabási–Albert initial graph must have between "
+                f"max(m1, m2) = {max(m1, m2)} and n = {n} nodes"
+            )
+        G = initial.copy()
+
     # Target nodes for new edges
-    targets = list(range(max(m1, m2)))
+    targets = list(G.nodes())
```
dschult, line 760:
```suggestion
    targets = list(G)
```
Or if you prefer `list(G.nodes)`

networkx/generators/random_graphs.py
```diff
@@ -631,6 +632,9 @@ def barabasi_albert_graph(n, m, seed=None):
     seed : integer, random_state, or None (default)
         Indicator of random number generation state.
         See :ref:`Randomness<randomness>`.
+    initial : Graph or None (default)
+        Initial network for Barabási–Albert algorithm.
```
dschult, line 636:
```suggestion
        Initial network for Barabási–Albert algorithm. A copy of `initial` is used.
```

networkx/generators/random_graphs.py
```diff
@@ -696,6 +707,9 @@ def dual_barabasi_albert_graph(n, m1, m2, p, seed=None):
...
```
dschult, line 711:
```suggestion
        Initial network for Barabási–Albert algorithm. A copy of `initial` is used.
```

dschult (commented):
I'm still reluctant to use `initial` here -- especially because I like this idiom and expect it to get used elsewhere. Does anyone have a preference between `initial_graph` and `starting_from`?

These comments are picking nits. If you don't have time, let me know.
I think this is very close to being merged.

networkx/generators/random_graphs.py
```diff
@@ -631,6 +632,9 @@ def barabasi_albert_graph(n, m, seed=None):
...
+        If None, starts from an star graph on (m+1) nodes.
```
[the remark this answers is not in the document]
Aufinal: Would `seed_graph` be better ? After all, it's also a way to control randomness.

networkx/generators/random_graphs.py
```diff
@@ -639,7 +643,7 @@ def barabasi_albert_graph(n, m, seed=None):
     Raises
     ------
     NetworkXError
-        If `m` does not satisfy ``1 <= m < n``.
+        If `m` does not satisfy ``1 <= m < n``, or the initial graph size m0 does not satisfy ``m <= m0 <= n``.
```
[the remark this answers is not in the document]
Aufinal: Good point ! The initial graph should be connected ; if not, all isolated vertices will stay isolated throughout the process. I can add that either as an error or a warning, depending on your opinion. As for the initial number of nodes, I forgot that networkx' `size` is the number of edges ; it's indeed better to change the formulation.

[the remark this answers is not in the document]
dschult: I think so long as you make a note in the docstring you shouldn't need to check for isolated nodes and raise a warning or an error. Most people won't provide a graph with isolated nodes, and some might want to do that on purpose. :}

networkx/generators/random_graphs.py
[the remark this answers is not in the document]
dschult: Yes, `seed_graph` would be good too. I'm OK with `seed_graph` or with `initial_graph`.  Thanks!

dschult (approved):
I think this is ready to merge

rossbar (approved):
LGTM - thanks @Aufinal !

PR #4661, opened 2021-03-08
Copy changes from https://github.com/networkx/networkx/pull/4573

robertclancy:


robertclancy:
Apologies for the PR noise. This was meant to be against the Intercom fork. We need to maintain a fork until https://github.com/networkx/networkx/pull/4573 is released.

PR #4664, opened 2021-03-10
Add removal of edges from planar embeddings

DrewSpringham:
I have added a way to remove half edges and node from planar embeddings in a way that preserves the structure.

Notes:

- I'm not a fan of the name remove_planar_node but couldn't think of anything better to differentiate from remove_node
- The use of itertools in remove_planar_node is probably a bit over the top, would appreciate feedback on what should be used

Thanks

### DIFF
--- a/networkx/algorithms/planarity.py
+++ b/networkx/algorithms/planarity.py
@@ -1001,6 +1001,47 @@ class PlanarEmbedding(nx.DiGraph):
         self[start_node][cw_reference]["ccw"] = end_node
         self[start_node][end_node]["ccw"] = reference_neighbor
 
+    def remove_half_edge(self, start_node, end_node):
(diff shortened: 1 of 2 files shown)


dschult:
For the name, you could go with something more verbose like: `remove_node_and_half_edges` Then your documentation doesn't have to describe what a planar node is.  I'm not sure what is best.

For the itertools, there is a networkx function `nx.all_neighbors(G, node)` which actually just uses `itertools.chain`. So your solution is good. `nx.all_neighbors` also checks if G is directed, and does the right thing in both cases. But you know `self` is a `DiGraph` so using chain makes sense.  

BUT, you should not convert it to a list and then iterate over it in the for loop.  Just leave out the list part.
```python
for v in nx.all_neighbors(self, u):
    self.remove_half_edge(u, v)
```

PR #4706, opened 2021-03-29
combinatorial_embedding_to_pos, specify outer face

ChristopheVuillot:
This PR adds the ability for the user to specify the triangulation and outer face when computing positions of the nodes for drawing.

The solution is the one discussed in https://github.com/networkx/networkx/discussions/4699#discussioncomment-542195

### DIFF
--- a/networkx/algorithms/planar_drawing.py
+++ b/networkx/algorithms/planar_drawing.py
@@ -2,7 +2,19 @@ import networkx as nx
 from collections import defaultdict
 
 
-__all__ = ["combinatorial_embedding_to_pos"]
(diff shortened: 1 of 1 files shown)


dschult:
This look great!   But still will need some basic tests before merging.
Can you look at the module (relative to the algorithms directory): `tests/test_planar_drawing.py`?
It uses `pytest` and is hopefully sufficiently readable that you could add a test to make sure
the new function works with a correct input...  
Probably good to also check that it gives a reasonable error if the input is not appropriate -- though I haven't thought about that for tis problem -- it might not make sense to do that...

Thank you

ChristopheVuillot:
Thanks for the feedback, I will have a look about adding tests.

PR #4713, opened 2021-03-30
Update bfs_search.py

a-vitale:
Aggiunta la stampa dei nodi processati (source + child)


dschult:
We don't want to print this often. For large Graphs it creates too much output.

Issue #4734, opened 2021-04-09
average_clustering() got an unexpected keyword argument 'trials'


DebarghaG:
Hey! 
I'm following [this](https://networkx.org/documentation/stable/_modules/networkx/algorithms/approximation/clustering_coefficient.html#average_clustering) and [this piece of documentation](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.approximation.clustering_coefficient.average_clustering.html#networkx.algorithms.approximation.clustering_coefficient.average_clustering)

### Current Behavior

On running : ```average_clustering(G, trials=1000, seed=None)```
```
TypeError                                 Traceback (most recent call last)
<ipython-input-24-7c16ef478e48> in <module>
----> 1 nx.average_clustering(G, trials=1000, seed=None)

TypeError: average_clustering() got an unexpected keyword argument 'trials'
```
However, running : ```average_clustering(G)```
It works as intended.

### Expected Behavior

average_clustering(G, trials=1000, seed=None) should be valid.

### Steps to Reproduce
Just make a Graph G and call nx.average_clustering(G, trials=1000, seed=None)

### Environment
Python version: 3.8.5
NetworkX version: 2.5.1

MridulS:
By default [this](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.average_clustering.html?highlight=average_clustering#networkx.algorithms.cluster.average_clustering) `avergage_clustering` is imported in the nx namespace, so to access the one in the `approximation` module you would need something like
``` python
>>> from networkx.algorithms import approximation
>>> approximation.average_clustering(G, trials=1000, seed=None)
......
```

We should definitely add this as an example in the [function documentation](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.approximation.clustering_coefficient.average_clustering.html#networkx.algorithms.approximation.clustering_coefficient.average_clustering).

DebarghaG:
Yeah, I guess that's a good addition to the documentation. Have made a PR.

PR #4735, opened 2021-04-09
Updating average_clustering() documentation - Issue #4734

DebarghaG:

### DIFF
--- a/networkx/algorithms/approximation/clustering_coefficient.py
+++ b/networkx/algorithms/approximation/clustering_coefficient.py
@@ -37,6 +37,11 @@ def average_clustering(G, trials=1000, seed=None):
     c : float
         Approximated average clustering coefficient.
 
+    Examples
+    --------
+    >>> from networkx.algorithms import approximation
+    >>> approximation.average_clustering(G, trials=1000, seed=None)
+
     References
     ----------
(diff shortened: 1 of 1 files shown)


MridulS:
Thanks for this @DebarghaG!
As part of our tests we run the examples too, so could you update the code to define a graph `G`?
``` python
>>> import networkx as nx
>>> from networkx.algorithms import approximation
>>> G = nx.erdos_renyi_graph(10, 0.2)
>>> approximation.average_clustering(G, trials=1000, seed=None)
```

DebarghaG:
@MridulS A little confused here. What's going wrong here?

MridulS:
ahh, it expects to check the output too, so do put in the output too and change the seed from `None` otherwise we can't guarantee the reproducibility of the output.

``` python
>>> import networkx as nx
>>> from networkx.algorithms import approximation
>>> G = nx.erdos_renyi_graph(10, 0.2)
>>> approximation.average_clustering(G, trials=1000, seed=42)
0.085
```

DebarghaG:
Hey @MridulS , Thanks for all the handholding! :hugs: :smiley:

MridulS (approved):

dschult (approved):
Very Nice!
Thank you!

PR #4739, opened 2021-04-11
Remove decorator dependency

boothby:
Pursuant to removing the dependency on `decorator`, I've made a general decorating class which applies functions to specified arguments.  See discussion in #4732.


networkx/utils/decorators.py
```diff
@@ -450,22 +359,388 @@ def random_array(dims, random_state=1):
...
-        except IndexError as e:
-            raise nx.NetworkXError("random_state_index is incorrect") from e
+    return argmap(create_py_random_state, random_state_index)
+
+
+_tabs = " " * 512
+
+
+class argmap:
+    """A decorating class which calls specified transformations on a function's
```
dschult, line 369:
```suggestion
    """Decorate a function by applying a map to its arguments
    
     A decorating class which calls specified transformations on a function's
```

dschult (commented):
This looks great... none of these comments are big deals either way.
One other comment I didn't find a place for here is:
- should we add `argmap` to the decorator section of the utils docs?  If so an entry in the decorator section of `doc/reference/utils.rst` should be included.

Thanks!

boothby:
Thanks Dan!  My next steps on this PR are (1) proper formatting for the docstrings, then (2) tests.  I appreciate the head-start :)

networkx/algorithms/smallworld.py
```diff
@@ -160,11 +157,10 @@ def lattice_reference(G, niter=1, D=None, connectivity=True, seed=None):
     import numpy as np
     from networkx.utils import cumulative_distribution, discrete_sequence
 
+    print(lattice_reference._code)
```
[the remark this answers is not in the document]
boothby: lol, nope.

networkx/utils/tests/test_decorators.py
```diff
@@ -18,10 +18,36 @@
...
+    with pytest.raises(nx.NetworkXNotImplemented):
+        test_m(nx.MultiGraph())
+
+    @not_implemented_for("graph")
+    def test_g(G):
+        pass
+
+    test_g(nx.MultiGraph())
+    with pytest.raises(nx.NetworkXNotImplemented):
+        test_g(nx.Graph())
```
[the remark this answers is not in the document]
boothby: Oh, great catch!  Apparently, I broke the "AND" behavior.  Fixed!

networkx/utils/decorators.py
```diff
@@ -53,51 +57,40 @@ def sp_function(G):
...
+        assert (
+            "graph" not in graph_types
+        ), "Function not implemented on graph AND multigraphs?"
+    if not set(graph_types) < {"directed", "undirected", "multigraph", "graph"}:
+        raise KeyError(
+            "use one or more of "
+            f"directed, undirected, multigraph, graph, {graph_types}"
+        )
+
+    dval = ("directed" in graph_types) or not ("undirected" in graph_types) and None
```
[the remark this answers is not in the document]
boothby: Yep, you got it... and yup, that needs a comment.

boothby:
@dschult I added `argmap` to the docs as you suggested

networkx/utils/tests/test_decorators.py
```diff
@@ -298,3 +356,137 @@ def make_random_state(rs):
...
+        def contextmanager(x):
+            nonlocal container
+            return x, lambda: container.append(x)
+
+        @argmap.try_finally(contextmanager, 0, 1, 2)
+        def do_not_much(x, y, z):
+            return x, y, z
+
+        x, y, z = do_not_much("a", "b", "c")
+        print(do_not_much._code)
```
dschult, line 397: Spurious print statement of `_code`

networkx/utils/decorators.py
```diff
@@ -450,22 +361,536 @@ def random_array(dims, random_state=1):
...
+        real_func = func.__argmap__.compile(func.__wrapped__)
+        func.__code__ = real_func.__code__
+        func.__globals__.update(real_func.__globals__)
+        func.__dict__.update(real_func.__dict__)
+        return func
+
+    def __call__(self, f):
+        """Construct a lazily decorated wrapper of f.  The decorated function
+        will be compiled when it is called for the first time, and it will
+        replace its own __code__ object so subsequent calls are fast.
```
dschult, line 514:
```suggestion
        """Construct a lazily decorated wrapper of f.  
        
        The decorated function will be compiled when it is called for the first time,
        and it will replace its own __code__ object so subsequent calls are fast.
```

networkx/utils/decorators.py
```diff
@@ -450,22 +361,536 @@ def random_array(dims, random_state=1):
...
+            The mangled version of f.__name__ (if f.__name__ exists) or f
+
+        """
+        f = f.__name__ if hasattr(f, "__name__") else f
+        fname = re.sub(cls._bad_chars, "_", f)
+        return f"argmap_{fname}_{cls._count()}"
+
+    def compile(self, f):
+        """Called once for a given decorated function -- collects the code from all
+        argmap decorators in the stack, and compiles the decorated function.
```
dschult, line 604:
```suggestion
        """Compile the decorated function.
        
        Called once for a given decorated function -- collects the code from all
        argmap decorators in the stack, and compiles the decorated function.
```

networkx/utils/decorators.py
```diff
@@ -450,22 +361,536 @@ def random_array(dims, random_state=1):
...
+        compiled = compile(code, filename, "exec")
+        exec(compiled, globl, locl)
+        func = locl[sig.name]
+        func._code = code
+        return func
+
+    def assemble(self, f):
+        """Collects the requisite data to compile the decorated version of f.
+        Note, this is recursive, and all argmap-decorators will be flattened
+        into a single function call
```
dschult, line 638:
```suggestion
        """Collects the requisite data to compile the decorated version of f.

        Note, this is recursive, and all argmap-decorators will be flattened
        into a single function call
```

networkx/utils/decorators.py
```diff
@@ -450,22 +361,536 @@ def random_array(dims, random_state=1):
...
+            mapblock.extend(
+                f"{name} = {fname}({name})" for name in map(get_name, self._args)
+            )
+
+        return sig, wrapped_name, functions, mapblock, finallys, mutable_args
+
+    @classmethod
+    def signature(cls, f):
+        """Compute a Signature so that we can write a function wrapping f with
+        the same signature and call-type.
```
dschult, line 746:
```suggestion
        """Compute the signature for f
        
        Compute a Signature so that we can write a function wrapping f with
        the same signature and call-type.
```

dschult (approved):
My comments are definitely nits.  Take them or leave them.
But I think this is ready to be merged -- and we should do so soon so we can do any needed clean-up before releasing v2.6.

I think t his is the best decorator solution of the 3 I've seen discussed:  
Kelly's two ideas and the decorator code itself.

networkx/utils/decorators.py
```diff
@@ -450,22 +361,536 @@ def random_array(dims, random_state=1):
...
+        -------
+        func : callable
+            The same function, with a new __code__ object.
+
+        """
+        real_func = func.__argmap__.compile(func.__wrapped__)
+        func.__code__ = real_func.__code__
+        func.__globals__.update(real_func.__globals__)
+        func.__dict__.update(real_func.__dict__)
+        return func
```
rossbar, line 509: I have to admit I have very little experience with using/modifying these components of the object model directly.

Does the compilation have a significant impact on performance? Do we have a sense of what the impact of `_lazy_compile` is on a per-call basis vs. call number? My thinking here is if it's small (say O(10's of microseconds), arbitrarily) then maybe it's not worth the extra complexity.

rossbar (commented):
I have to say that my review here is of very limited value, mostly because I struggle with penetrating the control flow within the `argmap` class for various use-cases. The *results* are good though!

My comments are all nits as well - though I did have a question about the `_lazy_compile`, mostly just to see if there was anything quantitative there that we could include in the discussion for the sake of future reference.

networkx/utils/decorators.py
```diff
@@ -53,51 +57,42 @@ def sp_function(G):
...
-
-    return bz2.BZ2File(path, mode=mode)
+    if "directed" in graph_types:
+        assert (
+            "undirected" not in graph_types
+        ), "Function not implemented on graph AND multigraphs?"
+    if "multigraph" in graph_types:
+        assert (
+            "graph" not in graph_types
+        ), "Function not implemented on graph AND multigraphs?"
```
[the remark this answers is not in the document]
boothby: I have a slight preference for `assert` in this case, but I'll change them to `raise`.  The reasoning is that if these asserts pass under testing, then they'll pass at runtime and it's safe to disable them.  But, the reason I'm willing to change is that we aren't guaranteed that the public won't use these decorators themselves.

And yeah, that's the wrong message.

networkx/utils/decorators.py
boothby: The issue highlighted in #4732 is that calling `compile` upon decoration significantly increases the import time.  The lazy strategy moves that cost to the first call of the decorated function.  Monkeypatching like this is not strictly necessary, and only serves to reduce the call-stack depth.  I'd be willing to back off on that if folks are uncomfortable with this.

networkx/utils/tests/test_decorators.py
boothby: definitely not a nit :)

networkx/utils/decorators.py
boothby: @dschult a second opinion, please?

rossbar: Ah - right I knew that conversation was somewhere - thanks for the link to the issue. All I was really looking for was the analysis that's already in that discussion :)

I'm happy with it as-is, I had been worried about other what happens when accessing/modifying these components of the object model for different Python implementations, but the PyPy tests in CI don't seem to mind so that significantly alleviates my concerns.

dschult: I like the current implementation. Monkeypatching is always a little strange, but I think in this case the "lazy" really pays off for most people because it is rare that you use most of the functions in networkx and this cuts out all those compiles for the functions a user doesn't use. The first time compile is "paying the piper" but not with any significant interest in that only the code to set up the monkeypatch and do the compiling is the extra cost of delaying compiling.  After that it shouldn't make any difference (unless the monkeypatch call is slower than the automated lookup which I haven't checked, but I think is not slower).
:)

dschult:
In reading through this, I saw that one of the nice additions this allows is using a string to indicate which argument to map.  I know that our current codebase is all written with an index (number) to indicate which argument to map, but it "feels right" to use the name instead.  In fact, if it simplifies the interface I would be in favor of requiring people to use the name rather than the index of the argument.

@boothby, what do you think about restricting in the interface to using a string instead of "string or index"?

I would be willing to put in some cycles into this. Based on what looks like a 1-2 week delay in 2.6, maybe we could make a push to get this in before 2.6 so users could be encouraged to "just use `pip install networkx--no-deps`" when they want to avoid numpy.

networkx/utils/decorators.py
```diff
@@ -454,22 +365,539 @@ def random_array(dims, random_state=1):
...
+                finally:
+                    close()
+
+        """
+        dec = cls(func, *args)
+        dec._finally = True
+        return dec
+
+    @staticmethod
+    def _lazy_compile(func):
```
boothby, line 490: Justify the performance tradeoff, reference discussion in #4732

networkx/utils/decorators.py
```diff
@@ -454,22 +365,539 @@ def random_array(dims, random_state=1):
...
+            _return = "return"
+
+        call_sig = f"{_return} {{}}({', '.join(call_sig)})"
+
+        return cls.Signature(fname, sig, def_sig, call_sig, names, npos, args, kwargs)
+
+    Signature = collections.namedtuple(
+        "Signature",
+        [
+            "name",
```
boothby, line 826: document the purpose of these fields

networkx/utils/decorators.py
```diff
@@ -454,22 +365,539 @@ def random_array(dims, random_state=1):
...
+
+    is equivalent to
+
+        def foo(file):
+            file, close_file = open_file(file)
+            try:
+                print(file.read())
+            finally:
+                close_file()
+
```
boothby, line 439: add a historical note about why this class exists, and see-also the decorators which are built on it

networkx/utils/decorators.py
```diff
@@ -454,22 +365,539 @@ def random_array(dims, random_state=1):
...
+    """
+
+    def __init__(self, func, *args):
+        self._func = func
+        self._args = args
+        self._finally = False
+
+    @classmethod
+    def try_finally(cls, func, *args):
+        """Alternative decorator-constructor which first transforms one or more
```
boothby, line 449: see-also `open_file`

networkx/utils/decorators.py
```diff
@@ -454,22 +365,539 @@ def random_array(dims, random_state=1):
...
+            fname, _ = functions[id(self._func)]
+        else:
+            fname, _ = functions[id(self._func)] = self._name(self._func), self._func
+
+        # this is a bit complicated -- we can call functions with a variety of
+        # arguments, so long as their input and output are tuples with the same
+        # structure
+        applied = set()
+
+        def get_name(arg, first=True):
```
boothby, line 710: reference the `is_partition` decorator which forced the existence of this feature

networkx/utils/decorators.py
```diff
@@ -454,22 +365,539 @@ def random_array(dims, random_state=1):
...
+                    )
+                return f"{sig.kwargs}[{arg!r}]"
+            else:
+                if sig.args is None:
+                    raise nx.NetworkXError(
+                        f"index {arg} not a parameter index and this function doesn't have args"
+                    )
+                mutable_args = True
+                return f"{sig.args}[{arg-sig.n_positional}]"
+
```
boothby, line 733: reference the `open_file` decorator which forced the existence of this feature

boothby (commented):
Comments mostly by @stefanv -- please add more requests if you like.  @dschult might take a stab at some of this?

dschult:
Some questions have come upon while enhancing docstrings:

1) The `n_positional` and `mutable_args` variables are effective only for very strange cases. For example, `@argmap(func, ("A", ("B", "C")))`. Or for using an index that only works when the var_positional argument is provided and long enough to provide that size index. Do we need this? Or is it just for completeness?  Could we simplify significantly by removing that feature and only providing at most a tuple of arguments (I think that is all that is needed for `G, partition`)?

See what you think of my attempt to describe the generality of argument specifying syntax. (line 485 or so -- search for "overflow").

2) The `argmap.try_finally` method seems like it is only used to construct decorators in utils/misc.py.  If I understand correctly it is actually only used in the `open_file` decorator.  Would removing it simplify the interface to make it easier for future developer to understand?  We would replace:  `return argmap.try_finally(_openfile, path_arg)` with
```python
    result = argmap(_open_file, path_arg)
    result._finally=True
    return result
```
I'm thinking that makes the description of this feature much simpler:  to wrap the decorated function inside a try/finally block, set the private attribute argmap._finally to True.  
We can then remove a lot of description about an alternate way to instantiate an argmap object.

boothby:
1. Yes, those features are a little complex, because I wanted to future-proof the code as much as possible.  One point of discomfort for me is that when errors occur in the `argmap`-generated code, they're pretty hard to track down, so I've gone through some pains to cover the bases.   The specific failure mode I have in mind is if a networkx developer introduces their own decorator that does clobber the function signature -- without making `args` mutable, this would blow up (note to self, this should be a unit test):

```python
def print_graph(f):
    def decorated(G, *args, **kwargs):
        print(G)
        return f(G, *args, **kwargs)
    return decorated

@open_file(1)
@print_graph
def do_something(G, file):
    pass
```

And yes, I do like the `overflow` example.

2. I dislike the notion of setting a private field externally like that.  I recognize that it's all encapsulated in the `decorators` file, but if `argmap` ends up getting used to do more, then it goes from a whiff to a stench pretty quickly.  IMO, a more palatable alternative would be to accept `try_finally` as a keyword-only argument to `__init__`.

3. Yeah, I can go either way on `argmap` showing up in the docs.  On one hand, it's really only meant for networkx developers, on the other hand, I use the online docs all the time.

dschult (approved):
I think @boothby 's changes now address the comments we have received about this PR. I hope my doc_string expansion attempts are 1) correct and 2) somewhat clear. 

I approve this PR and think it is ready to merge. Please take a look if you can. We've got two dev's but this is pretty complicated so any suggestions are helpful before we merge.

boothby:
Thanks for the docs, @dschult.  I believe that they're both correct and clear.  I've got a couple quibbles; I'll try and get to those tonight.

stefanv:
Thanks @boothby @dschult!  Sorry for being slow to review; I plan to get to that tonight.

networkx/utils/decorators.py
```diff
@@ -911,10 +909,9 @@ def compile(self, f):
         return func
 
     def assemble(self, f):
-        """Collects the requisite info to compile the decorated version of f.
-
-        Note, this is recursive, and all nested argmap-decorators for a
-        specific decorated function will be flattened into a single function.
+        """Collects components of the source for the decorated function wrapping f.
+        If f is argmap-decorated, then we recursively assemble the stack of
+        decorators into a single flattened function.
```
dschult, line 914: I think sphinx needs a blank line after the initial one-line description (used for documentation links *to* this method).  Hence the suggestion of a blank line.

The recursion occurs only for nested-argmap-decoration, correct?
```suggestion

        If f has multiple argmap decorators, we recursively assemble the stack of
        decorators into a single flattened function.
```

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+    func : callable
+        The function to apply to arguments
+
+    *params: iterable of (int, str or tuple)
+        A list of parameters, specified either as strings (their names), ints
+        (numerical indices) or arbitrarily nested tuples thereof.
+
+    Examples
+    --------
+    Let's start with a caveat. Most of these examples will use `@argmap(...)`
```
stefanv, line 438: Maybe not a caveat, as much as it is just a different way to use a decorator?

stefanv (commented):
I started the review; more soon!

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+    or by index, specifying the item in the args list.
+
+    Parameters
+    ----------
+    func : callable
+        The function to apply to arguments
+
+    *params: iterable of (int, str or tuple)
+        A list of parameters, specified either as strings (their names), ints
+        (numerical indices) or arbitrarily nested tuples thereof.
```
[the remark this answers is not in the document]
boothby: I had to make a choice: handle arbitrarily-nested tuples once and cleanly, only handle tuples once and break compatibility with `try_finally`, or handle tuples in multiple places in the block below this.

Right now, we only use 2-element tuples, but the documentation says what the code does.  There's an example in the class docstring.

networkx/utils/decorators.py
```diff
@@ -122,140 +123,93 @@ def open_file(path_arg, mode="r"):
...
 
-        return result
+        fobj = _dispatch_dict[ext](path, mode=mode)
+        return fobj, lambda: fobj.close()
 
-    return _open_file
+    return argmap(_open_file, path_arg, try_finally=True)
 
 
 def nodes_or_number(which_args):
```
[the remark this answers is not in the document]
boothby: I'm not sure what you mean by this, can you clarify?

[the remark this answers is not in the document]
dschult: We don't allow specifying a range of integers. Only the number of integers.  That is, the input `path_graph(5)` is allowed, but not `path_graph(2, 7)` the way you might expect it to work if it was called range in the name.  I guess the name `nodes_or_number_of_nodes` is most descriptive, but too long. :}

networkx/utils/decorators.py
```diff
@@ -267,48 +221,53 @@ def nodes_or_number(which_args):
     --------
     Decorate functions like this::
 
+       @nodes_or_number("nodes")
```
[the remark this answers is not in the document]
dschult: We don't provide a default for the argument of the decorator.  Better to be explicit I think.... at least more readable...

networkx/utils/decorators.py
```diff
@@ -357,19 +316,19 @@ def wrapper(*args, **kwargs):
...
 
 
-def random_state(random_state_index):
+def random_state(random_state_argument):
     """Decorator to generate a numpy.random.RandomState instance.
 
-    Argument position `random_state_index` is processed by create_random_state.
+    The decorator processes the argument indicated by `random_state_argument`
+    using :func:`nx.utils.create_random_state`.
+    The argument value can be a seed (integer), or a random number generator.
```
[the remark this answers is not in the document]
dschult: In general we support both `random.Random` and `numpy.random.RandomState`, but I think in this case it has to be `numpy.random.RandomState`. Either way -- I agree... we should say what is can be.

networkx/utils/decorators.py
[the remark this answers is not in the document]
boothby: I feel like changing the name of this 6 year-old decorator is out of scope for this issue?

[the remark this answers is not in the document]
stefanv: `nodes_or_number_of_nodes` is indeed a bit long :)

networkx/utils/decorators.py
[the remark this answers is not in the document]
stefanv: I think the main issue I wanted to highlight was that it is not immediately intuitive, based on the text here, what params should be.  I.e., what happens when you specify a nested tuple of ints, strings, parameters, etc.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+        A list of parameters, specified either as strings (their names), ints
+        (numerical indices) or arbitrarily nested tuples thereof.
+
+    Examples
+    --------
+    Let's start with a caveat. Most of these examples will use `@argmap(...)`
+    to apply the decorator immediately to the function defined on the next line.
+    In the NetworkX codebase however, `argmap` is used within a function to
+    construct a decorator. That is, the decorator defines a mapping function
+    and then uses `argmap` to build the decorator from the mapping.
```
stefanv, line 442: I wasn't 100% sure what this meant.  Is this a correct phrasing?

The decorator calls the original function, after transforming the input arguments.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+    A simple example is a decorator that specifies which currency to report money.
+    The decorator (named `convert_to`) would be used like::
+
+        @convert_to("US_Dollars", "income")
+        def show_me_the_money(name, income):
+            print(f"{name} : {income}")
+
+    And the code to create the decorator might be::
+
+        def convert_to(currency, which_arg):
```
stefanv, line 452: This is very helpful in understanding how to build an argmap!

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+                    amount = amount.to_currency(currency)
+                return amount
+            return argmap(_convert, which_arg)
+
+    Despite this common idiom for argmap, most of the following examples
+    use the `@argmap(...)` idiom to save space here.
+
+    As an example, the decorated function::
+
+        @argmap(sum, "x", "z")
```
stefanv, line 464: This example is a bit confusing, because you initially think it means `sum(x, z)`.  Maybe `square` or something that is more likely to be called on individual arguments separately would be more intuitive?

networkx/utils/decorators.py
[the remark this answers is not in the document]
stefanv: (It could simply be a sentence referring to the examples below.)

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+
+        def foo(a, b, c):
+            a, b = swap(a, b)
+            return a / b * c
+
+    More generally, the applied arguments can be nested tuples of strings or ints.
+    The syntax `@argmap(some_func, ("a", ("b", "c")))` would expect `some_func` to
+    accept 2 inputs with the second expected to be a 2-tuple. It should then return
+    2 outputs with the second a 2-tuple. The returns values would replace input "a"
+    "b" and "c" respectively. Similarly for `@argmap(some_func, (0, ("b", 2)))`.
```
stefanv, line 507: This is great; maybe an example here to illustrate the text?

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+
+    Also, note that an index larger than the number of named parameters is allowed
+    so long as a VAR_POSITIONAL input appears, e.g. `*args`. In that case the index
```
stefanv, line 510: What is VAR_POSITIONAL? Will most readers know?

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+
+    For example this map opens a file and then makes sure it is closed::
+
+        def open_file(fn):
+            f = open(fn)
+            return f, lambda: f.close()
+
+    The decorator applies that to the function `foo`::
+
+        @argmap(open_file, "file", try_finally=True)
```
stefanv, line 539: I'm curious about this choice of API.  I find it confusing that the `open_file` function does not transform arguments, as it does elsewhere with `@argmap`.  I'm wondering about the feasibility of the following:

```
def open_file(fn):
    return open(fn)

@argmap(open_file, "file", try_finally=lambda f: f.close())
```

Also, I'm curious whether this can be handled with something like:

```
def open_file(fn):
    with open(fn) as f:
        yield f
    # here, any other finalization operations needed
```

Then `@argmap(open_file, "file")`.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+            print(file.read())
+
+    Here's an example of the try_finally feature used to create a decorator::
+
+        def my_closing_decorator(which_arg):
+            def _opener(path):
+                if path is None:
+                    path = open(path)
+                    fclose = path.close
+                else:
```
stefanv, line 565: Closing a closed file is OK.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+            # this code doesn't need to worry about closing the file
+            print(file.read())
+
+    Notes
+    -----
+    An object of this class is callable and intended to be used when
+    defining a decorator. Generally, a decorator takes a function as input
+    and constructs a function as output. Specifically, an `argmap` object
+    returns the input function decorated/wrapped so that specified arguments
+    are mapped (transformed) to new values before the decorated function is called.
```
stefanv, line 584: Perfect.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+    As an overview, the argmap object returns a new function with all the
+    dunder values of the original function (like `__doc__`, `__name__`, etc).
+    Code for this decorated function is built based on the original function's
+    signature. It starts by mapping the input arguments to potentially new
+    values. Then it calls the decorated function with these new values in place
+    of the indicated arguments that have been mapped. The return value of the
+    original function is then returned. This new function is the function that
+    is actually called by the user.
+
+    Three additional features are provided.
```
stefanv, line 595: This overview is :+1:

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+          set of finally clauses if needed.
+
+      mutable_args : a bool indicating whether the `sig.args` tuple should be
+          converted to a list so mutation can occur.
+
+    After this recursive assembly process, the `argmap.compile` method
+    constructs code (as strings) to convert the tuple `sig.args` to a list
+    if needed. It joins the defining code with appropriate indents and
+    compiles the result. The compiled code is `exec`uted defining the
+    new function upon which the code string is monkey-patched
```
stefanv, line 674: If the function was just built from scratch, why does it still need to be patched again?  And what type of patching helps with introspection?

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+        Notes
+        -----
+        It was observed in NetworkX issue #4732 [1] that the import time of
+        NetworkX was significantly bloated by the use of decorators: over half
+        of the import time was being spent decorating functions.  This was
+        somewhat improved by a change made to the `decorator` library, at the
+        cost of a relatively heavy-weight call to `inspect.Signature.bind`
+        for each call to the decorated function.
+
+        The workaround we arrived at is to do minimal work at the time of
```
stefanv, line 738: Thanks for this thorough motivation!

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+
+        This method is part of the `compile` method's process yet separated
+        from that method to allow recursive processing. The outputs are
+        strings, dictionaries and lists that collect needed info to
+        flatten any nested argmap-decoration.
+
+        Parameters
+        ----------
+        f : callable
+            The function to be decorated.  If f is argmapped, we assemble it.
```
stefanv, line 924:
```suggestion
            The function to be decorated.  If `f` is argmapped, we assemble it.
```

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+        -------
+        sig : argmap.Signature
+            The function signature as an `argmap.Signature` object.
+        wrapped_name : str
+            The mangled name used to represent the wrapped function in the code
+            being assembled.
+        functions : dict
+            A dictionary mapping id(g) -> (mangled_name(g), g) for functions g
+            referred to in the code being assembled. These need to be present
+            in the `globals` scope of `exec` when defining the decorated function.
```
stefanv, line 936: I think we use single quotes for variables, and double quotes for Python codes (but not 100% sure if that convention is also followed by NetworkX.)

```suggestion
            in the ``globals`` scope of ``exec`` when defining the decorated function.
```

stefanv (commented):
@boothby Thank you for all the documentation you added; incredibly helpful!  It also gives me confidence that this can be maintained.  The structure is now really clear, and the inline comments guides a reader through the steps.

This was a huge amount of work, but the result looks great.  Thank you very much!

networkx/utils/decorators.py
```diff
@@ -122,140 +123,93 @@ def open_file(path_arg, mode="r"):
...
+    Notes
+    -----
+    Note that this decorator solves the problem when a path argument is
+    specified as a string, but it does not handle the situation when the
+    function wants to accept a default of None (and then handle it).
+
+    Here is an example of how to handle this case:
+
+      @open_file("path")
+      def some_function(arg1, arg2, path=None):
```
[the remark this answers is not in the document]
dschult: I think this alternative creates a single tempfile that gets reused rather than constructing a new one for each call of the function.  But I think the example could avoid the `close_fobj` flag by just testing `if path is None:` at the end instead of testing `close_fobj`.  But I'm not it makes much difference -- does the logic flow better?  maybe I little...  I'll make that change to the example.  

Check this to see if it gets better.

dschult:
I push some doc clarifications and marked comments as resolved. There are some comments that I left unresolved -- 
@boothby can you take a look at those (and any of the others that I made more confusing too)

networkx/utils/decorators.py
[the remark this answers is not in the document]
stefanv: This is fine.  Again, I don't think the check is necessary.  The file can simply be closed.  The real logic you want to show is that you get to do something special if the input is None.

I like the `default=` API better, I think, but that's for you and @boothby to decide.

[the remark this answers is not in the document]
dschult: I think we want to leave the file open when it has been opened by the caller. They may not be done with it.
We do want to close it when the decorator opened it.  But we don't want to close everything we use.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
 
 
-def py_random_state(random_state_index):
+def py_random_state(random_state_argument):
     """Decorator to generate a random.Random instance (or equiv).
 
-    Argument position `random_state_index` processed by create_py_random_state.
+    The decorator processes the argument indicated by `random_state_argument`
+    using :func:`nx.utils.create_py_random_state`.
+    The argument value can be a seed (integer), or a random number generator.
```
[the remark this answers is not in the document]
dschult: I added more to maybe make this clearer, but there were 9 cases to cover so it might be too-much-info.

networkx/utils/decorators.py
```diff
@@ -392,84 +355,868 @@ def random_array(dims, random_state=1):
...
+    in many functions to determine whether an argument is the number of nodes
+    to be created, or a list of nodes to be handled. The decorator provides
+    the code to accept either -- transforming the indicated argument into a
+    list of nodes before the actual function is called.
+
+    This decorator class allows us to process single or multiple arguments.
+    Or to process arguments that e.g. might need to be closed (or not) depending
+    on how they are input (used for file handle arguments that may either be
+    a string naming the file or an existing file handle). If the decorator opens
+    the file it should close it when done. Otherwise it should not close the file.
```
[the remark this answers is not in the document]
dschult: I moved this text to a new paragraph that describes the `try_finally` parameter (which we hadn't added to the class docs yet).

networkx/utils/decorators.py
dschult: I renamed "x" and "z" as "xlist" and "zlist" to make this clearer, but I would be fine with leaving as "x" and "z" and switching the function. The nice thing about `sum` is that it doesn't need to be defined.

networkx/utils/decorators.py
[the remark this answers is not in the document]
stefanv: All our decorator does is close it, but I guess as a "general principle"?

networkx/utils/decorators.py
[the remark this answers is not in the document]
stefanv: Yikes, yes I see!  Now I wonder: how many of those cases do we use and need?

For the current implementation, the docs are good.  But the implementation itself may be too magical.  Maybe that's for another PR.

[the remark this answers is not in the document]
dschult: Yes -- that is probably a good PR for v3.0.  We included all those versions because at the time we had code that used each of those mechanisms. Maybe that has changed...  Our goal was to make it as easy as possible for people to submit code to the project. At the same time we wanted to let users provide a single RNG for everything. So we constructed an interface to make the `random.Random` API available even with a numpy RandomState RNG passed in.

networkx/utils/decorators.py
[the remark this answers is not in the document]
dschult: No -- our `open_file` sometimes returns `lambda: None` for the second return value (the try_finally function).  If path is a string or a pathlib instance we provide the close function.
Otherwise, `path` is probably `None`, or a file handle so we `return path, lambda: None`  (don't close).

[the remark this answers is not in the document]
stefanv: Aha, thanks, I didn't realize that!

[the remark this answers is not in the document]
stefanv: The more I think about it, I think this decorator should go.

Having a `with nx_open(x)` is as short, less confusing, and more explicit.

networkx/utils/decorators.py
boothby: That's elaborated at the end of the sentence: "e.g., `*args`".  Is that too unclear?  Can you think of a better way to phrase this?  I read a ton of documentation, tutorials, etc. about the `*args` and `**kwargs` syntax in function calls & definitions, and there doesn't seem to be a consistent way of talking about those argument containers.

networkx/utils/decorators.py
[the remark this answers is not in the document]
dschult: But doesn’t that close the file handle?  (…or rather, what is the way to avoid closing it while using “with”?)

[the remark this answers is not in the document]
boothby: If you want to deprecate that function, please bring it up in another issue.  We'll need to retain this functionality until at least 3.0, so I won't be able to simplify `argmap` until then.

networkx/utils/decorators.py
boothby: The function shell was constructed in `__call__`.  We replace its `__code__` object with the one produced in `__compile__`.

networkx/utils/decorators.py
stefanv: I just had never heard the word VAR_POSITIONAL, and presume others wouldn't have had either.  Are there others than `*args`?

I think these are officially called "variadic arguments" or variable length arguments.

networkx/utils/decorators.py
stefanv: That is nice and explicit!  I prefer that to the more vague "monkey-patched".

networkx/utils/decorators.py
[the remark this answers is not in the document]
stefanv: I meant the standard straightforward way, without any decoration:

```
with nx_open(filename):
    ... do the work ...
```

And `nx_open` can open a file, create temporaries, whatever.

@boothby No problem holding off on a refactor like that until 3.0, if everyone agrees it's a good step.

stefanv (approved):

networkx/utils/decorators.py
boothby: Previously, I used an alternative constructor for the class.  Dan didn't like that so I moved it to a keyword argument.  Your proposal here doesn't capture whether or not the file needs to be closed, and may close a filehandle that the user wants to keep open.

stefanv: That's right.  My later suggestion was to simply use a context manager that does the opening, or that creates a temporary file (or whatever is needed).

dschult: The point of the `try_finally` logic is to close the file when it is appropriate. Opening the file is not the issue.

Also, see #2298 which discusses removing `open_file`.  
But that’s a separate issue from this PR I think…  The discussion there left off with: open_file is too complicated and confusing, but we still want to keep strings (and pathlib.Path objects) as valid ways to specify what file to use…
The tricky part is making sure we close files that we create and leave open files the user creates.

stefanv: Still, this should be doable with a custom context manager.  You can, e.g., return a NetworkXFile object that has a .close method that gets called when exiting the context.  The close method can be customized to do whatever (leave the file alone, close files open by nx, etc.).  And agreed that the issue is not relevant to this specific PR.

boothby: Please move this conversation/duplicate your suggestions to #2298 as it will be easier to follow & preserve there.

boothby: Oh, but I should address one bit I missed:

> I'm curious about this choice of API. I find it confusing that the open_file function does not transform arguments

It _does_ transform arguments.  It takes a filename argument, and transforms it to an open file (or does nothing to filehandles).

jarrodmillman (commented):
LGTM.  A few items that should be updated (I would prefer in this PR, but I can do it in a separate PR as well).

1. In `.github/workflows/test.yml`, change
```
pip install decorator . --no-deps
```
to
```
pip install . --no-deps
```

2. In `INSTALL.rst` change
```
$ pip install decorator networkx --no-deps
```
to
```
$ pip install networkx --no-deps
```

dschult:
OK.. I removed decorator from github workflows and install instructions.
:)

jarrodmillman (approved):

dschult:
Thanks @boothby !!

stefanv:
Great work @boothby @dschult — thank you!

PR #4741, opened 2021-04-13
Use pydata sphinx theme

jarrodmillman:
NumPy, SciPy, and other projects are moving (or have moved) to using the [pydata sphinx theme](https://github.com/pydata/pydata-sphinx-theme) for documentation.  I believe it is nicer looking and, more importantly, will help us create a more unified scientific Python ecosystem.  We also know the maintainers, so we can easily help fix and improve the theme.

I plan to clean up a few things, but I wanted to get some feedback before putting too much more time in to this.  We could merge this as is and I could do the clean up in a separate PR or I could make the changes to this PR.  I slightly prefer merging this first, but I don't have a strong preference.


stefanv:
I'm happy to see this; I'd like to do the same for scikit-image.  I like the idea that we can all work together on one theme, instead of everyone struggling separately to figure out Sphinx's arcane CSS style structure.

PR #4742, opened 2021-04-14
WIP: layered layout drawing algorithm

Adrien-Luxey:
This is still work in progress, but working to the point where I'd like NetworkX community's reviews.

I want to focus on solving glitches for some edge cases (edges drawn awkward when they could be straight, too big spacing between groups of nodes...). If someone has time, notably to fix the display of long edges (see below), I'd be glad :)

* Is anyone knowledgeable on `FancyArrowPatch`? Using `path` parameter hides the arrow for long edges, probably because `FancyArrowPatch` discards `shrinkA` and `shrinkB` parameters when `path` is given. See `ny_pylab.py:794`.
* `import numpy as np` appeared 17 times in `layout.py`. I made it a global import in this file.
* Why keep the `dim` parameter when most layouts don't support `dim != 2`? A lot of layout functions do that.
* Tests still need to be written. 

Hail NetworkX! :D

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -15,6 +15,7 @@ For the other layout routines, the extent is
 Warning: Most layout routines have only been tested in 2-dimensions.
 
 """
+import numpy as np
 import networkx as nx
 from networkx.utils import random_state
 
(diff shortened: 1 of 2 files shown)


Adrien-Luxey:
Worry, I'm resubmitting with commits squashed.

PR #4757, opened 2021-04-25
[Feature] Topological Generations

ArtinSarraf:
Please see the following NetworkX discussion for details:
https://github.com/networkx/networkx/discussions/4755

In short:  an algorithm, mainly based on the existing topological sort, to break a DAG up into layers, such that all dependencies of a node in a layer appear in a layer before it.

```
# Imagine all arrows pointing downwards in this DAG

  F   G
  |\ /|
D | E |
 \|/  |
  B   C
   \ /
    A

Result:
[[D, F, G],
 [C, E],
 [B],
 [A]]
```


networkx/algorithms/dag.py
```diff
@@ -170,10 +180,12 @@ def topological_sort(G):
 
     indegree_map = {v: d for v, d in G.in_degree() if d > 0}
     # These nodes have zero indegree and ready to be returned.
-    zero_indegree = [v for v, d in G.in_degree() if d == 0]
+    zero_indegree = deque(v for v, d in G.in_degree() if d == 0)
 
+    generation = 0
+    last_node_in_generation = zero_indegree[-1]
```
rossbar, line 186: I believe this line is responsible for the test failures - this will raise `IndexError`s if `zero_indegree` is empty.

ArtinSarraf: Thanks for following up @rossbar. Updated. If the zero-indegree is empty I decided to have it raise a separate explicit error from the one at the bottom, since we know that this is specifically due to a cycle in the graph, and not from the graph changing during iteration.

networkx/algorithms/dag.py
```diff
@@ -170,10 +180,16 @@ def topological_sort(G):
...
     indegree_map = {v: d for v, d in G.in_degree() if d > 0}
     # These nodes have zero indegree and ready to be returned.
-    zero_indegree = [v for v, d in G.in_degree() if d == 0]
+    zero_indegree = deque(v for v, d in G.in_degree() if d == 0)
+
+    generation = 0
+    if zero_indegree:
+        last_node_in_generation = zero_indegree[-1]
+    else:
+        raise nx.NetworkXUnfeasible("Graph contains a cycle.")
```
rossbar, line 189: Is this true? This seems to be a departure from the original implementation, which only raises `if indegree_map`.

rossbar (commented):
The new changes are resulting in different test failures - from a preliminary glance it looks like there is a change in behavior surrounding when a graph contains a cycle.

networkx/algorithms/dag.py
ArtinSarraf: I believe before the first yield, if there are no zero-indegrees, this can only happen if there is a cycle in the graph. 

```
Note: the above statement is not the only way a cycle can exist, e.g.

B > A
C > A
A > B

This will have a cycle, but still contain a zero-indegree
```
Despite the above, if an error can be confidently raised concerning the infeasibility of a dag in a specific condition, I think it would make sense to do it asap. Especially since identifying that case doesn't add any extra time/space complexity to the algorithm, and reduces unnecessary processing of the dag.

ArtinSarraf: 👍  I'll take a look over the weekend.

ArtinSarraf: so that's what all those tests are for! 😃

ArtinSarraf: Updated. Passing all tests run with `pytest networkx/`. Also added a few tests for the new function/topo-sort updates.

networkx/algorithms/dag.py
```diff
@@ -170,10 +180,13 @@ def topological_sort(G):
 
     indegree_map = {v: d for v, d in G.in_degree() if d > 0}
     # These nodes have zero indegree and ready to be returned.
-    zero_indegree = [v for v, d in G.in_degree() if d == 0]
+    zero_indegree = deque(v for v, d in G.in_degree() if d == 0)
+
+    generation = 0
+    last_node_in_generation = zero_indegree[-1] if zero_indegree else None
```
ArtinSarraf, line 186: setting last_node_in_generation to None could be an issue if None was a valid node. However, this led me to discover another existing bug, which makes this decision irrelevant.

None if a valid node allowed in a graph. However if you try to topo_sort a graph with None in it (on main branch) it will fail, because when it looks for edges of node=None, it will instead return all edges in the graph, instead of the None-node. 

I think 1 of the following actions should probably be taken:
1) None shouldn't be allowed as a node in a graph (seems like overkill potentially)
2) The edges method should do something like this instead of current. With this suggestion None can be used as a regular node.
```
__empty = object()
def edges(self, nbunch=__empty, data=False, default=None):
    ...
```
3) don't allow topo_sort when None exists as a node.

ArtinSarraf: @rossbar @dschult  ^

dschult: Nodes can be "any hashable except the object `None`"

ArtinSarraf: Makes sense. Has it ever been discussed to explicitly forbid `None`, rather than by convention? Since it seems as of now nothing stops you from using it as a node, and can lead to ambiguous errors in algorithms. In any case this is a discussion for another thread. If you think it's worth pursuing I can open a separate issue for it.

rossbar: +1 for opening a new [discussion topic](https://github.com/networkx/networkx/discussions) so we can discuss this issue separately, otherwise it we'll be juggling too many separate convos in this PR :).

Given that `None` can't be a node in a graph, the way `last_node_in_generation` is handled here seems fine since it's really just a state variable and is never added to a Graph object or returned from the function AFAICT.

ArtinSarraf:
@rossbar any ideas on what's causing this circleCI sphinx error?
image

rossbar:
No worries, it doesn't have anything to do with these changes: it's a documentation dependency issue. It's fixed now, so rebasing on `main` would fix it - feel free to do so or one of the maintainers will :)

jarrodmillman:
The doc build error should be fixed by #4794.

networkx/algorithms/tests/test_dag.py
```diff
@@ -478,6 +478,69 @@ def sorting_key(node):
...
+    )
+    assert result == []
+
+
+def test_topological_sort_with_generation_cycle():
+    G = nx.DiGraph([[2, 1], [3, 1], [1, 2]])
+    pytest.raises(
+        nx.NetworkXUnfeasible,
+        lambda: list(nx.topological_sort(G, with_generation=True)),
+    )
```
rossbar, line 513: Just another nit - I think using `pytest.raises` as a context manager is cleaner and easier to read:

```suggestion
    with pytest.raises(nx.NetworkXUnfeasible):
        nx.topological_sort(G, with_generation=True)
```

rossbar (approved):
This LGTM - I went ahead and reorganized the tests to use a "flatter" organization - the tests themselves look good! 

A couple other minor suggestions but on the whole I think this is a nice improvement.

networkx/algorithms/tests/test_dag.py
rossbar: There are a couple other instances below that should also be updated.

ArtinSarraf:
Thanks @jarrodmillman - merged main into the branch. 
@rossbar - incorporated your changes and updated the tests with the context manager style.

ArtinSarraf:
@rossbar anything else to review/consider?

networkx/algorithms/dag.py
```diff
@@ -113,11 +114,19 @@ def topological_sort(G):
     ----------
     G : NetworkX digraph
         A directed acyclic graph (DAG)
+    with_generation : bool, optional
```
dschult, line 117:
```suggestion
    with_generation : bool, optional (default: False)
```

networkx/algorithms/dag.py
```diff
@@ -113,11 +114,19 @@ def topological_sort(G):
...
+        Whether to also yield the earliest possible generation of the node, such
```
dschult, line 118:
```suggestion
        If True, yield 2-tuples `(node, generation)` where `generation` is an
        integer representing the earliest possible generation of the node such
```

networkx/algorithms/dag.py
```diff
@@ -113,11 +114,19 @@ def topological_sort(G):
...
+        that all ancestors are in a previous generation, and all descendants
+        are in a following generation.
 
     Returns
     -------
-    iterable
-        An iterable of node names in topological sorted order.
+    iterable | iterable[tuple[node, int]]
```
dschult, line 124:
```suggestion
    iterator of either nodes or (node, generation) 2-tuples
```

networkx/algorithms/dag.py
```diff
@@ -113,11 +114,19 @@ def topological_sort(G):
...
+        If with_generation=False: An iterable of node names in topological sorted order.
+
+        If with_generation=True: An iterable of tuples where the first
+        element is the node and the second element is the generation. In
+        topologically sorted order.
```
dschult, line 129:
```suggestion
    iterable
        If with_generation is False, iterates over nodes in topological sorted order.
        If with_generation is True, iterates over (node, generation) 2-tuples with
        node in topologically sorted order and generation indicating the earliest
        possible generation of that node such that all ancestors are in a previous
        generation and all descendants are in a following generation.
```

networkx/algorithms/dag.py
```diff
@@ -170,10 +179,13 @@ def topological_sort(G):
 
     indegree_map = {v: d for v, d in G.in_degree() if d > 0}
     # These nodes have zero indegree and ready to be returned.
-    zero_indegree = [v for v, d in G.in_degree() if d == 0]
+    zero_indegree = deque(v for v, d in G.in_degree() if d == 0)
+
+    generation = 0
+    last_node_in_generation = zero_indegree[-1] if zero_indegree else None
```
dschult, line 185:
```suggestion
    zero_indegree = [v for v, d in G.in_degree() if d == 0]

    generation = -1
```

networkx/algorithms/dag.py
```diff
@@ -170,10 +179,13 @@ def topological_sort(G):
...
 
     while zero_indegree:
-        node = zero_indegree.pop()
+        node = zero_indegree.popleft()
         if node not in G:
             raise RuntimeError("Graph changed during iteration")
         for _, child in G.edges(node):
```
dschult, line 191:
```suggestion
      generation += 1
      this_generation = zero_indegree
      zero_indegree = []
      for node in this_generation:
        if node not in G:
            raise RuntimeError("Graph changed during iteration")
        for child in G.neighbors(node):
```

networkx/algorithms/dag.py
```diff
@@ -185,7 +197,11 @@ def topological_sort(G):
                 zero_indegree.append(child)
                 del indegree_map[child]
 
-        yield node
+        yield (node, generation) if with_generation else node
+
+        if zero_indegree and node == last_node_in_generation:
+            generation += 1
+            last_node_in_generation = zero_indegree[-1]
```
dschult, line 204:
```suggestion

```

networkx/algorithms/dag.py
ArtinSarraf: So I actually like this approach, but this then deviates from the referenced algorithm described in the docstring:
```
    .. [1] Manber, U. (1989).       *Introduction to Algorithms - A Creative Approach.* Addison-Wesley.
```
Screenshot from the book below.
image

Are we ok deviating from this?

dschult: Does it deviate from the referenced algorithm?
I am thinking that this just changes the type of queue we are using (to a sequence of lists).
That is "put w in Queue" is changed to use a sequence of lists to provide the queue. Then "remove vertex v from Queue" removes the nodes from consideration one by one off the queue (doesn't remove them from the data structure until all nodes from that part of the queue are processed, but it does remove them from consideration).

As I was thinking about the bigger structure of these two functions last night, I notice that your new function recreates each of these lists. Would it be more effective to rearrange the computation to have the second (new) function generate the sets of nodes for each generation and then have the first function use that and then yield each of the nodes? Thoughts?

ArtinSarraf: The rearranged pattern you describe is actually the way I've done this in the past and what I would prefer too(the building the generations after the fact was solely to not disturb the original implementation, since I wasn't sure it would be appropriate for me to change a fundamental function that's been around for several years). If we're ok with updating the topological_sort, then I would prefer the way you described here: So would essentially look like this:

```
def topological_generations(G):
    indegree_map = {v: d for v, d in G.in_degree() if d > 0}
    zero_indegree = [v for v, d in G.in_degree() if d == 0]

    generation = -1

    while zero_indegree:
        generation += 1
        this_generation = zero_indegree
        zero_indegree = []
        for node in this_generation:
            if node not in G:
                raise RuntimeError("Graph changed during iteration")
            for child in G.neighbors(node):
                try:
                    indegree_map[child] -= 1
                except KeyError as e:
                    raise RuntimeError("Graph changed during iteration") from e
                if indegree_map[child] == 0:
                    zero_indegree.append(child)
                    del indegree_map[child]
        yield this_generation


def topological_sort(G):
    for generation in nx.topological_generations(G):
        for node in generation:
            yield node

```

dschult: Yes -- That makes more sense to me than the first implementation. I'm glad your instincts are similar -- that makes it more likely to be a good way to design it. :}

we don't even need the two lines of `generation = -1` and `generation += 1`, because the user can get the generation number from the iterator with `enumerate(topological_generations)`. 

I think this is also significantly faster than the previous implementation (3-4x) -- and I think it leaves the API the same for the original function.
:}

ArtinSarraf: Cool, pushed the changes to the branch. All tests passing. Needed to change G.neighbors back to G.edges, since it was failing in cases of multiple edges between the same node.

networkx/algorithms/dag.py
```diff
@@ -102,6 +102,81 @@ def is_directed_acyclic_graph(G):
...
+    indegree_map = {v: d for v, d in G.in_degree() if d > 0}
+    zero_indegree = {v for v, d in G.in_degree() if d == 0}
+
+    while zero_indegree:
+        this_generation = zero_indegree
+        zero_indegree = set()
+        for node in this_generation:
+            if node not in G:
+                raise RuntimeError("Graph changed during iteration")
+            for _, child in G.edges(node):
```
dschult, line 164:
```suggestion
            for child in G.neighbors(node):
```
The neighbors function is the right way to find the other end of the edges from a given node. I know it wasn't that way in the old code, but it is the better way to do it and it helps performance quite a bit too.

networkx/algorithms/dag.py
```diff
@@ -174,39 +249,9 @@ def topological_sort(G, with_generation=False):
...
-            generation += 1
-            last_node_in_generation = zero_indegree[-1]
-
-    if indegree_map:
-        raise nx.NetworkXUnfeasible(
-            "Graph contains a cycle or graph changed " "during iteration"
-        )
+    for generation in nx.topological_generations(G):
+        for node in generation:
+            yield node
```
dschult, line 254:
```suggestion
        yield from generation
```
I **think** this is a case where `yield from` makes sure the exception handling etc are done well.
Even if it isn't, it is probably good practice to use `yield from` in cases like this where you are just yielding from a generator.

networkx/algorithms/dag.py
dschult: I see now your comment that this doesn't work with multigraph edges...  e.g. it doesn't repeat the same node twice, so the degree is not reduced by the right amount. I think we could fix this, but it is not the point of this PR. So leaving it as is should be fine.  Sorry for the noise...

dschult:
I'll try to get this to pass the tests again.... 
I think the version of black was increased so we can prepare for the v2.6 release and that is causing this error.
I'll go ahead and change to the `yield from generation` idiom too.  If you can indicate somehow that you are OK with this @ArtinSarraf then we can get @rossbar to look one more time before merging.

dschult (approved):
I think this is ready to merge.
@rossbar can you take a re-look and merge if OK?

ArtinSarraf:
@dschult awesome thanks for getting that working, lgtm!

dschult:
Thanks for the `Returns` -> `Yields` 
I couldn't remember that option... :}

And +1 for getting rid of the `" "` that black leaves in the strings.

rossbar (approved):
This is very nice - I was hoping there was a solution with a smaller API surface (i.e. either a new function *or* an addition to the topological_sort signature, but preferably not both) but I failed to see a better approach.

I will merge when CI completes since I made a couple minor changes. Thanks @ArtinSarraf & @dschult !

ArtinSarraf:
👍 Thanks for your the time and effort @rossbar and @dschult

PR #4768, opened 2021-04-29
Improve reproducibilty of geometric graphs

### LINKED ISSUE #4767: Unable to get the same network from `soft_random_geometric_graph` on different machines
I'm unable, with version 2.5 of networkx, to get the same graph from `soft_random_geometric_graph` (with same parameters of course, and same seed of random number generator)

A cause of the problem is the fact that `scipy.spatial.cKDTree` doesn't return the edges in the same order. A simple pass of "sorted(...)" solves the probelm

### Current Behavior
The generated graph has different edges on different machines, even with the same seed

### Expected Behavior
The generated graph should have the same edges, given the same seed

### Steps to Reproduce
```
import numpy as np
import networkx as nx

rng = np.random.RandomState()
rng.seed(4)
N =100
scale = 2
x_pos = np.sqrt(N)*rng.random(N)
y_pos = np.sqrt(N)*rng.random(N)
# for soft geometric graph generation
pos  = {i: (x, y) for i, (x, y) in enumerate(zip(x_pos,y_pos))}

rng = np.random.RandomState()
rng.seed(2)
G = nx.generators.soft_random_geometric_graph(N,4, pos=pos, seed=rng )
```
This snippet is sufficient to generate graphs with different edges on different machines, provided that scipy is installed

### Environment

Python version: 3.8, 3.7
NetworkX version: 2.5

fabmazz:
Since on different versions of python `sp.spatial.cKDTree` gives the edge indices in different order, it becomes impossible to reproduce exactly the same graph, even when passing the same random number generator state.

This PR solves the problem by sorting the`edge_indexes` after getting the result from cKDTree.

Closes #4767

### DIFF
--- a/networkx/generators/geometric.py
+++ b/networkx/generators/geometric.py
@@ -52,7 +52,7 @@ def geometric_edges(G, radius, p):
     nodes, coords = list(zip(*nodes_pos))
     kdtree = sp.spatial.cKDTree(coords)  # Cannot provide generator.
     edge_indexes = kdtree.query_pairs(radius, p)
-    edges = [(nodes[u], nodes[v]) for u, v in edge_indexes]
+    edges = [(nodes[u], nodes[v]) for u, v in sorted(edge_indexes)]
     return edges
 
 


rossbar (approved):
LGTM, thanks @fabmazz and nice work chasing this down!

I suppose there's the potential for a performance hit here, but IMO it's worth it for better reproducibility.

MridulS (approved):

PR #4790, opened 2021-05-11
imread(url) is deprecated, use pillow + urllib to load image from URL

MridulS:


bkbncn:
I like this idea, but I have to say this code doesn't work.  The error message is `AttributeError: module 'urllib' has no attribute 'requrest'`.

Fix above at #4793.

PR #4794, opened 2021-05-12
Update documentation dependencies

jarrodmillman:


jarrodmillman:
The doc build is currently broken.  This updates the doc dependencies, fixes the build system, and speeds up the doc build.  Since the main branch is broken, I am self merging this.

I will be doing more doc work, so if there is a problem arises w/ this I will fix it later.

Issue #4322, opened 2020-11-06
New algorithm for ranking spanning arborescences


edxu96:
I'd like to contribute by implementing the first algorithm for ranking spanning arborescences (RANK). It's based on [camerini1980ranking]. According to the paper, given a network, RANK computes K most weighted spanning arborescenceos, rooted at a specified node.

The reason for this algorithm is discussed in the paper:

> Given a combinatorial optimization problem P, it may happen that P is too simple a mathematical model of a real problem, so that having at hand a number of alternative solutions is helpful in choosing the most suitable one in the real environment. It may also happen that it is too difficult to incorporate in P some given set of constraints, while it is an easy matter to solve P: in such cases it can be advisable (or even unavoidable) to rank the solutions until one satisfying the extra constraints is found. 

As mentioned in the paper, formal proofs can be found in [camerini1980k].

---

- [camerini1980ranking] Camerini, P. M., Fratta, L., & Maffioli, F. (1980). Ranking arborescences in O (Km log n) time. European Journal of Operational Research, 4(4), 235-242.
- [camerini1980k] Camerini, P. M., Fratta, L., & Maffioli, F. (1980). The k best spanning arborescences of a network. Networks, 10(2), 91-109.

dschult:
Thanks!

edxu96:
Sorry for the delay. I have made some progress today. Now it is able to:

- Rank spanning arborescences (SAs) in descending order.
- Allow the user to specify any numeric edge attribute for determining optimality.

I still have some tasks to make it more complete and user-friendly:

- [ ] Rank SAs in ascending order.
- [ ] Remove dependencies on `pandas` and `loguru`.
- [ ] Use decorator for accepted graph type.
- [ ] Label edges automatically with `_label` edge attribute instead of `label`, to avoid conflicts.
- [ ] Pass code coverage goal.

Should be able to write a PR before next next Monday.

Also, I have implemented another algorithm to count the number of SAs for a directed graph. Will open another issue later.

edxu96:
Just opened a PR (#4796).

edxu96:
This issue has been resolved by [`ArborescenceIterator`](https://github.com/networkx/networkx/blob/2b258dfd59ff69b54428126decfb59aadfd94189/networkx/algorithms/tree/branchings.py#L831).

PR #4796, opened 2021-05-13
New algorithm for ranking spanning arborescences for (#4322)

edxu96:
I have done:

- Use `black`.
- Run all the tests for algorithms and docs locally.
- Unit tests for core classes and functions used by these algorithms.
- Use decorator for graph type.

As far as I know, there are problems:

- All the edges are supposed to have the specified numerical attribute.
- There is only one test case with 5 spanning arborescences.
- The passed graph is deep copied and frozen inside the generator, in case it is modified later.


networkx/algorithms/tree/rank.py
```diff
@@ -0,0 +1,1686 @@
...
+            g_inv.edges[u, v][self.attr] = self.attr_max - g_inv.edges[u, v][self.attr]
+        return g_inv
+
+    def __next__(self) -> Tuple[Arborescence, int]:
+        a_current, _ = super().__next__()
+        a_current_inv = self.inverse_attr(a_current)
+        return a_current_inv, self.rank
+
+
+@nx.not_implemented_for("undirected", "multigraph")
```
dschult, line 1619:
```suggestion
@nx.utils.not_implemented_for("undirected", "multigraph")
```

networkx/algorithms/tree/rank.py
```diff
@@ -0,0 +1,1686 @@
...
+def ascend_spanning_arborescences(
+    g: nx.DiGraph,
+    root: str,
+    attr: Optional[str] = _ATTR,
+    attr_label: Optional[str] = _ATTR_LABEL,
+):
+    return AscendSpanningArborescences(g, root, attr, attr_label)
+
+
+@nx.not_implemented_for("undirected", "multigraph")
```
dschult, line 1629:
```suggestion
@nx.utils.not_implemented_for("undirected", "multigraph")
```

dschult (commented):
These changes might get this through the tests...

networkx/algorithms/tree/rank.py
edxu96: Thanks. Just did it.

networkx/algorithms/tree/rank.py
edxu96: Just did it.

dschult:
This looks like it is going to be very hard to review. 
You've got Type Hinting which we don't use in the package itself.
You are subclassing the base classes with multiple subclasses or subclasses when all it seems like you are doing is adding data structures and methods to to find the source and target of edges. My first quick look suggests that all of this can be handled using NetworkX's paradigm of edges as 2-tuples `(source, target)`.  No complex class structure is needed. The type hinting suggests that nodes and edges are going to be strings, which makes it hard to incorporate into the NetworkX environment where nodes are any non-None hashable and edges are 2-tuples (or 3-tuples for MultiEdges). It would also be helpful if you didn't put what are essentially doc_strings in the middle of code as if they are comments. Put the docstrings just after the viewable function or class signature so they get processed by our autodoc sphinx tools into the reference manual for the package. docstrings on `__init__` are ok... but will never get seen, so you should put most everything into the class doc_string, especially the parameters that the user should provide. It might also help to have a general overview of the whole module in the docstring at the top of the module so someone reading the code (or looking at this module to try to figure out whether it is what they want) can get a handle on the almost 1700 lines of code.  There seems to be a lot of unnecessary code in this module. Subclasses that only provide an `__init__` method that only calls the `__init__` from the base class aren't needed. Also, use Python and NetworkX data structures where possible -- like 2-tuples for edges.

I guess a summary of all these comments is a request that you go through the code again with the goal of refactoring it into a minimalist, sleak set of functions and data structures. 

The code clearly is a lot of work and it passes the tests and provides a nice feature. Thank you for this!

edxu96:
Thanks for the reply.

> You've got Type Hinting which we don't use in the package itself.

- [ ] No problem. Will adapt it to `networkx`'s style.

> You are subclassing the base classes with multiple subclasses or subclasses when all it seems like you are doing is adding data structures and methods to to find the source and target of edges. My first quick look suggests that all of this can be handled using NetworkX's paradigm of edges as 2-tuples (source, target).

This is the first thing I thought about when implementing this algorithm. The data structure with three dictionaries for sources, targets, weights of edges is used in the paper [camerini1980ranking] all along. There is no other paper or code implementation in the past 41 years, as far as I know. I learnt basic graph algorithms systematically before and (I think) am familiar with `networkx`. The way I see it, to redesign and code this algorithm in a more efficient way with a better data structure will take a long time.

> The type hinting suggests that nodes and edges are going to be strings, which makes it hard to incorporate into the NetworkX environment where nodes are any non-None hashable and edges are 2-tuples (or 3-tuples for MultiEdges).

I would suggest users to stick to string as node name. This is not an algorithm that someone can play around ... and at least I am not able to design any unit tests for more generic cases.

> It would also be helpful if you didn't put what are essentially doc_strings in the middle of code as if they are comments. Put the docstrings just after the viewable function or class signature so they get processed by our autodoc sphinx tools into the reference manual for the package. docstrings on __init__ are ok... but will never get seen, so you should put most everything into the class doc_string, especially the parameters that the user should provide. It might also help to have a general overview of the whole module in the docstring at the top of the module so someone reading the code (or looking at this module to try to figure out whether it is what they want) can get a handle on the almost 1700 lines of code.

- [ ] No problem. Will do that.

> There seems to be a lot of unnecessary code in this module. Subclasses that only provide an __init__ method that only calls the __init__ from the base class aren't needed. 

- [ ] Will do that. (I didn't implement this algorithm alone. It was integrated in a huge program in my thesis, and it did took me a long time to decouple it from the rest.)

> refactoring it into a minimalist, sleak set of functions and data structures.

Will see what I can do after solving those easy ones. It might be very hard to break them even smaller. In [camerini1980ranking], three giant algorithms, written in extremely brief mathematical terms, are present directly, along with some weird definitions ...

I did learn a lot by implementing it. It works okay with a larger case in my thesis, so I pass the defense :-)

PR #4800, opened 2021-05-13
Remove mayavi and cartopy dependencies

jarrodmillman:


rossbar (approved):
LGTM - I think this covers all of the mayavi-specific installation/configuration we had in CI.

It's a shame that we can't keep this, but mayavi still isn't pip-installable in Python 3.9. The good news is the example should still work locally for users who have managed to install mayavi in their environments, so the only real loss is the thumbnail/image reverts back to the default.

PR #4801, opened 2021-05-13
Improve documentation look

jarrodmillman:
@rossbar Could you take a quick look at this?

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -1,3 +1,35 @@
+.version-stable {
+    background-color: #2c7fb8ff;
+    color: #2c7fb8ff;
+}
+
+.version-alert a {
+    text-decoration: underline;
+    background-color: #2c7fb8ff;
+    color: #ddd;
+}
+
+.version-alert {
+    background-color: #ff7f0eff;
+    cff7f0effolor: #fff;
+    font-family: "Work Sans", sans-serif;
+    font-weight: 600; font-size: 16px;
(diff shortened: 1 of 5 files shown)


dschult:
Do the tabs across the top allow center alignment within each tab?  
Minor nit-pick but it is not clear that the words at the top are tabs you can use for navigating.  Maybe if they had lines or a graphic to show they are tabs. The left side bar of the page also seems quite wide to me for nothing other than a search dialog box. Maybe make search be one of the tabs... or make the sidebars slightly gray background...  But these are style issues and I'm not very good at style choices. :}   Also if my comments are too little too late, I perfectly understand. 

Thanks for taking on a new documentation theme!

jarrodmillman:
@dschult  I am trying to use the default theme as much as possible for now.  I decided to make a series of smaller PRs after getting carried away in #4744 (I plan to close that w/o merging after this and one more PR).

This PR is just to improve the development banner and to introduce the version switcher.  I plan to improve them both in a future PR---e.g., so that whether it a development version or an old version is handle through one master file.  But that may take time, so for now I just want to get the basic functionality in (even if it is slightly hard-coded).

The second PR will restructure the documentation and the headers will change substantially during that point.  That PR will likely involve a lot more discussion.  We could include more style changes in that PR.  Although, it may make sense to make more substantial aesthetic changes in a third PR (and I may need someone else to take the lead on that).

PR #4818, opened 2021-05-19
Link to guides

jarrodmillman:
This PR adds a link to https://networkx.org/nx-guides/ with the name `NX Guides` to the navbar:
screenshot

Is there a better name?  E.g., `Tutorials and Guides`, `Interactive Tutorials and Guides`, `Interactive Guides`, etc.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -138,7 +138,7 @@ html_theme_options = {
             "icon": "fab fa-github-square",
         },
     ],
-    "external_links": [{"name": "NX Guides", "url": "https://networkx.org/nx-guides/"}],
+    "external_links": [{"name": "Guides", "url": "https://networkx.org/nx-guides/"}],
     "navbar_end": ["navbar-icon-links", "version"],
     "page_sidebar_items": ["search-field", "page-toc", "edit-this-page"],
 }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -138,6 +138,7 @@ html_theme_options = {
             "icon": "fab fa-github-square",
         },
     ],
+    "external_links": [{"name": "NX Guides", "url": "https://networkx.org/nx-guides/"}],
     "navbar_end": ["navbar-icon-links", "version"],
     "page_sidebar_items": ["search-field", "page-toc", "edit-this-page"],
 }
--- a/doc/developer/index.rst
+++ b/doc/developer/index.rst
@@ -1,7 +1,7 @@
 .. _developer:
 
-Developer Guide
-***************
+Developer
+*********
 
 .. only:: html
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -118,7 +118,7 @@ Bibliography
 
    install
    tutorial
-   auto_examples/index
    reference/index
-   developer/index
    release/index
+   developer/index
+   auto_examples/index
--- a/doc/release/index.rst
+++ b/doc/release/index.rst
@@ -1,5 +1,5 @@
-API changes
-***********
+Releases
+********
 
 We don't use semantic versioning.  The first number indicates that we have
 made a major API break (e.g., 1.x to 2.x), which has happened once and probably


dschult:
I wonder if we should shorten the name `Developer Guide` to the single word `Developer`. In the same way that `install` is a good representation of what is linked to in a very terse, grammatically incorrect, but never-the-less clear and concise communication, I think that `Developer` does what we want and reduces confusion.  

I am a little concerned about `API Changes` as well since that is also 2 words (and it's not clear whether each word is on its own, or if we're supposed to read them together).  But the `API` part is short so it seems to hold together OK.

`NX Guides` is better than the longer words suggested.  But to further allow people to understand what it refers to, it might be better to locate it near the tutorial and gallery links.  I would put it between `Gallery` and `Reference`.  But I'm not sure that is possible with sphinx.  I suppose a word like `Library` might be good too.  Is there a word for a collection of guidebooks?  I actually keep coming back to `NX Guides`... :}

It looks nice!

rossbar (approved):
I also like shorter names in general, though I think the ones chosen here are good.

> Is there a better name? E.g., Tutorials and Guides, Interactive Tutorials and Guides, Interactive Guides, etc.

I wouldn't say "better", but in keeping with the short-names theme, we could name the link something like `Learn` (though that might be a bit of an overstatement until we add more articles :) ).

> it might be better to locate it near the tutorial and gallery links. I would put it between Gallery and Reference. But I'm not sure that is possible with sphinx.

I'm not sure it's possible either, but one thing that you could try would be to insert the link directly in the toctree, something like:

```rst

.. toctree::
   :maxdepth: 1
   :hidden:

   install
   tutorial
   auto_examples/index
   NX Guides <https://networks.org/nx-guides>
   ...
```

I'm not sure if this is something that the theme recommends against, but external links in toctrees do work generally.

jarrodmillman:
How does this look:
screenshot

I
- changed `Developer Guide` to `Developer`
- changed `API Changes` to `Releases`
- reordered menu

I can move `Gallery` and `NX Guides` to the middle of the menu, but I slightly prefer them at the end.   It seem more sensible to keep all the internal links together to me.  It is also simpler to implement, but not by much.  We can replace the menu with whatever we want by creating our own custom version of
- https://github.com/pydata/pydata-sphinx-theme/blob/master/pydata_sphinx_theme/_templates/navbar-nav.html

For example, something like this should work:
```
<ul id="navbar-main-elements" class="navbar-nav">

   <li class="toctree-l1 current active nav-item">
    <a class="current reference internal nav-link" href="/install.html">
     Install
    </a>
   </li>

  ...

</ul>
```
That is, we can just explicitly specify exactly what links in what order we want.
But I suspect it is more likely to break (I don't think it is documented).

Ross' suggestion should also work, but I am not sure it will be as clearly identified as an external link (I would need to try it and see).  But, if we decide to move the Guides link to the middle of the menu, I will try both options to see what works best.

dschult:
I like the switch to one word labels -- `Developer` and `Releases` fit well.  And now I am thinking that the `NX` in `Nx Guides` is redundant and could become just `Guides`.  

Jarrod, you mention that putting external links at the end makes more sense than switching between them. Can you explain that more? I was thinking documentation stuff first and then developer stuff later (releases and developer being the developer stuff).  I guess taking that to an extreme would put `Install - Tutorial - Gallery - Guides - Reference` in the order of sophistication of the user with our tools (though I always seem to go straight to reference for other libraries).  Then probably `Developer - Releases`.  But that is content based, not structure based.  And it is very left-to-right based... 

Anyway... this second example verifies to me that I like one word links.  :}

jarrodmillman:
The "putting external links at the end makes more sense" is based on two things:

1.  I like putting similar things close to one another.  Putting an external link in the middle of internal links seems slightly unusual to me.  This isn't a big deal and may be unique to me.  I have the impression that all the internal links should be clumped and the external links should be clumped.  There is only one external link at the moment, but there are several internal links.
2. The second, and perhaps more compelling, issue is that the default template puts all the internal links first and then adds external links.  If you look at https://github.com/pydata/pydata-sphinx-theme/blob/master/pydata_sphinx_theme/_templates/navbar-nav.html you will see that all the internal links are added and then external links are added.  So doing something different requires replacing the template.  Since replacing the template isn't described in the configuration options, I assume there is no reason that they won't completely change things going forward.  I suspect we can fix things, if they change things.  But using undocumented configuration options may require us to make more changes.  But maybe not.

That said, if two people support a different order of menu items, I will quickly drop my position.  (I don't have a strong opinion.)

dschult:
That makes sense, Thanks...
The distinction between internal and external is whether the content is created together or separately. We're trying to mimic cooperatively created content while keeping the actual files separate. So it's kind of a boundary case. The tipping point for me is that we'd need to either 1) hard code the link using @rossbar rst file, or 2) commit ourselves to maintaining a template file. I'd prefer to use a stock template. And option 1) would mean a different hard link for each version, etc and would likely become a headache.

So, let's put the external link on the right...
How about putting Gallery next to Guides -- like you demonstrated but without the NX?
That put the tutorial/reference docs together and the more in-depth gallery and guides together.

jarrodmillman:
screenshot

PR #4826, opened 2021-05-20
Deprecate preserve_random_state

### LINKED ISSUE #4775: Remove function utils.preserve_random_state
Remove and appropriately deprecate utils.preserve_random_state which as mentioned in #4732 looks like it was moved to utils in the same PR that its use was removed from networkx.

Needs:  (anything else?)
- [ ] Deprecate warning in the function itself. 
- [ ] Note in the list of deprecations to handle for v3.0.

jarrodmillman:
Close #4775.


dschult (approved):
Looks good to me.  Thanks!

rossbar (approved):

PR #4831, opened 2021-05-23
DOC: Clarify behaviour of k_crust(G, k)

### LINKED ISSUE #3944: k_crust(G, k) behavior is misleading when combined to k_core(G, k)
With reference to the documentation, k_crust(G, k) returns in fact the complement of k_core(G, k+1).
But this behavior is very misleading.
What is the reason of such behavior for k_crust when k is specified?
Thanks

Bharat123rox:
Fix #3944 by clarifying exact behaviour of `k_crust(G, k)`
Tagging @dschult for review


dschult (approved):
Thanks!  This looks good.

PR #4833, opened 2021-05-24
Deprecate run

jarrodmillman:
As discussed in #4829, use pytest directly instead:  we don't use `run` in the CI; the core developer team doesn't use this function when developing; and it isn't tested.


dschult (approved):
Looks good to me. :}

PR #4837, opened 2021-05-25
edge always accepted, tmp fix, better solution needs to be made

lkora:
Pull request for the following issue:
[https://github.com/networkx/networkx/issues/4836](url)


dschult:
I'm going to close this in favr of a larger-scale solution being discussed in #4836

PR #4841, opened 2021-05-27
Default to NumPy for simrank_similarity

jarrodmillman:

### DIFF
--- a/doc/developer/deprecations.rst
+++ b/doc/developer/deprecations.rst
@@ -88,3 +88,4 @@ Version 3.0
 * In ``algorithms/community/quality.py`` remove ``coverage`` and ``performance``.
 * Remove ``testing``.
 * In ``linalg/graphmatrix.py`` remove ``adj_matrix``.
+* In ``algorithms/similarity.py`` replace ``simrank_similarity`` with ``simrank_similarity_numpy``.
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -78,6 +78,8 @@ Improvements
   Adds ``forest_str`` for string representation of trees.
 - [`#4319 <https://github.com/networkx/networkx/pull/4319>`_]
   pagerank uses scipy by default now.
+- [`#4841 <https://github.com/networkx/networkx/pull/4841>`_]
+  simrank_similarity uses numpy by default now.
 - [`#4317 <https://github.com/networkx/networkx/pull/4317>`_]
   New ``source`` argument to ``has_eulerian_path`` to look for path starting at
   source.
@@ -237,6 +239,8 @@ Deprecations
   Deprecate ``assert_nodes_equal``, ``assert_edges_equal``, and ``assert_graphs_equal``.
 - [`#4850 <https://github.com/networkx/networkx/pull/4850>`_]
   Deprecate ``adj_matrix``.
+- [`#4841 <https://github.com/networkx/networkx/pull/4841>`_]
+  Deprecate ``simrank_similarity_numpy``.
 
 Contributors
 ------------
--- a/networkx/algorithms/similarity.py
+++ b/networkx/algorithms/similarity.py
@@ -1464,7 +1464,7 @@ def _simrank_similarity_numpy(
     # column-normalize the ``adjacency_matrix``
     s = np.array(adjacency_matrix.sum(axis=0))
     s[s == 0] = 1
-    adjacency_matrix /= s #adjacency_matrix.sum(axis=0)
+    adjacency_matrix /= s  # adjacency_matrix.sum(axis=0)
 
     newsim = np.eye(len(G), dtype=np.float64)
     for its in range(max_iterations):
--- a/networkx/algorithms/similarity.py
+++ b/networkx/algorithms/similarity.py
@@ -1201,7 +1201,7 @@ def simrank_similarity(
     source=None,
     target=None,
     importance_factor=0.9,
-    max_iterations=100,
+    max_iterations=1000,
     tolerance=1e-4,
 ):
     """Returns the SimRank similarity of nodes in the graph ``G``.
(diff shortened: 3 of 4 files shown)


dschult:
Well... this runs now... but it still gives errors because the type of the return value is different I suspect.
I'll leave that to someone closer to the problem than i am. :}

jarrodmillman:
Here are some examples of the different return types for the Python and NumPy versions:
```
    >>> G = nx.cycle_graph(2)

    >>> nx.simrank_similarity(G)
    {0: {0: 1, 1: 0.0}, 1: {0: 0.0, 1: 1}}
    >>> nx.simrank_similarity(G, source=0)
    {0: 1, 1: 0.0}
    >>> nx.simrank_similarity(G, source=0, target=0)
    1

    >>> nx.simrank_similarity_numpy(G)
    array([[1., 0.],
           [0., 1.]])
    >>> nx.simrank_similarity_numpy(G, source=0)
    array([1., 0.])
    >>> nx.simrank_similarity_numpy(G, source=0, target=0)
    1.0
```

In  f6d83da, I transformed the result of the NumPy implementation into the old dictionary-style return type of the Python implementation.  I also used `pytest.approx` in the tests b/c of different numerical results between the two methods.

jarrodmillman:
@dschult I think this is ready to merge.  If you see anything that should be fixed, feel free to push to my branch before you merge.

dschult:
This is actually complicated -- `simrank_similarity_numpy` doesn't work with non-integer nodes.
So, not only does the output need to be morphed, but the input `source` and `target` do too.

Also... a question:  Shouldn't we set this up so we only have to remove the `simrank_similarity_numpy` function when we deprecate?  As it is, we'll have to rewrite the `simrank_similarity` function **and** remove the deprecated function.

Another question: Aren't we were keeping the python versions in the main namespace? They have the `_python` suffix so they shouldn't be run by mistake. I haven't kept up with the pagerank namespace shifts. So I could just be wrong here.

jarrodmillman:
> I'm not sure if I'm following the protocol for deprecation correctly. In particular, with deprecation, aren't we supposed to leave the function that changed in tact until the deprecation? In this PR we are changing `simrank_similarity` which breaks backward compatibility.

I was basically trying to follow the page_rank and hits example.  But I think we are following the "rules".  In our documentation, we state

`A deprecation cycle is necessary for any breaking API change, meaning a change where the function, invoked with the same arguments, would return a different result after the change.`

So the question is what does `different result` mean.  My expectation is that this change should produce mostly the same result with small numerical differences.  I also expect that the NumPy / SciPy code should produce more accurate results.  So I don't think we need to a deprecation to `simrank_similarity`.  We do need to deprecate  `simrank_similarity_numpy` because that function will be removed in 3.0.

Am I underestimating how different the results of the NumPy and Python implementations are?  If so, can we should add a test as an example where they significantly differ.

jarrodmillman:
This LGTM in general.   But I am confused why f093500 add `_simrank_similarity_numpy`.  Won't we just delete that in the next release?  I thought the idea was after 2.6, we would just merge `simrank_similarity_numpy` into `simrank_similarity`.  So there would just be `simrank_similarity` and `_simrank_similarity_python` in the 3.0 release.

jarrodmillman:
If you want to make it so that we just delete `simrank_similarity_numpy`, we could move the functionality to `simrank_similarity` now.  But then we would have to convert the dictionary back to an array, which seems like more work.

dschult:
The deprecation warning for `simrank_similarity_numpy` shows up when we run `simrank_similarity`.
So people will get warnings even if they do the right thing. I think the private `_numpy` function takes care of that problem.

Is it OK to have the extra private function `_simrank_similarity_numpy` so we don't have warnings for people who call `simrank_similarity`?

dschult:
Ack -- I don't like this whole implementation... :{
Both versions stop after max_iterations without any warning the the method didn't converge.

The `_is_close` function doesn't treat tolerance correctly. So the Python version is more accurate than the numpy version simply because it effectively set the tolerance to 0.  I fixed `_is_close` but it is still maxing out the max_iterations without notification on most of the tests.

Should I put in a feature to raise an exception when it doesn't converge? I guess so.... That might trip up users, but perhaps for the better... they'll know their previous results weren't at any set tolerance level.

dschult:
I'm going to remove `_is_close`. It is a private function, used only once, and implemented in a bad way.
I'll put a more correct version in the function.

Do I need to deprecate it ?   It is a private function.

jarrodmillman:
Feel free to just remove `_is_close`.

jarrodmillman:
I wonder if `adjacency_matrix = nx.to_numpy_array(G)` should be `adjacency_matrix = nx.adjacency_matrix(G)`.

dschult:
It probably should be....

=== rabbit hole alert   
But that does lose the transparency of the name indicating that it is an adjacency matrix stored as a numpy array.  danger of a rabbit hole here as I start revising the name "adjacency_matrix" to mimic the pandas version: "to_pandas_adjacency".

We want to be able to make an incidence matrix, adjacency matrix and an edgelist. and store each of them as numpy array, scipy sparse matrix, or pandas dataframe (and maybe an Xarray?).

I think we should stick with our current convention that names that don't say otherwise return numpy arrays.  And then they should all have `nodelist` parameters to indicate node order, etc. 
============= back out of rabbit hole
Anyway... I'll make that change.

dschult:
Rats -- that changes it to a sparse matrix instead of a numpy array.
We'd have to change the code some...  

e.g.  sparse matrices are all 2dim.   use `identity` instead of `eye`, etc.

jarrodmillman:
Yes, I was implicitly wondering if we should use sparse matrices.  Most of the other examples use sparses matrices and are much faster.  I was imaging something like what I did in #4847.  There `hits` is reimplemented using `scipy.sparse` and I deprecate the existing implementations.  The sparse matrix stuff is annoying, but it may be years before things change.  And even then, I suspect we may not be able to immediately switch to a sparse array unless we want to drop old version of our dependencies.

jarrodmillman:
I am also considering using a new implementation in #4849.

dschult (approved):

PR #4847, opened 2021-05-27
Use scipy to compute eigenvalues

jarrodmillman:
hits


jarrodmillman:
The SVD version is fastest.
finalhits

dschult (approved):
Looks good to me! :}

PR #4848, opened 2021-05-27
Default to SciPy for eigenvector centrality

jarrodmillman:


jarrodmillman:
eigenvector_centrality

PR #4850, opened 2021-05-27
Add dtype argument to adjacency_matrix

jarrodmillman:
@dschult This adds a dtype argument to `adjacency_matrix` and deprecates `adj_matrix`.


dschult (approved):
Looks good!  Thanks!

PR #4864, opened 2021-06-01
DOC: point towards web archive link in GML docs

MridulS:
The GML specification documentation is not available anymore on the linked website, use archive to points towards an older version of the website.


jarrodmillman (approved):

rossbar (approved):
Thanks @MridulS !

PR #4867, opened 2021-06-02
spring_layout: ignore 'fixed' nodes not in the graph nodes

dizcza:
This pull request addresses the issue when `fixed` nodes passed to the `spring_layout` function are not in the original graph nodes. The current behavior: an unintuitive error is shown.

### Steps to reproduce

```python
import networkx as nx

graph = nx.Graph()
graph.add_nodes_from([1])
pos = {n: (0, 1) for n in [1, 2]}

nx.spring_layout(graph, fixed=[1, 2], pos=pos)
  File "<input>", line 7, in <module>
  File "<decorator-gen-736>", line 2, in fruchterman_reingold_layout
  File "/home/dizcza/miniconda3/envs/networkx/lib/python3.9/site-packages/networkx/utils/decorators.py", line 408, in _random_state
    return func(*new_args, **kwargs)
  File "/home/dizcza/miniconda3/envs/networkx/lib/python3.9/site-packages/networkx/drawing/layout.py", line 461, in fruchterman_reingold_layout
    fixed = np.asarray([nfixed[node] for node in fixed])
  File "/home/dizcza/miniconda3/envs/networkx/lib/python3.9/site-packages/networkx/drawing/layout.py", line 461, in <listcomp>
    fixed = np.asarray([nfixed[node] for node in fixed])
KeyError: 2

```

You may argue that the fix could be another raise error statement, however, it wouldn't be convenient because the user thinks in the following way: "`pos` dict can be a subset of the original nodes as well as it can have keys that are not in the original nodes. Therefore, I expect `fixed` parameter to behave in a similar manner."

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -185,6 +185,8 @@ API Changes
   from ``communicability_betweeness_centrality``
 - [`#4850 <https://github.com/networkx/networkx/pull/4850>`_]
   Added ``dtype`` parameter to adjacency_matrix
+- [`#4867 <https://github.com/networkx/networkx/pull/4867>`_]
+  The function ``spring_layout`` now ignores 'fixed' nodes not in the graph
 
 Deprecations
 ------------
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -390,6 +390,7 @@ def spring_layout(
 
     fixed : list or None  optional (default=None)
         Nodes to keep fixed at initial position.
+        Nodes not in ``G.nodes`` are ignored.
         ValueError raised if `fixed` specified and `pos` not.
 
     iterations : int  optional (default=50)
@@ -447,7 +448,7 @@ def spring_layout(
             if node not in pos:
                 raise ValueError("nodes are fixed without positions given")
         nfixed = {node: i for i, node in enumerate(G)}
-        fixed = np.asarray([nfixed[node] for node in fixed])
+        fixed = np.asarray([nfixed[node] for node in fixed if node in nfixed])
 
     if pos is not None:
         # Determine size of existing domain to adjust initial positions


dschult:
It is convenient to assume that the additional nodes should be ignored, but it is perhaps worse -- surprise! -- if they think the nodes should be in the graph and be fixed in position, but they aren't and no error is provided. Which is worse, an unreported node-not-in-graph error or a inconvenience of having to use `fixed={n:p for n,p in pos.items if n in G}`?  Which is better?

dizcza:
Well, in my humble opinion the latter is better.

rossbar:
I can certainly see both sides to the argument. I tend to agree with @dschult that the downside of silently ignoring extra nodes passed in via `fixed` is greater than the downside of forcing users to be explicit. I think different handling of the extra-nodes case between the `pos` and `fixed` kwargs is justified in that `pos` sets an initial condition, whereas `fixed` specifies a hard constraint. I think forcing users to be intentional about hard constraints is generally safer than making assumptions about their intentions, which can lead to situations that are confusing and hard to debug. This is just an opinion though, not a blocker - if there is a strong preference for the "convenience" approach that'd be fine with me as long as the behavior was well-documented!

I also agree @dizcza that the current error message is uninformative - it should definitely be improved if we decide to stick with the raising-an-exception approach.

dschult:
I'm not invested in either side of this...  I want to know what is the better API choice.
Certainly the current error message is not a good option.

Thanks for this!   
Would you be willing to expand on your opinion? What use case do you have with fixed nodes not in the graph?

dizcza:
I'm visualizing two assemblies of neurons, loosely speaking two groups of nodes, where each node can belong to one or multiple classes (labels). With learning/training, certain neurons pop out of a group, and a few new neurons (that were not in the previous iteration) replace them. If I call `nx.spring_layout` on each iteration from a scratch, the nodes' positions will be different each time I call the function. To substitute this, I pass `fixed` and `pos` from the previous iteration.

Before calling the spring layout, I do `fixed = set(nodes_current).intersection(fixed)`, which is totally fine if you know that you must do this prior to calling the function.

Iteration 1

download

Iteration 2

download (1)

boothby:
> What use case do you have with fixed nodes not in the graph?

This came up for me a few months ago, when I was trying to draw a graph (of an absurdly large data structure) progressively.  It's convenient to store positions for a set of nodes, and use those positions to guide the spring-layout of the next 'level' of nodes.  In the end, I want a single dict that contains all the final positions.  It isn't difficult to do those separately and merge them as necessary, or filter the big dictionary for each call, but it's more convenient to use a big dict and not filter it.

rossbar (approved):
Thanks for sharing further details on use-cases @boothby & @dizcza --- it looks like @dizcza 's original diagnosis was correct and in line with what (at least some) users expect. Given that the behavior is documented I think this is a perfectly good approach. Thanks all for weighing in!

dizcza:
It ain't much but it's honest work.

Thank you for looking into this. So much for the discussion about the change in one line of code! Whatever the final decision is, I appreciate that you look thoroughly into the details.

dschult (approved):
Excellent!   Thanks everyone -- especially @dizcza 

I'll add the API change to the release notes and then merge.

PR #4872, opened 2021-06-04
Deserializing custom default properties graph ml

### LINKED ISSUE #4778: Deserializing custom default properties from yEd graphML files
When importing a yEd created graphml file with custom properties on the 
nodes I found that the default values for boolean properties was always True.

After some browsing through the code of GraphMLReader I found that while
the properties of the nodes themselves are passed through the `decode_data_elements`
method (so that GraphML booleans are parsed correctly), the items in the 
node_default dictionary are not. 

For the defaults, the value of each key is just cast to the appropriate `python_type`
In the case of booleans with value='false' this results in `bool('false') == True`  :(

### Expected Behavior
The default attribute values should also be passed through `decode_data_elements`

### Steps to Reproduce
Make a yed graphml file with some custom node property of value 'boolean' with default 'false'
import the file using GraphMLReader
The resulting default node property has bool value True.

### Environment
Python version:
Python 3.9.4 (default, Apr  6 2021, 11:23:37) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin

NetworkX version:
2.5.1

### Additional context
As posted in google discussion group
[https://groups.google.com/g/networkx-discuss/c/XbFTyx-PKqc/m/pD2_WAdYAQAJ](https://groups.google.com/g/networkx-discuss/c/XbFTyx-PKqc/m/pD2_WAdYAQAJ)

bergmansj:
Added a simple test (analogous to existing tests)
And tried to fix the issue with minimal changes.

Hopefully fixes #4778


dschult (approved):
This all looks good to me!
Thanks!!

rossbar (approved):
Thanks @bergmansj !

PR #4874, opened 2021-06-05
Add a drawing example regarding the solution of traveling salesman problem

eskountis:
Me and @Mdontas created this pull request in order to add a drawing example for the tsp. It's been tested locally and it follows the Black code format.

### DIFF
--- a/examples/drawing/plot_tsp.py
+++ b/examples/drawing/plot_tsp.py
@@ -31,7 +31,7 @@ for i in range(len(pos)):
         dist = dist
         G.add_edge(i, j, weight=dist)
 
-cycle = nx_app.christofides(G, weight='weight')
+cycle = nx_app.christofides(G, weight="weight")
 edge_list = [(x1, x2) for x1, x2 in nx.utils.pairwise(cycle)]
 
 # Draw closest edges on each node only
--- a/examples/drawing/plot_tsp.py
+++ b/examples/drawing/plot_tsp.py
@@ -5,58 +5,49 @@ Traveling Salesman Problem
 
 This is an example of a drawing solution of the traveling salesman problem
 
-The function is used to produce the solution is greedy_tsp,
-where given a set of nodes and a source, it calculates the route of the nodes
+The function is used to produce the solution is greedy_tsp (Nearest Neighbor),
+where given a set of nodes, it calculates the route of the nodes
 that the traveler has to follow in order to minimize the total cost.
-The criteria is that in every iteration choose the next node to be the closest one
-to the latest on the current route (Nearest Neighbor).
 """
 
 import matplotlib.pyplot as plt
 import networkx as nx
 import networkx.algorithms.approximation as nx_app
-import random
 import math
 
-G = nx.DiGraph()
+G = nx.random_geometric_graph(20, radius=0.4, seed=3)
+pos = nx.get_node_attributes(G, "pos")
 
-total_nodes = 10
-pos = {}
-random.seed(1)
-for n in range(0, total_nodes):
-    x = random.randint(0, 100)
-    y = random.randint(0, 100)
-    pos.update({n: (x, y)})
-G.add_nodes_from(pos.keys())
+# Depot should be at (0,0)
+pos[0][0], pos[0][1] = 0.5, 0.5
 
-for i in range(0, total_nodes):
-    for j in range(0, total_nodes):
-        if i == j:
-            continue
-        dist = math.sqrt(
-            math.pow(pos[i][0] - pos[j][0], 2) + math.pow(pos[i][1] - pos[j][1], 2)
-        )
-        dist = round(dist)
+H = G.copy()
+
+
+# Calculating the distances between the nodes as edge's weight.
+for i in range(len(pos)):
+    for j in range(i + 1, len(pos)):
+        dist = math.hypot(pos[i][0] - pos[j][0], pos[i][1] - pos[j][1])
+        dist = dist
         G.add_edge(i, j, weight=dist)
-nodes_labels = {x: x for x in pos.keys()}
-nx.draw_networkx_nodes(G, pos, node_size=200)
-nx.draw_networkx_labels(G, pos, labels=nodes_labels)
 
 cycle = nx_app.greedy_tsp(G, source=0)
-
 edge_list = [(x1, x2) for x1, x2 in nx.utils.pairwise(cycle)]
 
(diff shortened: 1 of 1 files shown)


examples/drawing/plot_tsp.py
```diff
@@ -0,0 +1,62 @@
...
+    pos.update({n: (x, y)})
+G.add_nodes_from(pos.keys())
+
+for i in range(0, total_nodes):
+    for j in range(0, total_nodes):
+        if i == j:
+            continue
+        dist = math.sqrt(
+            math.pow(pos[i][0] - pos[j][0], 2) + math.pow(pos[i][1] - pos[j][1], 2)
+        )
```
dschult, line 38:
```suggestion
        dist = math.hypot(pos[i][0] - pos[j][0], pos[i][1] - pos[j][1])
```

dschult (commented):
This drawing might look better if G was constructed as a geometric graph. As is, it is hard to see any structure from the underlying graph that the TSP is pulled from.

Perhaps:  `G=nx.random_geometric_graph(20,.4)`
You would then have to use the main tsp function with the method set whatever you choose.

Also, I wonder why you chose nearest neighbor greedy instead of christofides?

Thanks for this!

eskountis:
We aren't sure, what you mean with the structure from the underlying graph that the TSP is pulled from. We created the same drawing with `random_geomentric_graph`, but the output was exactly the same just with different nodes each time. Maybe could you guide us even more on that?

Regarding the Nearest Neighbor method, we chose that because we wanted to show an example of tsp solution, no matter the quality of that solution. Any algorithms would be fine.

dschult:
The geometric graph you construct connects every node to every other node. That makes it hard to see the geometry.  
(It also takes a few lines of code that aren't part of the example really -- and NetworkX has a function to do this.)
By setting a threshold for edges to exist, you can get a graph that looks like map that people would use to navigate the domain.  (Using a "seed" makes the random process provide the same nodes every time, so pick one that gives a nice picture for the example. :)
```python
G=nx.random_geometric_graph(20,.4, seed=1)
```
The long edges do not appear! That makes it easier for the eye to "see" the image as a map of the cities rather than a jumbled mess of connections.  That is what I meant by "graph structure".. better words would have been "geometric structure".

If you are only going to show one method of TSP, you should go ahead and use the "best method" which is christofides (as determined by theoretical limits on the approximation of the optimal tour). If you want to compare many methods that would be fine too, but I think it is beyond what this example is aiming for.  That might be better as part of the [nx-guides](https://github.com/networkx/nx-guides) repository that came online recently and would be appropriate for a more in depth look at the TSP algorithm -- similar to the blog you have mentioned before -- only in a repository of guidebooks for NetworkX.

:)

eskountis:
Something like that would be appropriate (like the screenshot below)?

It's definetely more accurate and visually better than the first thought.

image

dschult:
Nice!!   Yes, this is what I was thinking of. 
You can push to the same branch on your repo and the PR will automatically update.
Thanks!

eskountis:
We think it's all been set up. However, we weren't able to run christofides locally to test the output because something were missing, the min_weight_matching method that christofides is calling, so maybe it's not in the latest release of NetworkX.

dschult:
Hmmm... christofides is being tested and doesn't have that trouble.  When you say the "latest release" do you mean 2.5.1?  Because for sure, you need to have the github repo (or 2.6rc1) to make christofides work.

eskountis:
Yes that's exactly what I meant, I installed the 2.6rc1 version. Everything works as it should be!
Thanks!

rossbar:
I went ahead and merged with upstream to get rid of the unrelated circleci failure so the previews can work. I will try to circle back for a review soon, thanks @eskountis

examples/drawing/plot_tsp.py
```diff
@@ -0,0 +1,53 @@
...
+
+H = G.copy()
+
+
+# Calculating the distances between the nodes as edge's weight.
+for i in range(len(pos)):
+    for j in range(i + 1, len(pos)):
+        dist = math.hypot(pos[i][0] - pos[j][0], pos[i][1] - pos[j][1])
+        dist = dist
+        G.add_edge(i, j, weight=dist)
```
rossbar, line 32: Alternatively you could explicitly create the complete graph (required by Christofides) explicitly and iterate through the edges there, e.g.

```python
G = nx.complete_graph(20)
for u, v in G.edges():
    G[u][v]['weight'] = math.hypot(pos[v][0] - pos[u][0], pos[v][1] - pos[u][1])
```

examples/drawing/plot_tsp.py
```diff
@@ -0,0 +1,53 @@
...
+
+cycle = nx_app.christofides(G, weight="weight")
+edge_list = [(x1, x2) for x1, x2 in nx.utils.pairwise(cycle)]
```
rossbar, line 35: Passing the generator to the list constructor will do the same thing without the explicit list comprehension:

```suggestion
edge_list = list(nx.utils.pairwise(cycle))
```

examples/drawing/plot_tsp.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Draw closest edges on each node only
+nx.draw_networkx_edges(H, pos, edge_color="blue", width=0.5)
+
+# Draw the route
+nx.draw_networkx(
+    G,
+    pos,
+    with_labels=True,
+    nodelist=set(cycle),
```
rossbar, line 45:
```suggestion
```

This shouldn't be necessary as the cycle should contain all the nodes in G.

rossbar (commented):
Thanks @eskountis , I think this is a nice example. I had a couple suggestions re: various code chunks in the example but generally I think this is a nice illustration of TSP.

rossbar (approved):
Thanks @eskountis, this LGTM - this slipped through the cracks on my end because I thought there were unresolved comments but in the end it was just a suggestion and not a blocker.

I have a sense that this example might actually be a better fit in the `Algorithms` category rather than `Drawing`, but we can always worry about moving it around later. The most important thing is getting it in. Thanks for the nice example!

dschult (approved):
This look good to me.
It hasn't been tested for a while. Should we run the tests on the latest main before merging?
I don't know of any potential trouble, but it seems like there have been many chages in the last 5 months.

rossbar:
Good idea - I'll rebase on main and force push which will retrigger CI. Assuming everything goes well I'll merge afterwards

PR #4875, opened 2021-06-05
Reformatted  table to address issue #4852

salym:
The grid table was not properly formatted using reStructuredText Markup which is required by Sphinx.

### DIFF
--- a/networkx/generators/sudoku.py
+++ b/networkx/generators/sudoku.py
@@ -6,17 +6,19 @@ algorithms for solving or generating Sudoku puzzles.
 A completed Sudoku grid is a 9x9 array of integers between 1 and 9, with no
 number appearing twice in the same row, column, or 3x3 box.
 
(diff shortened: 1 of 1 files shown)


rossbar (approved):
This is an obvious improvement - the new table is formatted as expected: https://2122-890377-gh.circle-artifacts.com/0/doc/build/html/reference/generators.html#module-networkx.generators.sudoku.

In principle we could also try to improve the column-formatting but IMO that just makes the rst less readable for relatively minimal gain --- this seems like the right level of table formatting to me.

Thanks @salym !

jarrodmillman (approved):

salym:
> In principle we could also try to improve the column-formatting but IMO that just makes the rst less readable for relativelyminimal gain --- this seems like the right level of table formatting to me.

Hi @rossbar,
this would be an easy job but I'm afraid this depends on the Sphinx theme's CSS.

PR #4884, opened 2021-06-08
Fix conda instructions

### LINKED ISSUE #4870: Documentation for conda based dev environment setup fails
The new contributor documentation available at https://networkx.org/documentation/latest/developer/contribute.html#development-workflow doesn't work for conda based development environment setup.

### Current Behavior

While installing developer environment dependencies 
```
$ conda install -c conda-forge `for i in requirements/{default,developer,test}.txt; do echo -n " --file $i "; done`

InvalidVersionSpec: Invalid version '1.19;platform_python_implementation!='PyPy'andpython_version<'3.10'': invalid character(s)
```
it fails to parse the pip specific requirements.txt file.
### Expected Behavior

Users should be able to setup the development environment by following the documentation.

### Steps to Reproduce

```
$ conda install -c conda-forge `for i in requirements/{default,developer,test}.txt; do echo -n " --file $i "; done`

InvalidVersionSpec: Invalid version '1.19;platform_python_implementation!='PyPy'andpython_version<'3.10'': invalid character(s)
```

### Environment

Python version: 3.9.4
NetworkX version: main branch
conda version: 4.10.1

jarrodmillman:
Fix #4870.


dschult:
That's an interesting point...  NetworkX on pypy just needs decorator. So, even a one liner:
```
pip install decorator networkx --nodeps
```
should work. I'm +1 for removing the environment markers and 3.10 stuff from the requirements file. 
And +1 for taking out the shell tools from the conda (and pip, which you already did) install line and hardcoding the requirement files:
```
conda install -c conda-forge --file requirements.txt --file requirements/developer.txt
```

As for your first question about `pip install -e . --no-deps`, I don't think there is a way to install from `.` in conda. I always just set `PYTHONPATH`... though that doesn't check the installation process.  It looks like you could make a `tar` or `bz2` of the packages and then install from that, but I haven't tried it.  The `pip install -e . --no-deps` works just fine within conda and should install the result into the correct `site-packages` directory within the current conda environment. So I think that is OK.

jarrodmillman:
Ok, I think this is ready.  Someone should double-check that I didn't mess anything up.

dschult:
I was able to get the recipe for conda to work with adjustments.
But I can't figure out the wheels stuff within Github actions. The Pypy run on MacOS is not able to load the wheels for numpy and scipy...  And I think, now that we've removed the pypy stuff from default.txt, that we need to call pip to load decorator and requirements/test.txt.  But I can't find where we tell the CI to pull in the wheels so the tests ch=rash on loading the wheels for the requirements we aren't going to use. :}

dschult (approved):
Thanks @jarrodmillman 
This looks good to me.

MridulS (approved):
conda is happy now.

fmagin:
Well, I just found this issue after an hour of confusing debugging why some package that requires networkx didn't properly install in PyPy. It required a `git blame` on the `requirements/default.txt` to find this issue and in that commit I found the instructions to just install it with `pip install decorator networkx --nodeps`

So there are definitely people affected by this, though I guess not much? Some clearer error that networkx doesn't fully support PyPy by default would really be nice for anyone else running into this, the errors that appear are really confusing

dschult:
I'm  not sure why this was so hard to find. The release notes for v2.6 of NetworkX explain what is going on, and the discussion here is searchable.  Maybe we need more discussion about dependencies to make it more easily searchable, but it's hard to enforce that in any quantifiable way.

PyPy does work with NetworkX -- well, with the parts of NetworkX that don't require scipy, etc. To install networkx that way you need to impose the --no-deps option (or we need to remove all requirements from our `requirements.txt` file -- which is being strongly considered).  Best would be to allow other packages to specify something like `networkx[minimal]` as a requirement rather than `networkx` -- but my understanding is that pypi doesn't have a way to do that yet.

fmagin:
Sometimes the GitHub search is just surprisingly horrible, but I hope that at least this post will make it easier to find.

The actual thing that fails is building scipy, specifically it fails with: `LookupError: https://files.pythonhosted.org/packages/bb/bb/944f559d554df6c9adf037aa9fc982a9706ee0e96c0d5beac701cb158900/scipy-1.7.0.tar.gz#sha256=998c5e6ea649489302de2c0bc026ed34284f531df89d2bdc8df3a0d44d165739`

This was weird und unspecific enough that we tried reproducing this on other machines first, which took some time. Tried finding out why scipy wasn't working with PyPy (because it should), but didn't find anything obvious. Then we started wondering why exactly scipy was even needed, realized that networkx was pulling it in, and I checked what changed with the networkx dependencies and ended up here.

The package I am using really just needs a solid graph data structure for Python, and I assume that there are various other projects that use networkx in this way. I don't see a clearly correct solution to this situation either, explicitly specifying `networkx[minimal]` as a dependency seems like the most elegant solution, but if that isn't working I am not sure what the better way would be.

dschult:
We have pulled the 2.6.0 and 2.6.1 releases from pypi and are reconsidering the dependency structure.
We test the package on PyPy with our CI and we plan to continue that. We just want a good way to make the default installation also include a few bigger dependencies so more of our algorithms are available by default.

fmagin:
I do see a dilemma between "networkx as a pure python graph library" (which had nearly no dependencies previously) vs "networkx as a cutting edge library for graph analysis algorithms". One idea that comes to mind could be separating those concerns into separate packages, one for the basic data structure and algorithms that can be implemented in pure python with no further dependencies, and another one for research/scientific level analysis algorithms, that can pull in the typical ecosystem of scientific libraries. But that might be a non trivial amount of work and will need serious discussions about what the `networkx` package aims to provide.

PR #4889, opened 2021-06-10
Adding extendability problem

gsemer:


dschult:
Let's just use the same branch for all the commits you make. When you push to that branch github automatically updates the PR and runs the tests.  I'll close the other PR and we can use this one as "the branch" for this addition.

gsemer:
I would like to know out of curiosity the reason for which there were not all the checks succesful. Please inform me if something did not work properly.

MridulS:
It's a styling error, could you run `black` on the files in the PR?

https://github.com/psf/black#installation
```
$ pip install black
$ black networkx/algorithms/bipartite/extendability.py networkx/algorithms/bipartite/tests/test_extendability.py
```

gsemer:
I succesfully installed black. But i am facing some problem running the last command. It says that the path does not exist. But i give the right path and do not know the reason this happening. What does PR mean ?

gsemer:
Forget the last question. Where should i run the command ? I did it in command line on my pc.

dschult:
Yay -- the PR is created!
If you run black from the `networkx/algorithms/bipartite` directory it should find and check only the files that have been changed.  If you want to be more specific (so it doesn't check everything) you can specify the filenames relative to the current directory.   Typically I would do:
```
black extendability.py tests/test_extendability.py 
```
That will make the style changes for you. If you just want to see what the changes will be, use `black --diff filenames`.

dschult:
Then commit those changes and then push to the same branch!
Each time you push to the branch on your repo it will automatically update the PR and run the tests.

gsemer:
I am going to delete test_ext branch. Thus, please ignore this PR. Take under consideration the new PR i made with the name "adding extendability problem (2nd try)"

PR #4907, opened 2021-06-17
Multiply rooted graph similarity

antonpibm:
Adding support for multiple roots in all similarity algorithms.
 - `optimize_edit_paths` and `graph_edit_distance` had support for a single root, I've extended the functionality to support multiple roots (multiply vertex-rooted graphs)
 - `optimize_graph_edit_distance` and `optimal_edit_paths`  actually use `optimize_edit_paths` to search for results but didn't have support for roots, I've added it to those methods as well.

### DIFF
--- a/networkx/algorithms/similarity.py
+++ b/networkx/algorithms/similarity.py
@@ -141,11 +141,11 @@ def graph_edit_distance(
         cost of 1 is used.  If edge_ins_cost is not specified then
         default edge insertion cost of 1 is used.
 
-    roots : 2-tuple
-        Tuple where first element is a node in G1 and the second
+    roots : 2-tuple or list of 2-tuples
+        Tuples where first element is a node in G1 and the second
         is a node in G2.
         These nodes are forced to be matched in the comparison to
-        allow comparison between rooted graphs.
+        allow comparison between multiply vertex-rooted graphs.
 
     upper_bound : numeric
         Maximum edit distance to consider.  Return None if no edit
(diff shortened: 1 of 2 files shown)


networkx/algorithms/similarity.py
```diff
@@ -1049,13 +1068,14 @@ def get_edit_paths(
 
     initial_cost = 0
     if roots:
-        root_u, root_v = roots
-        if root_u not in pending_u or root_v not in pending_v:
-            raise nx.NodeNotFound("Root node not in graph.")
+        roots = [roots] if type(roots) == tuple else roots
```
dschult, line 1071:
```suggestion
        if isinstance(roots, tuple) and len(roots) == 2:
            roots = [roots]
```

networkx/algorithms/similarity.py
```diff
@@ -1080,7 +1101,8 @@ def get_edit_paths(
             ]
         ).reshape(m, n)
         if roots:
-            initial_cost = 1 - node_match(G1.nodes[root_u], G2.nodes[root_v])
+            initial_cost = sum(
+                [1 - node_match(G1.nodes[root_u], G2.nodes[root_v]) for root_u, root_v in roots])
```
dschult, line 1105: Should this be (1 - sum(node_match for all pairs of root nodes))  or  sum( 1 - nodematch)??

dschult (commented):
I think this looks pretty good.  Sorry for the delay in review while we focused on release of NX v2.6.  Can you add a paragraph to the doc_string explaining what it means to have multiply rooted graphs and how that affects the algorithm?

PR #4909, opened 2021-06-17
Initial setup of lazy_import functions.

dschult:
Lazy load of potentially uninstalled libraries. 
Also, a function to set up the internal sub-packages in a lazy-load fashion.

This first version will fail tests in environments without numpy because pytest.importorskip is not lazy yet. 

Still needs:
- way to handle pytest.importorskip
- Loader class instead of monkey patch on SourceFileLoader
- Way to identify a lazily-loaded-not-yet-used module

### DIFF
--- a/networkx/lazy_imports.py
+++ b/networkx/lazy_imports.py
@@ -1,30 +1,31 @@
+import importlib
 import importlib.util
 import types
 import os
 import sys
-import pytest
 
 
-__all__ = ["lazy_import", "lazy_package_import", "lazy_importorskip"]
+__all__ = ["attach", "load"]
 
 
-def lazy_package_import(module_name, submodules=None, submod_attrs=None):
-    """Install lazily loaded submodules, and functions or other attributes.
+def attach(module_name, submodules=None, submod_attrs=None):
+    """Attach lazily loaded submodules, and functions or other attributes.
 
     Typically, modules import submodules and attributes as follows::
 
       import mysubmodule
       import anothersubmodule
+
       from .foo import someattr
 
-    The idea of the lazy package import is to replace the `__init__.py`
+    The idea of  this function is to replace the `__init__.py`
     module's `__getattr__`, `__dir__`, and `__all__` attributes such that
     all imports work exactly the way they normally would, except that the
     actual import is delayed until the resulting module object is first used.
 
     The typical way to call this function, replacing the above imports, is::
 
-      __getattr__, __lazy_dir__, __all__ = install_lazy(
+      __getattr__, __lazy_dir__, __all__ = lazy.attach(
           __name__,
           ['mysubmodule', 'anothersubmodule'],
           {'foo': 'someattr'}
@@ -34,19 +35,16 @@ def lazy_package_import(module_name, submodules=None, submod_attrs=None):
 
     Parameters
     ----------
-
     module_name : str
         Typically use __name__.
     submodules : set
         List of submodules to lazily import.
     submod_attrs : dict
         Dictionary of submodule -> list of attributes / functions.
-        These attributes are created the first time the submodule
-        is used.
+        These attributes are imported as they are used.
 
     Returns
     -------
-
     __getattr__, __dir__, __all__
 
     """
@@ -83,6 +81,75 @@ def lazy_package_import(module_name, submodules=None, submod_attrs=None):
     return __getattr__, __dir__, list(__all__)
 
 
+def load(fullname):
+    """Return a lazily imported proxy for a module or library.
+
+    We often see the following pattern::
+
+      def myfunc():
+          import scipy as sp
+          sp.argmin(...)
+          ....
+
+    This is to prevent a library, in this case `scipy`, from being
+    imported at function definition time, since that can be slow.
+
+    This function provides a proxy module that, upon access, imports
+    the actual module.  So the idiom equivalent to the above example is::
+
+      sp = lazy.load("scipy")
+
+      def myfunc():
+          sp.argmin(...)
+          ....
+
+    The initial import time is fast because the actual import is delayed
+    until the first attribute is requested. The overall import time may
+    decrease as well for users that don't make use of large portions
+    of the library.
+
+    Parameters
+    ----------
+    fullname : str
+        The full name of the package or subpackage to import.  For example::
+
+          sp = lazy.load('scipy')  # import scipy as sp
+          spla = lazy.load('scipy.linalg')  # import scipy.linalg as spla
+
+    Returns
+    -------
+    pm : importlib.util._LazyModule
+        Proxy module. Can be used like any regularly imported module.
+        Actual loading of the module occurs upon first attribute request.
+
+    """
+    try:
+        return sys.modules[fullname]
+    except:
+        pass
+
+    # Not previously loaded -- look it up
+    spec = importlib.util.find_spec(fullname)
+
+    if spec is None:
+        # module not found - construct a DelayedImportErrorModule
+        spec = importlib.util.spec_from_loader(fullname, loader=None)
+        module = importlib.util.module_from_spec(spec)
+        tmp_loader = importlib.machinery.SourceFileLoader(module, path=None)
+        loader = DelayedImportErrorLoader(tmp_loader)
+        loader.exec_module(module)
+        # dont add to sys.modules. The module wasn't found.
+        return module
+
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[fullname] = module
+
+    loader = importlib.util.LazyLoader(spec.loader)
+    loader.exec_module(module)
+
+    return module
+
+
 class DelayedImportErrorLoader(importlib.util.LazyLoader):
     def exec_module(self, module):
         super().exec_module(module)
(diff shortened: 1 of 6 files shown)


dschult:
I believe the test failures here are due to not replacing `pytest.importorskip` uniformly throughout the library. So some modules are using `nx.lazy_importorskip` and others aren't.  I've tested it locally on utils/tests/test_misc.py and on `linalg/tests/test_algebraic_connectivity.py` and it works for those.

I've settled on mimicking the `importlib.LazyLoader` and `_LazyModule` code as much as possible. Two new subclasses of these enable introspection of the "spec" without triggering the `ModuleNotFoundError`. That allows a function `lazy_importorskip` in addition to the two functions from scikit-image (morphed accordingly) `lazy_import()` and I renamed the other to `lazy_package_import()`.

They should be fairly easy to read now.  I'll start workign on figuring out how much of the code has to be changed to enable `lazy_importorskip` and I'll think a little about whether we can suggest this change to the original `pytest.importorskip`.  Probably no new pushed from me until Sunday.  Feedback welcome. :}

networkx/lazy_imports.py
```diff
@@ -0,0 +1,179 @@
...
+    def __getattribute__(self, attr):
+        spec = super().__getattribute__("__spec__")
+        raise ModuleNotFoundError(
+            f"Delayed Report: module named '{spec.name}' not found.\n"
+            "Reporting was Lazy -- delayed until module attributes accessed.\n"
+            f"Most likely, {spec.name} is not installed"
+        )
+
+
+def lazy_importorskip(modname, minversion=None, reason=None):
```
stefanv, line 117:
```suggestion
def lazy_importerskip(modname, minversion=None, reason=None):
```

stefanv: It's unclear why pytest made this spelling mistake in their code!  I guess we can just fix ours.

networkx/lazy_imports.py
```diff
@@ -0,0 +1,179 @@
...
+        loader.exec_module(module)
+        return module
+
+    # package not found - construct delayed error module
+    spec = importlib.util.spec_from_loader(fullname, loader=None)
+    module = importlib.util.module_from_spec(spec)
+    tmp_loader = importlib.machinery.SourceFileLoader(module, path=None)
+    loader = DelayedImportErrorLoader(tmp_loader)
+
+    sys.modules[fullname] = module
```
stefanv, line 177: Do we want to insert modules not found into `sys.modules`?

networkx/tests/test_lazy_imports.py
```diff
@@ -0,0 +1,57 @@
...
+    assert isinstance(anything_not_real, DelayedImportErrorModule)
+    # see if it changes for second access
+    try:
+        anything_not_real.pi
+        assert False  # Should not get here
+    except ModuleNotFoundError:
+        pass
+
+
+def test_lazy_package_import():
```
stefanv, line 31: This is nice; would be good to get something like that over to the skimage PR too!

networkx/tests/test_lazy_imports.py
```diff
@@ -0,0 +1,57 @@
...
+        "name": name,
+        "submods": submods,
+        "myall": myall,
+        "__getattr__": None,
+        "__lazy_dir__": None,
+        "__all__": None,
+    }
+    assert locls.keys() == expected.keys()
+    for k, v in locls.items():
+        if k not in ("__getattr__", "__lazy_dir__", "__all__"):
```
stefanv, line 56: Maybe `if not expected[k] is None: assert v == expected[k]`, instead of repeating the keys without values here.

networkx/lazy_imports.py
```diff
@@ -0,0 +1,179 @@
...
+        for attr in set(attr_to_modules.keys()) | submodules:
+            __getattr__(attr)
+
+    return __getattr__, __dir__, list(__all__)
+
+
+class DelayedImportErrorLoader(importlib.util.LazyLoader):
+    def exec_module(self, module):
+        """Make the module load lazily."""
+        module.__spec__.loader = self.loader
```
stefanv, line 89: Curious where this came from; is this the part you said you took from LazyLoader?

stefanv: I guess what I'm curious about is whether you can't:

```python
super().exec_module(module)
module.__class__ = DelayedImportErrorModule
```

stefanv (commented):
Nice work, Dan!  This looks great already.  I just had a few questions (mainly for my own enlightenment).

networkx/lazy_imports.py
dschult: The spelling of `importorskip` bothered me when I first encountered pytest too.
But think of it as:  `import or skip`    It should probably be:   `import_or_skip`.
I think we should keep it the same as the word that pytest uses.
Though we could suggest the underscores to pytest.

networkx/lazy_imports.py
dschult: This is the tricky part -- I couldn't figure out another way to make future imports work. But I thought of something yesterday which make removing this more feasible:

Perhaps we should make delayed_not_found_errors only work for the current module's scope. That is, they return a module object (lazily loaded) but don't change `sys.modules`.  Then `np` is available in the module as a trigger for a delayed error report, but it doesn't cause delayed error reports for other modules (or other libraries who don't know anything about networkx's lazy load setup -- which I think is a much bigger problem).  

It does mean that we have to lazy load, check for `spec`, catch `spec is None`, and create this DelayedModuleNotFound object inside each module where we use `numpy` when `numpy` is not installed. But I'm guessing that is not going to be a performance problem.

Not adding to `sys.modules` might also remove any need to overcome `importorskip`.
Yes!  I think this will be a cleaner approach -- if I'm not missing anything.  I'll try it. :}

networkx/lazy_imports.py
dschult: Yes!  I should use `super` and then fix the class.  Nice!

networkx/lazy_imports.py
stefanv: 🤯 I would never have guessed the *or*! Yes, the underscores would be most helpful.

dschult:
Some of these ideas might be helpful in the PR over at:  scikit-image/scikit-image#5101

networkx/tests/test_lazy_imports.py
dschult: > This is nice; would be good to get something like that over to the skimage PR too!

I'm going to try to get something like these tests of `attach` over in the skimage PR as suggested. 

Should I also update the `lazy.load` function there to allow "delayed failed modules"?  It currently raises a `ModuleNotFoundError` immediately upon `lazy.load` when a library is not found. That works well for many cases, but not if you want to import optional libraries. Thoughts @stefanv ?

stefanv: I think that would be a good feature to have, thanks Dan!

dschult:
Looks like the differences between the NX lazy-load and scikit lazy-load are:
- `attach` first argument name is `module_name` rather than `package_name`
- `load` raises `ModuleNotFoundError` for uninstalled libraries in both, but this version delays the ImportError until it is used. The skimage version raises upon the calling of `load`.

What are the next steps?
1) replace all non-builtin `import`s with lazy imports.  Should they be moved out of the functions to the top of the module? We lose easy search for where they are used, but gain that all import related commands appear at top of module.
2) load the NetworkX subpackage code in a lazy way.  It may be worth going through our prior choices of what gets put in the top level namespace as part of this.

Note: number 1) is done to `utils/misc.py` in this PR to show what it might look like.

Finally, shall I make these "next steps" in this PR?  I'm thinking maybe making another PR that relies on this one. It will touch a lot of code to change the imports.

networkx/tests/test_lazy_imports.py
```diff
@@ -0,0 +1,96 @@
...
+    math = lazy.load("math")
+    anything_not_real = lazy.load("anything_not_real")
+
+    # Now test that accessing attributes does what it should
+    assert math.sin(math.pi) == pytest.approx(0, 1e-6)
+    # poor-mans pytest.raises for testing errors on attribute access
+    try:
+        anything_not_real.pi
+        assert False  # Should not get here
+    except ModuleNotFoundError:
```
stefanv, line 19: I guess these are not raised any more

stefanv (approved):
This looks good to me.

As @dschult mentioned, the biggest change from the skimage approach here is that import exceptions are delayed.  This means that a user may be able to import a subpackage successfully, but that code will break once it is used. In practice (a) it will allow more code to execute successfully [if your code doesn't use a specific import, it won't fail] (b) raise errors at a different time to what users are currently accustomed to (on function execution rather than module import).  I think (b) is not an issue for two reasons: (1) these imports used to reside inside functions before load anyway and (2) the difference will only really be noticeable in interactive work, where raising the exception later makes more sense.

OK, I may have convinced myself to port that change over to skimage as well ;)

dschult:
Thanks @Stefan,   
There is another trait of this way to lazy-load.  It means that new contributors who are used to using `import numpy as np` at the top of the module will need to prompted to change that syntax to `np = nx.lazy_imports.load("numpy")`.

But I guess they are currently being prompted to move the import inside their functions. So that's likely to be about the same maintenance burden.  The CI will check it either way too.
:}

stefanv:
> But I guess they are currently being prompted to move the import inside their functions. So that's likely to be about the same maintenance burden. The CI will check it either way too. :}

True :)  I commented on the other PR that slightly nicer syntax may be something like `nx.lazy_import(x)` or `nx.lazy_load(x)`.

dschult:
I was using the function name "load" because of the version in skimage. But maybe that's not the best name...  

Why is the term "load" used?  That seems to normally refer to bringing data from a file into a variable.  The standard term in Python seems to be "import". Is there a reason to avoid the term "import" here,  e.g.  `nx.lazy_import("numpy")`?  What was the thought behind using the name `load`?

stefanv:
We went through various discussions on naming and that is what we settled on, but I don't think there's any particularly good reason for it other to distinguish it from a traditional import. `lazy_import` should be fine.

networkx/tests/test_lazy_imports.py
dschult: I think the tests pass, so these ModuleNotFoundError errors are raised.
Are you saying that they shouldn't be raised? 
Once we try to access an attribute of a lazily loaded module the module must finish the loading process right?

stefanv: Ah, woops, I thought you had changed it to `DelayedImportError`, my bad.

PR #4911, opened 2021-06-20
Update graphML docstring to include examples

berlincho:
I include examples for `named_key_ids` and `edge_id_from_attribute` parameters in the docstring since the two parameters are hard to understand without specific examples. 

Also, I add an escape character to prevent  `\n` from taking action.
From
```python
>>> linefeed = chr(10)  # linefeed = \n
```
To
```python
>>> linefeed = chr(10)  # linefeed = \\n
```

### DIFF
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -81,7 +81,7 @@ def write_graphml_xml(
        For example, if edges have both int and float 'weight' attributes,
        we infer in GraphML that both are floats.
     named_key_ids : bool (optional)
-       If True use attr.name as value for key elements' id attribute.
+       If True use attr.name as value for node and edge data elements' key attribute.
     edge_id_from_attribute : dict key (optional)
         If provided, the graphml edge id is set by looking up the corresponding
         edge data attribute keyed by this parameter. If `None` or the key does not exist in edge data,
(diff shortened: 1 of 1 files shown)


berlincho:
Is there anyone who can check the docstring and decide if they are ok?

rossbar (commented):
The examples seem technically correct, but don't really illustrate what these parameters actually do.

Writing good examples here will be tricky I think --- one way to improve the examples would be to use `BytesIO` as a stand-in for a file object, as that allows introspection via `getvalues()`. This is useful because it allows people reading the examples to actually see the *output* from the various `write_gml` calls.

dschult:
I agree @rossbar Also, one of the confusing things about this module's docstrings is that words like `attr`, `keys` and `id` are used interchangeably without description of their distinctions or what they even refer to.  I'd prefer not to use `attr.name` at all unless we explain what `attr` and `.name` refer to.  The dot makes `attr.name` look like a data structure that presumably is defined somewhere in the module, but it isn't.  I believe it is a way to refer to the name of an attribute. But even then -- is it a GraphML special attribute, or an arbitrary attribute?

I know you are just using terminology that was there before and this is definitely better than it was. But maybe we could take a bigger look at the doc_string from the perspective of someone who doesn't know GraphML and is looking to use this module.  That should include: 1) saying that GraphML provides the ability to store node and edge attributes while introducing language for the name and value of the attribute (will it be "key":"value" or "name":"value" or do we need both a key and a name), 2) describing what attribute names are treated specially and how to handle these, 3) the default names for node attributes and how to override that (I think this is in there already), 4) the reason for needing edge_id.

Also, as far as example attributes, it might be better to use attribute examples that are not likely to be treated as special attributes. "eid" or "edge_id" are quite possibly handled specially (e.g. required for each node). While "color" or "shape" or "foo" are much less likely to be treated specially.

Thanks for this!

berlincho:
@dschult @rossbar Thanks for the constructive feedback. I agree the terms in the default docstring are misleading (e.g. keys, id, attribute...). That's one of the reasons I would like to provide clear examples here. I will take a look at GraphML spec and think about how to present the description.

PR #4912, opened 2021-06-20
In `DiGraph.to_undirected`, remove not accessed variable

europeanplaice:
In `DiGraph.to_undirected`, local variable `graph_class` was defined but not used in this function. 
Though it is not a serious problem, this should be removed.

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -1177,10 +1177,11 @@ class DiGraph(Graph):
         >>> list(G2.edges)
         [(0, 1)]
         """
+        graph_class = self.to_undirected_class()
         if as_view is True:
             return nx.graphviews.generic_graph_view(self, Graph)
         # deepcopy when not a view
-        G = Graph()
+        G = graph_class()
         G.graph.update(deepcopy(self.graph))
         G.add_nodes_from((n, deepcopy(d)) for n, d in self._node.items())
         if reciprocal is True:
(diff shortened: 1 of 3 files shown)


dschult:
Actually, this is a bug and the fix is to replace `Graph` with ` graph_class` in lines 1181 and 1183.  similar to line 1595 or so in graph.py 

Thanks for finding this!!

europeanplaice:
Thanks for the reply!
I fixed it to be the same as `graph.py`.
Is it what you meant? Sorry if it's wrong.

dschult:
That’s one of the changes.  The other one that is needed is two lines prior where Graph should be replaced by graph_class.
`generic_graph_view(self, Graph)`   ->   `generic_graph_view(self, graph_class)`

ii wonder if there is a test we could add that would check this…

dschult:
I moved your test to the `test_graph.py` module so it could be applied to graph, digraph, multigraph and multidigraph.
I also added a little to it and made a second test of the `to_directed_class` feature (which does not seem to be tested anywhere).  Then I made undirected versions of directed tests and vice versa.

See what you think...  Anything else you'd like to include in this PR?

europeanplaice:
You made a great improvement on my test!

I don't have anything to add, and you can finish this pull request.

dschult (approved):

rossbar (approved):
LGTM - thanks @europeanplaice & @dschult !

PR #4916, opened 2021-06-22
Update read_gml docstring with destringizer ex

randallwvr90:
Responding to issue #4877 by adding a simple destringizer example to the
read_gml docstring.

### DIFF
--- a/networkx/readwrite/gml.py
+++ b/networkx/readwrite/gml.py
@@ -17,9 +17,9 @@ interact with different languages and even different Python versions.
 Re-importing from gml is also a concern.
 
 Without specifying a `stringizer`/`destringizer`, the code is capable of
-handling `int`/`float`/`str`/`dict`/`list` data as required by the GML
-specification.  For other data types, you need to explicitly supply a
-`stringizer`/`destringizer`.
+writing `int`/`float`/`str`/`dict`/`list` data as required by the GML 
+specification.  For writing other data types, and for reading data other
+than `str` you need to explicitly supply a `stringizer`/`destringizer`.
 
 For additional documentation on the GML file format, please see the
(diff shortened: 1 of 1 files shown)


dschult:
Thanks for this!!

What do you think of the comment in the "Notes" section of the doc_string that:

```
Without specifying a `stringizer`/`destringizer`, the code is capable of
handling `int`/`float`/`str`/`dict`/`list` data as required by the GML specification.
For other data types, you need to explicitly supply a `stringizer`/`destringizer`.
```

It seems like this isn't quite correct if we need to supply `int` to convert to ints.
Shall we change it to something like:
```
Without specifying a `stringizer`/`destringizer`, the code is capable of
writing `int`/`float`/`str`/`dict`/`list` data as required by the GML 
specification.  For writing other data types, and for reading data other
than `str` you need to explicitly supply a `stringizer`/`destringizer`.
```

If people approve, this paragraph appears in 4 places in the file and could be changed in each place. I can do that if its too much trouble for you.

randallwvr90:
Right, it looks like `parse_kv(curr_token)` will not cast the value if `destringizer` is not specified. Would it be better to use `literal_eval` to get the functionality that is in that comment block you quoted?

Also, it's no problem for me to change all those comments if that's what is decided.

dschult:
I think we should change the doc_string to match the code -- not change the code to match the doc_string. And reading to strings by default is reasonable for a string based format like GML.

What do you think of the new wording I proposed? Any suggestions? Go ahead and make the change to what you think it should be. Once the two of us agree we can request another set of eyes take a look.  Thanks!

randallwvr90:
@dschult I think your wording was good. I replaced it and pushed the changes. Thanks for catching that and suggesting the additional change!

dschult (approved):
Thanks for this @randallwvr90 !

jarrodmillman (approved):

PR #4927, opened 2021-06-24
Add 3.0 migration guide

jarrodmillman:
I am not sure how much more detail to add.  Maybe just pointing to the release notes for details about the deprecated code would be enough for now.

### DIFF
--- a/doc/release/migration_guide_from_2.x_to_3.0.rst
+++ b/doc/release/migration_guide_from_2.x_to_3.0.rst
@@ -1,0 +1,44 @@
+:orphan:
+
+*****************************
+Preparing for the 3.0 release
+*****************************
+
+.. note::
+   Much of the work leading to the NetworkX 3.0 release will be included
+   in the NetworkX 2.6 and 2.7 releases.  For example, we are changing our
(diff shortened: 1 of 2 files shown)


doc/release/migration_guide_from_2.x_to_3.0.rst
```diff
@@ -0,0 +1,44 @@
...
+We plan to release 2.7 near the end of summer and 3.0 near the end of the year.
+
+Default dependencies
+--------------------
+
+We no longer depend on the "decorator" library.
+
+Given their central position in the scientific Python ecosystem and the fact
+that they provide self-contained, easy to install wheels, NumPy, SciPy,
+Matplotlib, and Pandas are now default dependencies.
```
dschult, line 31: Maybe we need to turn this sentence around -- putting the list of libraries first so they don't get lost.

```
NumPy, SciPy, Matplotlib, and Pandas are now default dependencies due to their central position in the scientific Python ecosystem and the fact that they provide self-contained, easy to install wheels for pip.
```

dschult (approved):

PR #4928, opened 2021-06-25
Fix numeric and degree assortativity coefficient calculation

### LINKED ISSUE #4776: numeric assortativity coefficient only works for nonnegative numeric attributes

vpozdnyakov:
It fixes #4776 and wrong assortativity coefficient calculation in some cases that follow PR #4851.

A cause of the issue is in `numeric_ac` in [correlation.py](https://github.com/networkx/networkx/blob/main/networkx/algorithms/assortativity/correlation.py). Before my fixes in `numeric_ac`
```python
x = np.arange(nx)
y = np.arange(ny)
```
we create supports for random variables x and y. We assume that node values are integer and coincide with mixing matrix indices — 0 is in the first row/column, 1 is in the second row/column and so on. But after PR #4851, node values can also be negative or float. In such cases we need to know a mixing matrix indices for each unique pair of values. After my fixes, supports of x and y contain all node values, not only integers. The same for degree assortativity coefficient calculation.

Note that the PR #4851 version also correctly calculates assortativity coefficient in the case of *equal intervals* between support values due to normalization (in fact, this is a correlation between x and y, not covariance). For example, it correctly calculates a degree assortativity coefficient for float node degrees `[0, 0.5, 1, 1.5]` or an assortativity coefficient for mixed node values `[-1, 0, 1, 2]`.  But in #4776 we have a graph with node values `[-1, 1, 2]` and then the old coefficient is incorrect. I added tests that check cases with non-equal intervals between support values — one for degree and one for node values.

### DIFF
--- a/networkx/algorithms/assortativity/correlation.py
+++ b/networkx/algorithms/assortativity/correlation.py
@@ -73,8 +73,12 @@ def degree_assortativity_coefficient(G, x="out", y="in", weight=None, nodes=None
     .. [2] Foster, J.G., Foster, D.V., Grassberger, P. & Paczuski, M.
        Edge direction and the structure of networks, PNAS 107, 10815-20 (2010).
     """
-    M = degree_mixing_matrix(G, x=x, y=y, nodes=nodes, weight=weight)
-    return numeric_ac(M)
+    if nodes is None:
+        nodes = G.nodes
+    degrees = set([G.degree(weight=weight)[n] for n in nodes])
+    mapping = {d: i for i, d, in enumerate(degrees)}
+    M = degree_mixing_matrix(G, x=x, y=y, nodes=nodes, weight=weight, mapping=mapping)
+    return numeric_ac(M, mapping=mapping)
 
 
 def degree_pearson_correlation_coefficient(G, x="out", y="in", weight=None, nodes=None):
@@ -223,8 +227,12 @@ def numeric_assortativity_coefficient(G, attribute, nodes=None):
     .. [1] M. E. J. Newman, Mixing patterns in networks
            Physical Review E, 67 026126, 2003
     """
-    a = numeric_mixing_matrix(G, attribute, nodes)
-    return numeric_ac(a)
+    if nodes is None:
+        nodes = G.nodes
+    vals = set(G.nodes[n][attribute] for n in nodes)
+    mapping = {d: i for i, d, in enumerate(vals)}
+    M = attribute_mixing_matrix(G, attribute, nodes, mapping=mapping)
+    return numeric_ac(M, mapping)
 
 
 def attribute_ac(M):
@@ -254,7 +262,7 @@ def attribute_ac(M):
     return r
 
 
-def numeric_ac(M):
+def numeric_ac(M, mapping):
     # M is a numpy matrix or array
     # numeric assortativity coefficient, pearsonr
     import numpy as np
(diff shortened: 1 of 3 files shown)


networkx/algorithms/assortativity/correlation.py
```diff
@@ -254,20 +262,21 @@ def attribute_ac(M):
...
-    y = np.arange(ny)
+    x = np.array(list(mapping.keys()))
+    y = x  # x and y have the same support
+    idx = list(mapping.values())
     a = M.sum(axis=0)
     b = M.sum(axis=1)
-    vara = (a * x ** 2).sum() - ((a * x).sum()) ** 2
-    varb = (b * x ** 2).sum() - ((b * x).sum()) ** 2
+    vara = (a[idx] * x ** 2).sum() - ((a[idx] * x).sum()) ** 2
+    varb = (b[idx] * x ** 2).sum() - ((b[idx] * x).sum()) ** 2
```
dschult, line 279: I know the values are the same, but could you change “x” to “y” in “varb”?
This also makes me wonder if the mixing matrix for a directed graph ought to have in_degree for rows and out_degree for columns.  But I think that is beyond this issue.  It might be a topic of a whole paper on mixing over graphs (if it hasn’t been written already).

vpozdnyakov: Ok, I will change it.
In the current implementation, a mixing matrix is squared and supports (sets of possible values) for in_degree and out_degree are the same. But there is another way to represent a mixing matrix in a non-squared view. That is as you said, in_degree is for rows and out_degree is for columns (or vice versa). Both are correct, and do not affect assortativity coefficient. The disadvantage of a non-squared view is that it requires separate mappings for rows and columns, say two variables `out_mapping` and `in_mapping` instead of a single `mapping`. The advantage of a non-squared view is that a matrix is much smaller in some limit cases, say in a graph with many trees. For example, a graph



has a 3x3 degree mixing matrix 
```
mapping={1: 0, 3: 1, 4: 2}
```
```
[[0.   0.   0.  ]
 [0.43 0.   0.  ]
 [0.57 0.   0.  ]]
```
or a 2x1 degree mixing matrix
```
out_mapping = {3: 0, 4: 1}
in_mapping = {1: 0}
```
```
[[0.43]
 [0.57]]
```
As we see, if we increase the number of such trees with different number of leafs, the non-squared mixing matrix will still be just a one-column matrix.

dschult: Nice description -- Thanks! 

I think this is ready to merge.  Anything else you know of?

dschult (approved):

networkx/algorithms/assortativity/correlation.py
vpozdnyakov: Yes, it is ready to merge

rossbar (approved):
LGTM, thanks for the followup @vpozdnyakov !

Issue #3745, opened 2019-12-10
Price-matching iGraph's Community Deteection Algorithm


DonaldTsang:
Would like to see more support for Community Detection
See https://www.r-bloggers.com/summary-of-community-detection-algorithms-in-igraph-0-6/
- [ ] Edge-Betweenness
- [ ] Leading Eigenvector
- [ ] Fast-Greedy
- [ ] Multi-Level (also https://github.com/igraph/igraph/issues/890)
- [ ] Walktrap
- [ ] Label Propagation
- [ ] InfoMap

dschult:
[We have some measures](https://networkx.github.io/documentation/stable/reference/algorithms/centrality.html) that are not on that list. And some that are. and some I'm not familiar with. Are you able to make a pull request with any that you feel most important?

DonaldTsang:
@dschult if reading the iGraph source code can provide context, https://igraph.org/c/doc/igraph-Community.html has some documentation for https://github.com/igraph/igraph 
Unfortunately there is no pure Python implementation

DonaldTsang:
Here is a "bible" on many different algorithms and its classification https://arxiv.org/pdf/0906.0612.pdf

ezhil-mathy:
Is this not the implementation of fast-greedy algorithm? [https://github.com/networkx/networkx/blob/master/networkx/algorithms/community/modularity_max.py](url)

guyroznb:
Hi, 
I want to try this topic.
@dschult  which algorithm do you think I should start with?

dschult:
The "bible" paper listed above reports that in their comparisons an algorithm called InfoMap did best.
How about starting with that one?

bsh90:
I would like to work on this. i will try coding the infomap algorithm that i found.
in long-term If i could be on this project, i will read the papers above to be able to implement some other algorithms, etc. (since they require much time)

aaronzo:
@dschult @guyroznb I'd be very happy to implement InfoMap, I'm quite familiar with it. Is there any ongoing work on this?

dschult:
There is no ongoing work on InfoMap that I know of.  But @guyroznb should update us if they are still working on it.

BrandonKMLee:
You guys might want to check this project https://cdlib.readthedocs.io/en/latest/reference/cd_algorithms/node_clustering.html
It is written in Python and integrates iGraph with NetworkX. But some of the algorithms are not mainlined into NetworkX.
For Infomap:
[...]

PR #4929, opened 2021-06-25
Louvain Community Detection Algorithm

z3y50n:
This is the PR regarding the implementation of Louvain Community Detection Algorithm.

Work remaining:
- [x] Write more extensive documentation
- [x] Add more tests
- [x] Add resolution parameter
- [x] Add support for directed graphs (if doable)
- [x] Improve performance for large graphs
- [x] Improve `_gen_graph` function either by improving `nx.quotient_graph` or by manually writing code to generate the new graph based on the communities
- [x] Add `weights` label inside the `quotient_graph`


networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _one_level(G, partition, weight="weight"):
+    """Calculate one level of the tree"""
+    node2com = {u: i for i, u in enumerate(G.nodes())}
+    total_weights = {i: G.degree(u, weight=weight) for i, u in enumerate(G.nodes())}
+    degrees = G.degree(weight=weight)
+    rand_nodes = sample(G.nodes, len(G.nodes))
+    nb_moves = 1
+    m = G.size(weight=weight)
```
dschult, line 65: Minor points that probably won't change much... but might:

1) The `degrees` object is still a DegreeView of the graph, so it still has to compute the degree each time you look it up.  It might be faster to make that a dict:  `dict(G.degree(weight=weight)`.
2) Your `total_weights` can be initialized from `degrees` saving one pass of computing the degrees:
```python
degrees = dict(G.degree(weight=weight))
total_weights = {i: deg for i, deg in enumerate(degrees.values())}
```

3) The `rand_nodes` ordering can be constructed by [random.shuffle](https://docs.python.org/3/library/random.html#random.shuffle) which is built for this kind of operation, so should be faster. You have to form a list first because it changes the order in-place:  
```python
rand_nodes = random.shuffle(list(G.nodes))
```

The size of G (held in variable `m`) can be computed from `degrees` saving one more computation of the degree of each node.
```python
m = sum(degrees.values()) / 2  # for undirected graph "size" is half the sum of degrees
```

Also -- does m change as the communities change (and as we go from level to level?) It might be the same in which case we don't have to recompute it.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+    return weights
+
+
+def _gen_graph(G, partition):
+    """Generate induced graph from given partition"""
+    H = nx.quotient_graph(G, partition, relabel=True)
+    for u in H.nodes:
+        if H.nodes[u]["nedges"] != 0:
+            H.add_edge(u, u, weight=H.nodes[u]["nedges"])
```
dschult, line 109: Should the self_loop weight be “number of edges” between nodes in block “u”,  or “total weight of the edges”?
If it should be total weight, let’s give `quotient_graph` the input node_data(b) the stores b itself (instead of “graph”) and the total weight of edges from nodes in both nodes in b (instead of “nedges”). I think that will speed it up too.

networkx/algorithms/community/louvain.py
z3y50n: I did make the proposed changes thanks.

Indeed the size `m` will always stay the same

networkx/algorithms/community/louvain.py
z3y50n: self_loops are the number of edges between nodes inside the same community only

dschult: Hmmm… then maybe the “size” of the graph does change. We lose information about the weights of the edges within a community if we make the self-loop weight be the number of edges when we are using edge weights for other edges.

z3y50n: Oh there was some misunderstanding, when I say number of edges I mean the weights between nodes of the same community. I interpreted wrongly what you said.

dschult: OK... good -- then we should definitely define `node_data` to retain the original graph nodes in that partition as well as the sum of the edge weights. I think that will be faster than the current code anyway which stores the subgraph induced by the original nodes, the number of nodes, the number of edges and the density.  :}

z3y50n: I tested with defining the `node_data` as you suggested but the timing was still pretty bad so I decided to write my own function to generate the new graph.

The new version takes about 0.5 seconds to generate a new graph of 27k nodes and 1500 communities.

dschult: What was the timing before?  (using quotient_graph)

z3y50n: About 200 seconds

dschult: Wow!   Great job! :}  
That should speed things up a bit!  :}

Makes me want to revisit `quotient_graph`.  I'll open an issue about that.

z3y50n:
There is one roadblock with the `weight` parameter.

If the user inputs `weight=None` then the algorithm won't work because after the first pass when we create a new graph based on the partition there will be weights between the new edges. As a way to bypass this I assign `weight` to be equal to `"weight"` after the first `_one_level` call but this can still cause trouble if the original graph contains `"weights"` that the user didn't want to take into consideration.

One thing that @MridulS suggested was to assign a very unique name that would be very distinguishable, which again bares the minimum risk of having some edge attribute with the same name, or we can create a copy of the original graph at the beginning of the algorithm and naming the attribute that holds the `weights` however we see fit which can add some overhead and complexity.

dschult:
I think instead of copying the graph at the beginning, we can make a new graph with only the weight attribute. Then we don't have to treat each level differently.  Something like this:
```python
graph = G.__class__()
graph.add_weighted_edges(G.edges(data=weight, default=1))
graph.add_nodes_from(G)
```
This should work for multigraphs too, but we should include a test to make sure that is maintained.

dschult:
Some comments on the tests:
- we should set a seed for any random aspects of the tests. That way they don't create intermittent failures that are hard to track down. So, if you are going to use `randint` to generate random weights, you should also import `random.seed` and set a seed for that test.
- The test for `weight=none` should also test that the weights change the partition. Otherwise we can't tell whether using "None" is any different from using "weight". Perhaps it would also be good to also test a name for weight other than "weight".  This would require calling louvain twice -- once with weight="foo" and once with `weight=None`. Then assert that "foo" makes it not match `part` and `None` makes it match `part`.
- The `nx.set_edge_attributes` call doesn't need to be inside a loop. The loop can be put inside the input parameter. 
```python
random.seed(1234)
nx.set_edge_attributes(graph, {edge: randint(1, 20) for edge in graph.edges}, name="foo") 
```

z3y50n:
Yeah I had some similar thoughts, thanks for the suggestions. I am making the changes right away!

z3y50n:
I have a confusion regarding `MultiGraphs`
If I create a `MultiGraph` as follows:
```
G = nx.petersen_graph(create_using=nx.MultiGraph)
print(G.edges)
```
I get the following edges:
```
[(0, 1, 0), (0, 1, 1), (0, 4, 0), (0, 4, 1), (0, 5, 0), (0, 5, 1), (1, 2, 0), (1, 2, 1), (1, 6, 0), (1, 6, 1), (2, 3, 0), (2, 3, 1), (2, 7, 0), (2, 7, 1), (3, 4, 0), (3, 4, 1), (3, 8, 0), (3, 8, 1), (4, 9, 0), (4, 9, 1), (5, 7, 0), (5, 7, 1), (5, 8, 0), (5, 8, 1), (6, 8, 0), (6, 8, 1), (6, 9, 0), (6, 9, 1), (7, 9, 0), (7, 9, 1)]
```
while if I use this way:
```
G = nx.MultiGraph(nx.petersen_graph())
print(G.edges)
```
this is the output:
```
[(0, 1, 0), (0, 4, 0), (0, 5, 0), (1, 2, 0), (1, 6, 0), (2, 3, 0), (2, 7, 0), (3, 4, 0), (3, 8, 0), (4, 9, 0), (5, 7, 0), (5, 8, 0), (6, 8, 0), (6, 9, 0), (7, 9, 0)]
```
So as you can see in the first case each edge is counted twice. Is that normal that these two ways produce different results?

dschult:
The Petersen Graph constructor (generator?) works from an adjacency list data structure that contains both directions for each edge. When create_using is MultiGraph, this will double each edge. When create_using is a DiGraph, it puts edges with both directions. With Graph is creates a single edge -- though it actually does this twice.  Constructing the graph using a nx.Graph puts a single edge. So `nx.MultiGraph(nx.petersen_graph())` has only single edges. While `nx.petersen_graph(create_using=nx.MultiGraph)` has double edges.

Fun fact:  `nx.DiGraph(nx.path_graph(5))` makes an edge going each direction, while `nx.DiGraph(nx.path_graph(5).edges)` makes a single directed edge for each step along the path.

dschult (commented):
Just a couple more suggestions to look at for speed and readability.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,206 @@
...
+    seed.shuffle(rand_nodes)
+    nb_moves = 1
+    improvement = False
+    while nb_moves > 0:
+        nb_moves = 0
+        for u in rand_nodes:
+            best_mod = 0
+            best_com = node2com[u]
+            partition[best_com].difference_update(G.nodes[u].get("nodes", {u}))
+            inner_partition[best_com].remove(u)
```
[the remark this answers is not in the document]
z3y50n: Yeah I can move these statements inside this `if best_com != node2com[u]:` statement in order to make the removal only when there is actually a move to make

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,206 @@
...
+    Input `m` is the size of the graph `G`.
+    Input `partition` is a valid partition of the graph `G` given as a list of sets of
+    nodes.
+    """
+    node2com = {u: i for i, u in enumerate(G.nodes())}
+    inner_partition = [{u} for u in G.nodes()]
+    degrees = dict(G.degree(weight="weight"))
+    total_weights = [deg for deg in degrees.values()]
+    nbrs = {u: dict(G[u]) for u in G.nodes()}
+    rand_nodes = list(G.nodes)
```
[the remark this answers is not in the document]
z3y50n: Good idea I have added the changes

networkx/algorithms/community/louvain.py
```diff
@@ -204,3 +207,16 @@ def _gen_graph(G, partition):
...
         H.add_edge(com1, com2, **{"weight": wt + temp})
     return H
+
+
+def _convert_multigraph(G, weight):
+    """Convert a Multigraph to normal Graph"""
+    H = nx.Graph()
+    H.add_nodes_from(G)
+    for u, v, data in G.edges(data=True):
+        wt = data.get(weight, 1)
```
dschult, line 217: The `G.edges` method can do the data lookup. Not sure that it is any faster, but it's a cleaner interface, maybe:
```
    for u, v, wt in G.edges(data=weight, default=1):
        if H.has_edge(u, v):
            H[u][v]["weight"] += wt
        else:
            H.add_edge(u, v, weight=wt)
    return H
```

z3y50n: Good catch thanks!

dschult:
There is another issue #3112 that asks for community detection algorithms for directed graphs.
The [pdf linked to from that issue ](https://github.com/elplatt/Paper-CNM-Modularity/blob/master/paper.pdf) is not a refereed paper, but gives a reasonable overview with definitions of modularity and directed graph treatments.

Before jumping into that too far, it might be reasonable to think about whether to separate that feature as a separate PR, or just build it in to this one.

z3y50n:
There is also [this paper](https://hal.archives-ouvertes.fr/hal-01231784/document) which is more recent and explains the directed method and provides some good formulas.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,289 @@
...
+        graph = _gen_graph(graph, inner_partition)
+        yield partition
+
+
+def _one_level(G, m, partition, resolution=1, is_directed=False, seed=None):
+    """Calculate one level of the tree
+
+    Input `m` is the size of the graph `G`.
+    Input `partition` is a valid partition of the graph `G` given as a list of sets of
+    nodes.
```
dschult, line 178:
```suggestion
    """Calculate one level of the Louvain partitions tree

    Parameters
    ----------
    G : NetworkX Graph/DiGraph
        The graph from which to detect communities
    m : number
        The size of the graph `G`.
    partition : list of sets of nodes
        A valid partition of the graph `G`
    resolution : positive number
        The resolution parameter for computing the modularity of a partition
    is_directed : bool
        True if `G` is a directed graph.
    seed : integer, random_state, or None (default)
        Indicator of random number generation state.
        See :ref:`Randomness<randomness>`.

```

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,289 @@
...
+    inner_partition = list(filter(len, inner_partition))
+    return partition, inner_partition, improvement
+
+
+def _neighbor_weights(nbrs, node2com):
+    """Calculate weights between node and its neighbor communities.
+
+    Input `nbrs` should be a dict of the node's neighbors.
+
+    `node2com` is a dict with nodes as keys and community index as values.
```
dschult, line 249: Formalize the parameters docs for this function.  Say that 'nbrs' is keyed by neighbor to the edge weight as value.

networkx/algorithms/community/tests/test_louvain.py
```diff
@@ -0,0 +1,115 @@
...
+        {32, 33, 8, 14, 15, 18, 20, 22, 26, 29, 30},
+    ]
+    partition = louvain_communities(G, seed=2)
+
+    assert part == partition
+
+
+def test_none_weight_param():
+    G = nx.karate_club_graph()
+    seed(1234)
```
dschult, line 52: This use of `seed` is not needed because you set `seed=2` below.

And you shouldn't actually need to import seed for this module (I think) because all the seed setting is done in the function calls.

networkx/algorithms/community/tests/test_louvain.py
```diff
@@ -0,0 +1,115 @@
...
+    nx.set_edge_attributes(G, {edge: randint(1, 20) for edge in G.edges}, name="foo")
```
dschult, line 53: This might be a better test by not using `randint`, but instead use `enumerate` or some non-random way to generate the weights. That eliminates any possible intermittent test failures due to random results (which shouldn't happen here, but it's good to eliminate that source of debugging effort to make sure it isn't the part causing the errors).    With this and seed removed you can remove the `import` statement at the top of the module.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,289 @@
...
+        \Delta Q = \frac{k_{i,in}}{m}
+        - \gamma\frac{k_i^{out} \cdot\Sigma_{tot}^{in} + k_i^{in} \cdot \Sigma_{tot}^{out}}{m^2}
+
+    where $k_i^{out}$, $k_i^{in}$ are the outer and inner weighted degrees of node $i$ and
+    $\Sigma_{tot}^{in}$, $\Sigma_{tot}^{out}$ are the sum of in-going and out-going links incident
+    to nodes in $C$.
+
+    The first phase continues until no individual move can improve the modularity.
+
+    > Note: The order in which the nodes are considered can affect the final output.
```
dschult, line 48: Should we add to this "note" a short description of how node order is done in this implementation?

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,289 @@
...
+
+    The second phase consists in building a new network whose nodes are now the communities
+    found in the first phase. Once this phase is complete it is possible to reapply the
```
dschult, line 51: Maybe state how the edge weights are determined for this new network?

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,289 @@
...
+    seed : integer, random_state, or None (default)
+        Indicator of random number generation state.
+        See :ref:`Randomness<randomness>`.
+
+    Returns
+    -------
+    list
+        A list of sets (partition of `G`). Each set represents one community and contains
+        all the nodes that constitute it.
+
```
dschult, line 79: Add a "See Also" section with the `generate_dendogram` function here.  And then in `generate_dendogram` add a "See Also" section with the `louvain_communities` function listed there.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,289 @@
...
+__all__ = ["louvain_communities", "generate_dendrogram"]
+
+
+@py_random_state("seed")
+def louvain_communities(
+    G, weight="weight", resolution=1, threshold=0.0000001, seed=None
+):
+    r"""Find the best partition of G using the Louvain Community Detection
+    Algorithm.
+
```
dschult, line 19: Maybe add a first paragraph description of what Louvain tries to do, with citation to the papers [1] and [2].
A similar description (maybe the exact wording so readers know they've already read it) could go into `generate_dendogram`.

networkx/algorithms/community/tests/test_louvain.py
```diff
@@ -0,0 +1,115 @@
+from random import randint, seed
+
+import pytest
```
dschult, line 3: Not needed?

dschult (commented):
Maybe we should rename `generate_dendogram` since that could be used to describe many approaches and many types of dendograms (e.g. the evolutionary tree-of-life people use dendograms a lot... some for community/species representation, but not Louvain communities).  I'm open to lots of choices...  Maybe `louvain_partitions`?

aaronzo:
Maybe also worth thinking about the thread on #3745 looking at implementing [InfoMap](https://arxiv.org/pdf/0906.1405.pdf) community detection. The optimization algorithm for InfoMap follows the Louvain algorithm (both stages), except that it minimizes 'the map equation' (heuristically this is something like the information flow between communities) rather than maximising modularity.

Perhaps it is worth exploring which pieces of this PR could be decoupled from the modularity equation, if at all. However, performance is quite key with these algorithms and I'm aware that many time gains are achieved by exploiting the mathematical structure of the target equation - as the `_one_level` function does here` - so a naïve approach of allowing a developer to input a cost function to optimise against would be no good.

I see for example, that the 'generate_dendrogram` function could be used for infomap - if we could swap the `_one_level` and the `modularity` function out.

networkx/algorithms/community/louvain.py
z3y50n: Is it suffice to say that the order is determined using networkx's `seed.shuffle`?

dschult: Yes -- though the word `seed` is local to this function, so maybe better to say `random.shuffle` or even leaving out the name of the function and just say:  "... using a random shuffle."

z3y50n:
@aaronzo You are quite right. In order for the algorithm to perform in reasonable times, I had to precompute and store a lot of things and not use other `NetworkX` existing functions, or I couldn't see a good way to generalize some of the functions used by the algorithm and export them to `NetworkX`.

That being said I hope that the structure can guide someone (even myself if/when I get time) with minimum changes to implement other similar algorithms such as InfoMap.

z3y50n:
I observed one minor issue, if there is no modularity gain on the first pass, then the algorithm will yield nothing and the queue will throw an exception because it will be empty.

The easiest fix is to yield the partition where each node is its own community after the following line.

https://github.com/z3y50n/networkx/blob/4c82fa6aa974de7f254b53173603a7071582c27e/networkx/algorithms/community/louvain.py#L158

dschult:
It might be worth unrolling the `while` loop slightly, use`while improvement:` and use `return` instead of `break` inside the loop. 

I also notice that you currently don't yield the last partition computed if the change in modularity is below threshold. It might be good to go ahead and yield that last partition since it is already computed. 

What do you think of something like this? (untested)
```python
    partition, inner_partition, improvement = _one_level(
        graph, m, partition, resolution, is_directed, seed
    )
    yield partition
    while improvement:
        new_mod = modularity(
            graph, inner_partition, resolution=resolution, weight="weight"
        )
        if new_mod - mod <= threshold:
            return
        graph = _gen_graph(graph, inner_partition)
        partition, inner_partition, improvement = _one_level(
            graph, m, partition, resolution, is_directed, seed
        )
        yield partition
```

z3y50n:
Using the proposed way it would yield the final partition twice, once when it is computed and then on the next pass where the will be no improvement and that's why we need to check if there was any improvement before we yield.

However, as I think I have seen in other implementations, if we call _one_level at least once we will solve the above issue that I mentioned.

dschult:
Yup!   You are right.... :}  Many ways to get it right I suspect...

z3y50n:
I made the requested changes. I think the PR is good to go!

dschult (approved):
This looks good to me:}

aaronzo:
@z3y50n I actually think yielding the 'base' partitions is more expected behaviour for a user anyway - I would guess `louvain_partitions` would have more niche usage than `louvain_communities`, I would use it in the case I'm investigating how the algorithm built partitions and I think it's nice to see the first layer combines single nodes.

If you really wanted to avoid it you could do on https://github.com/networkx/networkx/blob/9629af4cdc9757368cd44bea3365ea1fe04b6a2d/networkx/algorithms/community/louvain.py#L110
```python
return d.pop() if d else [{u} for u in G.nodes()]
```
but I think yielding the base case is better as you have done

z3y50n:
@aaronzo Yeah, according to the references and some implementations the base partition is yielded only in the case that there is no modularity gain on the first pass of the algorithm. With the latest changes I made in the `while` loop the algorithm will yield at least once so I think it is the same result as the example you wrote.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,325 @@
...
+__all__ = ["louvain_communities", "louvain_partitions"]
+
+
+@py_random_state("seed")
+def louvain_communities(
+    G, weight="weight", resolution=1, threshold=0.0000001, seed=None
+):
+    r"""Find the best partition of G using the Louvain Community Detection
+    Algorithm.
```
MridulS, line 18: Just some documentation nit picks here:
Could you change this to `partition of a graph` instead of `partition of G`.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,325 @@
...
+    >>> import networkx.algorithms.community as nx_comm
+    >>> G = nx.petersen_graph()
+    >>> nx_comm.louvain_communities(G, seed=123)
+    [{0, 4, 5, 7, 9}, {1, 2, 3, 6, 8}]
+
+    References
+    ----------
+    .. [1] Blondel, V.D. et al. Fast unfolding of communities in
+       large networks. J. Stat. Mech 10008, 1-12(2008)
+    .. [2] Traag, V.A., Waltman, L. & van Eck, N.J. From Louvain to Leiden: guaranteeing
```
MridulS, line 98: This reference isn't referenced anywhere in the text. We should either add a link to it somewhere in the function doc above or remove it.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,325 @@
...
+       well-connected communities. Sci Rep 9, 5233 (2019). https://doi.org/10.1038/s41598-019-41695-z
+    .. [3] Nicolas Dugué, Anthony Perez. Directed Louvain : maximizing modularity in directed networks.
+        [Research Report] Université d’Orléans. 2015. hal-01231784
```
MridulS, line 101: Could you also add the DOI links and the explicit URL for HAL archive.
[1] https://doi.org/10.1088/1742-5468/2008/10/P10008
[3] https://hal.archives-ouvertes.fr/hal-01231784

PS: We should formalise referencing in our documentation.

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,325 @@
...
+    Yields
+    ------
+    list
+        A list of sets (partition of `G`). Each set represents one community and contains
+        all the nodes that constitute it.
+
+    References
+    ----------
+    .. [1] Blondel, V.D. et al. Fast unfolding of communities in
+       large networks. J. Stat. Mech 10008, 1-12(2008)
```
MridulS, line 157: https://doi.org/10.1088/1742-5468/2008/10/P10008

networkx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,325 @@
...
+        \Delta Q = \frac{k_{i,in}}{m}
+        - \gamma\frac{k_i^{out} \cdot\Sigma_{tot}^{in} + k_i^{in} \cdot \Sigma_{tot}^{out}}{m^2}
+
+    where $k_i^{out}$, $k_i^{in}$ are the outer and inner weighted degrees of node $i$ and
+    $\Sigma_{tot}^{in}$, $\Sigma_{tot}^{out}$ are the sum of in-going and out-going links incident
+    to nodes in $C$.
+
+    The first phase continues until no individual move can improve the modularity.
+
+    > Note: The order in which the nodes are considered can affect the final output. In the algorithm
```
MridulS, line 51: This note should probably live under 
```
 Notes
 -----
```
so sphinx picks this up properly. For an example look at https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.approximation.clique.large_clique_size.html?highlight=notes

MridulS (approved):
Looks great to me! Just a little bit of nits picks here and there.

z3y50n:
Thanks for your suggestions, I made the changes

dschult:
This looks good!  Thanks @z3y50n !!

MridulS:
Yay! This is amazing :)

PR #4930, opened 2021-06-25
quotient_graph doc fix

awohns:
Small fix to the documentation of `networkx.algorithms.minors.quotient_graph` which I think makes it a bit clearer. The docs refer to `node_relation` which isn't a parameter and isn't referred to elsewhere. Basically it should just be referring to the partition.


dschult:
Thanks!  This is helpful!!

dschult (approved):

rossbar (approved):
Thanks @awohns

PR #4931, opened 2021-06-26
move partition checking outside private _quotient_graph function

dschult:
Making the private function free from argument checking may allow Louvain method code to call the private version to avoid argument checking because we know the inputs are ok.

### DIFF
--- a/networkx/algorithms/minors/contraction.py
+++ b/networkx/algorithms/minors/contraction.py
@@ -51,6 +51,12 @@ def equivalence_classes(iterable, relation):
         Duplicate elements will be ignored so it makes the most sense for
         `iterable` to be a :class:`set`.
 
+    Notes
+    -----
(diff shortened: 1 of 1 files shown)


networkx/algorithms/minors/contraction.py
```diff
@@ -325,9 +338,7 @@ def _quotient_graph(
     relabel=False,
     create_using=None,
 ):
-    # Each node in the graph must be in exactly one block.
-    if any(sum(1 for b in partition if v in b) != 1 for v in G):
-        raise NetworkXException("each node must be in exactly one block")
+    """Construct the quotient graph assuming inout has been checked"""
```
z3y50n, line 341: It should be "input"

rossbar (approved):
LGTM, thanks @dschult & @z3y50n !

PR #4932, opened 2021-06-26
Page number for Katz centrality reference

ankona:
Minor fix for Katz centrality documentation. Page referenced in `Networks, An Introduction` appears wrong when I went to read the reference:

image


dschult (approved):
Yes!  Thanks -- the right page number is helpful. :}

Issue #4936, opened 2021-06-28
refactor assortativity subpackage


dschult:
The `assortativity` subpackage is poorly tests, uses poorly chosen function names and given the number of issues raised about it, should be refactored.  The distinction between `numeric` and `attribute` should be made clear (numeric seems to have been chosen to mean "positive integer valued attributes" and attribute to mean "scalar valued attributes") and, if reasonable, `numeric` codes removed (`numeric_mixing_matrix` is currently slated to be removed). The helper functions `attribute_ac` and `numeric_ac` also need a careful look and perhaps removal -- they aren't tested and recent issues show that there were assumptions made for `numeric_ac` that were not made clear enough for us to see easily. 

Another feature of this subpackage is the dual data structures of dicts (as a pure-python sparse matrix storage object) and `numpy.array`s.  This is reasonable, but there is a lot of going back and forth between the two which should probably be reduced. Maybe only one should be dominantly used or else the docs should be expanded to describe to users when to use one over the other.

PR #4937, opened 2021-06-28
Deprecate k_nearest_neighbors

dschult:
The function name `k_nearest_neighbors` isn't used n any of the cited papers, is the same name as used for commonly used machine learning algorithms (which are not related to this degree_correlation work.  Perhaps it came from "nearest neighbor degree" which **is** used in the papers, and does depend on "k", but importantly, is not called "k nearest neighbors".  Since this function name is simply an alias we should just redirect users to `average_degree_connectivity` which is the name used for the actual function definition.  That name is not great either given that there is a function `average_neighbor_degree`, but that naming discussion can be delayed to another PR. (see issue #4936)

This PR also fixes handling of degree/in_degree/out_degree for average_neighbor_degree and adds some very simple tests for it.   More testing is needed in this assortativity package.

### DIFF
--- a/doc/developer/deprecations.rst
+++ b/doc/developer/deprecations.rst
@@ -89,3 +89,4 @@ Version 3.0
 * In ``linalg/graphmatrix.py`` remove ``adj_matrix``.
 * In ``algorithms/similarity.py`` replace ``simrank_similarity`` with ``simrank_similarity_numpy``.
 * In ``algorithms/assortativity/mixing.py`` remove ``numeric_mixing_matrix``.
+* In ``algorithms/assortativity/connectivity.py`` remove ``k_nearest_neighbors``.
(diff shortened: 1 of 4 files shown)


rossbar (approved):
+1 for deprecating `k_nearest_neighbors` - I took the liberty of adding a bullet to `doc/developer/deprecations.rst` as well.

I would have a slight preference for splitting the deprecation and testing + `average_neighbor_degree` improvements into separate PRs or at least not squashing the commits here just to keep the history granular enough in case something needs to be reverted. That said, my approval applies to the additional changes as well!

dschult:
#4948 now has the `average_neighbor_degree` part of this PR.  So, this PR is now only the deprecation of `k_nearest_neighbors`.

jarrodmillman:

jarrodmillman:

PR #4945, opened 2021-06-30
fix spelling in docstring of conftest.py

dschult:


rossbar (approved):
Nice catch, thanks @dschult !

PR #4948, opened 2021-06-30
Fix neighbor degree for directed graphs

dschult:
Make `average_neighbor_degree` work for directed graphs, add tests and inline one function to simplify.

### DIFF
--- a/networkx/algorithms/assortativity/neighbor_degree.py
+++ b/networkx/algorithms/assortativity/neighbor_degree.py
@@ -118,17 +118,3 @@ def average_neighbor_degree(G, source="out", target="out", nodes=None, weight=No
         else:
             avg[n] = sum(G_n[nbr].get(weight, 1) * tgt_deg[nbr] for nbr in G_n) / deg
     return avg
-
-
-# obsolete
-# def average_neighbor_in_degree(G, nodes=None, weight=None):
-#     if not G.is_directed():
-#         raise nx.NetworkXError("Not defined for undirected graphs.")
-#     return _average_nbr_deg(G, G.in_degree, G.in_degree, nodes, weight)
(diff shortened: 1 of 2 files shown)


jarrodmillman:

dschult:
PR 4946 ran the circleCI hook without trouble. But that  isn't a branch on my repo.  Maybe all  the PRs from my repo are giving trouble. I used to have circleCI as a hook on my repo. But I disabled the hook earlier this year because we thought it might be causing me trouble. There are no hooks appearing on the repo at the moment.  Maybe I should check the circleCI side... but I don't know what to even check for there.

dschult:
When I look at the NetworkX repo on circleci.com there is a report of all pipelines. You can select a branch to look at, and there are many to select from. None of the PRs I have opened are on that list (e.g. this one: 4948).  The webhooks log on GitHub shows this pr creation event was sent to circleci at 17:07 on June 30 (my time I guess), and the “response” if very similar to the response for other PR creation events.  But clearly circleci is handling it differently.  

Anyway, I’m not sure why circleci doesn’t seem to have this pr branch on its list of branches to see the pipeline log.

dschult:
OK.... Something changed -- though I can see many of you are active and creating "Actions" for circleci so I don't know if it is something someone else did.  But the circleCI is now working after my latest push of a spurious commit.

I **did** log into circleci.com and stopped "following" networkx. While looking around to start that process I noticed that actions for circleci were being reported against github.com/dschult/networkx when I had made the previous pushes.  That was why I stopped following networkx. I'm not sure what "following" does on circleci.com and I don't know if it is following my fork (dshcult/networkx) or the main repo (networkx/networkx).

Anyway, I hope it is all working now.

PR #4952, opened 2021-07-06
`descendants_at_distance` also for non-DiGraphs

michaeldorner:
I think there is no reason to limit `descendants_at_distance` to DiGraphs. My change affects the documentation only.  If I am wrong, maybe a check would make sense.

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -374,7 +374,7 @@ def descendants_at_distance(G, source, distance):
 
     Parameters
     ----------
-    G : NetworkX Graph
+    G : NetworkX graph
         A graph
     source : node in `G`
     distance : the distance of the wanted nodes from `source`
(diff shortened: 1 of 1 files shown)


rossbar (approved):
There's definitely a mismatch here as the function does indeed work (at least for simple cases) with undirected graphs. I think you're correct that the docs should be updated to reflect this fact, rather than changing the function to disallow undirected graphs.

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -374,8 +374,8 @@ def descendants_at_distance(G, source, distance):
 
     Parameters
     ----------
-    G : NetworkX DiGraph
-        A directed graph
+    G : NetworkX Graph
```
dschult, line 377:
```suggestion
    G : NetworkX graph
```
lower case on "graph" removes confusion with the `Graph` object type.

dschult (approved):
Thanks!

PR #4965, opened 2021-07-09
modularity_max: provide labels to get_edge_data

mathilde-leval:
Fixes [bug](https://github.com/networkx/networkx/issues/4966) which makes `greedy_modularity_communities` crash if the graph nodes have been relabeled.

### DIFF
--- a/networkx/algorithms/community/modularity_max.py
+++ b/networkx/algorithms/community/modularity_max.py
@@ -92,7 +92,9 @@ def greedy_modularity_communities(G, weight=None, resolution=1):
     a = [k[i] * q0 for i in range(N)]
     dq_dict = {
         i: {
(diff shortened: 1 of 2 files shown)


dschult:
Looks like this builds on #4690   Thanks for this!

rossbar (commented):
Instead of modifying `test_modularity_communities_weighted` in-place, it might be better to add an additional test, say `test_modularity_communities_weighted_relabeled` so that there are tests for both cases.

rossbar (approved):
I went ahead and added a quick note to the new test linking it back to gh-4966. LGTM - thanks @mathilde-leval !

mathilde-leval:
Ok, thanks @rossbar !

dschult (approved):

PR #4976, opened 2021-07-16
Add multigraph betweenness

pinselimo:
This PR adds betweenness centrality (BC) for multigraphs. The function ```betweenness_centrality``` is currently marked as ```@not_implemented_for("multigraph")``` however ```edge_betweenness_centrality``` is not even though it suffers from the same shortcomings. With this PR both functions should yield correct results.

The PR consists of three commits addressing issues with BC for multigraphs:

### Issues

**1. Edge betweenness centrality lacks ```not_implemented``` decorator**
The first commit marks ```èdge_betweenness_centrality``` as not implemented for multigraphs.
It suffers from the issues described below and should at least be marked correctly.

**2. Weighted betweenness centrality disregards weight of parallel edges**
The second commit resolves an issue regarding both betweenness for nodes and edges. For multigraphs ```G[v].items()``` returns ```w, edgedata``` like usual Graphs. However, ```edgedata``` differs in that it is first indexed by edge keys instead of weight keys. The implementation hence always calculates unweighted betweenness.

~~~python
>>> G = MultiGraph()
>>> G.add_edges_from(
            ((0, 1, {'w':20.8}), (1, 2, {'w':1.5}), (0, 2, {'w':1.8}))
            )
>>> edge_betweenness_centrality(G, weight='w')
{(0, 1): 0.3333333333333333,
(0, 2): 0.3333333333333333,
(1, 2): 0.3333333333333333}   
>>> # Correct would be: 0.0, 0.66.., 0.66...
~~~

For parallel edges we always have to consider the lowest weight. Fortunately, the weighted shortest paths module provides ```_weight_function```. Utilizing this import solves the issue.

However, this fix can only enable correct betweenness centrality for nodes. Edge betweenness centrality still suffers from the following issue.

**3. Edge betweenness centrality does not return edge keys**
Calculating edge betweenness centrality gives the values but not which edge key to apply them to.
This is especially a problem with the example given in [```set_edge_attributes```](https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.set_edge_attributes.html). It does not work with multigraphs because of this design, however it would not fail to calculate the betweenness.

Also, parallel edges of equal weight should divide their value.

### Correctness

Correctness was checked using property based tests with a sample size of 1000. They are contained in my [other branch](https://github.com/pinselimo/networkx/tree/check-multigraph-betweenness) in ```examples/algorithms/betweenness/check_properties.py```.

```pytest``` has also been checked, although I haven't added additional tests so far. Give me a heads up if you deem them necessary.
And of course all code has been formatted with ```black```.

Thanks,
Simon

### DIFF
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -5,12 +5,12 @@ import warnings
 
 from networkx.utils import py_random_state
 from networkx.utils.decorators import not_implemented_for
+from networkx.algorithms.shortest_paths.weighted import _weight_function
 
 __all__ = ["betweenness_centrality", "edge_betweenness_centrality", "edge_betweenness"]
 
 
 @py_random_state(5)
-@not_implemented_for("multigraph")
 def betweenness_centrality(
     G, k=None, normalized=True, weight=None, endpoints=False, seed=None
 ):
@@ -147,7 +147,6 @@ def betweenness_centrality(
 
 
 @py_random_state(4)
-@not_implemented_for("multigraph")
 def edge_betweenness_centrality(G, k=None, normalized=True, weight=None, seed=None):
     r"""Compute betweenness centrality for edges.
 
@@ -277,6 +276,7 @@ def _single_source_shortest_path_basic(G, s):
 
 
 def _single_source_dijkstra_path_basic(G, s, weight):
+    weight = _weight_function(G, weight)
     # modified from Eppstein
     S = []
     P = {}
@@ -299,7 +299,7 @@ def _single_source_dijkstra_path_basic(G, s, weight):
         S.append(v)
         D[v] = dist
         for w, edgedata in G[v].items():
-            vw_dist = dist + edgedata.get(weight, 1)
+            vw_dist = dist + weight(v,w,edgedata)
             if w not in D and (w not in seen or vw_dist < seen[w]):
                 seen[w] = vw_dist
                 push(Q, (vw_dist, next(c), v, w))
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -236,6 +236,11 @@ def edge_betweenness_centrality(G, k=None, normalized=True, weight=None, seed=No
     betweenness = _rescale_e(
         betweenness, len(G), normalized=normalized, directed=G.is_directed()
     )
+    if G.is_multigraph():
+        betweenness = _add_edge_keys(G, betweenness, weight=weight)
+        for e in G.edges():  # Remove edges without key
+            if e in betweenness:
+                del betweenness[e]
     return betweenness
 
 
@@ -299,7 +304,7 @@ def _single_source_dijkstra_path_basic(G, s, weight):
         S.append(v)
         D[v] = dist
         for w, edgedata in G[v].items():
-            vw_dist = dist + weight(v,w,edgedata)
+            vw_dist = dist + weight(v, w, edgedata)
             if w not in D and (w not in seen or vw_dist < seen[w]):
                 seen[w] = vw_dist
                 push(Q, (vw_dist, next(c), v, w))
@@ -394,4 +399,20 @@ def _rescale_e(betweenness, n, normalized, directed=False, k=None):
             scale = scale * n / k
         for v in betweenness:
             betweenness[v] *= scale
+    return betweenness
+
+
+def _add_edge_keys(G, betweenness, weight=None):
+    _weight = _weight_function(G, weight)
+    betweenness.update(dict.fromkeys(G.edges(keys=True), 0.0))
+
+    get_keys = lambda u, v, d: [
+        k for k in d if not weight or d[k].get(weight, 1) == _weight(u, v, d)
+    ]
+
+    for u, v in G.edges():
+        keys = get_keys(u, v, G[u][v])
+        for k in keys:
+            betweenness[(u, v, k)] = betweenness[(u, v)] / len(keys)
+
     return betweenness
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -404,6 +404,34 @@ def _rescale_e(betweenness, n, normalized, directed=False, k=None):
 
 @not_implemented_for("graph")
 def _get_edge_keys(G, weight):
+    r"""Similar to `_weight_function`. Returns a function that
+    returns the keys of edges with the lowest weight between
+    two nodes.
+
+    This function is only suitable for multigraphs.
+
+    Parameters
+    ----------
+    G : NetworkX graph.
+
+    weight : string or function
+        If it is callable, `weight` itself is returned. If it is a string,
+        it is assumed to be the name of the edge attribute that represents
+        the weight of an edge. In that case, a function is returned that
+        gets the edge weight according to the specified edge attribute.
+
+    Returns
+    -------
+    function
+        This function returns a callable that accepts exactly three inputs:
+        a node `u`, a node adjacent to the first one `v` and the edge attribute
+        dictionary `d` for the edge joining those nodes. That function returns
+        a list with integers representing the keys of those edges matching
+        the lowest `weight` of edges between `u` and `v`.
+
+    If any edge does not have an attribute with key `weight`, it is assumed to
+    have weight one.
+    """
     _weight = _weight_function(G, weight)
 
     def get_keys(u, v, d):
@@ -414,6 +442,26 @@ def _get_edge_keys(G, weight):
 
 
 def _add_edge_keys(G, betweenness, weight=None):
+    r"""Adds the corrected betweenness centrality (BC) values for multigraphs.
+
+    Parameters
+    ----------
+    G : NetworkX graph.
+
+    betweenness : dictionary
+        Dictionary mapping adjacent node tuples to betweenness centrality values.
+
+    weight : string or function
+        See `_get_edge_keys` for details. Defaults to `None`.
+
+    Returns
+    -------
+    edges : dictionary
+        The parameter `betweenness` including edges with keys and their
+        betweenness centrality values.
+
+    The BC value is divided among edges of equal weight.
+    """
     _keys = _get_edge_keys(G, weight)
     betweenness.update(dict.fromkeys(G.edges(keys=True), 0.0))
 
--- a/networkx/algorithms/centrality/betweenness.py
+++ b/networkx/algorithms/centrality/betweenness.py
@@ -402,16 +402,23 @@ def _rescale_e(betweenness, n, normalized, directed=False, k=None):
     return betweenness
 
 
-def _add_edge_keys(G, betweenness, weight=None):
+@not_implemented_for("graph")
+def _get_edge_keys(G, weight):
     _weight = _weight_function(G, weight)
+
+    def get_keys(u, v, d):
+        w = _weight(u, v, d)
+        return [k for k in d if d[k].get(weight, 1) == w]
+
+    return get_keys
+
+
+def _add_edge_keys(G, betweenness, weight=None):
+    _keys = _get_edge_keys(G, weight)
     betweenness.update(dict.fromkeys(G.edges(keys=True), 0.0))
 
-    get_keys = lambda u, v, d: [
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by dschult** [comment]
Thank you very much for this!!

- The `edge_betweenness_centrality` function should definitely use `not_implemented_for`.
- Using the `_weight_function` to process the keydict correctly is precisely the way I'd suggest to handle it.
- Converting the edge 2-tuples to edge 3-tuples with form (node, nbr, key) should be able to be done by finding the minimum weight edge -- just as you do in the `_weight_function` only returning the edge key instead of the weight.

It might actually be better to just convert the multigraph to a graph so you only compute the shortest edge between two nodes once.  But that presumably uses slightly more memory to construct this smaller graph. 
Thoughts?

And, of course, it would be good to have tests to make sure the MultiGraph case isn't ignoring weights. But let's decide which way to present this example first (building _weight_function and something like _find_edge vs. converting G to a Graph with only the shortest edges between each pair).

**Comment by pinselimo** [comment]
The conversion is not necessary as ```_single_source_dijkstra_path_basic``` already only computes every neighbor node only once. Using ```_weight_function``` it now defaults to the minimal weight between any two nodes.

Merely adding the keys with the lowest weight to the dictionary won't respect cases where multiple edges bear the same weight. In that case the edge BC needs to be divided among those edges. Therefore, ```_add_edge_keys``` reuses ```_weight_function``` to find those keys and adjust their weights accordingly.
The overhead is pretty minimal, but could be removed by turning the ```get_keys``` lambda expression into a named function. This would call ```_weight``` only once per neighbor instead of for each edge.

**Comment by dschult** [comment]
Yes, Could you make the `get_keys` lambda expression into a named function?
Thanks!

**Comment by dschult** [comment]
Can you add some simple test?
Also for line 429, the edge keys are not necessarily integers. so just call them the edge keys.
Thanks!

**Comment by pinselimo** [comment]
> Can you add some simple test?

Added for BC and edge BC. Also saw that there were some tests without docstrings and added those.

> Also for line 429, the edge keys are not necessarily integers. so just call them the edge keys.
> Thanks!

Ah yes, got a little misguided there. Fixed.

File: networkx/algorithms/centrality/tests/test_betweenness_centrality.py [hunk 1]
```diff
@@ -655,3 +696,80 @@ def test_normalized_weighted_graph(self):
...
             assert b[n] == pytest.approx(b_answer[n] / norm, abs=1e-7)
+
+    def test_weighted_multigraph(self):
+        """Edge betweenness centrality: weighted multigraph"""
+        eList = [
+            (0, 1, 5),
+            (0, 2, 4),
+            (0, 3, 3),
+            (0, 3, 3),
+            (0, 4, 2),
```
dschult, line 707: This test uses a graph with identical multiedge from `(0, 3)` but those two edges tie with each other and no other shortest path.  Is there a way to test a multiedge that ties not only with its parallel twin, but also with a path that doesn't traverse this node-pair?  Otherwise we never fully test the correct splitting of ties.  I'm getting muddled trying to come up with an example, but is this testing idea clear?

pinselimo: Adding ```(0, 1, 4)``` should do, as currently the shortest weighted path from 0 to 1 is of length 4 via path ```[0,3,1]```.
Then we have two paths from 0 to 1 using different edges. This is the request, right?

networkx/algorithms/centrality/betweenness.py
```diff
@@ -236,6 +236,11 @@ def edge_betweenness_centrality(G, k=None, normalized=True, weight=None, seed=No
     betweenness = _rescale_e(
         betweenness, len(G), normalized=normalized, directed=G.is_directed()
     )
+    if G.is_multigraph():
+        betweenness = _add_edge_keys(G, betweenness, weight=weight)
+        for e in G.edges():  # Remove edges without key
```
dschult, line 241: Can you help me understand why we need to remove the edges in `betweenness` that don't have a key?
Do some edges have keys and other don't?  Are we sure we don't throw out any edges that we might actually want?  e.g. what if there is only one edge between a pair of nodes. Will this throw out the betweenness for that edge?

pinselimo: The reasoning behind removing those edges lies in my aim to comply with the first example given for [```set_edge_attributes```](https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.set_edge_attributes.html). I think it shouldn't be possible (especially for novices) to encounter an example that throws an error. Leaving them in would cause ```
ValueError: not enough values to unpack (expected 3, got 2)``` in ```networkx.functions``` [line 811](https://github.com/pinselimo/networkx/blob/add-multigraph-betweenness/networkx/classes/function.py#L811).

There are no edges without a key. Adding an edge on a ```MultiGraph``` using ```add_edge``` with ```key=None``` (the default value) will create and add a new key. This is also the return value of ```add_edge``` in ```MultiGraph```s. (Same goes for ```MultiDiGraph```. ```add_edges_from``` uses ```add_edge``` internally).

I initially added it outside the function because in [line 234](https://github.com/pinselimo/networkx/blob/add-multigraph-betweenness/networkx/algorithms/centrality/betweenness.py#L234) nodes are removed in the same scope. But I will move the code inside the function call.

networkx/algorithms/centrality/betweenness.py
```diff
@@ -394,3 +397,78 @@ def _rescale_e(betweenness, n, normalized, directed=False, k=None):
...
+    betweenness.update(dict.fromkeys(G.edges(keys=True), 0.0))
+
+    for u, v in G.edges():
+        keys = _keys(u, v, G[u][v])
+        for k in keys:
+            betweenness[(u, v, k)] = betweenness[(u, v)] / len(keys)
+
+    for e in G.edges():  # Remove edges without key
+        if e in betweenness:
+            del betweenness[e]
```
dschult, line 472: Nice! Thanks....
And  now it becomes more clear to me that we seem to be doing extra work here... We update the betweenness dict with 3-tuple edges and values 0 and then we update each of those. Finally we remove any dict entry with a 2-tuple edge.  BUT.... all the original entries will be deleted and all the new entries will be kept. So we could just make a new dict and avoid the extra adds and deletes.  Perhaps something like this?

```python
return {(u, v, k): betweenness[(u, v)] / len(keys) 
    for u, v in G.edges()
    for keys in [_keys(u, v, G[u][v])] 
    for k in keys
}
``` 
Notice the `for keys in ` loop uses brackets around the _keys to avoid having to call it twice inside a dict comprehension.

It might also be significantly faster to in-line the `_get_edge_keys` function since it is a private function and only used in this one loop.  Something like:
```python
    _weight = _weight_function(G, weight)
    result = {}
    for u, v in G.edges():
        d = G[u][v]
        wt = _weight(u, v, d)
        keys = [k for k in d if d[k].get(weight, 1) == wt]
        for k in keys:
            result[(u, v, k)] = betweenness[(u, v)] / len(keys)
    return result
```
If the timing isn't much different then use which ever is easier to read.

pinselimo: Going from an empty dict would "forget" parallel edges with ```weight(e) > min(weights[(u,v)])``` . I also found that ```G.edges()``` will go over **all** edges but omit keys. If there are two edges ```(0,1,0), (0,1,1)``` the loop would go over ```(0,1)``` twice.

My solution now will create a new dict from the edges with keys and assign their BCs. I tried using a dict-comprehension because I'd favor its readability but it is vastly slower since we need to check for each key if we need to assign ```0.0``` or ```bc/len(keys)```. The variant submitted is the fastest alternative.

dschult: Excellent!   Thanks!

networkx/algorithms/centrality/betweenness.py
```diff
@@ -394,3 +397,39 @@ def _rescale_e(betweenness, n, normalized, directed=False, k=None):
...
+
+    The BC value is divided among edges of equal weight.
+    """
+    _weight = _weight_function(G, weight)
+
+    edge_bc = dict.fromkeys(G.edges, 0.0)
+    for u, v in betweenness:
+        d = G[u][v]
+        wt = _weight(u, v, d)
+        keys = [k for k in d if d[k].get(weight, 1) == wt]
```
dschult, line 430: Are we being consistent here in how we handle the `weight` parameter?  Are we going to allow it to be "a function or a string" or just "a string".  I think the current code works if it is a string, but doesn't work if `weight` is a function.  That's because the `d[k].get(weight, 1)` won't work correctly if `weight` is a function. 

If we restrict `weight` to be a string, we should probably avoid using `_weight_function` and just use a loop over all edges between the node-pairs, computing the minimal weight *and* the number of edges with that weight.  But it'd be nice to have this all work with a `weight` function too.  I'll think about how to do that and am open to suggestions.

dschult: I think we should build our own _weight_function here and prepare for using a more flexible `_weight_function` type decorator like the one discussed in #4085 and #4114.  Something like this:
```python
If not callable(weight):
    _weight = weight
    def weight(u, v, key_dict):
        min_wt, keys = float(‘inf’), [] 
        for k, data_dict in key_dict.items():
            wt = data_dict.get(_weight, 1)
            if wt < min_wt:
                min_wt, keys = wt, [k]
            elif wt == min_wt:
                keys.append(k)
        return min_wt, keys 
```

pinselimo: Good catch! I think having a global solution definitely favors maintainability. However, I don't think until then our solution should be polymorphic in the return type. If a user hands us a callable weight we'd fail trying to query the keys.

An interim solution would be to create a dict with just the edge value:

~~~
keys = [k for k in d if _weight(u, v, {k: d[k]}) == wt]
~~~

This minimizes code duplication and is consistent with the usage of current callable weights. I've tested it already.

If other parts of networkx also need to be revisited regarding how they act on multigraphs, I don't think inlining getting those keys has such a big impact on performance. So this solution could be reused without having to impose changes to any callable weights.
That said, I think having a polymorphic return type like proposed in the global solution is highly advisable. This would enforce authors to think about what cases with parallel edges of equal weight.

dschult (approved):
I approve this PR.... Thanks!

networkx/algorithms/centrality/betweenness.py
dschult: Nice trick for making the weight function treat a multiedge like it is the only edge between those nodes. :}

rossbar (commented):
LGTM, just one comment about a linked function in a docstring that doesn't seem to exist.

From reading through the conversation in the PR, I was confused by one thing: it seems like the original consensus was that `edge_betweenness_centrality` should be marked with `not_implemented_for` multigraphs, but it looks like the opposite happened (some new tests were added for EBC w/ multigraph inputs). I was just curious if/where the change in decision was discussed - I might just be missing something in a "resolved conversation" somewhere...

networkx/algorithms/centrality/betweenness.py
```diff
@@ -394,3 +397,39 @@ def _rescale_e(betweenness, n, normalized, directed=False, k=None):
...
+
+    Parameters
+    ----------
+    G : NetworkX graph.
+
+    betweenness : dictionary
+        Dictionary mapping adjacent node tuples to betweenness centrality values.
+
+    weight : string or function
+        See `_get_edge_keys` for details. Defaults to `None`.
```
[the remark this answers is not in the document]
pinselimo: Good catch, thanks!

pinselimo:
> From reading through the conversation in the PR, I was confused by one thing: it seems like the original consensus was that `edge_betweenness_centrality` should be marked with `not_implemented_for` multigraphs, but it looks like the opposite happened (some new tests were added for EBC w/ multigraph inputs). I was just curious if/where the change in decision was discussed - I might just be missing something in a "resolved conversation" somewhere...

The initial problem was indeed the missing ```not_implemented_for``` so I added an initial commit to fix it. From there I went and enabled multigraph betweenness for both nodes and edges. This was all part of my initial proposal. 
I wanted to make sure the issue is fixed even if my proposal of enabling multigraphs is rejected.

rossbar (approved):
> The initial problem was indeed the missing not_implemented_for so I added an initial commit to fix it. From there I went and enabled multigraph betweenness for both nodes and edges. This was all part of my initial proposal.

Got it, thanks for the clarification. LGTM!

PR #4984, opened 2021-07-26
Remove mentioning of gdal

felixonmars:
It doesn't seem to be used anywhere now. Also 2.5's changelog says it was removed.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -78,9 +78,9 @@ Extra packages
 --------------
 
 .. note::
-   Some optional packages (e.g., `gdal`) may require compiling
-   C or C++ code.  If you have difficulty installing these packages
-   with `pip`, please consult the homepages of those packages.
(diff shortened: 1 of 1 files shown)


jarrodmillman:
In 2.5, I removed gdal from the requirements file (see https://github.com/networkx/networkx/commit/5638e1ff).  These instructions tell you how to install gdal manually.  In 2.6, we deprecated gdal.  But our support for it will not be removed until the 3.0 release.

So this shouldn't be merged until after we remove the use of gdal before the 3.0 release and after the 2.7 release.

jarrodmillman:
It is used here:  networkx/readwrite/nx_shp.py.

felixonmars:
Ah I see. Thanks for the info!

jarrodmillman:
Thanks for this PR!  We are planning to release 2.7 in a couple of months and 3.0 near the end of the year.  We will merge this after 2.7 and before 3.0.

PR #5022, opened 2021-08-14
Style changes

dschult:
I'm pulling these style changes out of PR #4740 to ease review of both PRs.  Thanks to @mjschwenne 

As far as I can tell, these are all changes which do not affect the results of `black` but do shorten the code and make it easier to read.  Most simply merge lines multiline expressions into a single line.


dschult:
Something didn't work right... :} closing...

dschult:
OK... I think it is fixed now.  The only changes should be pulling together lines that would work for black.

PR #5041, opened 2021-08-28
Added leverage centrality algorithm

nirav0999:
1. Added leverage centrality algorithm for weighted and unweighted undirected graphs. <br>
2. Added the leverage centrality documentation. <br>
3. Added the leverage centrality test cases. <br>

All test cases are successful including the previously maintained ones and the newly added ones for leverage centrality,


z3y50n (commented):
Very nice work. I have made some minor suggestions that could slightly improve performance.

networkx/algorithms/centrality/leverage.py
```diff
@@ -0,0 +1,108 @@
+"""Functions for computing eigenvector centrality."""
+
+from math import sqrt
```
z3y50n, line 3: Unused Import

networkx/algorithms/centrality/leverage.py
```diff
@@ -0,0 +1,108 @@
...
+    # Iterating over all the nodes
+    for n in G.nodes():
+        # Initial leverage centrality
+        LCsum_n = 0
+        degree_n = degrees[n]
+        # Leverage centrality is undefined for a node with degree 0
+        if degree_n == 0:
+            raise nx.NetworkXPointlessConcept(
+                "cannot compute centrality for a node with no degree = 0", n
+            )
```
z3y50n, line 97: You can check for zero degree outside the `for` loop on line 89
```Python
if 0 in degrees:
    raise nx.NetworkXPointlessConcept(
        "cannot compute centrality for a node with no degree = 0"
    )
```
That way we won't do unnecessary computations until that node's turn.

networkx/algorithms/centrality/leverage.py
```diff
@@ -0,0 +1,108 @@
...
+        # Getting all the neighbors of the node n
+        neighbors = G.neighbors(n)
+        # Iterating through all the neighbors and getting their 'share' from the algo
+        for neighbor in neighbors:
+            degree_neighbor = degrees[neighbor]
+            LCsum_n += (degree_n - degree_neighbor) / (degree_n + degree_neighbor)
```
z3y50n, line 103: You could use `sum` like this:
```Python
LCsum_n = sum((degree_n - degrees[nbr]) / (degree_n + degrees[nbr]) for nbr in neighbors)
```

Of course since it is Python you can write the entire `for` loop in one line like this:
```Python
leverage_cent = {u: sum((degrees[u] - degrees[nbr]) / (degrees[u] + degrees[nbr]) for nbr in G[u]) / degrees[u] for u in G.nodes()}
```
but it would certainly decrease readability imo

networkx/algorithms/centrality/tests/test_leverage_centrality.py
```diff
@@ -0,0 +1,53 @@
...
+        G = nx.path_graph(3)
+        G.edges[0, 1]["weight"] = 3
+        G.edges[1, 2]["weight"] = 5
+        b_answer = {0: -0.15151, 1: 0.0856643, 2: -0.0461538}
+        b = nx.leverage_centrality(G, weight="weight")
+        for n in sorted(G):
+            assert b[n] == pytest.approx(b_answer[n], abs=1e-4)
+
+
+class TestLeverageCentralityExceptions:
```
z3y50n, line 38: We should add a test to check a non-empty Graph that has a node with 0 degree to check the other exception case.

PR #5086, opened 2021-09-15
Fix lowest_common_ancestors (issue #4942)

jamestrimble:
This fixes the bug in `lowest_common_ancestors` using the [change](https://github.com/networkx/networkx/issues/4942#issuecomment-919675945) proposed by @dschult.

The discussion in issue #4942 includes an extra (naive) algorithm. Should we add this as an additional option, or perhaps just use it for testing?

We should also add a test case to make sure the bug doesn't reappear.


jamestrimble:
Thanks @rossbar for the review. I agree with your nits and I've finally got round to adding the fixes :-)

MridulS (approved):
Added @dschult as a co-author to https://github.com/networkx/networkx/commit/9a7633f3fda15ceb0c1566f1a2770283e782aeab

PR #5095, opened 2021-09-21
Get number of edges by calling the proper method (issue #5092)

SimoneGasperini:
Minor improvement related to the issue #5092.

All the occurences of `sum(1 for _ in G.edges())` in ./networkx/generators/tests/test_random_graphs.py have been replaced with `G.number_of_edges()`.


rossbar (approved):
LGTM - IMO `G.number_of_edges()` is more readable than `sum(1 for _ in G.edges())`. Thanks @SimoneGasperini !

dschult (approved):
Looks Good!   Thanks!

PR #5104, opened 2021-09-26
Added condition to count only once self-loop edges

zozo123:
The vertex_cover algorithm updated the cost of u,v twice
in case node v and node u are the same nodes. As a result
self-edges nodes are overlooked. Here we fixed the cost map upating
to occur only once if v==u.

#5103


dschult:
It seems that changing `==` to `<=` in the `return` line allows self-loops to be handled too. 

But, it appears to me that the algorithm is implemented inefficiently for python. We are subtracting the cost to indicate membership in a set.  We then subtract 0 from future lookups involving that node.  Better to add the minimal cost node to the set right away and replace the future subtractions with a check for membership in the cover.  In my simple tests, this takes less than half the time and, of course, fixes the self-loop issue as well.  

```python
    cost = dict(G.nodes(data="weight", default=1))
    cover = set()
    for u, v in G.edges:
        if u in cover or v in cover:
            continue
        if cost[u] <= cost[v]:
            cover.add(u)
            cost[v] -= cost[u]
        else:
            cover.add(v)
            cost[u] -= cost[v]
    return cover
```

Either way, we need some tests to make sure the successful treatment of self-loops is not broken by later changes.
Can you add a test to `tests/test_vertex_cover.py`?

zozo123:
@dschult, thank you for contributing the snippet. I have updated the PR accordingly.

PR #5114, opened 2021-10-05
Deprecate union name param

z3y50n:
As discussed on #5099 I created a separate PR to deprecate the unused `name` parameter from `union` function


rossbar (approved):
Thanks for splitting this out @z3y50n - LGTM!

MridulS (approved):

PR #5121, opened 2021-10-09
Faster operators in algorithms/operators/all.py

kpberry:
TL;DR: I rewrote the operators in algorithms/operators/all.py to take linear time instead of quadratic time, and I rewrote the corresponding operators in algorithms/operators/binary.py in terms of the *_all operators instead of the other way around.

Hey everyone! I noticed that the operators in algorithms/operators/all.py (e.g., union_all, intersection_all, etc.), simply call the corresponding operators from algorithms/operators/binary.py repeatedly, which causes their run times to all be quadratic with respect to the number of graphs. However, they can all be implemented easily in linear time, so I went ahead and did that. Since the binary operations are also special cases of the *_all operations and there's not a significant overhead to process multiple graphs vs. a pair of graphs, I rewrote the binary operators in terms of the *_all operators (which keeps the duplicated code to a minimum).

This refactor doesn't change the signatures/outputs of these functions, so I don't think any documentation change is directly necessary for this (though there are other documentation changes for these files in #5099).

### DIFF
--- a/networkx/algorithms/operators/all.py
+++ b/networkx/algorithms/operators/all.py
@@ -53,7 +53,7 @@ def union_all(graphs, rename=(None,)):
     U = graphs[0]
 
     if any(G.is_multigraph() != U.is_multigraph() for G in graphs):
-        raise nx.NetworkXError('All graphs must be graphs or multigraphs.')
+        raise nx.NetworkXError("All graphs must be graphs or multigraphs.")
 
     # rename graph to obtain disjoint node labels
     def add_prefix(graph, prefix):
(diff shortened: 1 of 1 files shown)


kpberry:
I noticed this issue when performing the union of hundreds of small, mostly disjoint graphs (graphs for which the size of their unions will most often be the sum of their individual sizes). The operation was taking in excess of 30 seconds originally, but was almost instant (much less than 1 second) after this refactor.

The formal argument for the run time is based on the size of the output graphs at each step. I'll make the argument for union/union_all, since the argument for disjoint_union and compose is similar.
  
If you have N graphs with V nodes and E edges (on average), then calling the union operation repeatedly to perform union_all will take O((V + E) N^2) steps. The first union will result in a graph with between V and 2V nodes and between E and 2E edges. Assuming the worst case of 2V + 2E, the second union will be combining a graph of size 2V + 2E with a graph of size V + E, which can result in a graph of size 3V + 3E. Continuing like this, the output sizes after each union will be V + E, 2V + 2E, 3V + 3E, ..., NV + NE. Since the original implementation of the union operation constructs a new graph of size at most 2V + 2E, it must take at least 2V + 2E steps to do so. Thus, the total number of steps will be at least (V + E) + (2V + 2E) + (3V + 3E) + ... + NV + NE = (V + E) * N (N + 1) / 2, which is in  O((V + E) N^2).

The original intersection algorithm was actually linear because the size of the output graph was non-increasing, so we only need to consider the best case output size in the argument above, yielding O((V + E) N) steps. However, it still seems best to me to include union_all in the refactor in order to be consistent with the other implementations.

The new union algorithm avoids the quadratic run time by constructing one new graph and adding each node/edge from each of the input graphs exactly once, which should result in a run time which is proportional to the sum of the number of input edges and vertices (assuming vertex/edge insertion are O(1)).

PR #5131, opened 2021-10-15
Allow edge style to be a list of styles for DiGraphs

mgbckr:
Fixes issue where `DiGraph`s fail to allow a list of edge styles (or when enabling `arrows=True` for `Graph`):

```python
import networkx as nx

g = nx.DiGraph()
g.add_nodes_from([0,1,2])
g.add_edges_from([(0,1), (0,2)])

nx.draw_networkx_nodes(g, pos=[(0,0), (0,1), (1,0)])
nx.draw_networkx_edges(g, pos=[(0,0), (0,1), (1,0)], style=["-", "--"],)
```

This throws an error:
```
ValueError: Unrecognized linestyle: ['-', '--']
```

The proposed pull request fixes this issue.

### DIFF
--- a/networkx/drawing/tests/test_pylab.py
+++ b/networkx/drawing/tests/test_pylab.py
@@ -145,7 +145,12 @@ def test_linestyle():
     np = pytest.importorskip("numpy")
 
     def test_styles(edges, style="solid"):
-        """Function to test actually set styles"""
+        """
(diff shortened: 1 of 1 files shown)


dschult:
Thanks for this idea and for the code!   :}

Could you make the if/else logic match that for `width` which is just before your code for `linestyle`?
Also, could you add tests for ensure that it is working?

PR #5184, opened 2021-11-17
Implement Dial's algorithm

AndrewEckart:
I would like to contribute an implementation of [Dial's algorithm](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm#Specialized_variants), which computes single-source shortest paths in graphs with non-negative integer edge weights. 

I have drafted a simple implementation which runs in *O(E+WV)* time using a bucket queue (where *W* is the maximum edge weight). This works well when *W* is a small constant.

Currently, the public API for shortest paths includes about a dozen different signatures for Dijkstra's algorithm. I have included just one for Dial's algorithm, `nx.dial_predecessors_and_distance`, but as with Dijkstra this relies on an internal function which can easily be used to support alternate signatures.

Open to feedback!

### DIFF
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -4,7 +4,7 @@ Shortest path algorithms for weighted graphs.
 
 from collections import deque
 from heapq import heappush, heappop
-from itertools import count
+from itertools import count, starmap
 import networkx as nx
 from networkx.algorithms.shortest_paths.generic import _build_paths_from_predecessors
 
@@ -35,6 +35,7 @@ __all__ = [
     "find_negative_cycle",
     "goldberg_radzik",
     "johnson",
+    "dial_predecessor_and_distance",
 ]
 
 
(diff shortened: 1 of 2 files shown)


networkx/algorithms/shortest_paths/weighted.py
```diff
@@ -2396,3 +2397,200 @@ def dist_path(v):
...
+    ValueError
+        If any edge weight in `G` is not a non-negative integer.
+
+
+    Notes
+    -----
+    The optional predecessor and path dictionaries can be accessed by
+    the caller through the original pred and paths objects passed
+    as arguments. No need to explicitly return pred or paths.
+
```
dschult, line 2456: Please add a reference to a description of Dial's algorithm.  Here and in the public-facing function.

dschult (commented):
Can you add this function to the docs by adding it to `doc/reference/algorithms/shortest_paths.rst`?
Then we can also look at the doc formatting in the "details" link of the last test (ci-circleci: Build Artifact).

PR #5232, opened 2021-12-17
Allowing None edges in weight function of bidirectional Dijkstra

NikHoh:
added following feature also to bidirectional Dijkstra: The weight function can be used to hide edges by returning None.

### DIFF
--- a/networkx/algorithms/shortest_paths/weighted.py
+++ b/networkx/algorithms/shortest_paths/weighted.py
@@ -2288,16 +2288,11 @@ def bidirectional_dijkstra(G, source, target, weight="weight"):
             return (finaldist, finalpath)
 
         for w, d in neighs[dir][v].items():
-            if dir == 0:  # forward
-                cost = weight(v, w, d)
-                if cost is None:
-                    continue
-                vwLength = dists[dir][v] + cost
-            else:  # back, must remember to change v,w->w,v
-                cost = weight(w, v, d)
-                if cost is None:
-                    continue
-                vwLength = dists[dir][v] + cost
+            # weight(v, w, d) for forward and weight(w, v, d) for back direction
+            cost = weight(v, w, d) if dir == 0 else weight(w, v, d)
+            if cost is None:
+                continue
+            vwLength = dists[dir][v] + cost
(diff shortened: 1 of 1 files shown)


networkx/algorithms/shortest_paths/weighted.py
```diff
@@ -2285,9 +2289,15 @@ def bidirectional_dijkstra(G, source, target, weight="weight"):
...
+                cost = weight(v, w, d)
+                if cost is None:
+                    continue
+                vwLength = dists[dir][v] + cost
             else:  # back, must remember to change v,w->w,v
-                vwLength = dists[dir][v] + weight(w, v, d)
+                cost = weight(w, v, d)
+                if cost is None:
+                    continue
+                vwLength = dists[dir][v] + cost
```
dschult, line 2300: Nice idea!  I hadn't realized this feature wasn't available.

The code might be more readable by reducing duplication. What do you think of:
```python
    cost = weight(v, w, d) if dir == 0 else weight(w, v, d)
    if cost is None:
        continue
    vwLength = dists[dir][v] + cost
```

NikHoh: That should also do the job ;)

dschult (approved):
This looks good.  We should also make a note in the docstring that this feature is available. The docstring for dijkstra may be good enough:
```
    Notes
    -----
    Edge weight attributes must be numerical.
    Distances are calculated as sums of weighted edges traversed.
    
    The weight function can be used to hide edges by returning None.
    So ``weight = lambda u, v, d: 1 if d['color']=="red" else None``
    will find the shortest red path.
    
    Based on the Python cookbook recipe (119466) at
    https://code.activestate.com/recipes/119466/
    
    This algorithm is not guaranteed to work if edge weights
    are negative or are floating point numbers
    (overflows and roundoff errors can cause problems).
```
If that's too much for this PR, just say so and we'll merge this and put in the docs separately.

NikHoh:
The adapted docstring is already part of the 1st commit

jarrodmillman (approved):

PR #5238, opened 2021-12-20
Allows dictionary keys to have dots (.) in their keys during GML I/O

aanastasiou:
While trying to [figure out GML serialisation](https://github.com/networkx/networkx/discussions/5233), I realised that GML dictionaries could not have keys that contained dots. This PR attempts to fix that.

### DIFF
--- a/networkx/readwrite/gml.py
+++ b/networkx/readwrite/gml.py
@@ -300,7 +300,7 @@ def parse_gml_lines(lines, label, destringizer):
 
     def tokenize():
         patterns = [
-            r"[A-Za-z][0-9A-Za-z_]*\b",  # keys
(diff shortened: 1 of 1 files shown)


jarrodmillman:
According to the GML format, it looks like periods are not allowed.  Is there a newer specification of the format?  Do you know if other packages that support the GML format allow periods?

See page 2 of this file for more details:
- https://web.archive.org/web/20190303094704/http://www.fim.uni-passau.de:80/fileadmin/files/lehrstuhl/brandenburg/projekte/gml/gml-technical-report.pdf

aanastasiou:
Indeed, spotted it later too, trying to access some of these fields from yEd :/

PR #5242, opened 2021-12-27
Update imports to use lazy loading

dschult:
This PR should only be merged **after** #4909 

This PR changes the `import numpy as np` by moving it to the top of the module and using `np = nx.lazy_imports.load("numpy")`, with similar treatment for the other libraries.

Another PR will change the NetworkX loading process to ensure lazy loading of all internal packages (subpackages).


networkx/__init__.py
```diff
@@ -51,7 +51,9 @@ def __getattr__(name):
     raise AttributeError(f"module {__name__} has no attribute {name}")
 
 
-# These are import orderwise
+# These are imported in order as listed
+from networkx import lazy_imports
```
stefanv, line 55: Why not use lazy imports to make this lazily accessible? :smile:

Maybe expose `nx.lazy_load` as a function?

PR #5247, opened 2022-01-05
Minor changes to speed up asynchronous label propagation for community detection.

leojklarner:
The innermost loop of the asynchronous label propagation algorithm counts the frequencies of labels among a node's neighbours.

The current implementation initialises a collections.Counter object, loops through all neighbouring nodes, and adds each of them to the counter:

```
label_freq = Counter()
for v in G[node]:
        label_freq.update({labels[v]: G.edges[node, v][weight] if weight else 1})
```

In cases where the counts are not weighted, it is much more efficient to directly update the counter with a list of labels:

```
label_freq = Counter()
label_freq.update([labels[v] for v in G[node]])
```

A dummy script shows that the direct update from a list is more than 10x faster than the loop.

```
import random
from collections import Counter
random_list = [random.randrange(100) for _ in range(1000)]
simple_map = {v: v+1 for v in random_list}
```
```
%%timeit
counter1 = Counter()
counter1.update([simple_map[v] for v in random_list])
```
69.9 µs ± 994 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
%%timeit
counter2 = Counter()
for v in random_list:
    counter2.update({simple_map[v]:1})
```
826 µs ± 23.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

I also modified the skip-condition for nodes with zero neighbours to just use the clearer and more efficient boolean value of the adjacency list and re-ordered/adjusted the comments to better reflect what's happening in the code.

### DIFF
--- a/networkx/algorithms/community/label_propagation.py
+++ b/networkx/algorithms/community/label_propagation.py
@@ -81,7 +81,7 @@ def asyn_lpa_communities(G, weight=None, seed=None):
             # making the algorithm asynchronous.
             label_freq = Counter()
             if weight is None:
-                label_freq.update([labels[v] for v in adj])
+                label_freq.update(map(labels.get, adj))
             else:
                 for v in adj:
                     label_freq.update({labels[v]: G.edges[node, v][weight]})
--- a/networkx/algorithms/community/label_propagation.py
+++ b/networkx/algorithms/community/label_propagation.py
@@ -71,20 +71,24 @@ def asyn_lpa_communities(G, weight=None, seed=None):
         seed.shuffle(nodes)
 
         for node in nodes:
-            adj = G[node]
-            if not adj:
+
+            if not G[node]:
                 continue
 
             # Get label frequencies among adjacent nodes.
             # Depending on the order they are processed in,
             # some nodes will be in iteration t and others in t-1,
             # making the algorithm asynchronous.
-            label_freq = Counter()
             if weight is None:
-                label_freq.update(map(labels.get, adj))
+                # initialising a Counter from an iterator of labels is
+                # faster for getting unweighted label frequencies
+                label_freq = Counter(map(labels.get, G[node]))
             else:
-                for v in adj:
-                    label_freq.update({labels[v]: G.edges[node, v][weight]})
+                # updating a defaultdict is substantially faster
+                # for getting weighted label frequencies
+                label_freq = defaultdict(float)
+                for _, v, wt in G.edges(node, data=weight, default=1):
+                    label_freq[labels[v]] += wt
 
             # Get the labels that appear with maximum frequency.
             max_freq = max(label_freq.values())
--- a/networkx/algorithms/community/label_propagation.py
+++ b/networkx/algorithms/community/label_propagation.py
@@ -64,31 +64,38 @@ def asyn_lpa_communities(G, weight=None, seed=None):
 
     labels = {n: i for i, n in enumerate(G)}
     cont = True
+
     while cont:
         cont = False
         nodes = list(G)
         seed.shuffle(nodes)
-        # Calculate the label for each node
+
         for node in nodes:
-            if len(G[node]) < 1:
+            adj = G[node]
+            if not adj:
                 continue
 
-            # Get label frequencies. Depending on the order they are processed
-            # in some nodes with be in t and others in t-1, making the
-            # algorithm asynchronous.
+            # Get label frequencies among adjacent nodes.
+            # Depending on the order they are processed in,
+            # some nodes will be in iteration t and others in t-1,
+            # making the algorithm asynchronous.
             label_freq = Counter()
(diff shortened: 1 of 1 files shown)


dschult:
That makes good sense.  It is faster (even for the cases where counts are weighted) to use an iterator rather than a loop over `update` calls, or even a single update over a constructed dict (or list).  We want to both reduce the calls to `update` and reduce creating dicts or lists which just get sent to `update`.

How about this:
```python
if weight is None:
    edge_info = iter(G[node])
else:
    edge_info = ((v, edge_wt) for _, v, edge_wt in G.edges(node, data=weight, default=1))
label_freq = Counter()
label_freq.update(edge_info)
```
I get this as much faster for both weighted and unweighted cases.  Do you get it faster for the unweighted case?
And more importantly, did I do it right? Are we still getting the same answers?

leojklarner:
Thank you for the reply.

Using an iterator instead of an explicitly created list is a good idea. However, the counter needs to use the nodes' current labels as keys, rather than the node IDs, so you'd need to add an additional call to the labels dict for each neighbour.

Using an iterator for the unweighted case gives an additional 25% speed-up over the list comprehension:
```
edge_info = map(labels.get, G[node])
```
however the [Counter's update method](https://docs.python.org/3/library/collections.html#collections.Counter) does not accept iterables of key-value pairs, so
```
edge_info = ((label[v], wt) for _, v, wt in G.edges(node, data=weight, default=1))
label_freq.update(edge_info)
```
would just count the frequencies of (label, edge_weight) pairs, while
```
edge_info = {label[v]: wt for _, v, wt in G.edges(node, data=weight, default=1)}
label_freq.update(edge_info)
```
would just return the edge weight of the last occurrence of each label.

So given my understanding that the edge weights should be used as multipliers for the count of the corresponding label, it seems that either an incremental for-loop or a list with `wt` repetitions of the corresponding label is needed, of which the former is preferable in my opinion.

networkx/algorithms/community/label_propagation.py
```diff
@@ -81,7 +81,7 @@ def asyn_lpa_communities(G, weight=None, seed=None):
             # making the algorithm asynchronous.
             label_freq = Counter()
             if weight is None:
-                label_freq.update([labels[v] for v in adj])
+                label_freq.update(map(labels.get, adj))
```
dschult, line 84: I prefer `label_freq.update(labels[v] for v in adj)` for readability. Does it make a difference for speed?

networkx/algorithms/community/label_propagation.py
```diff
@@ -64,31 +64,38 @@ def asyn_lpa_communities(G, weight=None, seed=None):
...
-                label_freq.update(
-                    {labels[v]: G.edges[node, v][weight] if weight else 1}
-                )
-            # Choose the label with the highest frecuency. If more than 1 label
-            # has the highest frecuency choose one randomly.
+            if weight is None:
+                label_freq.update(map(labels.get, adj))
+            else:
+                for v in adj:
+                    label_freq.update({labels[v]: G.edges[node, v][weight]})
```
dschult, line 87: I see what you mean about `Counter` treating an iterator differently than I was expecting. It seems that Counter is slower than a dict for weighted counting.  We should be able to speed up what we've got (my tests with 2000 nodes give >3 times speedup after switching to a dict).
```python
label_freq={labels[v]:0 for v in G[node]}
for _, v, wt in G.edges(node, data=weight, default=1):
    label_freq[labels[v]] += wt
```

networkx/algorithms/community/label_propagation.py
leojklarner: The generator expression is 30% slower than the list comprehension for me, which is in turn 30% slower than the map expression. I agree that the list comprehension is much cleaner, but also think that 30% is a relatively sizeable speed-up.

networkx/algorithms/community/label_propagation.py
leojklarner: Great point, the defaultdict approach is also much faster for me when using weighted counts. For unweighted counts it's still faster to initialise a Counter from an iterator/list, but as both are dict subclasses it shouldn't be an issue to just use a Counter if `weight=None` and a dict otherwise.

networkx/algorithms/community/label_propagation.py
dschult: I am also surprised that the generator is slower than the list comprehension.
That might depend on the Python version. I'm using 3.8
In any case, the `map` version is indeed faster.  I don' t get 30% faster. 
But I'm testing the whole function, not just the update line.

Thanks!

leojklarner:
I implemented the defaultdict approach mentioned above. Thank you for the feedback and taking the time to review!

dschult (approved):
Thanks!   LGTM

PR #5258, opened 2022-01-14
Update relabel.py to preserve node order

SultanOrazbayev:
This preserves the original order of the nodes, as discussed in https://stackoverflow.com/a/70706215/10693596

### DIFF
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -5,6 +5,7 @@ __all__ = ["convert_node_labels_to_integers", "relabel_nodes"]
 
 def relabel_nodes(G, mapping, copy=True):
     """Relabel the nodes of the graph G according to a given mapping.
+    
     The original node ordering is not preserved if the mapping includes overlap
     between old and new labels. 
 
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -134,8 +134,7 @@ def _relabel_inplace(G, mapping):
         D = nx.DiGraph(list(mapping.items()))
         D.remove_edges_from(nx.selfloop_edges(D))
         try:
-            # only keep the nodes in the original graph
-            nodes = [n for n in nx.topological_sort(D) if n in mapping.keys()]
+            nodes = reversed(list(nx.topological_sort(D)))
         except nx.NetworkXUnfeasible as err:
             raise nx.NetworkXUnfeasible(
                 "The node label sets are overlapping and no ordering can "
@@ -149,8 +148,14 @@ def _relabel_inplace(G, mapping):
     directed = G.is_directed()
 
     for old in nodes:
-        new = mapping.get(old, old)
-        G.add_node(new, **G.nodes.get(old, {}))
+        # Test that old is in both mapping and G, otherwise ignore.
+        try:
+            new = mapping[old]
+            G.add_node(new, **G.nodes[old])
+        except KeyError:
+            continue
+        if new == old:
+            continue
         if multigraph:
             new_edges = [
                 (new, new if old == target else target, key, data)
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -118,7 +118,8 @@ def relabel_nodes(G, mapping, copy=True):
     if not hasattr(mapping, "__getitem__"):
         m = {n: mapping(n) for n in G}
     else:
-        m = mapping
+        # enforce a complete mapping (if only partial mapping is defined)
+        m = {n: mapping.get(n, n) for n in G}
     if copy:
         return _relabel_copy(G, m)
     else:
@@ -126,9 +127,6 @@ def relabel_nodes(G, mapping, copy=True):
 
 
 def _relabel_inplace(G, mapping):
-    # enforce a complete mapping (if only partial mapping is defined)
-    mapping = {n: mapping.get(n, n) for n in G}
-
     if len(mapping.keys() & mapping.values()) > 0:
         # labels sets overlap
         # can we topological sort and still do the relabeling?
--- a/networkx/relabel.py
+++ b/networkx/relabel.py
@@ -126,13 +126,17 @@ def relabel_nodes(G, mapping, copy=True):
 
 
 def _relabel_inplace(G, mapping):
+    # enforce a complete mapping (if only partial mapping is defined)
+    mapping = {n: mapping.get(n, n) for n in G}
+
     if len(mapping.keys() & mapping.values()) > 0:
         # labels sets overlap
         # can we topological sort and still do the relabeling?
         D = nx.DiGraph(list(mapping.items()))
         D.remove_edges_from(nx.selfloop_edges(D))
         try:
-            nodes = reversed(list(nx.topological_sort(D)))
+            # only keep the nodes in the original graph
+            nodes = [n for n in nx.topological_sort(D) if n in mapping.keys()]
         except nx.NetworkXUnfeasible as err:
             raise nx.NetworkXUnfeasible(
                 "The node label sets are overlapping and no ordering can "
@@ -146,14 +150,8 @@ def _relabel_inplace(G, mapping):
     directed = G.is_directed()
 
     for old in nodes:
-        # Test that old is in both mapping and G, otherwise ignore.
-        try:
-            new = mapping[old]
-            G.add_node(new, **G.nodes[old])
-        except KeyError:
-            continue
-        if new == old:
-            continue
+        new = mapping.get(old, old)
+        G.add_node(new, **G.nodes.get(old, {}))
         if multigraph:
             new_edges = [
                 (new, new if old == target else target, key, data)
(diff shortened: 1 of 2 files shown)


dschult:
This look good.  To avoid constructing these node sets `old_labels` and `new_labels` can you change the start of this function to:
```
if len(mapping.keys() & mapping.values()) > 0:
```
It was originally written before we could take advantage of the set methods of `keys` and `values`.
Thanks!

dschult (approved):
This Looks good to me!
Thanks!!

SultanOrazbayev:
Thank you, great to contribute a tiny bit to networkx! :)

rossbar (commented):
The change LGTM, it might be nice to add a test for this to ensure any future changes don't cause a regression. It might also be good to add a quick note to the docstring that the original node ordering is guaranteed to be preserved.

SultanOrazbayev:
Sounds good, I will add a test and updated doc-string before next Monday.

SultanOrazbayev:
While working on the test, it turns out that the current fix is an incomplete solution. It doesn't work if mapping keys are sorted in a different order OR if mapping is partial. I added two test cases, and have in mind one solution where `remapping` is sorted in the order of original nodes (to enforce the same order), but this doesn't feel like a nice solution somehow...

SultanOrazbayev:
A bit confused why this doc test is failing:

```python
import networkx as nx

G = nx.MultiGraph()

G.add_edge(0, 1, value="a")
G.add_edge(0, 2, value="b")
G.add_edge(0, 3, value="c")

mapping = {1: 4, 2: 4, 3: 4}
H = nx.relabel_nodes(G, mapping, copy=False) # using modified function
print(H[0])
#Expected:
#   {4: {0: {'value': 'a'}, 1: {'value': 'b'}, 2: {'value': 'c'}}}
#Got:
#   {4: {1: {'value': 'a'}, 2: {'value': 'b'}, 3: {'value': 'c'}}}
```

networkx/relabel.py
```diff
@@ -116,44 +118,38 @@ def relabel_nodes(G, mapping, copy=True):
...
-    if len(old_labels & new_labels) > 0:
+    if len(mapping.keys() & mapping.values()) > 0:
         # labels sets overlap
         # can we topological sort and still do the relabeling?
         D = nx.DiGraph(list(mapping.items()))
         D.remove_edges_from(nx.selfloop_edges(D))
         try:
-            nodes = reversed(list(nx.topological_sort(D)))
+            # only keep the nodes in the original graph
+            nodes = [n for n in nx.topological_sort(D) if n in mapping.keys()]
```
dschult, line 137: This change should not be made. The topological sort is needed to handle the order of the changes that allow in-place relabeling to work.  e.g. if we change `1 -> 3` and then later change `3->4`, that's not what the mapping asked for.  I think we cannot leave the node order unchanged if we have an in-place relabeling with overlapping node sets.

networkx/relabel.py
```diff
@@ -116,44 +118,38 @@ def relabel_nodes(G, mapping, copy=True):
...
-        # Test that old is in both mapping and G, otherwise ignore.
-        try:
-            new = mapping[old]
-            G.add_node(new, **G.nodes[old])
-        except KeyError:
-            continue
-        if new == old:
-            continue
+        new = mapping.get(old, old)
+        G.add_node(new, **G.nodes.get(old, {}))
```
dschult, line 152: This change should also be backed out to make it work with overlapping node sets.

networkx/relabel.py
```diff
@@ -4,7 +4,9 @@
 
 
 def relabel_nodes(G, mapping, copy=True):
-    """Relabel the nodes of the graph G according to a given mapping.
+    """Relabel the nodes of the graph G according to a given mapping. Note
+    that the original node ordering is preserved (even if mapping is defined
+    partially).
```
dschult, line 9: Keep the first line as it was. Add the note as a paragraph separated from the first line by a blank line.
That's because the Sphinx documentation system uses the first line (up to the first blank line) as a short description of the function when it is referred to in the docs.

dschult:
> A bit confused why this doc test is failing:

It looks like the original edge had key 0. 
Then the new edges got keys 1, 2, 3 and the original edge is then removed.
I haven't worked through all the logic of the new code for this case, but that's what is happening.

SultanOrazbayev:
I added three tests, and the code should preserve order if there is a complete mapping OR copy is set to True.

networkx/tests/test_relabel.py
```diff
@@ -299,3 +299,27 @@ def test_relabel_circular(self):
...
+        H = nx.relabel_nodes(G, mapping, copy=False)
+        new_order = list(H.nodes())
+        assert [mapping.get(i, i) for i in original_order] == new_order
+
+    def test_relabel_preserve_node_order_partial_mapping_with_copy_true(self):
+        G = nx.path_graph(3)
+        original_order = list(G.nodes())
+        mapping = {0: "a", 1: "b"}  # partial mapping on purpose
+        H = nx.relabel_nodes(G, mapping, copy=True)
+        new_order = list(H.nodes())
```
dschult, line 324: Maybe make the partial order test also have keys out of order?
And consider changing `list(G.nodes())` to either `list(G)` or `list(G.nodes)`, but I'll leave that up to you.

dschult (commented):
Just a couple of nits to pick... :}  Thanks!

networkx/tests/test_relabel.py
SultanOrazbayev: Ah, interesting, I didn't know I could do `list(G)`. I changed both suggestions (out of order mapping and `list(G)`).

networkx/relabel.py
```diff
@@ -4,9 +4,9 @@
 
 
 def relabel_nodes(G, mapping, copy=True):
-    """Relabel the nodes of the graph G according to a given mapping. Note
-    that the original node ordering is preserved (even if mapping is defined
-    partially).
+    """Relabel the nodes of the graph G according to a given mapping.
+    The original node ordering is not preserved if the mapping includes overlap
+    between old and new labels.
```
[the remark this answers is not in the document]
SultanOrazbayev: Looks good, I wasn't sure about the blank line, I think it's already there, so hopefully this is ok.

SultanOrazbayev:
Adding tests was very valuable as it wasn't clear initially that this wouldn't always work for partial remapping.

dschult:
Yes -- and adding tests will make sure we don't change that partial behavior unintentionally.
Would it make sense to have a test for a case that the order is not preserved so we know if that changes?
I'm not sure how easy that would be.

SultanOrazbayev:
Hmm, let me think about this one more.

networkx/tests/test_relabel.py
```diff
@@ -299,3 +299,27 @@ def test_relabel_circular(self):
...
+        new_order = list(H)
+        assert [mapping.get(i, i) for i in original_order] == new_order
+
+    def test_relabel_preserve_node_order_partial_mapping_with_copy_true(self):
+        G = nx.path_graph(3)
+        original_order = list(G)
+        mapping = {1: "a", 0: "b"}  # partial mapping and keys out of order on purpose
+        H = nx.relabel_nodes(G, mapping, copy=True)
+        new_order = list(H)
+        assert [mapping.get(i, i) for i in original_order] == new_order
```
dschult, line 325: It looks likie this could work for a test where the node order changes with `copy=False` and a partial mapping.
```suggestion
    def test_relabel_preserve_node_order_partial_mapping_with_copy_true(self):
        G = nx.path_graph(3)
        original_order = list(G)
        mapping = {1: "a", 0: "b"}  # partial mapping and keys out of order on purpose
        H = nx.relabel_nodes(G, mapping, copy=True)
        new_order = list(H)
        assert [mapping.get(i, i) for i in original_order] == new_order

    def test_relabel_preserve_node_order_partial_mapping_with_copy_false(self):
        G = nx.path_graph(3)
        original_order = list(G)
        mapping = {1: "a", 0: "b"}  # partial mapping and keys out of order on purpose
        H = nx.relabel_nodes(G, mapping, copy=False)
        new_order = list(H)
        assert [mapping.get(i, i) for i in original_order] != new_order
```

SultanOrazbayev:
Thank you for the suggestion and apologies for the delay in responding!

networkx/relabel.py
```diff
@@ -139,8 +140,8 @@ def _relabel_inplace(G, mapping):
                 "resolve the mapping. Use copy=True."
             ) from err
     else:
-        # non-overlapping label sets
-        nodes = old_labels
+        # non-overlapping label sets, sort them in the order of G nodes
+        nodes = list(n for n in G if n in mapping.keys())
```
dschult, line 144:
```suggestion
        nodes = [n for n in G if n in mapping]
```
I think this will be faster and easier to read.

dschult (commented):
Nice test.  I have one performance/readability suggestion. 
Then this LGTM!

dschult (approved):

jarrodmillman (approved):

MridulS (approved):

PR #5259, opened 2022-01-14
Modified the isomorphvf2.py so that it recognizes some subgraph matching edges cases that weren't recognized by the algorithm

Melissa1412:
The original version of isomorphvf2 was not able to find the matching in subgraph matching cases like the following: 
It is obvious that the query graph (right) is a subgraph of the target graph (left), but the subgraph_isomorphisms_iter object returns nothing. 
target_graph                query_graph


dschult:
A subgraph must be an "induced subgraph".  So your example is not a subgraph isomorphism.
What you want is a "subgraph monomorphism" and there is a method to iterate over those as well.

Melissa1412:
> A subgraph must be an "induced subgraph". So your example is not a subgraph isomorphism.
> What you want is a "subgraph monomorphism" and there is a method to iterate over those as well.

Melissa1412:
nice to know :)

Issue #54, opened 2012-07-20
draw(G,pos) doesn't always create figures with the nodes in exactly the same place (migrated from Trac #46)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/46
Reported 2006-06-07 by jomiller, assigned to unknown.

I've defined G through erdos_renyi_graph.
I then define pos by spring_layout(G).

I then use the commands:
In [234]:draw(G,pos, nodelist=G.nodes(), node_color='b')

In [235]:savefig('bugtest1.png')

In [236]:show()

In [237]:draw(G,pos, nodelist=G.nodes(), node_color='b')

In [238]:draw(G,pos, nodelist=[1], node_color='r')

In [239]:savefig('bugtest2.png')

In [240]:show()

The nodes in the two png files do not lie in the same place.  I'm using the version currently the default on t7 machines.

PR #5271, opened 2022-01-18
add draw_sudoku function

mjhajharia:
Refer to this PR in nx-guides https://github.com/networkx/nx-guides/pull/54

Relevant comments 

- https://github.com/networkx/nx-guides/pull/54#pullrequestreview-799305610
- https://github.com/networkx/nx-guides/pull/54#issuecomment-1015906468

About this PR
- This PR adds a function which can create neat sudoku visualisations for n=3 and even n=4, they can mostly be of pedagogical value but I think it'll be a cool thing to have in the library for people who are fond of graph theory, it's like a nice quirk(!!!!) 
- It's primarily for nx-guides so it might or might not be a good fit here, do let me know!
- It isn't polished up yet so please feel free to comment/give feedback/correct etc
- Some other ideas
1. maybe I should only export the separate edges or provide customisation on how users want to colour them?
2. Maybe exporting the graph in gexf or something would be nice? but i dont think any of them preserve edge color etc
3. Atleast add an option to save the matplotlib figure with some customizations

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -17,6 +17,9 @@ Warning: Most layout routines have only been tested in 2-dimensions.
 """
 import networkx as nx
 from networkx.utils import np_random_state
+import matplotlib.pyplot as plt
+from random import sample
+from math import sqrt
 
 __all__ = [
     "bipartite_layout",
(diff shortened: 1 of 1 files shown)


mjhajharia:
Reference to understand the code with comments -> https://nbviewer.org/gist/mjhajharia/5a9ede19745f54f8b390680bff1a0439 it's hacky but yeah

mjhajharia:
I thought about this for a while, and I think I should close this in favour of [#54](https://github.com/networkx/nx-guides/pull/54) because although this can be a part of the source code, it isn't a suitable fit here as:
- the code is too hacky
- this is of pedagogical value and not so much about standard core graph algorithms
- it distorts the general API structure, I don't think it will fit into the layouts.py file and this doesn't seem like the right place

So I'll go ahead with the notebook in nx-guides

cc: @MridulS @rossbar

PR #5275, opened 2022-01-22
Update gexf website link in documentation

mbastian:
Hi, we've recently put the GEXF website again into its own domain http://gexf.net/ so this documentation should be updated. Thanks!


jarrodmillman (approved):

dschult (approved):
Thanks!

PR #5280, opened 2022-01-25
Update imports to use lazy imports

dschult:
This PR changes the idiom `import numpy as np` within each function by moving it to the top of the module and using `np = nx.lazy_import("numpy")`, with similar treatment for the other libraries.

Another PR will change the NetworkX loading process to ensure lazy loading of all internal packages (subpackages).

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -259,20 +259,27 @@ Guidelines
   and widely-used.  New default dependencies should be easy to install on all
   platforms, widely-used in the community, and have demonstrated potential for
   wide-spread use in NetworkX.
-* Use the following import conventions::
+* We use a lazy import system so users without e.g. numpy can still use NetworkX.
+  Use the following import conventions::
 
-   import numpy as np
-   import scipy as sp
-   import matplotlib as mpl
-   import matplotlib.pyplot as plt
-   import pandas as pd
    import networkx as nx
+   np = nx.lazy_import("numpy")  # instead of import numpy as np
+   sp = nx.lazy_import("scipy")
+   mpl = nx.lazy_import("matplotlib")
+   plt = nx.lazy_import("matplotlib.pyplot")
+   pd = nx.lazy_import("pandas")
+
+* For subpackage importing we use notation to indicate the source of the subpackage.
+  For example, many libraries have a ``linalg`` subpackage: ``nx.linalg``,
+  ``np.linalg``, ``sp.linalg``, ``sp.sparse.linalg``.
+  We use an import pattern that makes the origin of a particular ``linalg`` explicit.
 
   After importing `sp`` for ``scipy``::
 
-   import scipy as sp
+   sp = nx.lazy_import("scipy")
 
-  use the following imports::
+  use the following import idioms. These should be used **within**
+  the functions that they will be used or the lazy loading expires::
 
    import scipy.linalg  # call as sp.linalg
    import scipy.sparse  # call as sp.sparse
@@ -280,9 +287,6 @@ Guidelines
    import scipy.stats  # call as sp.stats
    import scipy.optimize  # call as sp.optimize
 
-  For example, many libraries have a ``linalg`` subpackage: ``nx.linalg``,
-  ``np.linalg``, ``sp.linalg``, ``sp.sparse.linalg``. The above import
-  pattern makes the origin of any particular instance of ``linalg`` explicit.
 
 * Use the decorator ``not_implemented_for`` in ``networkx/utils/decorators.py``
   to designate that a function doesn't accept 'directed', 'undirected',
--- a/networkx/algorithms/assortativity/correlation.py
+++ b/networkx/algorithms/assortativity/correlation.py
@@ -9,8 +9,8 @@ from networkx.algorithms.assortativity.pairs import node_degree_xy
 
 import networkx as nx
 
-np = nx.lazy_imports.load("numpy")
-sp = nx.lazy_imports.load("scipy")
+np = nx.lazy_import("numpy")
+sp = nx.lazy_import("scipy")
 
 __all__ = [
     "degree_pearson_correlation_coefficient",
--- a/networkx/algorithms/assortativity/correlation.py
+++ b/networkx/algorithms/assortativity/correlation.py
@@ -7,6 +7,11 @@ from networkx.algorithms.assortativity.mixing import (
 )
 from networkx.algorithms.assortativity.pairs import node_degree_xy
 
+import networkx as nx
+
+np = nx.lazy_imports.load("numpy")
+sp = nx.lazy_imports.load("scipy")
+
 __all__ = [
     "degree_pearson_correlation_coefficient",
     "degree_assortativity_coefficient",
@@ -150,7 +155,6 @@ def degree_pearson_correlation_coefficient(G, x="out", y="in", weight=None, node
     .. [2] Foster, J.G., Foster, D.V., Grassberger, P. & Paczuski, M.
        Edge direction and the structure of networks, PNAS 107, 10815-20 (2010).
     """
-    import scipy as sp
     import scipy.stats  # call as sp.stats
 
     xy = node_degree_xy(G, x=x, y=y, nodes=nodes, weight=weight)
@@ -283,8 +287,6 @@ def attribute_ac(M):
 def numeric_ac(M, mapping):
     # M is a numpy matrix or array
     # numeric assortativity coefficient, pearsonr
-    import numpy as np
-
     if M.sum() != 1.0:
         M = M / float(M.sum())
     nx, ny = M.shape  # nx=ny
--- a/networkx/algorithms/bipartite/matrix.py
+++ b/networkx/algorithms/bipartite/matrix.py
@@ -7,7 +7,7 @@ import itertools
 from networkx.convert_matrix import _generate_weighted_edges
 import networkx as nx
 
-sp = nx.lazy_imports.load("scipy")
+sp = nx.lazy_import("scipy")
 
 __all__ = ["biadjacency_matrix", "from_biadjacency_matrix"]
 
--- a/networkx/algorithms/bipartite/matrix.py
+++ b/networkx/algorithms/bipartite/matrix.py
@@ -8,6 +8,7 @@ from networkx.convert_matrix import _generate_weighted_edges
 import networkx as nx
 
 sp = nx.lazy_import("scipy")
+sp.sparse = nx.lazy_import("scipy.sparse")  # call as sp.sparse
 
 __all__ = ["biadjacency_matrix", "from_biadjacency_matrix"]
 
@@ -75,8 +76,6 @@ def biadjacency_matrix(
     .. [2] Scipy Dev. References, "Sparse Matrices",
        https://docs.scipy.org/doc/scipy/reference/sparse.html
     """
-    import scipy.sparse  # call as sp.sparse
-
     nlen = len(row_order)
     if nlen == 0:
         raise nx.NetworkXError("row_order is empty list")
--- a/networkx/algorithms/bipartite/matrix.py
+++ b/networkx/algorithms/bipartite/matrix.py
@@ -7,6 +7,8 @@ import itertools
 from networkx.convert_matrix import _generate_weighted_edges
 import networkx as nx
 
+sp = nx.lazy_imports.load("scipy")
+
 __all__ = ["biadjacency_matrix", "from_biadjacency_matrix"]
 
 
@@ -73,7 +75,6 @@ def biadjacency_matrix(
     .. [2] Scipy Dev. References, "Sparse Matrices",
        https://docs.scipy.org/doc/scipy/reference/sparse.html
     """
-    import scipy as sp
     import scipy.sparse  # call as sp.sparse
 
     nlen = len(row_order)
--- a/networkx/algorithms/bipartite/spectral.py
+++ b/networkx/algorithms/bipartite/spectral.py
@@ -3,7 +3,7 @@ Spectral bipartivity measure.
 """
 import networkx as nx
 
-sp = nx.lazy_imports.load("scipy")
+sp = nx.lazy_import("scipy")
 
 __all__ = ["spectral_bipartivity"]
 
--- a/networkx/algorithms/bipartite/spectral.py
+++ b/networkx/algorithms/bipartite/spectral.py
@@ -3,6 +3,8 @@ Spectral bipartivity measure.
 """
 import networkx as nx
 
+sp = nx.lazy_imports.load("scipy")
+
 __all__ = ["spectral_bipartivity"]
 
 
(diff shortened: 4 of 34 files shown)


dschult:
I think this one is ready for review. The name is now `nx.lazy_import` as it was made in #4909 .  This PR moves the imports of whole packages out of the functions.  It really isn't as much gain as we'd hoped because we still need to import subpackages inside the functions.  
```
sp = nx.lazy_import("scipy")
sp.sparse = nx.lazy_import("scipy.sparse")     <--  sp.sparse is not found inside the function.

sp_sparse = nx.lazy_import("scipy.sparse")     <-- does seem to work in basic tests...
```
I haven't tried the latter option on a large scale though because it would require changing a lot of code that uses `sp.sparse` to access the subpackage.  Thoughts @jarrodmillman ?

dschult:
I have changed how the lazy_import handles subpackages. 
We can now load the subpackages at the top of the module instead of inside a function. For example
```python
sp = nx.lazy_import("scipy")
sp.sparse.linalg = nx.lazy_import("scipy.sparse.linalg")
```
allows use of `sp.sparse.linalg.eigs` inside a function within that module. Even if scipy is not installed, the assignment to `sp.sparse.linalg` will maintain the delayed error until attributes are looked up within `sp.sparse` or `sp.sparse.linalg`.

The cost of doing this is a small pollution of `sys.modules`!!!

If `scipy` is not installed, I add a module to `sys.modules` named `__scipy__DelayedImportErrorModule`.  This means I can reuse a single delayed module instead of creating one for each lazy_import called with that module. I needed to do that so that `nx.lazy_import("scipy.sparse.linalg")` can add subpackages to the existing delayed import module rather than adding them to a new delayed import module. Adding them to a new module doesn't let us use `sp.sparse` later when `sp` is one module object and the `sparse` subpackage module was added to a different module object.

I think it is a bad idea to add `scipy` to `sys.modules` when scipy is a `DelayedImportErrorModule`.  Then **any** python code could inspect `sys.modules`, see "scipy" as a key and start acting as if was installed.  By renaming it `__scipy__DelayedImportErrorModule` I hope to alleviate that concern.

@stefanv and @jarrodmillman do you have thoughts on this approach?
Also, is there a way to avoid the IPython traceback errors processing the traceback of the Delayed Module?

dschult:
I'm not getting these errors locally, and the traceback doesn't work well with these DelayedImportErrorModule imports on CI.  The error reports indicate it is having trouble with a lazily-loaded pydot, but the trouble is happening when testing pandas.

I haven't uploaded locally to scipy 1.8.  But that doesn't seem to be the cause of these error. Anybody know what I might need to do to recreate these locally?

MridulS:
I am able to get these errors locally, with a clean environment install of networkx with python3.10

MridulS:
It seems to be erroring out because of a deprecation warning raised in Pandas (@dschult can you update your pandas version locally and try this out?), and it seems like it's transported up to the lazy_import chain.

My local traceback:
``` python
File ~/miniconda3/envs/nx5280/lib/python3.10/site-packages/pandas/core/frame.py:9020, in DataFrame.append(self, other, ignore_index, verify_integrity, sort)
   8917 def append(
   8918     self,
   8919     other,
   (...)
   8922     sort: bool = False,
   8923 ) -> DataFrame:
   8924     """
   8925     Append rows of `other` to the end of caller, returning a new object.
   8926
   (...)
   9013     4  4
   9014     """
   9015     warnings.warn(
   9016         "The frame.append method is deprecated "
   9017         "and will be removed from pandas in a future version. "
   9018         "Use pandas.concat instead.",
   9019         FutureWarning,
-> 9020         stacklevel=find_stack_level(),
   9021     )
   9023     return self._append(other, ignore_index, verify_integrity, sort)

File ~/miniconda3/envs/nx5280/lib/python3.10/site-packages/pandas/util/_exceptions.py:32, in find_stack_level()
     27 def find_stack_level() -> int:
     28     """
     29     Find the first place in the stack that is not inside pandas
     30     (tests notwithstanding).
     31     """
---> 32     stack = inspect.stack()
     34     import pandas as pd
     36     pkg_dir = os.path.dirname(pd.__file__)

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1678, in stack(context)
   1676 def stack(context=1):
   1677     """Return a list of records for the stack above the caller's frame."""
-> 1678     return getouterframes(sys._getframe(1), context)

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1655, in getouterframes(frame, context)
   1653 framelist = []
   1654 while frame:
-> 1655     frameinfo = (frame,) + getframeinfo(frame, context)
   1656     framelist.append(FrameInfo(*frameinfo))
   1657     frame = frame.f_back

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1629, in getframeinfo(frame, context)
   1627 start = lineno - 1 - context//2
   1628 try:
-> 1629     lines, lnum = findsource(frame)
   1630 except OSError:
   1631     lines = index = None

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:952, in findsource(object)
    949     if not (file.startswith('<') and file.endswith('>')):
    950         raise OSError('source code not available')
--> 952 module = getmodule(object, file)
    953 if module:
    954     lines = linecache.getlines(file, module.__dict__)

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:869, in getmodule(object, _filename)
    866 # Update the filename to module name cache and check yet again
    867 # Copy sys.modules in order to cope with changes while iterating
    868 for modname, module in sys.modules.copy().items():
--> 869     if ismodule(module) and hasattr(module, '__file__'):
    870         f = module.__file__
    871         if f == _filesbymodname.get(modname, None):
    872             # Have already mapped this module, so skip it

File ~/dev/networkx/networkx/lazy_imports.py:202, in DelayedImportErrorModule.__getattribute__(self, attr)
    199 except AttributeError:
    200     pass
--> 202 raise ModuleNotFoundError(
    203     f"Delayed Report: module named '{spec.name}' not found.\n"
    204     "Reporting was Lazy -- delayed until module attributes accessed.\n"
    205     f"Most likely, {spec.name} is not installed"
    206 )

ModuleNotFoundError: Delayed Report: module named 'pydot' not found.
Reporting was Lazy -- delayed until module attributes accessed.
Most likely, pydot is not installed
```

dschult:
I've got pandas 1.4.0 and scipy 1.8.0rc2.  But I don't get an error.
I'm using conda-forge with pip on top to get scipy 1.8.0rc2.   
Is it possible this is a unix-only problem?  
Do I need to set my handling of warnings differently?
Thanks very much for this --

MridulS:
[Disclaimer: this is the first time I'm doing a deep dive into the lazy_import module!]

Hmm, I'm not sure what's going on here:

I don't think it has anything to do with scipy or any other package.

This issue pops up from [a line in pandas `find_stack_level`]( https://github.com/pandas-dev/pandas/blob/c0bdeb4c10a74b7837820e564ed9cad406a83c5a/pandas/util/_exceptions.py#L32 ), where it's inspects the stack.
(that bit of code has been removed in https://github.com/pandas-dev/pandas/pull/45247)

If I try to inspect the stack after just importing `networkx` (assuming one of the `lazy_import`s module isn't present in the env)
``` python
In [1]: import networkx as nx

In [2]: import inspect

In [3]: inspect.stack()
ModuleNotFoundError                       Traceback (most recent call last)
Input In [3], in <module>
----> 1 inspect.stack()

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1678, in stack(context)
   1676 def stack(context=1):
   1677     """Return a list of records for the stack above the caller's frame."""
-> 1678     return getouterframes(sys._getframe(1), context)

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1655, in getouterframes(frame, context)
   1653 framelist = []
   1654 while frame:
-> 1655     frameinfo = (frame,) + getframeinfo(frame, context)
   1656     framelist.append(FrameInfo(*frameinfo))
   1657     frame = frame.f_back

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1625, in getframeinfo(frame, context)
   1622 if not isframe(frame):
   1623     raise TypeError('{!r} is not a frame or traceback object'.format(frame))
-> 1625 filename = getsourcefile(frame) or getfile(frame)
   1626 if context > 0:
   1627     start = lineno - 1 - context//2

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:829, in getsourcefile(object)
    827     return filename
    828 # only return a non-existent filename if the module has a PEP 302 loader
--> 829 module = getmodule(object, filename)
    830 if getattr(module, '__loader__', None) is not None:
    831     return filename

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:869, in getmodule(object, _filename)
    866 # Update the filename to module name cache and check yet again
    867 # Copy sys.modules in order to cope with changes while iterating
    868 for modname, module in sys.modules.copy().items():
--> 869     if ismodule(module) and hasattr(module, '__file__'):
    870         f = module.__file__
    871         if f == _filesbymodname.get(modname, None):
    872             # Have already mapped this module, so skip it

File ~/dev/networkx/networkx/lazy_imports.py:202, in DelayedImportErrorModule.__getattribute__(self, attr)
    199 except AttributeError:
    200     pass
--> 202 raise ModuleNotFoundError(
    203     f"Delayed Report: module named '{spec.name}' not found.\n"
    204     "Reporting was Lazy -- delayed until module attributes accessed.\n"
    205     f"Most likely, {spec.name} is not installed"
    206 )

ModuleNotFoundError: Delayed Report: module named 'pydot' not found.
Reporting was Lazy -- delayed until module attributes accessed.
Most likely, pydot is not installed
```

as soon as `pydot` is installed, everyone seem to be happy.
I'm not sure if this is intended behavior.

dschult:
The fact that it is looking for 'pydot' makes me think that one of the lazy_imports of pydot is leaking out into the wild.
So, it could be that pydot is being handled differently, or maybe pydot is just first on the stack and removing those lazy_imports will reveal the next lazy_imported library.

hmmm...

MridulS:
It's not just pydot, if I remove scipy, numpy it starts complaining about that too.
After removing `scipy` from the env
``` python
In [1]: import inspect

In [2]: import networkx as nx

In [3]: inspect.stack()
ModuleNotFoundError                       Traceback (most recent call last)
Input In [3], in <module>
----> 1 inspect.stack()

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1678, in stack(context)
   1676 def stack(context=1):
   1677     """Return a list of records for the stack above the caller's frame."""
-> 1678     return getouterframes(sys._getframe(1), context)

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1655, in getouterframes(frame, context)
   1653 framelist = []
   1654 while frame:
-> 1655     frameinfo = (frame,) + getframeinfo(frame, context)
   1656     framelist.append(FrameInfo(*frameinfo))
   1657     frame = frame.f_back

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:1625, in getframeinfo(frame, context)
   1622 if not isframe(frame):
   1623     raise TypeError('{!r} is not a frame or traceback object'.format(frame))
-> 1625 filename = getsourcefile(frame) or getfile(frame)
   1626 if context > 0:
   1627     start = lineno - 1 - context//2

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:829, in getsourcefile(object)
    827     return filename
    828 # only return a non-existent filename if the module has a PEP 302 loader
--> 829 module = getmodule(object, filename)
    830 if getattr(module, '__loader__', None) is not None:
    831     return filename

File ~/miniconda3/envs/nx5280/lib/python3.10/inspect.py:869, in getmodule(object, _filename)
    866 # Update the filename to module name cache and check yet again
    867 # Copy sys.modules in order to cope with changes while iterating
    868 for modname, module in sys.modules.copy().items():
--> 869     if ismodule(module) and hasattr(module, '__file__'):
    870         f = module.__file__
    871         if f == _filesbymodname.get(modname, None):
    872             # Have already mapped this module, so skip it

File ~/dev/networkx/networkx/lazy_imports.py:202, in DelayedImportErrorModule.__getattribute__(self, attr)
    199 except AttributeError:
    200     pass
--> 202 raise ModuleNotFoundError(
    203     f"Delayed Report: module named '{spec.name}' not found.\n"
    204     "Reporting was Lazy -- delayed until module attributes accessed.\n"
    205     f"Most likely, {spec.name} is not installed"
    206 )

ModuleNotFoundError: Delayed Report: module named 'scipy' not found.
Reporting was Lazy -- delayed until module attributes accessed.
Most likely, scipy is not installed
``

**Comment by MridulS** [comment]
> The cost of doing this is a small pollution of sys.modules!!!
I'm not a 100% sure but I think this is the issue!

**Comment by dschult** [comment]
I agree.  Something has changed for Modules.  And it is definitely from Python 3.10.   It all passes on 3.9.

**Comment by MridulS** [comment]
It's locally giving the stack error for me on py3.9 too :/ (I'm using MacOS btw)

In a clean python 3.9 env with nothing installed and running this bit of code inside networkx directory
``` python
Python 3.9.10 | packaged by conda-forge | (main, Feb  1 2022, 21:27:48)
[Clang 11.1.0 ] on darwin
>>> import inspect
>>> import networkx as nx
>>> inspect.stack()
  File "/Users/ms/miniconda3/envs/nx5280-py39/lib/python3.9/inspect.py", line 1554, in stack
    return getouterframes(sys._getframe(1), context)
  File "/Users/ms/miniconda3/envs/nx5280-py39/lib/python3.9/inspect.py", line 1531, in getouterframes
    frameinfo = (frame,) + getframeinfo(frame, context)
  File "/Users/ms/miniconda3/envs/nx5280-py39/lib/python3.9/inspect.py", line 1501, in getframeinfo
    filename = getsourcefile(frame) or getfile(frame)
  File "/Users/ms/miniconda3/envs/nx5280-py39/lib/python3.9/inspect.py", line 709, in getsourcefile
    if getattr(getmodule(object, filename), '__loader__', None) is not None:
  File "/Users/ms/miniconda3/envs/nx5280-py39/lib/python3.9/inspect.py", line 746, in getmodule
    if ismodule(module) and hasattr(module, '__file__'):
  File "/Users/ms/dev/networkx/networkx/lazy_imports.py", line 202, in __getattribute__
    raise ModuleNotFoundError(
ModuleNotFoundError: Delayed Report: module named 'numpy' not found.
Reporting was Lazy -- delayed until module attributes accessed.
Most likely, numpy is not installed
```

dschult:
Aha!   Thanks for this -- the `inspect.stack()` must try to access attributes on the modules. It is probably looking for something like `__file__`. But as soon as the module tries to access the attributes, the Delayed Import Error is initiated.  So, `sys.modules` is not a good place to store the modules that have been lazy_imported.  I knew that at some level, and I should have realized that other code would go through `sys.modules` checking each module. 

The difficulty then is how to store the delayed module in a central place so that subpackages can be added to it without triggering the ImportError.  I can think of two approaches: 1) make a networkx owned dict of delayed modules.  2) enhance the DelayedModule to handle introspection of dunders like `__file__` without triggering the ImportError. In creating the DelayedModule, I use the base class of a FileModule. I'll look around for others.

I am also beginning to realize that the delayed ImportError feature may not be worth the effort. The current system puts imports inside the functions. The is really not a large cost compared to creating a parallel module/import system. :}  Perhaps I should shelve the `lazy_import` feature and move to getting the `lazy.attach` feature implemented -- that is what the SPEC is more about anyway.

PR #5284, opened 2022-01-27
Fix: Update louvain_partitions for threshold (update mod to new_mod in each level)

cirusthenter:
Currently `threshold` in `louvain_partitions` does not work. This is because `mod` in the while loop (lines 181-191) is not updated since it is initialized with the modularity of the first partition (where each node is in a community that only has the node).

### DIFF
--- a/networkx/algorithms/community/louvain.py
+++ b/networkx/algorithms/community/louvain.py
@@ -185,6 +185,7 @@ def louvain_partitions(
         )
         if new_mod - mod <= threshold:
             return
+        mod = new_mod
         graph = _gen_graph(graph, inner_partition)
         partition, inner_partition, improvement = _one_level(
             graph, m, partition, resolution, is_directed, seed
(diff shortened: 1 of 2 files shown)


cirusthenter:
It seems that the PyPy failure comes from `LFR_benchmark_graph` and `louvain_communities` under PyPy have different results from under Python. Simply changing `assert mod1 < 0.512 < mod2` to `assert mod1 <  mod2` in `test_louvain.py` line 124 works?

By the way I am a first-time contributor and any comments or suggestions are appreciated.
Thanks!

z3y50n:
Nice catch, thank you!
You are right regarding the tests, depending on the python version you might get some slightly different results so you should change it to `assert mod1 < mod2`.

cirusthenter:
Thank you @z3y50n. FIixed!

z3y50n (approved):

dschult (approved):
Thanks for this bug report -- and the fix -- and for the test!!
This looks good to me.

PR #5285, opened 2022-01-28
Weighted karate club graph

kpberry:
Hello everyone! I noticed that the original paper for Zachary's karate club has weights on the graph, but the version that networkx loads is unweighted. I wasn't able to find a copy of the graph with the edge weights online, so I updated them by hand and updated karate_club_graph to load the weighted edges when passed the optional parameter weighted=True. 

Because I typed the graph myself while referring to the paper, I have removed the link to the UCI datasets from the karate_club_graph documentation.

In order to ensure the code is tested, I updated the weighted louvain test to load the karate graph as weighted and assert that the detected communities for graphs with different weights are different from each other.

It is also worth mentioning that I have set the default value of weighted to False. While I think that weighted=True is a better default, since it reflects the original context of the network and provides useful extra information, there are some test cases within networkx that implicitly rely on the karate graph being unweighted, and I would imagine there to be a decent amount of code/tests in other libraries that depends on the graph being unweighted as well. I think it would be reasonable to update those test cases, but I would prefer not to break other code unexpectedly without a strong enough reason.


networkx/generators/social.py
```diff
@@ -11,12 +11,18 @@
...
 
-def karate_club_graph():
+def karate_club_graph(weighted=False):
     """Returns Zachary's Karate Club graph.
 
     Each node in the returned graph has a node attribute 'club' that
     indicates the name of the club to which the member represented by that node
-    belongs, either 'Mr. Hi' or 'Officer'.
+    belongs, either 'Mr. Hi' or 'Officer'. Each has a weight based on the
+    number of contexts in which the correspnding members interacted.
```
dschult, line 20:
```suggestion
    belongs, either 'Mr. Hi' or 'Officer'. Each edge has a weight based on the
    number of contexts in which that edge's incident node members interacted.
```

networkx/generators/social.py
```diff
@@ -11,12 +11,18 @@
 ]
 
 
-def karate_club_graph():
+def karate_club_graph(weighted=False):
```
dschult, line 14: I would prefer to make it weighted  -- and to not make it optional (no keyword).
The routines can use `weighted=None` to handle the unweighted version.
This is a good addition.  It also simplifies the code a little below (when adding edges) 
if we remove the option.

kpberry: I started working on this, but, unfortunately, there are some complications in fixing two of the three broken tests:

A minor issue is that in [algorithms.community.tests.test_modularity_max.test_modularity_communities](https://github.com/networkx/networkx/blob/cceb43d15e1d01476c8c15ff273399dee0e3b1aa/networkx/algorithms/community/tests/test_modularity_max.py#L10), both greedy_modularity_communities and naive_greedy_modularity_communities are tested, but the latter does not take a `weight` argument since it is a naive reference implementation. The simplest fix is to set the edge weights in the graph to be 1 after loading it instead of trying to call one function with a `weight` argument and the other without.

A larger issue came up when trying to fix [algorithms.tests.test_non_randomness](https://github.com/networkx/networkx/blob/cceb43d15e1d01476c8c15ff273399dee0e3b1aa/networkx/algorithms/tests/test_non_randomness.py#L8). The [non_randomness](https://github.com/networkx/networkx/blob/cceb43d15e1d01476c8c15ff273399dee0e3b1aa/networkx/algorithms/non_randomness.py#L11) function doesn't take a weight parameter, but implicitly uses the weights from the adjacency matrix when computing the graph's eigenvalues by calling `nx.to_numpy_array` without setting a value for `weight` . The seemingly simple fix would be to add a `weight` argument to `non_randomness` and pass it along when getting the adjacency matrix. However, referring to [the paper](http://csce.uark.edu/~xintaowu/publ/299_Ying.pdf) where matrix non-randomness is defined, the graph is explicitly defined as unweighted. I think their calculation naively generalizes to weighted graphs, but I'm not an expert in spectral graph theory, so I'm not completely sure that it does. If it doesn't, then the fix is to always pass `weight=None` when loading the adjacency matrix. Either way, I think a proper fix for the broken test case involves fixing this function and not modifying the test case itself, though it should probably be a topic for a separate issue/pull request since it doesn't directly relate to the karate graph.

The third broken test case, `algorithms.community.tests.test_louvain.test_partition` can be fixed by simply setting `weight=None`, as you suggested.

dschult: Nice summary!  

The minor issue with `naive_greedy_modularity_communities` is a little embarrassing. That should definitely have a `weight` keyword argument.  I believe the history is that it was a private function for a long time and when we converted it to a public function we forgot to think about the general input structure. I think we should add a keyword with default:  `weight=None`.  

For the second `non_randomness` test, the naive extension from unweighted edges to weighted edges generalizes nicely in this case. Here, we should probably implement a default of `weight="weight"` to avoid backward incompatibility.  So the steps will be:  1) adding a keyword argument to `non_randomness` with default `weight="weight"`. 2) change the test to call it with `weight=None`.

I think these are aligned with your thoughts above.  If you'd rather they be done by someone who has experience with those two functions, I'll take a stab at the changes to `non_randomness` and `naive_greedy_modularity_comunities`. Just let me know.

kpberry: No worries, I went ahead and removed the weight parameter from karate_club_graph, added the weight parameters to non_randomness and naive_greedy_modularity_communities, and updated test cases/documentation accordingly. Hopefully, the changes should be easy to verify, and this is now safe to merge.

dschult (approved):
This looks great!
Thanks for the additional tests added to help ensure the new keywords are handled well..
 And Thanks for the submission!

rossbar (approved):
LGTM , thanks @kpberry !

kpberry:
Happy to contribute :)

PR #5287, opened 2022-01-28
Add exception for unconnected graph

### LINKED ISSUE #5286: `json_graph.tree_data` can cause maximum recursion depth error.
### Current Behavior

Currently the algorithm compares the `n_nodes` with `n_edges` to check if `G` is a tree. https://github.com/networkx/networkx/blob/0cc70051fa0a979b1f1eab4af5b6587a6ebf8334/networkx/readwrite/json_graph/tree.py#L74-L75 
This check can be bypassed with specific inputs and cause a recursion error.

### Expected Behavior

The code should check whether there are cycles with `root` as the source and raise an exception.
Another possible fix would be to check if the graph is not weakly connected.

### Steps to Reproduce

```Python3
>>> import networkx as nx
>>> G = nx.DiGraph([(1, 2), (2, 3), (3, 1)])
>>> G.add_node(4)
>>> data = nx.json_graph.tree_data(G, 1)
RecursionError: maximum recursion depth exceeded
```

### Environment

Python version: 3.8.10
NetworkX version: 2.7rc1.dev0

z3y50n:
Closes #5286.

This PR checks if the Graph is weakly connected to avoid the recursion error described in #5286 
Alternatively the issue could be solved by checking for cycles like this:
```Python3
try:
    nx.find_cycles(G, source=root)
    raise TypeError("G contains cycles")
except nx.NetworkXNoCycle:
    pass
```
but since it doesn't really make sense to create a tree from a graph that isn't connected I feel the `is_weakly_connected` choice is better.

### DIFF
--- a/networkx/readwrite/json_graph/tests/test_tree.py
+++ b/networkx/readwrite/json_graph/tests/test_tree.py
@@ -35,6 +35,11 @@ def test_exceptions():
     with pytest.raises(TypeError, match="is not directed."):
         G = nx.path_graph(3)
         tree_data(G, 0)
+    with pytest.raises(TypeError, match="is not weakly connected."):
+        G = nx.path_graph(3, create_using=nx.DiGraph)
+        G.add_edge(2, 0)
+        G.add_node(3)
+        tree_data(G, 0)
     with pytest.raises(nx.NetworkXError, match="must be different."):
         G = nx.MultiDiGraph()
         G.add_node(0)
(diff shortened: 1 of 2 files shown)


rossbar (approved):
LGTM - I have a vague sense that this would be better as a `ValueError` or maybe `NetworkxError`, but the other related exceptions are already `TypeError`s so matching the existing pattern makes sense. Thanks @z3y50n !

dschult (approved):
Nice!

PR #5297, opened 2022-02-03
DOC: change status to accepted for NXEP2, add resolution

MridulS:
Wasn't super sure about how to update this so added a new section `Resolution` in the NXEP

### DIFF
--- a/doc/developer/nxeps/nxep-0002.rst
+++ b/doc/developer/nxeps/nxep-0002.rst
@@ -5,7 +5,7 @@ NXEP 2 — API design of view slices
 ==================================
 
 :Author: Mridul Seth
-:Status: Draft
+:Status: Accepted
 :Type: Standards Track
 :Created: 2020-07-23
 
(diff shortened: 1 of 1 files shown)


dschult (approved):
Nice!

rossbar (approved):
+1 for the accepted status.

I think it might also be worth mentioning in the "Resolution" section that the decision is **not** final; i.e. if there's strong user/developer feedback to go with a different approach (e.g. supporting slicing as was originally proposed) then this can be done. IIRC we went with option 1 as a "provisional" step i.e. it was a way to improve the current situation without a major change to semantics. We definitely want to keep encouraging opinions/thoughts about the original proposal as well!

MridulS:
Just a meta question about NXEPs: Once a NXEP is changed to "Accepted" is it supposed to stay untouched for eternity?

https://networkx.org/documentation/latest/developer/nxeps/nxep-0000.html#review-and-resolution does talk about `replaced`, would that require a new NXEP?

jarrodmillman (approved):

Issue #5257, opened 2022-01-13
Distance_measures.py functions should work for weighted graphs too.


dschult:
Many of the functions in `distance_measures.py` work for unweighted graphs, but have straightforward definitions for weighted graphs too.  We should go through this module and upgrade each function to work for weighted graphs.  It should be a straightforward fix with a keyword "weight" input which defaults to the current unweighted treatment.  The `shortest_path` functions provide a nice interface for that to use as an example.

lucasmccabe:
To be clear, we mean `distance_measures.py`, right? Or is `distance_measurements.py` somewhere and I'm just not seeing it?

dschult:
Thanks -- Yes -- I mean `distance_measures.py`.  I'll correct it by editing my statement.

nishant42491:
Can i work on this issue

Alt-Shivam:
Hey @dschult 
I would like to work on it.
Thanks.

lucasmccabe:
Hi All, I would also be interested in contributing to this.

It looks like there are handful of functions to address - perhaps we can split the work among ourselves?

Alt-Shivam:
Yeah, @lucasmccabe 
let's do this. :)

nishant42491:
> Hi All, I would also be interested in contributing to this.
> 
> It looks like there are handful of functions to address - perhaps we can split the work among ourselves?

sure lets do it

Alt-Shivam:
Yeah @nishant42491 @lucasmccabe :)

BasLaa:
Hi guys,

Are you still working actively on this? Otherwise I can also take over some work.

haricash:
Hello, if you guys are still working on this and need an extra helping hand, do add me too!

MridulS:
Hello all! It's great to see so much interest to work on the project :)

You can read more about the development processes on our website:

- [New Contributor FAQ](https://networkx.org/documentation/latest/developer/new_contributor_faq.html)
- [Contributor Guide](https://networkx.org/documentation/latest/developer/contribute.html)

For the point about assigning issues we don't usually assign issues to anyone, feel free to open a PR when you are ready.
If you are worried it may conflict with what other people are working on please put in the method you are working on in this issue and open a PR.

Even if you are stuck please do open a PR (you can keep it as a draft) as we can help you review and improve the code in the PR :)

BasLaa:
[...]

PR #5298, opened 2022-02-03
Add weight to eccentricity

BasLaa:
## Motivation
As talked about in issue #5257 we want distance measurements to work for weighted graphs too. This is a draft I am working on.

## Explanation
Most distance measurements in `distance_measures.py` rely on the `eccentricity` method and the `extrema_bounding` method. Thus, the key is to incorporate weights into these methods. I have currently added weights to `eccentricity`. 

## Considerations

- Currently, I am using Bellman Ford for calculating weighted shortest paths to allow negative weights. Do we want to allow negative weights?
- The `weight` argument should be a string as the Bellman Ford method requires a string. However, if I am not mistaken, it can also take a weight function, as Dijkstras method implementation can, but this is not documented
- There is not much type checking happening for weights anywhere, should we incorporate it? If so, can someone explain or send me a link to read on how NetworkXErrors work?

## Plan
I want to get weights working for `eccentricity `first, and then add that to all the functions that use `eccentricity`. Afterwards, I will work on `extrema_bounding`

### DIFF
--- a/networkx/algorithms/distance_measures.py
+++ b/networkx/algorithms/distance_measures.py
@@ -212,7 +212,7 @@ def extrema_bounding(G, compute="diameter"):
     return None
 
 
-def eccentricity(G, v=None, sp=None):
+def eccentricity(G, v=None, sp=None, weight=None):
     """Returns the eccentricity of nodes in G.
 
     The eccentricity of a node v is the maximum distance from v to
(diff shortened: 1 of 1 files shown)


BasLaa:
This currently doesn't work I realized, I will commit something proper this weekend

BasLaa:
After some manual testing, the `weight` argument works properly in `eccentricities` for non-negative weights.
Something weird is going on in the `extrema_bounding` method though, it doesn't work yet with weights.
However, this `extrema_bounding` method is not mentioned in any of the documentation. 
Is it actually used?

PR #5306, opened 2022-02-06
MAINT: Cleanup link analysis module, remove unused variables

MridulS:
Remove more characters from our codebase.
Make linters like pylint, isort a bit more happier.

Tiny subjective nitpicks around the code.

### DIFF
--- a/networkx/algorithms/link_analysis/__init__.py
+++ b/networkx/algorithms/link_analysis/__init__.py
@@ -1,2 +1,2 @@
-from networkx.algorithms.link_analysis.pagerank_alg import *
 from networkx.algorithms.link_analysis.hits_alg import *
+from networkx.algorithms.link_analysis.pagerank_alg import *
--- a/networkx/algorithms/link_analysis/hits_alg.py
+++ b/networkx/algorithms/link_analysis/hits_alg.py
@@ -245,8 +245,9 @@ def hits_numpy(G, normalized=True):
        doi:10.1145/324133.324140.
        http://www.cs.cornell.edu/home/kleinber/auth.pdf.
     """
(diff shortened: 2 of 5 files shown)


rossbar (commented):
Most of the changes looke good, but I think we need to be careful about the in-place division operators on numpy arrays. If we're absolutely sure that the dtypes of the arrays being operated on are floating point then it's fine, but if not then you can run into the following situation:

```python
>>> a = np.array([1, 2, 3], dtype=int)
>>> a /= 2
   ...
UFuncTypeError: Cannot cast ufunc 'true_divide' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

FWIW this was my main hesitation back when that bot suggested we switch to in-place operators everywhere in the codebase :joy: I'm not confident that most linters are aware of these subtleties.

MridulS:
Stupid bots indeed. Atleast in this case I guess we can be sure that everything is `float`. Everything starts with a numpy float array `A`.

dschult:
Would the dtype error be avoided using:
```
a = np.array([1, 2, 3], dtype=int) / 2
```

MridulS:
Yeah the following bits of operation on `a` shouldn't be a problem as after `a = np.array([1, 2, 3], dtype=int) / 2` the `dtype` is `float`.

[In this bit of code we are explicit about the `float` dtype while creating the array]

rossbar (approved):
LGTM - the in-place division on ndarray's look good in these cases as it does seem they're always guaranteed to have floating point dtypes. The rest of the cleanups are obvious improvements IMO - thanks @MridulS !

jarrodmillman (approved):

PR #5307, opened 2022-02-06
DOC: Update documentation to include callables for weight argument

MridulS:
While going through some of the docs I saw that the documentation doesn't accurately represent what the `weight` argument can do. We can pass in a function callable to the `weight` argument too for these methods.


dschult (approved):
LGTM!!  Thanks!

rossbar (approved):
LGTM, thanks @MridulS !

PR #5308, opened 2022-02-06
MAINT: Cleanup centrality module, remove unused variables

MridulS:
Remove more characters from our codebase.
Make linters like pylint, isort a bit more happier.

Tiny subjective nitpicks around the code.


networkx/algorithms/centrality/dispersion.py
```diff
@@ -69,12 +69,12 @@ def _dispersion(G_u, u, v):
                 norm_disp = ((total + b) ** alpha) / (embededness + c)
             else:
                 norm_disp = (total + b) ** alpha
-            dispersion = norm_disp
+            disper = norm_disp
```
dschult, line 72: Why are we making this abbreviation? Is there a name collision problem here?
Seems better to write out the name, especially if the function name has an underscore in front.

networkx/algorithms/centrality/group.py
```diff
@@ -515,7 +513,7 @@ def _heuristic(k, root, DF_tree, D, nodes, greedy):
             zip(np.diag(DF_tree.nodes[node_p]["betweenness"]), nodes), reverse=True
         )
     ]
-    [CL.remove(m) for m in CL if m in DF_tree.nodes[node_p]["GM"]]
+    _ = [CL.remove(m) for m in CL if m in DF_tree.nodes[node_p]["GM"]]
```
dschult, line 516: This change (I find) makes it even harder to read than the poor readable version we had before.
Can we just use:
```suggestion
    added_nodes = DF_tree.nodes[node_p]["GM"]
    for m in CL:
        if m in added_nodes:
            CL.remove(m)  
```

This big loop could be sped up by preassigning `DF_tree.nodes[node_p]` to a variable and maybe `DF_tree.nodes[node_p]["GM"]` and `root_node["sigma"]`.  Maybe that's out of scope for this PR, but we are looking at this code, so if not here, we should make an issue to update this code.

dschult (commented):
Minor tweaks -- but LGTM!   Thanks!

networkx/algorithms/centrality/closeness.py
```diff
@@ -108,7 +109,7 @@ def closeness_centrality(G, u=None, distance=None, wf_improved=True):
         nodes = G.nodes
     else:
         nodes = [u]
-    closeness_centrality = {}
+    closeness_cen = {}
```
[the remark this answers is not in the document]
MridulS: yeah, apparently we shouldn't be using the a variable name inside the function of the same name. I'll update this to  `closeness_dict`

networkx/algorithms/centrality/dispersion.py
MridulS: The method `_dispersion` is inside `dispersion`, so we should probably avoid naming a variable the same as the outer function name.

jarrodmillman: How about:
```
        if normalized:
            if embededness + c != 0:
                disp = ((total + b) ** alpha) / (embededness + c)
            else:
                disp = (total + b) ** alpha

        else:
            disp = total

        return disp
```

networkx/algorithms/centrality/current_flow_betweenness.py
```diff
@@ -122,14 +122,14 @@ def approximate_current_flow_betweenness_centrality(
...
-    for i in range(k):
+    for _ in range(k):
         s, t = seed.sample(range(n), 2)
         b = np.zeros(n, dtype=dtype)
         b[s] = 1
         b[t] = -1
         p = C.solve(b)
         for v in H:
-            if v == s or v == t:
+            if v in (s, t):
```
dschult, line 132: This might be clearer with naming the pair `(s, t)` earlier. e.g.
```python
s, t = pair = seed.sample(range(n), 2)
...
...
    if v in pair:
```

networkx/algorithms/centrality/group.py
```diff
@@ -515,7 +513,10 @@ def _heuristic(k, root, DF_tree, D, nodes, greedy):
             zip(np.diag(DF_tree.nodes[node_p]["betweenness"]), nodes), reverse=True
         )
     ]
-    [CL.remove(m) for m in CL if m in DF_tree.nodes[node_p]["GM"]]
+    added_nodes = DF_tree.nodes[node_p]["GM"]
+    for m in CL:
+        if m in added_nodes:
+            CL.remove(m)
```
dschult, line 519: It looks like this would be clearer and probably more efficient to include in the previous list comprehension which defines the `CL` list we are removing from here.  If I understand correctly, it would just need an additional line after 514 with something like:
```python
if node not in DF_tree.nodes[node_p]["GM"]
```

dschult (commented):
These look good.  Thanks!  
I've put a few notes and hope to get back to it later today to actually make the changes.
I also noticed a comment misspelled `node` as `nude` in line 456 in group.py so I'll find that too.

dschult (approved):
I made a couple of tweaks. 
LGTM!

jarrodmillman (approved):

PR #5311, opened 2022-02-08
Added forceatlas2 layout

cvanelteren:
# Added forceatlas 2 layout
I've added the ForceAtlas2 layout from gephi that was implemented by [bhargavchippada](https://github.com/bhargavchippada) with linLog merge from [victorbenichoux](https://github.com/victorbenichoux). The ForceAtlas2 layout forms a fast alternative for spring layout and work particularly well for large graphs. It originates from gephi and is a good addition to the networkx toolkit.

Best wishes,
C

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "networkx/drawing/fa2"]
+	path = networkx/drawing/fa2
+	url = https://github.com/cvanelteren/forceatlas2/
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -32,6 +32,7 @@ __all__ = [
     "fruchterman_reingold_layout",
     "spiral_layout",
     "multipartite_layout",
+    "forceatlas2_layout",
 ]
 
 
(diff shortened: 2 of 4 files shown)


MridulS:
Thanks for this @cvanelteren!

There are merge conflicts in one of the file plus we don't setup git submodules while running the CI.

I'm generally -1 on adding code through submodules. It would be super great if we could get this in if someone can port
the code to python and networkx. We *could* add the package as a dependency and use it but I'm not too excited about new dependencies too.

cvanelteren:
Hi @MridulS! Thanks for clarifying, it was what I suspected. The dependency I understand. The original implementation seems to be pretty low on activity. There are some full features missing from forceatlas2 (judging from the python code) that I may develop at some future stage. However, in its current stage it is highly usable and functioning. I think it would be a valuable addition to networkx. 

What do you mean exactly with porting it to python and networkx? The code in its current format has an optional dependency for cython, but will otherwise use either the precompiled code or the cython code in pure mode (python mode). To clarify, the default mode is  pure python. Converting is loses this flexibility, but I could aid in this too.
Cheers,
C

jarrodmillman:
The code also appears to have a GPL license:
https://github.com/cvanelteren/forceatlas2/blob/master/LICENSE

MridulS:
> What do you mean exactly with porting it to python and networkx?

When I saw cython as a dependency I thought parts of it were written in Cython, but that doesn't seem to be the case. thanks for pointing it out.

We can't include GPL code with NetworkX (this would require re-licensing the code base from BSD to GPL, this is at least what I understand IANAL).
If it stays as an optional dependency I'm not sure how it works out (the dynamic linking of GPL code. I am no expert)

MridulS:
@cvanelteren I'm going to close this for now, but thanks for starting up the conversation! I would like to see ForceAtlas2 included, but implemented specifically for NetworkX. The original paper https://doi.org/10.1371/journal.pone.0098679 would be good starting point :)

cvanelteren:
@MridulS  ok deal, I have implemented the forceatlas2 based on the paper in native python with networkx. Will perform a pull request later this week when I cleaned up the code a bit.

PR #5344, opened 2022-02-19
Break astar ties using heuristic value

lior8:
I changed A* tiebreaking to rely on the heuristic value of a node, and only then on the counter. This makes it significantly faster when dealing with searches where the path is "clear" to the algorithm, upwards of X10 faster than the current tiebreaker.

An alternative could be to allow the user to choose the tiebreaking via an additional optional parameter specifying if to tiebreak via heuristic value or not. Or even give the user the option to list a function which will recieve the current node, the target node, and will return a tuple specifying the tiebreaking value (while still using the counter to make sure no two elements are the same).

### DIFF
--- a/networkx/algorithms/shortest_paths/astar.py
+++ b/networkx/algorithms/shortest_paths/astar.py
@@ -14,6 +14,7 @@ def astar_path(G, source, target, heuristic=None, weight="weight"):
     using the A* ("A-star") algorithm.
 
     There may be more than one shortest path.  This returns only one.
+    Ties are broken based on the heuristic value (the smaller, the better).
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


dschult:
This looks like a nice improvement to me! 
Could you add some words to the paragraph at line 16 (it is inside the doc_string)? That line makes it clear that we only return one path when there are more than one. That seems like a good place to describe how we break ties. Your description above is quite good so use something like that.

Thanks!

lior8:
I added a line that explains about the tiebreaking.
I would like to say that in general, I think the A* functions need a bit of an update in terms of variable names and terms used to be more consistent with heuristic search conventions. Will it be alright if I updated it, verify my implementation with the existing tests (and write more if there is a need), and open a new PR for it?

dschult:
Can you give an example of something that you would change? 
Or even just  a pointer to what you mean by "consistent with heuristic search conventions". What conventions? By whose interpretation? 
We don't have a lot of time to review changes to variable names, and prefer PRs like this one where it is pretty easy to see what is happening.  OTOH if the changes are going to make the code significantly more readable and easier to figure out what it is actually doing, then that would definitely be worthwhile. The problem is that there is a lot of grey area between those extremes. :}

lior8:
There is an error we the current code. I will fix it and open a new PR once its done

PR #5369, opened 2022-03-01
BUG: Explicity import importlib.machinery

MridulS:
fixes https://github.com/networkx/networkx/issues/5367

Tested this on fedora 35 and it works
```
# pip install https://github.com/mriduls/networkx/archive/importlib.zip
# python3 -c "import networkx"
```

We should probably have a 2.7.1 release too.

### DIFF
--- a/networkx/lazy_imports.py
+++ b/networkx/lazy_imports.py
@@ -1,4 +1,5 @@
 import importlib
+import importlib.machinery
 import importlib.util
 import types
 import os


jarrodmillman:
@MridulS @dschult  I want Stefan to take a look at this b/c it may impact the lazy loading SPEC and I am not sure how he would like to fix this.  I am planning to release 2.7.1 next Monday.  So we should be careful with what we merge over the next week.  Straightforward bug fixes and documentation improvements would be good candidates.

stefanv:
I'd like to propose https://github.com/networkx/networkx/pull/5371 as an alternative way to address the issue, but it would need @dschult's review—he wrote that piece of code.

jarrodmillman:
See #5371.

PR #5397, opened 2022-03-17
Multigraph bridges

lucasmccabe:
This PR modifies `nx.bridges` to accommodate multigraphs. 

The fast bridge-finding algorithm used by `nx.bridges` ([Algorithm 1](https://arxiv.org/abs/1209.0700)) requires a simple graph. A bridge in a multigraph must have multiplicity one, so running the algorithm on the simple version of a multigraph and verifying that any discovered bridges are not multi-edges should be sufficient.

### DIFF
--- a/networkx/algorithms/bridges.py
+++ b/networkx/algorithms/bridges.py
@@ -13,7 +13,8 @@ def bridges(G, root=None):
 
     A *bridge* in a graph is an edge whose removal causes the number of
     connected components of the graph to increase.  Equivalently, a bridge is an
-    edge that does not belong to any cycle.
+    edge that does not belong to any cycle. Bridges are also known as cut-edges,
+    isthmuses, or cut arcs.
 
     Parameters
     ----------
--- a/networkx/algorithms/bridges.py
+++ b/networkx/algorithms/bridges.py
@@ -57,16 +57,14 @@ def bridges(G, root=None):
     ----------
     .. [1] https://en.wikipedia.org/wiki/Bridge_%28graph_theory%29#Bridge-Finding_with_Chain_Decompositions
     """
-    multi_edges = []
-    if isinstance(G, nx.MultiGraph):
-        multi_edges = [
-            (e[0], e[1]) for e in G.edges if G.number_of_edges(e[0], e[1]) > 1
-        ]
-        G = nx.Graph(G)
-    chains = nx.chain_decomposition(G, root=root)
+    multigraph = G.is_multigraph()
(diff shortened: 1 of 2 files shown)


dschult (commented):
Thanks for this improvement to bridges so it works with a multigraph!

If I understand the tests right, you make a double edge for (1,2) to see if an edge that would not be a bridge in a simple graph is still not a bridge as a double. And you make the double edge of (3, 4) to make sure an edge that would be a bridge in a simple graph is no longer a bridge when doubled. Finally, you leave (2,3) as a single edge to make sure it remains a bridge.  So I think that's a complete test.

For the algorithm code, it'd be nice to avoid making the set `set(G.edges())` because that  could be *very* big.  Avoiding holding all multiedges in a list (or set) would help too.  How about:
```python
    multigraph = G.is_multigraph()
    H = nx.Graph(G) if multigraph else G
    chains = nx.chain_decomposition(H, root=root)
    chain_edges = set(chain.from_iterable(chains))
    for u, v in H.edges():
        if (u, v) not in chain_edges and (v, u) not in chain_edges:
            if multigraph and len(G[u][v]) > 1:
                continue
            yield u, v
```

lucasmccabe:
> If I understand the tests right, you make a double edge for (1,2) to see if an edge that would not be a bridge in a simple graph is still not a bridge as a double. And you make the double edge of (3, 4) to make sure an edge that would be a bridge in a simple graph is no longer a bridge when doubled. Finally, you leave (2,3) as a single edge to make sure it remains a bridge. So I think that's a complete test.

Yes, that's exactly what I had in mind :)

> For the algorithm code, it'd be nice to avoid making the set set(G.edges()) because that could be very big. Avoiding holding all multiedges in a list (or set) would help too. 

Ah, good point. I can see how this is preferable, for instance, if there are a lot of multi-edges.

I've incorporated this into the most recent commit. Thank you!

dschult (approved):
I approve this PR.  I just thought of another documentation suggestion which I'll leave up to you.  Should we put a sentence in the bridges doc_string telling how we handle multiedges?

Thanks very much!

lucasmccabe:
> I approve this PR.

Thank you @dschult - your earlier review helped!

> Should we put a sentence in the bridges doc_string telling how we handle multiedges?

Good point. In the most recent commit, I've added a bit about this to the Notes section.

rossbar (approved):
LGTM, thanks @lucasmccabe !

lucasmccabe:
Great! Thank you @dschult and @rossbar

PR #5401, opened 2022-03-19
DOC: examples code blacks needs a blank line

Carreau:
Otherwise they are just seen as text
Screen Shot 2022-03-19 at 10 02 35


MridulS:
Thanks!

PR #5411, opened 2022-03-23
Added Lemke graph

cvanelteren:
Added Lemke graph (https://mathworld.wolfram.com/LemkeGraph.html)

### DIFF
--- a/networkx/generators/small.py
+++ b/networkx/generators/small.py
@@ -573,3 +573,32 @@ def tutte_graph(create_using=None):
     ]
     G = make_small_undirected_graph(description, create_using)
     return G
+
+
+def lemke_graph():
+    """Lemke graph
+
+    Lemke  graph consists  of 8  nodes with  13 edges  which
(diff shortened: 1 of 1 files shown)


rossbar (commented):
In principle this seems like a nice addition, though it will need a bit of polishing (adding it to the `__all__`, updating docstring, etc.)

Most of the other `small` graphs support `create_using`, but I don't necessarily know if it's worth it to support it here for the sake of consistency alone.

dschult:
I would like to see more information about the graph and maybe a sentence or two on why anyone would be interested. For example, the link you provide in your comment gives *very* brief reasons why it might be interesting.  Why did you get interested in it? Provide a citation or two?

Also, the example is not very helpful. I agree with @rossbar that some polishing is needed.

cvanelteren:
The Lemke graph here is the smallest graph that does not satisfy the 2-pebble problem. The create using would not be sensible I think for this generator, but could be added for consistency, let me know what works.

PR #5425, opened 2022-03-25
Keep node properties of original graph when creating bipartite clique graph

jpauwels:
When calling `make_clique_bipartite`, the resulting graph does not contain the properties of the original graph in the bottom nodes. This trivial PR fixes this.


dschult:
I recommend that the user add the node properties themselves.  In fact it is sometimes better to look up the node properties in the original graph rather than duplicating that storage. 

Some people want node attributes and others don't. In any case, it's much better for us to keep the API simple and it is easy to add them yourself. And the extra time and memory of copying them means we don't want that as the default.

Thanks for the suggestion though... :}

Issue #5433, opened 2022-03-28
DOC: scipy conf. citation links spin forever


ChrisKeefe:
The PDF and BibTeX links [here](https://networkx.org/documentation/latest/#citing) point to conference.scipy.org endpoints that never resolve. This is probably someone else's problem ™️, and I'm hoping you folks know who to check in with for a new URL or an upstream fix.

### Current Behavior
When I click the link, the target page never resolves, resulting in a Connection Timed Out error after a couple of minutes.

### Expected Behavior
I get a PDF or a BibTex file.

### Steps to Reproduce
Visit the citations section linked above, and click one of the affected links.

### Environment
OpenSUSE, Firefox/Chromium

MridulS:
Yeah that's not good. Thanks for the report!

aterrel:
Server has been restored. We should move these though as that server is on an old cluster that we want to migrate away from.

MridulS:
Thanks @aterrel!

PR #5441, opened 2022-03-29
Use https links for conference.scipy.org

ChrisKeefe:
The server reset from https://github.com/networkx/networkx/issues/5433 fixed the resolution issues, but both Firefox and Chromium refuse to play nice with the .`bib` file on this machine, apparently because they're concerned with the insecurity of the `http:` target. 

ff_not_downloaded

Both are happy to download the .bib from `https:...`. I've changed the PDF and .bib links to `https`.


jarrodmillman (approved):

dschult:
Thanks @ChrisKeefe !

PR #5451, opened 2022-04-01
Adding more examples for to_numpy_array method's usage

dtekinoglu:


dschult:
It looks like the output from your last example is written as floating point (with the decimal point) but the function is actually using integers. That is because the datatype (dtype) of the "nonedge" value is an integer and the whole matrix takes the dtype of the nonedge value unless it is specified elsewhere.  (Notice that the default for nonedge is `0.0` which has type `float64`, while `nonedge=-1` produces dtype `int64`).

You should either change your expected answer to the last example to report integer values, or make the nonedge value 1.
I think your indentation of the array output is also missing some spaces to align the first line with the rest.

You can see the error messages if you dig through the "details" link next to a failed test.  Search for "error" or "fail" or scroll up and down until you find where an error is reported.

dtekinoglu:
Thank you @dschult ! After changing nonedge value to a float and fixing the indentation issue, all checks were succesful.

dschult (commented):
Hurray for passing the tests.

Two other comments:  
- I liked the example of a negative value for the `nonedge`. The value `0.1` is too close to a "normal" edge weight.  Could you change it back to `-1.` which is still `-1` but as a float instead of an int?
- Our style checker (black) doesn't check example code. So could you update the second example to match the PEP 8 python style standard?:  no spaces before or after `=` when used as a keyword assignment, and a space after each comma between function arguments.

Thanks...

dtekinoglu:
Thank you, @dschult for the feedback.  I made the changes you suggested.

dschult (approved):
This Looks good to me.

MridulS (approved):
Looks good to me!

Thanks for contributing to @dtekinoglu!

PR #5452, opened 2022-04-01
Add examples for the condensation function

kpetridis24:
Added two examples for the better understanding of the condensation funciton use

### DIFF
--- a/networkx/algorithms/components/strongly_connected.py
+++ b/networkx/algorithms/components/strongly_connected.py
@@ -373,6 +373,7 @@ def condensation(G, scc=None):
 
     Examples
     -----
+    # Contracting two sets of strongly connected nodes, into two distinct SCC using the barbell graph.
     >>> G = nx.barbell_graph(4, 0)
     >>> G.remove_edge(3, 4)
     >>> G = nx.DiGraph(G)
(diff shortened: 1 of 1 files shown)


kpetridis24:
I fixed the new line issues and re-created the previously submitted PR with no issues in the current version.

networkx/algorithms/components/strongly_connected.py
```diff
@@ -371,6 +371,26 @@ def condensation(G, scc=None):
     NetworkXNotImplemented
         If G is undirected.
 
+    Examples
+    -----
+    >>> G = nx.barbell_graph(4, 0)
+    >>> G.remove_edge(3, 4)
+    >>> G = nx.DiGraph(G)
```
MridulS, line 378: Thanks for this!
Could you also add a quick comment about the graph construction. I wrote down the example
quickly without a comment. Could you put something like `# create 2 distinct SCC using the barbell graph`

networkx/algorithms/components/strongly_connected.py
```diff
@@ -371,6 +371,26 @@ def condensation(G, scc=None):
...
+    >>> H = nx.condensation(G)
+    >>> H.nodes.data()
+    NodeDataView({0: {'members': {0, 1, 2, 3}}, 1: {'members': {4, 5, 6, 7}}})
+    >>> H.graph['mapping']
+    {0: 0, 1: 0, 2: 0, 3: 0, 4: 1, 5: 1, 6: 1, 7: 1}
+
+    >>> G = nx.complete_graph(7)
+    >>> G = nx.DiGraph(G)
```
MridulS, line 386: There is no need to cast the `DiGraph`, for complete graphs we can directly get a `DiGraph` with `create_using`.
`G = nx.complete_graph(7, create_using=nx.DiGraph)` should do the job :)

kpetridis24: Oh nice, i haven't used this function ever. Thanks for the clarification.

MridulS (approved):

dschult (approved):

PR #5454, opened 2022-04-01
Make sympy extra dep

jarrodmillman:
I made sympy an extra (not default) dependency as we had previously discussed not making things default dependencies until they are used a lot or for core functionality.

I also updated the minimum dependency from 1.0 to 1.10. This is a new dependency used by a single function that didn't exist before. I don't think there is any need to support a release of sympy that came out in 2016 and only claims to work on Python 3.5 or older.  (We only run on 3.8 or more recent).

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,4 +2,3 @@ numpy>=1.19
 scipy>=1.8
 matplotlib>=3.4
 pandas>=1.3
-sympy>=1.0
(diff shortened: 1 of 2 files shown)


rossbar (approved):
I have no strong opinions about whether sympy should be a `default` or `extra` dependency. @jarrodmillman makes a good point that it's only used in one function unlike the other default dependencies which are more integrated throughout the codebase (or provide a commonly-used feature like plotting).

I think the original thinking on adding it as a default dependency was that `sympy` is a pure-Python package and thus unlikely to cause e.g. transitive dependency issues. No strong feelings either way, +0.5 for moving to `extra` for now.

PR #5456, opened 2022-04-02
Fixed wrong dict factory usage on MultiDiGraph

FabianBall:
Fixed the issue that the wrong dict factory on a MultiDiGraph was used for edge attributes (edge_key_dict_factory instead of edge_attr_dict_factory)

Extended tests to typecheck the dict factories and added a test that incorporates custom dict factories on a MultiDiGraph

### DIFF
--- a/networkx/classes/multidigraph.py
+++ b/networkx/classes/multidigraph.py
@@ -481,7 +481,7 @@ class MultiDiGraph(MultiGraph, DiGraph):
             key = self.new_edge_key(u, v)
         if v in self._succ[u]:
             keydict = self._adj[u][v]
-            datadict = keydict.get(key, self.edge_key_dict_factory())
+            datadict = keydict.get(key, self.edge_attr_dict_factory())
             datadict.update(attr)
             keydict[key] = datadict
         else:
--- a/networkx/classes/tests/test_graph.py
+++ b/networkx/classes/tests/test_graph.py
@@ -26,6 +26,11 @@ class BaseGraphTester:
 
     def test_nodes(self):
         G = self.K3
+        assert isinstance(G._node, G.node_dict_factory)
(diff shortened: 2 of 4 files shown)


dschult (approved):
This looks great @FabianBall !   Thanks very much!

The test for Mypy relates to changing the type of the default factories and following #5127 shows that mypy doesn't deal with these well. I'll leave it to @rossbar @stefanv and @NeilGirdhar to figure out whether we should set this up to be ignored or perhaps there is a better way to finesse the type checking rules. 

I believe this is ready to go other than mypy -- so I'm approving it.

rossbar (approved):
I spent some time looking at this and couldn't find any obvious, non-invasive way to make mypy happy so I went with the solution in #5127 and added `# type: ignore` to all the subclass `_factory` attributes. There may be a way to "turn off" the attribute-type-inference-from-parent feature in mypy globally, but it shouldn't be a blocker here.

Everything else LGTM, thanks @FabianBall that was a very nice catch and the updated/new tests are excellent!

FabianBall:
Perfect, glad I could help. I surely would not have stumbled across the issue if I had not have used the feature...

NeilGirdhar:
> I spent some time looking at this and couldn't find any obvious, non-invasive way to make mypy happy so I went with the solution in https://github.com/networkx/networkx/pull/5127 and adde

I annotated them here: https://github.com/neilgirdhar/networkx/tree/annotations

However, the real problem is that callable class variables is just awkward design.  And setting those class variables as member variables is the constructor is even more awkward.  The ordinary way to accomplish this is with ordinary methods that subclasses can overload.  This seems to me like over-optimization, but perhaps there examples where it makes a significant difference.

danieleades:
see #5890

PR #5460, opened 2022-04-03
JSON graph documentation

aliveevie:
From Networkx documentation this was they wrote on the issues:


MridulS:
Hi @aliveevie! It's not clear from these changes what's happening here. Also there is already an open Pull Request that fixes the issue #5277 -> https://github.com/networkx/networkx/pull/5323 (We should try to get that merged in!).

But please do review the code in https://github.com/networkx/networkx/pull/5323 too! It would be helpful to get another set of eyes on it :)

PR #5488, opened 2022-04-06
added a new centrality: decay centrality

unna97:

### DIFF
--- a/networkx/algorithms/centrality/__init__.py
+++ b/networkx/algorithms/centrality/__init__.py
@@ -17,3 +17,4 @@ from .percolation import *
 from .second_order import *
 from .trophic import *
 from .voterank_alg import *
+from .decay import *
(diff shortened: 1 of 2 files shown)


unna97:
To Do:
-[ ] Add tests for the decay centrality

- [ ] Currently doesn't throw errors for weighted graphs. Couldn't find the definition for weighted or directed graphs, figure out how to handle it.

unna97:
@MridulS  could you help me out with how to approach the second problem, please?

unna97:
@MridulS  I found this implementation for weighted and directed graph: https://rdrr.io/cran/centiserve/man/decay.html

I am going to implement it as defined above, is that okay?

BrandonKMLee:
noob question:
```python
    if u is None:
        nodes = G.nodes
    else:
        nodes = [u]
```
can be rewritten as `nodes = [u] if u is None else G.nodes`?

PR #5497, opened 2022-04-07
Tests for decay centrality

unna97:


dschult:
The "style" error is to use double-quotes instead of single-quotes on your text strings. You can get that fixed automatically by installing and running "black" on the file.  "black --diff" displays the changes without changing the file.

It would also be good to make the doc_strings wrap their text so that lines are ~80 chars or less.

What do you think about putting this new function into the `degree_alg.py` module rather than starting a new module? I'm fine either way, but wonder if you thought of the best place for this function. :}

unna97:
@dschult  thank you for the suggestions! I did think about it putting as a function but couldn't figure out the module to put it in. I will shift it to degree_alg.py.  Also, should I implement this centrality for weighted and directed graphs as well from the definition?
https://github.com/networkx/networkx/pull/5488

unna97:
@dschult  Can you help figure out why this is failing? It doesn't fail on my local system. Is it maybe because I used pytest exact instead of approx?

dschult:
If you click on the "Details" link for the test that is failing you can see the exact error message. (I usually look at one of the non-coverage test failures first because they give slightly less output to scroll through.) 

I suspect that if you test locally using the doctest feature the error will show up.
```pytest --doctest-modules decay.py tests/test_decay.py```

It indicates that you didn't leave a blank line after the Examples part of the doc_string. So the test expects to get the output along with all the text of the references section of the doc_string.

**Comment by unna97** [comment]
Thank you, @dschult.  Is there a way to automate the docstring formatting?

To do tasks:

- [ ]  Add tests for weighted graphs
- [ ]  Change exact check for tests to approx for float

**Comment by unna97** [comment]
@dschult  Will this pull request be reviewed?

**Review by dschult — COMMENTED** [review]
Here are some things to take a look at:

code module:
- temp_G is not actually used in your code. It only points to the original G. It doesn't save the original G. But luckily you don't need it. The line `G = G.to_undirected()` doesn't change G. It only returns an undirected version of G which you then assign to the name G. So you are good to go and you don't have to do any cleanup for G near the end.
- Your input `u` should allow a set of nodes as well -- like `nbunch` for the base graph class methods. You can call it `nbunch` (a common name used in other functions) and use it to set up a variable `nodes` saving `nbunch` so you can use it near the end to check whether to return a single value or a dict. Check if nbunch is in G and if so wrap it in a list the way you do with `u`. Also check if nbunch is None as you do. If neither of these then set nodes to nbunch. Does that API make sense to you?  Of course the doc_string has to be adjusted accordingly.
- Instead of setting path_length and then putting it inside some logic in the loop, set up the function ahead of time inside the logic, and just use path_length in the loop without the logic.  Something like:
```python
if mode == "in" and G.is_directed():
    path_length = lambda n : nx.shortest_path_length(G, target=n, weight=weight)
else:
    path)_length = lambda n : nx.shortest_path_length(G, source=n, weight=weight) 
```
- There is no need to set decay_centrality to 0 inside the loop.
- I think you don't need to check `v == n` in line 102. If they are equal the distance will be zero. 

Tests:
- You've got the tests all within a single class. It is now deemed better to use functions one at a time rather than a class. We are moving that direction and would like your code in that style. Basically, your setup creates all the graphs needed for all the tests, and it gets run before each test. But each test only uses one of the graphs. So you should shift the methods to be function with the first part of each function creating the graph.
- Your tests check the values for a number of graphs, but it is not clear which are exact values you know to be correct, and which are simply checking values that were run on the graph once-upon-a-time so you are checking that the values haven't changed.  Name them so they describe what aspect of the function you are checking. If the function tests directed versions of the code, then name the function that.
- You don't seem to have tests that make sure the special cases are treated the way you expect. For example, you have logic about `mode=='all' and G.is_directed()`. So test with e very simple graph that this returns the value you expect. For example, a 2 or maybe 3 edge graph with edge weights 1 should be used for the two sides of this logic to make sure both sides are computed correctly.  Similarly test inputing a single node as well as a subset of nodes and all nodes. Test the corner cases like empty graphs. And test that the right exceptions are raised when they should be.  You know that *your* code will do this, so you are building a system to make sure someone else (who could be you six months from now) doesn't accidentally change something without realizing it. :}
- Is there any danger of the tests being affected by round-off error? It looks like it is possible. You should consider using `pytest.approx` and choosing the precision for which you want the answers to agree.

The doc_string:
- much of the doc_string is not rendering correctly by the sphinx documentation building system. You can see the documentation in the **ci/circleci: documentation artifact** automagic testing system checks in your PR.  That one shows up near the bottom -- maybe 35 out of 37 or something. Click on "Details" to link to a page with the new docs. Then navigate to your function and make sure it looks right. The needed changes are mostly spacing issues. Blank lines after the first line and between every section.
- The value after each colon in the parameter section should be the type of the output. For example `delta : float (default=0.5)` Then call it the decay parameter in the sentence that describes what it means.
- Check that the See Also section links actually link to those functions.

You can also build the documentation locally if you have sphinx and a few other packages listed in `requirements/doc.txt`. Go to the `doc` directory and use `make html` to actually do the build locally. Then you can check that they render well and don't have to wait for te CI testing system to finish.

:)

unna97:
@dschult  Thanks for the feedback, I will work on this. Could you maybe point to tests in function style (if they are already present)?

unna97:
- [ ]  Allow set of nodes and change doc accordingly
- [ ] add tests for edge cases and exception
- [ ]  Change tests from class to function
- [ ]  Doc Changes

dschult:
There might be a better example, but `networkx/algorithms/tests/test_swap.py` is reasonable and uses functions.

unna97:
@dschult  I think I have finished all the tasks you have highlighted. Let me know if any other changes are required.

networkx/algorithms/centrality/decay.py
```diff
@@ -0,0 +1,115 @@
...
+    weight : string or function, optional (default = None)
+      If None, every edge has weight/distance/cost 1.If a string, use this edge
+      attribute as the edge weight.Any edge attribute not present defaults
+      to 1. If this is a function, the weight of an edge is the value returned
+      by the function. The function must accept exactly three positional
+      arguments:
+      1. the two endpoints of an edge
+      2. the dictionary of edge attributes for that edge.
+      The function must return a number.
+    Returns
```
dschult, line 49: The doc formatter wants a blank line before each "section" (underlined..  like Returns, Examples, etc).
We use 4 space indentation.
The "weight" parameter should use the same doc_string description as for `shortest_path_length`.
Each reference should have the two dots in front of it. 

You can check how all these documentation changes look in the generated html docs (one of the "tests" near the bottom of the list is called "documentation artifact" and if you click on the details it shows the html docs.

networkx/algorithms/centrality/decay.py
```diff
@@ -0,0 +1,115 @@
...
+        nodes = G.nodes
+    else:
+        if type(u) is list:
+            nodes = u
+        elif u in G:
+            nodes = [u]
+        else:
+            raise nx.NetworkXError("u is not a node in the graph")
+
+    for n in nodes:
```
dschult, line 100: Variable name: Can you use "node" instead of "n" since the formula contains "n" as the number of nodes.

networkx/algorithms/centrality/decay.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if u is None:
+        nodes = G.nodes
+    else:
+        if type(u) is list:
+            nodes = u
+        elif u in G:
+            nodes = [u]
+        else:
+            raise nx.NetworkXError("u is not a node in the graph")
```
dschult, line 98: Sorry -- It looks like I will have to backtrack on what I suggested before. The other centrality measures don't seem to use `nbunch` like I thought they did. The module `closeness.py` is fairly representative. It allows input parameter `u` as an optional node. But not a list of nodes. And the return value is a dict when `u is None`, but it just returns the centrality value (not in a dict) when `u` is specified.  If you decide to change this back, you'll need to change the returns section of the doc_string too.

networkx/algorithms/centrality/tests/test_decay_centrality.py
```diff
@@ -0,0 +1,327 @@
...
+import pytest
+
+import networkx as nx
+
+
+def test_decay_centrality_K():
+    """
+    Test the decay centrality of the krackhardt kite graph.
+
+    """
```
dschult, line 15: Can you put the comments on all one line with triple " on that same line? (trying to reduce the number of blank lines)

networkx/algorithms/centrality/tests/test_decay_centrality.py
```diff
@@ -0,0 +1,327 @@
...
+Unit tests for decay centrality
+
+"""
+
+import pytest
+
+import networkx as nx
+
+
+def test_decay_centrality_K():
```
dschult, line 11: rename the tests to describe what they are testing: e.g. `test_decay_centrality_kite_graph`
`test_decay_centrality_tree`
`test_decay_centrality_mode_inputs`
etc

dschult (commented):
The main things I see are minor formatting changes of the doc_string, variable names in the code and changing names of the tests to indicate what feature they are testing. These are all minor. Push back if you'd prefer not to make a change/suggestion, or to change it in a different way.

The substantive issue I came up with is back-tracking on something I suggested before. The other centrality measures don't allow a list of nodes -- either a single node or all the nodes (when u is None). When a single node is requested, only the value is returned (not wrapped inside a dict).  I don't feel strongly on how this should work -- so maybe following the lead of a representative function like those inside `closeness.py`.

MridulS:
Thanks for your work on this @unna97 ! This hasn't been touched in a while, let me know if you still have the bandwidth to finish this up :)

unna97:
@MridulS  yes, I can finish this up. Will work on it in next few days!

PR #5502, opened 2022-04-08
Added example on efficiency measure

Lukong123:
Added example under efficiency measure


networkx/algorithms/chains.py
```diff
@@ -49,6 +49,11 @@ def chain_decomposition(G, root=None):
     The worst-case running time of this implementation is linear in the
     number of nodes and number of edges [1]_.
 
+    Examples
+    --------
+    >>> G = nx.path_graph(5)
+    >>> chain_d = nx.chain_decomposition(G)
```
jarrodmillman, line 55: I am not sure this example is good illustration.  Does a path graph have a meaningful chain decomposition?  I would also recommend showing what the chain decomposition is rather than assigning it to a variable.

Also, does anyone know if this reference makes sense here. I didn't look the paper up, but the title doesn't mention chain decompositions and it seems very recent.

Lukong123: Hello sir, @jarrodmillman  thanks for pointing that out. 
Please check out this recent example, I hope it is better than the previous one.

MridulS:
changes merged in https://github.com/networkx/networkx/pull/5641, closing this one

PR #5516, opened 2022-04-10
a hack to force self edges to be ignored on the first node inspected

sheldonkhall:
An attempt to resolve a dormant bug in the graph edit distance algorithm. The cited paper leaves the implementation of the edge matching up to the user. In the current implementation every time a new node in G1 is matched, self-edges on the newly matched node and every edge from the newly matched node to every other previously matched node in the search tree are inspected for potential edge matches in G2.

The issue arises when the first node matched has a self edge on it. The first node matched implicitly has no previous nodes matched and therefore no edges to any other nodes are inspected. This forces the self-edge to be deleted rather than being matched to an edge between two nodes.

This fix stops edge matching occurring on the first node match. Instead it is allowed only after at least one node match has already occurred.

For full disclosure one of the commented assert statements now fails but all tests pass. I think this is not quite ready to merge, but I was interested to get a second opinion.

### DIFF
--- a/networkx/algorithms/similarity.py
+++ b/networkx/algorithms/similarity.py
@@ -745,18 +745,30 @@ def optimize_edit_paths(
         N = len(pending_h)
         # assert Ce.C.shape == (M + N, M + N)
 
-        g_ind = [
-            i
-            for i in range(M)
-            if pending_g[i][:2] == (u, u)
-            or any(pending_g[i][:2] in ((p, u), (u, p)) for p, q in matched_uv)
-        ]
-        h_ind = [
-            j
-            for j in range(N)
-            if pending_h[j][:2] == (v, v)
-            or any(pending_h[j][:2] in ((q, v), (v, q)) for p, q in matched_uv)
-        ]
+        # only attempt to match edges after one node match has been made
+        # this will stop self-edges on the first node being automatically deleted
+        # even when a substitution is the better option
+        if matched_uv:
+            g_ind = [
+                i
+                for i in range(M)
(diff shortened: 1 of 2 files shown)


sheldonkhall:
It is strange. That failure seems to only occur on the base (ubuntu, 3.11-dev) image and it fails on an import numpy statement and it is hard to tell why it is related, but I will try to resolve it.

dschult:
Ahh...   I've found the trouble.  The other tests in this test module `importorskip` numpy at the start of the test class.  This new test is outside the test class, so you'll need to call `importorskip` yourself, or make the test a method inside the class.

```python
        pytest.importorskip("numpy")
        pytest.importorskip("scipy")
```
(I dont think you need the scipy... but I wasn't sure and too lazy to check..:)
If you put that line in the test function, it will try to import numpy and if that fails it will skip the test.

dschult (approved):
This change fixes the issue and I think it should be merged.

But -- looking through this code, I believe we should look through for a rewrite to take advantage of current Python features. For example, there are many `list` functions calls which should be made into list comprehensions. Similarly the `zip(range(m), g_ind)` and similar structures should 1) use `enumerate` and avoid indexing completely. e.g. `for g in g_ind`. Furthermore, the function being tweaked here would be easier to read IMO if we handled the case of an empty `matched_uv` up front.  That case falls through the logic, executing various `else` clauses when we know what will happen to it.  But all of this is beyond the scope of this more immediate fix.

I approve this PR and I will open an issue to suggest a rewrite/update of this code.

jarrodmillman (approved):

PR #5517, opened 2022-04-10
Solves Issue #5486 partially

dtekinoglu:
I modified documentation of bridges algorithms in the following ways:

- Although `@not_implemented_for("multigraph")` was not stated for `bridges()` and `has_bridges()` methods, if the input graph is multigraph, they raise `NetworkXNotImplemented`. Considering this fact, I added  `@not_implemented_for("multigraph")` statements for this functions. This might be not only a problem with documentation, but also a defect.
- I added `Raise` fields to the documentation parts of all three bridge algorithms.

### DIFF
--- a/networkx/algorithms/bridges.py
+++ b/networkx/algorithms/bridges.py
@@ -7,6 +7,7 @@ from networkx.utils import not_implemented_for
 __all__ = ["bridges", "has_bridges", "local_bridges"]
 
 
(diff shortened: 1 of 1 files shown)


dschult:
I believe the bridges code was recently made to work with multigraph.  #5397 
So this PR is probably no longer needed...  

Comment or reopen if needed...

PR #5536, opened 2022-04-14
Added examples in tournament and tree functions

NikitaSharma1:
Added examples in tournament functions and is_branching() and is_aborescence() functions of tree.


rossbar (approved):
The added examples LGTM, concise and informative. Thanks @NikitaSharma1 !

PR #5549, opened 2022-04-17
Fix min_edge_cover in special cases (#5538)  and correct documentation

dschult:
Fixes the defect described in #5538 while maintaining the current discrepancy between `matching` APIs for bipartite and non-bipartite matching functions. bipartite returns a matching dict of mates, while non-bipartite returns the matching as a set of edges.  

That means that the corrected `min_edge_covering` still returns two types of sets -- if a bipartite matching algorithm is used, it returns a set with both 2-tuples for each edge; while if a non-bipartite matching algorithm is used, it returns a set with one 2-tuple for each edge.  It seems like this should be changed too. And I could add that to this PR or to a separate PR.

This PR also updates the documentation which e.g. currently says that this function only works for bipartite graphs.

### DIFF
--- a/networkx/algorithms/covering.py
+++ b/networkx/algorithms/covering.py
@@ -56,6 +56,8 @@ def min_edge_cover(G, matching_algorithm=None):
     Minimum edge cover for bipartite graph can also be found using the
     function present in :mod:`networkx.algorithms.bipartite.covering`
     """
+    if len(G) == 0:
+        return set()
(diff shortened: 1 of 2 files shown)


rossbar (approved):
LGTM, thanks @dschult ! The updated documentation certainly make it clear what was happening and why, and what users can expect depending on the provided `matching_algorithm`.

PR #5560, opened 2022-04-19
Adding an Example Map for GSoC Proposal

stanyas:
I am addressing the issue [5510](https://github.com/networkx/networkx/issues/5510).

Please, disregard the first two commits, this is my first time making a pull request, and I experimented.


stanyas:
graphs

This is how the example graph looks

MridulS:
Before the review just wanted to point out that the GSoC application deadline is under 4 hours so please submit the application before that, Google is pretty strict with deadlines :)

stanyas:
> Before the review just wanted to point out that the GSoC application deadline is under 4 hours so please submit the application before that, Google is pretty strict with deadlines :)

Thanks, I was working on it. I submitted.

NikitaSharma1:
@stanyas I faced a similar issue in style in the start as you are facing now. I was told to make the following changes and it worked for me. 
 `One thing that will need to be done is to ensure that the docstring examples are properly formatted, i.e. the code in the examples follows the black formatting rules. Since these examples are small, it may be easy enough to fix up the formatting manually. Alternatively, I believe a pre-commit hook has been recently added to handle doc formatting automatically, so updating your development environment with the latest pre-commit hooks would likely work too`
I think you should try this too. 
Please correct me if I'm wrong.

NikitaSharma1:
@stanyas I faced a similar issue in style in the start as you are facing now. I was told to make the following changes and it worked for me. 
 `One thing that will need to be done is to ensure that the docstring examples are properly formatted, i.e. the code in the examples follows the black formatting rules. Since these examples are small, it may be easy enough to fix up the formatting manually. Alternatively, I believe a pre-commit hook has been recently added to handle doc formatting automatically, so updating your development environment with the latest pre-commit hooks would likely work too`
I think you should try this too. 
Please correct me if I'm wrong.

stanyas:
> @stanyas I faced a similar issue in style in the start as you are facing now. I was told to make the following changes and it worked for me. `One thing that will need to be done is to ensure that the docstring examples are properly formatted, i.e. the code in the examples follows the black formatting rules. Since these examples are small, it may be easy enough to fix up the formatting manually. Alternatively, I believe a pre-commit hook has been recently added to handle doc formatting automatically, so updating your development environment with the latest pre-commit hooks would likely work too` I think you should try this too. Please correct me if I'm wrong.

Thanks, you were right!

MridulS:
Thanks for starting discussion on this one @stanyas, we can use https://github.com/networkx/networkx/issues/5824 for further discussion on this one :)

PR #5563, opened 2022-04-20
Gallery example for Maximum Independent Set

stanyas:


rossbar (commented):
Maximum independent set seems like a good candidate for a gallery example!

It looks like there are still some unintended modifications left over from some previous experiments, e.g. the `x` that's been added to the `mayavi2_spring.py` example - that should be removed from this PR.

There are a few other updates that need to be in place in order to have a working gallery example:
 1. The example needs to be moved to the appropriate place. Currently the gallery examples are derived from the `examples/` directory, so it should live there. Given the content of the example I'd suggest maybe putting it in the `Algorithms` subcategory.
 2. The filename has to start with `plot_` in order to be picked up by sphinx-gallery: see the other examples in the `examples/` folder for guidance.
 3. Similarly the example should have an rst-formatted explanatory string at the top to give a brief summary of what the example is illustrating. Again, the other gallery examples are a good place to look to get an idea of what this should look like.

Once these structural updates are in place I'd be happy to take a closer look at the example itself!

MridulS:
@stanyas let us know if you still have the bandwidth to finish up this PR :)
Thanks for opening this up!

rossbar (approved):
I think this is a nice example, so I went ahead and pushed up the final set of changes to get this fully plugged into the gallery machinery. I also took the liberty of simplifying the example a bit. Thanks for this @stanyas !

MridulS:
Thanks @stanyas @rossbar !

PR #5600, opened 2022-05-04
Cache `nodes` property on Graph

stefanv:
Closes https://github.com/ComplexGroupInteractions/xgi/issues/103

### DIFF
--- a/networkx/classes/graph.py
+++ b/networkx/classes/graph.py
@@ -8,6 +8,7 @@ Self-loops are allowed but multiple edges are not (see MultiGraph).
 For directed graphs see DiGraph and MultiDiGraph.
 """
 from copy import deepcopy
+from functools import cached_property
 
 import networkx as nx
 from networkx.classes.coreviews import AdjacencyView
@@ -658,7 +659,7 @@ class Graph:
             except KeyError:
                 pass
 
-    @property
+    @cached_property
     def nodes(self):
         """A NodeView of the Graph as G.nodes or G.nodes().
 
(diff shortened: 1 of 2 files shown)


jarrodmillman (approved):

dschult (commented):
This is very good to know about!  We should do this in more places in the Graph class:

`G.adj`, `G.edges`, `G.nodes` should each be `cached_property` instead of `property`.

 -- though not for `G.name`. That one should be calculated each time it is called.

dschult:
After fussing with this for a while I believe the cached_property is interfering with the current view handling of `__slots__`.
It's a nice rabbit hole to fall into, but I'm currently wondering if the need for `__slots__` goes away once we have `cached_property`.  The motivation for `__slots__` is speed and memory usage for objects that are created many times. If we are caching the reportviews like `G.edges`, that motivation is significantly less.  My experience with slots is that they work fine, but you have to be really careful with copy and deepcopy and pickle. So, it would reduce our maintenance of the Views significantly if we remove the slots and just use a dictionary worth of space for each view created. The cached properties that should be on the same order as the number of graphs rather than the number of time the user calls G.edges.

I'd love some thoughts from people who have used `__slots__` before. And also from those who have had difficulty maintaining the Views because of `__slot__` considerations.  I'm confident we can eventually implement both cached properties and `__slots__`. But would it be better to just get rid of the `__slots__`?

dschult (approved):
The cached_property for nodes works well. 
I'll make a separate one for edges/degree/adj and the others.

So this is ready to go.

PR #5602, opened 2022-05-04
Add generate / write "network text" (formerly graph_str)

Erotemic:
This is an extension of the `forest_str` that I previously submitted. I found myself working with `forest_str` very often. I've found it very useful to be able to print out the graph in an interpretable manner.

But recently I've been finding myself working with non-forest graphs or near-forest graphs, and the existing `forest_str` function is not setup to work with them. But after thinking about it, I realized it would still be useful to look at the DFS tree of a lot of the graphs, and given a DFS tree you could munge it to denote where backedges are, thus giving a complete representation of a general graph.

The way this works is that when an outgoing edge is connects a node already in the tree, there is a special suffix written to the right of node indicating what those non-tree edges are. When a node recuses into a child that already drew one of these implicit back-edges it emits an ellipsis (or lets the vertical line trail) for conciseness. 

The doctest illustrates 3 cases:

* The case of a regular forest, no backedges are produced
* The same forest but with one backedge introduced between 1 and 7 
* A fully connected clique.

And the unit test illustrates the case of incrementally going from a graph with no edges to a fully connected graph. This does show that there are a few cases that might be cleaned up. For self loops the current implementation seems to show the edges twice, so I'll see if I can fix that, but overall I think the idea works well and gives as good of a visualization as one might hope to get for a general graph in text form.

I coded this up as a variant of `forest_str`, and then replaced the logic of `forest_str` with a call to this new `graph_str` function as to not break compatibility.


MridulS:
I have restarted the CI, the tests shouldn't fail now :)

Erotemic:
@MridulS Thanks! It's working now. This PR is finished and ready for review: @rossbar @dschult @jarrodmillman

I've expanded the tests cases, which also demonstrate the sort of text visualizations `graph_str` generates.

Note that the new `nx.graph_str` function completely supersedes `nx.forest_str` (with the only difference being the former doesn't throw an error when it gets a non-forest graph), such that there is no point to `nx.forest_str` anymore. It may be a good idea to deprecate `forest_str`, (or perhaps not add a new function and just allow forest_str to work with any graph?).

networkx/readwrite/tests/test_text.py
```diff
@@ -1,3 +1,4 @@
+import itertools as it
```
dschult, line 1: The abbreviation `it` is too commonly used... That makes it hard to find where the itertools are used (though not that hard). It looks to me like you only use itertools.product.  Maybe you could just use `from itertools import product` or I'm OK with `import itertools` and then use the full name below (those lines aren't very long anyway).

networkx/readwrite/text.py
```diff
@@ -72,6 +75,131 @@ def forest_str(graph, with_labels=True, sources=None, write=None, ascii_only=Fal
...
+        write=write,
+        ascii_only=ascii_only,
+    )
+
+
+def graph_str(graph, with_labels=True, sources=None, write=None, ascii_only=False):
+    """
+    Creates a nice text representation of a graph
+
+    This works via a depth-first traversal the graph and writing a line for
```
dschult, line 95:
```suggestion
    This works via a depth-first traversal of the graph and writing a line for
```

networkx/readwrite/text.py
```diff
@@ -72,6 +75,131 @@ def forest_str(graph, with_labels=True, sources=None, write=None, ascii_only=Fal
...
+
+    >>> graph = nx.complete_graph(3, create_using=nx.DiGraph)
+    >>> print(nx.graph_str(graph))
+    ╙── 0 ╾ 1, 2
+        ├─╼ 1 ╾ 2
+        │   ├─╼ 2 ╾ 0
+        │   │   └─╼  ...
+        │   └─╼  ...
+        └─╼  ...
+    """
```
dschult, line 198: Is there a reference that describes this notation for graphs in detail?

networkx/readwrite/text.py
```diff
@@ -72,6 +75,131 @@ def forest_str(graph, with_labels=True, sources=None, write=None, ascii_only=Fal
...
+    from collections import defaultdict
+
+    import networkx as nx
+
```
dschult, line 202: Can we put these import statements at the top just after the module doc_string? 
They don't represent optional libraries (like nimpy and scipy) and so don't need to be imported inside the functions.

dschult (commented):
I like this! I think of it as an ASCII representation of the graph. (maybe unicode, maybe text...whatever). :}

The improvement from forest-only to any kind of graph is nice. And I agree that we should deprecate the `forest_str` name and point people to the new name.  

Can we find a name more informative than `graph_str`?  
Perhaps `generate_ascii_text` or `generate_unicode` or `generate_text`?
That helps this follow the "generate/write/parse/read" naming convention within the read/write package.  And the term `str` is too common and not informative enough.  

I think this module could also be comfortable in the `drawing` subpackage. But it is fine here too.

networkx/readwrite/text.py
Erotemic: I made this notation up, so I didn't follow any reference. I wouldn't be surprised if something like this does exist and has a detailed description, but I stumbled on it as a natural extension of the more obvious way to draw an "text forest". Let me see if I can summarize:

1. Given a set of source nodes (which can be specified, or automatically discovered via finding the CCs/SCCs and choosing one node with minimum degree from each), we traverse the graph in depth first order.

2. Each reachable node will be printed exactly once on it's own line.

3. Edges are indicated in one of three ways: 

    a. a parent "L-style" connection on the upper left. This corresponds to 
    b. a backref "<-style" connection shown directly on the right. For directed graphs, these are drawn for any incoming edges to a node that is not a parent edge. For undirected graphs, these are drawn for only the non-parent edges that have already been represented (The edges that have not been represented will be handled in the recursive case).
    c. a child "L-style" connection on the lower right. Drawing of the children are handled recursively.

4. The children of each node (wrt the directed DFS tree) are drawn underneath and to the left of it. In the case that a child node has already been drawn the connection is replaced with a "..." to indicate that there is one or more connections represented elsewhere. 
 

There might be a way to phrase this more completely / succinctly (or a reference to something that describes the same structure), but that's my stab at it.

Erotemic:
> The improvement from forest-only to any kind of graph is nice. And I agree that we should deprecate the forest_str name and point people to the new name.

I've added a template for a deprecation warning with exact versions TBD. 

> Can we find a name more informative than graph_str? Perhaps generate_ascii_text or generate_unicode or generate_text?
That helps this follow the "generate/write/parse/read" naming convention within the read/write package. And the term str is too common and not informative enough.

We can certainly re-implement this as a generator. 

The `generate_text` seems like it matches the other patterns, but its also about as non-descriptive. I think it would help to give a name to this format, that would make the method name more straightforward. One idea might be "visual text", so the methods would be `generate/write_visual_text`? 

Brainstorm of names for this format: 

* nice text
* graph text
* visual text
* textrepr
* nicerepr
* network_text (I kinda like this for the similarity to networkx)

> I think this module could also be comfortable in the drawing subpackage. But it is fine here too.

This does live in sort of a weird place between a serialization format and a visual drawing. The purpose is closer aligned to the latter, whereas the details are closer to the former. I have a slight preference towards `readwrite` given that it would also be possible to implement the "read/parse" methods for this format as well (although I don't propose to do so here).

networkx/readwrite/text.py
dschult: That text looks great!  Including the stuff at the top that we don't know a direct reference, but that there might be one. I guess you'd need to change the "I" to some sort of different voice like "This notation is made up and didn't follow any specific reference. Some document might exist somewhere for a basis for it, but this is a natural extension of one way to draw a text forest.  To summarize:"

dschult:
I like giving this format a name and "network_text" is my favorite of the suggestions above.
:}

MridulS:
I really like this!

Just wanted to put this bit of code here too, this uses the library [rich](https://github.com/Textualize/rich) which is great for *pretty* terminal pictures. They have an inbuilt tree structure but it has it's limitations. Is it worth investigating this further?

``` python
import networkx as nx
from rich.tree import Tree
def rich_tree(G, root_node, depth_limit=4):
    bfs_tree = nx.bfs_tree(G, root_node, depth_limit=depth_limit).to_directed()
    tree_nodes = {i: Tree(f'{str(i)}') for i in bfs_tree}
    for n in bfs_tree:
        current_node = tree_nodes[n]
        for children in bfs_tree.neighbors(n):
            tree_nodes[n].add(tree_nodes[children])
    return tree_nodes[root_node]
```

``` python
>>> G = nx.Graph()
>>> G.add_edge('hello', 'world')
>>> G.add_edge('world', 'tree')
>>> G.add_edge('hello', 'rich-tree')
>>> root_node = 'hello'
>>> rich_tree(G, root_node)

hello
├── world
│   └── tree
└── rich-tree

```

This could also be used for `forest`
``` python
def rich_forest(G, depth_limit=4):
    for n in nx.connected_components(G):
        yield rich_tree(nx.subgraph(G, n), n.pop(), depth_limit)
```
```python
>>> G = nx.erdos_renyi_graph(10, 0.15)
>>> from rich import print
>>> for i in rich_forest(G):
>>>     print(i)
0

1
├── 2
├── 5
│   └── 7
└── 6
    └── 3

4

8
└── 9


```

Erotemic:
### Changes

* I noticed I had `\xa0` bytes instead of spaces. I replaced those
* I added a `generate_network_text` function that yields the lines as they are generated. 

### Comments

The motivation for this is quickly getting something on the screen that gives the user an intuitive gist of the graph structure, so I think connecting this to rich is a great idea. Maybe @willmcgugan has thoughts? For this to work, the `rich.tree.Tree` would need to be expanded to incorporate glyphs (guides) for both directed and undirected graphs, as well as the notion of printing the "backrefs" to the right of the node (or maybe this could be coerced by modifying the label).  It might also be useful if it indicated that it was cutting itself off due to screen limitations (which I think would be one of the main benefits of including rich here). But I do think it would be nice if when I was drawing a graph to the screen if "knew" that it was too big, so it would only show me the part of it that it could. I think that if we include rich support, then that should definitely go in the "drawing" submodule. It shouldn't be too hard to make a POC that extends `rich.tree` to `rich.graph` or something like that. 

I've kept the `graph_str` invocation in for now, but I've changed it such that it works by calling a new function `generate_network_text`. We can remove `graph_str` when this PR starts to wind down, but ATM it makes development easier for me. 

Does it make sense for there to be a "readwrite" text generator for this process at all? Or does this only make sense in the context of "drawing"? My thought is what happens if we abandon the generator and go for a rich only solution? There are a few downsides I can think of: 

1. the additional dependency needed before this function can be used (which isn't a big deal, rich is small and commonly used 2. what happens if I just want the full text, and I don't want it to cut off based on the assumption it will be drawn to somewhere? For someone like me who likes working with text files and passing them around, that second point is a big deal. Rich obfuscates the raw text, and i don't know of an easy way to just have it write it while ignoring any console limitations. 
3. For rich to write the graph, there will have to be a conversion to its internal representation. That will take time. On the other hand, the current method has low latency and will start dumping the graph immediately, which can have benefits. 

So the questions are: 

1. does a "generate_network_text" function make sense?
2. Should there be some function that uses `generate_network_text` to just "dump" the text to the screen? Maybe `write_network_text` can do this by default by defaulting `file` to stdout? Or should there be something else?
3. Should a rich variant go in the "drawing" module? If so what it is called? `rich_graph`? Should that be in this PR or should we separate that out? (Unfortunately, I don't think we can directly reuse any functions here, the internal code will be roughly the same, but it needs to be written in the context of a rich Renderable).

Erotemic:
@dschult Looking at the other modules in networkx.readwrite, it seems to be the case that there are functions named `generator_*` that actually are generators and do yield lines or tokens. E.g. `generate_adjlist`, `generate_gml`, `generate_graphml`, `generate_pajek`. Note at least in the case of `generate_adjlist`, this is contrary to its docs, which states that it does return a string. That should probably be noted as an issue. (Also it seems like a lot of the `write_*` functions that claim you can pass "path" in as a file path actually don't let you do that). 

In any case, it make a lot of sense to have this functionality implemented as a generator, even if that is a non-public function. I was already doing something very similar to that anyway with the "write" argument. I'm fine with having the backend implemented in either way. I've updated the code to include a corresponding `write_networkx_text` function which works similarly to how I was using `graph_str`. The only difference is that it does not return the text, it just dumps it to wherever you point it at (which is stdout by default).

I'm going to put this PR down for the day and work on my day job stuff. Let me know how you want to proceed with the `generate_*` issue. I'm in favor of them actually being generators, especially given that most of them already seem to be.

dschult:
I was wrong about the functions `generate_*`. You are right that they yield one line at a time of a text representation of the graph. Sorry about the distraction...

Let's bring the discussion back to the rich text option.

networkx/readwrite/text.py
Erotemic: I've added a modified version of the text to the function.

dschult (commented):
Just a couple of questions on the format description recently added.

networkx/readwrite/text.py
```diff
@@ -70,14 +79,98 @@ def forest_str(graph, with_labels=True, sources=None, write=None, ascii_only=Fal
...
+        directed graphs, these are drawn for any incoming edges to a node that
+        is not a parent edge. For undirected graphs, these are drawn for only
+        the non-parent edges that have already been represented (The edges that
+        have not been represented will be handled in the recursive case).
+
+        c. a child "L-style" connection on the lower right. Drawing of the
+        children are handled recursively.
+
+    4. The children of each node (wrt the directed DFS tree) are drawn
+    underneath and to the left of it. In the case that a child node has already
```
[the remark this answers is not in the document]
dschult: Thanks -- I meant to also ask about the half-sentence in part `a.` of this description. I must not have hit the button to add my comment.  I think that half-sentence should be added to, or maybe just removed.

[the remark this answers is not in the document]
Erotemic: It should have read "This corresponds to a traversal in the directed DFS tree."

Erotemic:
I was playing around with the rich implementation, and based on that I made a few changes to this one. Namely, I moved all of the glyphs into a set of classes to simplify the logic of picking which ones to use inside the function. I also moved the code to find the minimal set of source nodes to a helper function. I think this makes the core logic more readable.

Additionally, based on @MridulS's example I realized it would be a good idea to add a "max_depth" parameter here, so I made a small change to accommodate that and added the associated tests / docs.

Erotemic:
I think the only outstanding issue is choosing a version to deprecate `forest_str` in favor of this. Then this should be all set to merge.

The current text is:

```
    msg = (
        "\nforest_str is deprecated as of version <TBD:3.x?> and will be removed "
        "in version <TBD:3.y?>. Use generate_network_text or write_network_text "
        "instead.\n"
    )
    warnings.warn(msg, DeprecationWarning)
```

suggestions on how to replace TBD?

Erotemic:
I've been using this a lot in my day-to-day and it's been super helpful. I put a wrapper around it to make it easier to return the whole thing as a string, which is a behavior I like, but there is another feature that I can't easily without an additional modification here (or vendoring the code, which I'd prefer to avoid):

I'd like to change the `with_labels` argument such that if it is passed as a string, it uses that attribute instead of "labels". I don't think this behavior exists elsewhere, although it might be a good NXEP. 

I've pushed up the patch to add this feature to generate_network_text (with an associated test), but we can revert it if it is out of scope.

Erotemic:
@dschult @MridulS Is there anything blocking this from getting merged? I've been vendoring this code into all of my libraries that use networkx, I'd love to just have it available (and be able to tell other people to use it).

WRT rich: 

While I like the idea of having some sort of rich support, I think having the barebones version first is more important. Also, Just using `from rich import print` and then `nx.write_network_text(tree, print, end='')` already works pretty well: 

image

Erotemic:
@MridulS @dschult Would this be merged if I rebased it on main and fixed conflicts?

networkx/readwrite/text.py
```diff
@@ -70,122 +533,31 @@ def forest_str(graph, with_labels=True, sources=None, write=None, ascii_only=Fal
         L-- 1
             L-- 2
     """
-    import networkx as nx
+    msg = (
+        "\nforest_str is deprecated as of version <TBD:3.x?> and will be removed "
+        "in version <TBD:3.y?>. Use generate_network_text or write_network_text "
+        "instead.\n"
```
MridulS, line 539: Let's go with this one.
```suggestion
        "\nforest_str is deprecated as of version 3.1 and will be removed "
        "in version 3.3. Use generate_network_text or write_network_text "
        "instead.\n"
```

MridulS (approved):
Thanks for this @Erotemic!

Could you also add text.rst to https://github.com/networkx/networkx/tree/main/doc/reference/readwrite so sphinx picks it up? Looks like we never really had it as part of our docs.

Could you also add `forest_str` to deprecations in `doc/developer/deprecations.rst` ? Thanks!

Erotemic:
@MridulS Requested tasks are finished and the branch is rebased on main. Not sure what the style issue is, seems like a CI build error. Is this something that will be fixed on main?

MridulS:
Yeah, it's a linting issue, ignore that check.

dschult (approved):
Thanks for this!

PR #5606, opened 2022-05-05
Improve the `DiGraph.reverse` Docstring

Pysics:


dschult:
This doesn't follow our usual doc_string tense.  Moving it to the "is returned" form does not change the meaning and it doesn't match the rest. I'm going to close this, but you can still leave comments and we'll open it again if needed.

PR #5608, opened 2022-05-06
Distance measures example

Lukong123:
Added examples under distance measures algorithms.

### DIFF
--- a/networkx/algorithms/distance_measures.py
+++ b/networkx/algorithms/distance_measures.py
@@ -541,9 +541,9 @@ def barycenter(G, weight=None, attr=None, sp=None):
 
     Examples
     --------
-    >>> G = nx.Graph([(0, 1), (1, 2), (1, 5), (5, 4), (2, 4), (2, 3), (4, 3), (3, 6)])
-    >>> nx.barycenter(G)
-    [2]
+    >>> G = nx.Graph([(0, 1), (1, 2), (1, 3), (3, 4), (5, 4), (5, 3), (4, 1), (3, 6)])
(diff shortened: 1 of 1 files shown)


networkx/algorithms/distance_measures.py
```diff
@@ -304,6 +304,17 @@ def eccentricity(G, v=None, sp=None):
     -------
     ecc : dictionary
        A dictionary of eccentricity values keyed by node.
+
+    Examples
+    --------
+    >>> G = nx.Graph([(1, 2), (2, 3), (3, 4), (2,5), (3,6), (5, 6),(5,7), (6, 8), (6,9), (8,9)])
```
dschult, line 310: Try to make the spacing follow PEP8 standards (basically what the `black` program gives for python code).  Unfortunately we don't have a way to make black reformat the examples in the doc_strings. So you have to
put the spaces after each comma yourself, or copy/paste out of the example, run black on that and then copy
back into the file.

Also, these lines are getting pretty long -- and will be longer after the spaces.  Perhaps you can use a shorter set of edges for the example. I don't see the pattern in the edges, so maybe just leave off the last edges? But maybe I'm just not seeing it.

Finally, you don't need to create the graph G for the next example. It is already constructed here. So you can call the next example to get eccentricity for 2 nodes without constructing the graph again.  Similar comments for the other examples.

These are really good -- simple -- straightfoward -- how to use this function kind of examples. Thanks!

Lukong123: Okay sir, thanks for the review. I will do the necessary adjustment.

dschult (approved):
Thanks!

PR #5614, opened 2022-05-08
Cache edges, degree, adj properties of Graph classes

dschult:
Use cached_property for caching the view properties of graph classes (not `G.name` which can be changed and thus should be a regular non-cached property).

The changed properties include: G.edges, G.degree, G.adj  (and G.in_edges, G.out_edges, G.in_degree, G.out_degree, G.pred, G.succ).  Added tests that multiple calls return the same object (caching is successful). 

Changes to Graph, DiGraph, MultiGraph, MultiDiGraph and AntiGraph.  Also reportviews.py had to change one `__setstate__` function.

### DIFF
--- a/networkx/algorithms/approximation/kcomponents.py
+++ b/networkx/algorithms/approximation/kcomponents.py
@@ -1,6 +1,7 @@
 """ Fast approximation for k-component structure
 """
 import itertools
+from functools import cached_property
 from collections import defaultdict
 from collections.abc import Mapping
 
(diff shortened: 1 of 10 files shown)


jarrodmillman (approved):

stefanv:
@dschult Just double checking: I assumed when I made my original PR that the AdjacencyViews update themselves dynamically (they are called views, after all!).  Is that assumption correct?

dschult:
Yes, that is correct.  The Views point to the graph data structures directly to report their information.  That means you are safe unless you e.g. replace the dict `G._adj`. Then you have to clear the `cached_property`.  Similarly, if someone changes `G._node`, they should `delattr(self, "nodes")` to reset the `cached_property`.

stefanv:
Thanks @dschult!

PR #5629, opened 2022-05-12
Add Mehlhorn Steiner approximations

GuyAglionby:
Re discussion #5352 

This PR adds two new approximation algorithms for the Steiner tree problem: one from Wu et al. (1986), and another from Mehlhorn (1988). Both are asymptotically faster than the existing approximation from Kou et al. (1981).

One thing I haven't done is change the default algorithm from Kou's, although I think this would be desirable. I suggest setting Wu's as the default, as although the worst-case runtime is slower than Mehlhorn's, in practice I've found it to be faster. This is probably down to Mehlhorn requiring in all cases the shortest paths from all terminal nodes to all other nodes to be calculated; this is not the case with Wu's algorithm.

Any comments and feedback appreciated.

### DIFF
--- a/networkx/algorithms/approximation/steinertree.py
+++ b/networkx/algorithms/approximation/steinertree.py
@@ -1,9 +1,8 @@
-import heapq
+from heapq import heappush, heappop
 from itertools import chain
 
-from networkx.utils import pairwise, not_implemented_for, UnionFind
-from heapq import heappush, heappop
 import networkx as nx
+from networkx.utils import pairwise, not_implemented_for, UnionFind
 
 __all__ = ["metric_closure", "steiner_tree"]
 
@@ -57,7 +56,7 @@ def metric_closure(G, weight="weight"):
     return M
 
 
-def mehlhorn_steiner_tree(G, S, weight):
+def _mehlhorn_steiner_tree(G, S, weight):
     paths = nx.multi_source_dijkstra_path(G, S)
 
     d_1 = {}
@@ -66,16 +65,16 @@ def mehlhorn_steiner_tree(G, S, weight):
         s[v] = paths[v][0]
         d_1[(v, s[v])] = len(paths[v]) - 1
 
-    new_edges = {}
-    for e in G.edges(data=True):
-        u, v, data = e
-        weight_here = d_1[(u, s[u])] + data.get(weight, 1) + d_1[(v, s[v])]
-        key = tuple(sorted([s[u], s[v]]))
-        new_edges[key] = min(new_edges.get(key, float("inf")), weight_here)
-
+    # G1-G4 names match those from the Mehlhorn 1988 paper.
     G_1_prime = nx.Graph()
-    for (n1, n2), edge_weight in new_edges.items():
-        G_1_prime.add_edge(n1, n2, weight=edge_weight)
+    for u, v, data in G.edges(data=True):
+        su, sv = s[u], s[v]
+        weight_here = d_1[(u, su)] + data.get(weight, 1) + d_1[(v, sv)]
+        if not G_1_prime.has_edge(su, sv):
+            G_1_prime.add_edge(su, sv, weight=weight_here)
+        else:
+            new_weight = min(weight_here, G_1_prime[su][sv][weight])
+            G_1_prime.add_edge(su, sv, weight=new_weight)
 
     G_2 = nx.minimum_spanning_edges(G_1_prime, data=True)
 
@@ -89,7 +88,7 @@ def mehlhorn_steiner_tree(G, S, weight):
     return G_4
 
 
-def kou_steiner_tree(G, terminal_nodes, weight):
+def _kou_steiner_tree(G, terminal_nodes, weight):
     # H is the subgraph induced by terminal_nodes in the metric closure M of G.
     M = metric_closure(G, weight=weight)
     H = M.subgraph(terminal_nodes)
@@ -100,7 +99,7 @@ def kou_steiner_tree(G, terminal_nodes, weight):
     return edges
 
 
-def wu_steiner_tree(G, terminal_nodes, weight):
+def _wu_steiner_tree(G, terminal_nodes, weight):
     # Step 1
     source = {q: q for q in terminal_nodes}
     length = {q: 0 for q in terminal_nodes}
@@ -112,7 +111,7 @@ def wu_steiner_tree(G, terminal_nodes, weight):
     for q in terminal_nodes:
         for r in G.neighbors(q):
             if r in terminal_nodes:
-                both_in_terminal_nodes.add((min(r, q), max(r, q)))
+                both_in_terminal_nodes.add(frozenset([q, r]))
                 continue
             d_q_r = G[q][r].get(weight, 1)
             heappush(priority_queue, HeapEntry(d_q_r, (r, d_q_r, q, None, None)))
@@ -157,7 +156,7 @@ def wu_steiner_tree(G, terminal_nodes, weight):
                 n_subtrees -= 1
             else:
                 # case 3.2
-                heapq.heappush(
+                heappush(
                     priority_queue,
                     HeapEntry(d + length[t], (source[t], d + length[t], s, p1, t)),
                 )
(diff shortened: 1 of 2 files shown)


rossbar (commented):
At first glance this looks really nice! I haven't looked at anything in detail yet, but the first question to me is: what should the API be? I think the `method` kwarg is a nice solution that maintains backward compatibility and maximizes the discoverability of the new methods, so I agree with @boothby that this is a good way to go. Does anyone have any other ideas for what the API could/should be?

dschult:
I also like the `method` kwarg approach. 

@GuyAglionby do you have a sense for whether and why we should include multiple methods?
For example, why not just implement the Wu method?

The default should to be Kou for backward compatibility -- at least for now. 
And perhaps the default should be `None` with some code to give a warning that the default will be changing in version 3.? and if they want to avoid that change, they should specify the method explicitly. Along with the warning the method could get set to 'Kou" for now. and then switch it later.

Thanks very much for this!!

boothby:
This looks pretty good, @GuyAglionby, I'll go over it in detail in the next few days.  Thanks for your contribution!

GuyAglionby:
Hi @boothby, thanks for taking a brief look! Just a gentle reminder in case this had slipped through

networkx/algorithms/approximation/steinertree.py
```diff
@@ -46,22 +57,172 @@ def metric_closure(G, weight="weight"):
...
+    for (n1, n2), edge_weight in new_edges.items():
+        G_1_prime.add_edge(n1, n2, weight=edge_weight)
+
+    G_2 = nx.minimum_spanning_edges(G_1_prime, data=True)
+
+    G_3 = nx.Graph()
+    for u, v, d in G_2:
+        path = nx.shortest_path(G, u, v, weight)
+        for n1, n2 in pairwise(path):
+            G_3.add_edge(n1, n2)
```
dschult, line 86: Can G_1_prime be called G_1?  not a big deal, but it is a little strange.

Should G_2 be called G at all?  It is a bunch of edges... not a graph.
Either way, you don't seem to use "data=True" set in the "d" variable in the next loop, but then not used.
So, maybe something like:

```python
    G_3 = nx.Graph()
    for u, v in nx.minimum_spanning_edges(G_1_prime, data=False):
        G_3.add_edges_from(pairwise(nx.shortest_path(G, u, v, weight)))

    return nx.minimum_spanning_edges(G_3, data=False)
```

networkx/algorithms/approximation/steinertree.py
```diff
@@ -46,22 +57,172 @@ def metric_closure(G, weight="weight"):
...
+    s = {}
+    for v in G.nodes():
+        s[v] = paths[v][0]
+        d_1[(v, s[v])] = len(paths[v]) - 1
+
+    new_edges = {}
+    for e in G.edges(data=True):
+        u, v, data = e
+        weight_here = d_1[(u, s[u])] + data.get(weight, 1) + d_1[(v, s[v])]
+        key = tuple(sorted([s[u], s[v]]))
```
dschult, line 73: In general, the nodes are not sortable, so you should find a way to avoid the sort on `key`. You can use a graph data structure to help with this.  Set the edge on G_1_prime and then check it the next time. The edge weight is connected to both (u, v) and (v, u) so the order doesn't matter.  Something like:

```python
    G_1_prime = nx.Graph()
    for u, v, data  in G.edges(data=True):
        su, sv = s[u], s[v]
        weight_here = d_1[(u, su)] + data.get(weight, 1) + d_1[(v, sv)]
        if not G_1_prime.has_edge(su, sv):
            G_1_prime.add_edge(su, sv, weight=weight_here)
        else:
            new_weight = min(weight_here, G_1_prime[su][sv][weight])
            G_1_prime.add_edge(su, sv, weight=new_weight)
```

networkx/algorithms/approximation/steinertree.py
```diff
@@ -46,22 +57,172 @@ def metric_closure(G, weight="weight"):
...
+    length = {q: 0 for q in terminal_nodes}
+    pred = {}
+
+    # Step 2
+    both_in_terminal_nodes = set()
+    priority_queue = []
+    for q in terminal_nodes:
+        for r in G.neighbors(q):
+            if r in terminal_nodes:
+                both_in_terminal_nodes.add((min(r, q), max(r, q)))
```
dschult, line 115: Avoid sorting/comparing the nodes here by using a frozenset... The two directions will combine into one in the set both_in_terminal_nodes. And it will still work to iterate `for q, r in both...`
```both_in_terminal_nodes.add(frozenset(q, r))```

File: networkx/algorithms/approximation/steinertree.py [hunk 4]
```diff
@@ -46,22 +57,172 @@ def metric_closure(G, weight="weight"):
     return M
 
 
+def mehlhorn_steiner_tree(G, S, weight):
```
dschult, line 60: Since these helper functions aren't exposed publicly in `__all__`, their name should start with an underscore (by convention in NetworkX) to indicate that they are private functions to this module.

networkx/algorithms/approximation/steinertree.py
```diff
@@ -86,15 +256,37 @@ def steiner_tree(G, terminal_nodes, weight="weight"):
...
-    # Use the 'distance' attribute of each edge provided by M.
-    mst_edges = nx.minimum_spanning_edges(H, weight="distance", data=True)
-    # Create an iterator over each edge in each shortest path; repeats are okay
-    edges = chain.from_iterable(pairwise(d["path"]) for u, v, d in mst_edges)
+    if method is None:
+        import warnings
+
+        msg = (
+            "steiner_tree will change default method from 'kou' to 'wu' "
+            "in version 3.0.\nSet the `method` kwarg to remove this warning."
```
dschult, line 278: I think this deprecation will need to get pushed out to v3.2 because we've taken so long to review this PR.

dschult (commented):
Sorry this took so long to review. We'll get it in soon I hope.
This looks nice!  Thank you. It will be helpful.

I have a couple of comments below -- mostly about networkx conventions. e.g. nodes can't be sorted or even compared (e.g. you might have a number and a string).  If any aren't clear or don't make sense, ask.

Thanks!

networkx/algorithms/approximation/steinertree.py
dschult: Ahh... I see.
Maybe add a comment at line 75 like: 
`# G1-G4 names match those from the Mehlhorn 1988 paper.`

It's a little strange to name a list of edges G1_prime as if it is a graph.  Does the paper also have G1? what about another G with a prime? What does the prime signify?

GuyAglionby: Kou and Mehlhorn's algorithms are basically the same, but with a different first step of the total five steps. This G1_prime of Mehlhorn is the replacement of Kou's G1.

With this in mind, I think the methods for the two of them could actually be merged, but maybe that's a matter for another PR?

dschult: Sounds good to postpone any decisions about merging the two methods.

GuyAglionby:
Thanks for your comments, I think I've addressed the majority of them.

A couple of things came up when I was looking through this:
- The existing Kou et al. implementation was incomplete, and missed steps 4 and 5 of the algorithm. I have added a failing test case and fixed it.
- Similarly, my implementation of the Mehlhorn algorithm missed step 5.
- The implementation of the Wu et al. algorithm fails on the example given in that paper (which I also added as a test case). I suspect this is to do with how `pred` is built, but unfortunately don't have time in the near future to investigate further. I've removed it from this PR to avoid stalling it, and will try to resend another time

networkx/algorithms/approximation/steinertree.py
```diff
@@ -1,11 +1,20 @@
 from itertools import chain
 
-from networkx.utils import pairwise, not_implemented_for
 import networkx as nx
+from networkx.utils import pairwise, not_implemented_for
```
dschult, line 4:
```suggestion
from networkx.utils import not_implemented_for, pairwise
```
To appease  the import-sort styling changes.

networkx/algorithms/approximation/steinertree.py
```diff
@@ -1,11 +1,20 @@
...
 
+class HeapEntry:
+    def __init__(self, comparator, data):
+        self.comparator = comparator
+        self.data = data
+
+    def __lt__(self, other):
+        return self.comparator < other.comparator
+
+
```
dschult, line 17:
```suggestion
```
I think the HeapEntry was only used in the Wu method so should be removed now.

dschult (commented):
Sorry to see the Wu method leave, but I understand. The two comments here will hopefully get the tests passing and the merge conflicts removed.

GuyAglionby:
Thanks!

dschult (approved):
This all looks good to me. I was trying to think more about tests, but I'm afraid I don't know enough about the steiner trees to think deeply about whether there are other good unit-like tests.  So, I approve this PR.

PR #5633, opened 2022-05-13
added example on moral graph

Lukong123:


Lukong123:
> This LGTM! Thanks

Thank you, sir.

PR #5643, opened 2022-05-16
added examples to efficiency_measures.py

Lukong123:

### DIFF
--- a/networkx/algorithms/efficiency_measures.py
+++ b/networkx/algorithms/efficiency_measures.py
@@ -137,6 +137,12 @@ def local_efficiency(G):
     float
         The average local efficiency of the graph.
 
+    Examples
+    --------
+    >>> G = nx.Graph([(0, 1), (0, 2), (0, 3), (1, 2), (1, 3)])
+    >>> nx.local_efficiency(G)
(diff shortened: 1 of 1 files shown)


dschult (commented):
Interesting... the global efficiency and the local efficiency return different results even though the graph is small enough that they should be the same. The difference is round-off error. We should probably reduce potential testing headaches later by removing any round-off troubles.

Perhaps `>>>  round(nx.global_efficiency(G), 12)` and then 12 decimal digits of accuracy in the answer.  No need to worry about the first example with 0.5, because 1/2 has no round-off in binary representation.

Lukong123:
> Interesting... the global efficiency and the local efficiency return different results even though the graph is small enough that they should be the same. The difference is round-off error. We should probably reduce potential testing headaches later by removing any round-off troubles.
> 
> Perhaps `>>> round(nx.global_efficiency(G), 12)` and then 12 decimal digits of accuracy in the answer. No need to worry about the first example with 0.5, because 1/2 has no round-off in binary representation.

Okay sir, thanks.

MridulS (approved):
Thanks @Lukong123!

PR #5648, opened 2022-05-17
DOC: remove a comment

tacoo:
Removed a comment because the argument `target` cannot be specified


dschult (approved):
Thanks!  You are correct!   :}

rossbar (approved):
Thanks @tacoo !

PR #5650, opened 2022-05-17
Update betweenness.py

Choah:
If True the betweenness values are normalized by 2/((n-1)(n-2)) for graphs, and 1/((n-1)(n-2)) for directed graphs where n is the number of nodes in G.
Normalization for undirected graphs should be added.


dschult:
The betweenness_centrality normalization is already adjusted for directed edges.
```python
G = nx.path_graph(5)
nx.betweenness_centrality(G)
nx.betweenness_centrality(G.to_directed())   # normalized to be the same values.

# Note: DiGraph with single direction for each edge gives value cut in half
nx.betweenness_centrality(nx.DiGraph(G.edges())   # BC is half of value for G
```
I'm going to close this, but you can comment here and we will see it. We can reopen if needed.

PR #893, opened 2013-06-26
Hamming Distance

jac2130:
I added a file that calculates the simple Hamming Distance between two (possibly directed) graphs.

PR #5653, opened 2022-05-18
Hamming distance code bits from PR #893

MridulS:
Opening this PR with some changes so we can go through this once again.

This supersedes #893


MridulS:
Hasn't been touched in a while, closing this for now.

PR #1826, opened 2015-10-30
Make dfs_predecessors and dfs_successors return iterator

MridulS:

PR #5654, opened 2022-05-18
Update dfs_test with more comprehensive tests

MridulS:
From https://github.com/networkx/networkx/pull/1826, the current tests for DFS may miss an edge case (have multiple successors) if there are any changes in the code in the future.


dschult (approved):
Looks good

rossbar (approved):
Thanks @MridulS !

Issue #5655, opened 2022-05-19
DiGraph serialization (get/setstate) triggers an AttributeError for "_adj"


marscher:
### Current Behavior

Passing a `DiGraph` to deepcopy uses the serialization protocol via get and set_state methods of `InEdgeView`. In the implementation there is an assignment to self._adj like this:
``` python
 self._adjdict = G._pred if hasattr(G, "pred") else G._adj
```
So the passed `Graph` does not have the `_adj` attribute. By inspection I figured, that the state dictionary contains the `_adjdict` property directly.
If I replace the code to set `_adjdict` from the state, everything works as expcted.

### Expected Behavior
Copying a `DiGraph` via deepcopy should work. This used to be the case for networkx<2.8.1

### Steps to Reproduce

```python
G = nx.MultiDiGraph()
edges = G.in_edges()
deepcopy(G)
```

### Environment

Python version: 3.8
NetworkX version: 2.8.1

### Additional context
A github action workflow triggering the error:

https://github.com/BAMWelDX/weldx/runs/6502339152?check_suite_focus=true#step:9:110

``` python
 File ~/micromamba/envs/weldx/lib/python3.8/site-packages/networkx/classes/reportviews.py:1290, in InEdgeView.__setstate__(self, state)
   1288 def __setstate__(self, state):
   1289     self._graph = G = state["_graph"]
-> 1290     self._adjdict = G._pred if hasattr(G, "pred") else G._adj
   1291     self._nodes_nbrs = self._adjdict.items

AttributeError: 'DiGraph' object has no attribute '_adj'
```

MridulS:
Thanks for the report! This is indeed a bug:

A minimal example I was able to come up with (this is a bug for both `DiGraph` and `MultiDiGraph`):
``` python
In [2]: class Something:
    ...:     def __init__(self):
    ...:         self.g = nx.MultiDiGraph()
    ...:         self.e = self.g.in_edges
    ...:

In [3]: s = Something()

In [4]: deepcopy(s)
...
...
AttributeError: 'MultiDiGraph' object has no attribute '_adj'
```

This should be fixed by https://github.com/networkx/networkx/pull/5657

dschult:
Here's another minimal example:
```python
G = nx.MultiDiGraph()
edges = G.in_edges()
deepcopy(G)
```

marscher:
That was a really quick response and fix. Thank you so much! Have a nice weekend

PR #5657, opened 2022-05-20
Fix __setstate__ for OutEdgeView subclasses, read _adjdict from state

MridulS:
This should fix https://github.com/networkx/networkx/issues/5655

The state written to disk was recently changes in https://github.com/networkx/networkx/pull/5614 for `OutEdgeView` but the sub-classes `InEdgeView` and `InMultiEdgeView` still depended on the old read mechanism from the Graph object itself, this PR changes it to read directly from the `_adjdict` state.

### DIFF
--- a/networkx/classes/tests/test_reportviews.py
+++ b/networkx/classes/tests/test_reportviews.py
@@ -1,3 +1,6 @@
+import pickle
+from copy import deepcopy
+
 import pytest
 
 import networkx as nx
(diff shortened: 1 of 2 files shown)


MridulS:
The MacOS failures are related to github packages being down right now.

dschult (commented):
Excellent -- thanks for digging through and finding this. 

I think this means that we should add some tests that could catch errors with deepcopy and/or pickle. That could be in this PR or another if we want to get this out quickly.

Also, this means that 2.8.1 has a Defect and perhaps we should release a 2.8.2
What do you think @jarrodmillman ?

dschult:
Perhaps a test could be something like (untested):
```python
pytest.mark.parametrize("graph", [nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph])
def test_slots_in_views(graph):
    G = graph()
    (G.adj, G.degree, G.edges)
    if G.is_directed():
        (G.pred, G.succ, G.in_edges, G.out_edges, G.in_degree, G.out_degree)
    # raises error if slots not working for setstate and getstate
    deepcopy(G)

```

networkx/classes/tests/test_reportviews.py
```diff
@@ -1388,3 +1391,29 @@ def test_slicing_reportviews(reportview, err_msg_terms):
...
+        assert "pred" in cached_dict
+        assert "succ" in cached_dict
+        assert "in_edges" in cached_dict
+        assert "out_edges" in cached_dict
+        assert "in_degree" in cached_dict
+        assert "out_degree" in cached_dict
+
+    # Raises error if the cached properties and views do not work
+    pickle.loads(pickle.dumps(G, -1))
+    deepcopy(G)
```
MridulS, line 1419: @dschult added a pretty pedantic test just because of the rabbit hole of caches and slots I went into 😄 

This test successfully fails in the main branch

dschult (approved):
This LGTM!   Thanks Mridul!

Issue #5658, opened 2022-05-24
arg margins missing from valid_node_kwds in nx_pylab.py


harristeague:
margins is a new arg for draw_networkx_nodes(), but does not pass through from draw_networkx() since it is missing from the valid args list in nx_pylab.py:271

### Current Behavior

"invalid argument" exception

### Expected Behavior

it should accept the argument

### Steps to Reproduce

        g = nx.gnm_random_graph(10, 15)
        nx.draw_networkx(g, pos=nx.planar_layout(g), margins=(1,1))

### Environment

Python version: 3.8.12
NetworkX version: 2.8

MridulS:
fixed in https://github.com/networkx/networkx/pull/5660

PR #5660, opened 2022-05-24
Extract valid kwds from the function signature for draw_networkx_*

MridulS:
This makes sure we don't miss the cases where we add a new
argument to a `draw_networkx_*` and not update the valid kwargs
in `draw_networkx`.

An extremely explicit way of dealing with https://github.com/networkx/networkx/issues/5658
(I can go back to old way of checking valid kwds if this doesn't look good :) )

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -268,58 +268,25 @@ def draw_networkx(G, pos=None, arrows=None, with_labels=True, **kwds):
     draw_networkx_labels
     draw_networkx_edge_labels
     """
+    from inspect import signature
+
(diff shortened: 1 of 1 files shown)


dschult (approved):
This looks like a good approach.

I think we may also have issues with matplotlib providing keyword arguments that we don't support. The method used here could provide a template for how to handle updates to matplotlib signatures or support of multiple versions of matplotlib when they add/remove keyword options.  Nice.

Also, and I think unrelated to this Pr except its in the same file. It looks like the import of the networkx layout functions is not used in this module (lines 21-29)

jarrodmillman (approved):

PR #5666, opened 2022-05-26
fix bug for directed graphs

thaumkid:
if you read the original document for directed louvain (https://hal.archives-ouvertes.fr/hal-01231784/document), you'll see that in section 2, the equation for gain in the undirected case (delta Q) has a factor of 2 divided into both sides of the equation, whereas the directed case has that factor missing from both sides. In the previous code here, the factor was missing from only the first fraction, so I also "remove" it here to balance things out. Equivalently, this makes "resolution" twice as effective as before, thereby matching its strength in the undirected case

### DIFF
--- a/networkx/algorithms/community/louvain.py
+++ b/networkx/algorithms/community/louvain.py
@@ -250,7 +250,7 @@ def _one_level(G, m, partition, resolution=1, is_directed=False, seed=None):
                             out_degree * Stot_in[nbr_com]
                             + in_degree * Stot_out[nbr_com]
                         )
-                        / m
+                        / (2*m)
                     )
                 else:
                     gain = 2 * wt - resolution * (Stot[nbr_com] * degree) / m


dschult:
The tricky part of these formulas is the symbol `m` which represents the number of edges. That value `m` is in some sense twice as big for the directed case as for the (at least one way of making an) analogous undirected  case. The safest way to check the Delta Q formulas is to derive it from the Q formula. 

Are you getting a result from an example that you think is wrong? Or are you simply comparing the formulas in the code wit those in the paper.  If the formulas differ we should make sure that the definitions of the symbols are the same. In other words we should check the formulas. :}     The the code is correct but the formulas look different than a comment in the code should be added to help the reader understand those differences.

thaumkid:
Earlier I took the karate club graph and ran the undirected graph with the code, and then converted it to a directed graph by turning every original edge into two, and compared the outcomes with the louvain algorithm kind of like this:
```
import networkx as nx
G = nx.karate_club_graph()
DG = nx.DiGraph()
for (x,y) in G.edges():
   DG.add_edge(x,y)
   DG.add_edge(y,x)
c = nx.algorithms.community.louvain_communities(G,resolution=1,seed=1234)
dc = nx.algorithms.community.louvain_communities(DG,resolution=1,seed=1234)
```
The outcome made it clear that resolution wasn't working the same in the two cases, whereas
```
dc = nx.algorithms.community.louvain_communities(DG,resolution=0.5,seed=1234)
```
led to a result much more in line with what was expected. Which led to me to find the algorithm in the above-mentioned paper and find the difference as described.

z3y50n:
First of all, in the undirected case the reason we divide by `2*m` in the second fraction is because in the full formula the first term is divided by `2*m` and the second by `4*m^2` and since we only care whether the modularity gain is positive or negative we can simplify the formula to reduce unnecessary calculations (I have seen other implementations do that).

Your change makes sense in the case @dschult is describing because sometimes it isn't very clear whether we need `m` or `2*m` (here `m` is calculated using NetworkX's built in `graph.size`). I believe the same formula though is being used in the C++ version of the paper's authors you cited which can be found [here](https://github.com/nicolasdugue/DirectedLouvain). You can observe [here](https://github.com/nicolasdugue/DirectedLouvain/blob/23d511c1182a71b30335d2193797ecbc2f9dfb18/src/community_friend_static.cpp#L48-L54) and [here](https://github.com/nicolasdugue/DirectedLouvain/blob/23d511c1182a71b30335d2193797ecbc2f9dfb18/src/graph.cpp#L125-L128) that it seems we use the same `m` version (although I am not a main C++ coder so I could very well be missing something in the code)

Besides that, @rossbar I am not 100% if we want to have the same results in the directed and the undirected case (I admit your example has made me even a little in favour of this PR). In the cited paper they even go ahead and compare the two formulas for the directed and undirected case (which is a different case because they convert the directed graph to an undirected by ignoring the edge direction).

Lastly, the best way to verify the formula is working is to use `nx.algorithms.community.quality.modularity` function to measure the difference and the formula used for the modularity gain. They should be the same (I remember I had tested that for the undirected case and probably for the directed as well but it doesn't harm to redo the check one more time)

P.S. One really helpful explanation of the modularity formula is in [this](https://qr.ae/pvF8oG) quora answer if someone wants to dig a little deeper.

thaumkid:
awesome, thanks!

PR #5668, opened 2022-05-26
equitable_coloring: Get lazily first item instead of creating whole list

matusvalo:
This PR adjusts equitable_coloring to compute first value lazily using iterator instead of computing whole list of values. This yields not negligible performance increase:

```bash
# main branch
$ python -m timeit -s 'import networkx as nx; G = nx.circular_ladder_graph(5000)' -- 'd = nx.coloring.equitable_color(G, num_colors=4)'
1 loop, best of 5: 939 msec per loop

# PR
$ python -m timeit -s 'import networkx as nx; G = nx.circular_ladder_graph(5000)' -- 'd = nx.coloring.equitable_color(G, num_colors=4)'
2 loops, best of 5: 142 msec per loop
```

### DIFF
--- a/networkx/algorithms/coloring/equitable_coloring.py
+++ b/networkx/algorithms/coloring/equitable_coloring.py
@@ -109,7 +109,7 @@ def move_witnesses(src_color, dst_color, N, H, F, C, T_cal, L):
     while X != dst_color:
         Y = T_cal[X]
         # Move _any_ witness from X to Y = T_cal[X]
-        w = [x for x in C[X] if N[(x, Y)] == 0][0]
+        w = next(x for x in C[X] if N[(x, Y)] == 0)
         change_color(w, X, Y, N=N, H=H, F=F, C=C, L=L)
         X = Y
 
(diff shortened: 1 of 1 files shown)


dschult (commented):
This looks right -- it replaces lists which only have their first element accessed with `next` of an iterator.  

Can you tell me why in one case is contained in a try/except for StopIteration (which makes sense to me) and the others aren't inside a try/except?

networkx/algorithms/coloring/equitable_coloring.py
```diff
@@ -109,7 +109,7 @@ def move_witnesses(src_color, dst_color, N, H, F, C, T_cal, L):
     while X != dst_color:
         Y = T_cal[X]
         # Move _any_ witness from X to Y = T_cal[X]
-        w = [x for x in C[X] if N[(x, Y)] == 0][0]
+        w = next(x for x in C[X] if N[(x, Y)] == 0)
```
matusvalo, line 112: This line is not protected by try/except block catching `StopIteration` because previous implementation is not catching possible `IndexError` exception. It means that algorithm guarantees (assuming the implementation in main is correct) that it is guaranteed that the list/iterator contains at least one item which can be accessed

networkx/algorithms/coloring/equitable_coloring.py
```diff
@@ -218,15 +218,16 @@ def procedure_P(V_minus, V_plus, N, H, F, C, L, excluded_colors=None):
...
+                        try:
+                            # Finding the solo neighbor of w in X_prime
+                            y = next(
+                                node
+                                for node in L[w]
+                                if F[node] == X_prime and N[(node, W_1)] == 1
+                            )
+                        except StopIteration:
+                            pass
+                        else:
```
matusvalo, line 230: In following block, we have a check `len(y_candidates) > 0` which means that the algorithm allows to list `y_candidates` to be empty. Hence we need to catch `StopIteration` and manage it accordingly. Using the syntax of `try-except-else` syntax [1]:

main branch:
```python
y_candidates = [...]

if len(y_candidates) > 0:
    y = y_candidates[0]
    # list is not empty use first value from list
```

pr:
```python
try:
    y = next(...)
except StopIteration:
    pass
    # iterator is empty skip processing `y`
else:
    # iterator is not empty (exception was not raised) process `y`
```

[1] https://docs.python.org/3/tutorial/errors.html#handling-exceptions

matusvalo:
Thank you for review @dschult. To answer your question, see my comments above.

dschult (approved):
Thanks for the comments -- it makes perfect sense to me now.

I don't see any reason for additional tests or changes to the doc_string.
So this looks good to me! :}

MridulS (approved):
Thanks @matusvalo!

PR #5673, opened 2022-05-28
More tests for clustering (upstreaming from graphblas-algorithms)

eriknw:
These improve coverage of `graphblas-algorithms`.  CC @MridulS 

Also, some floating point comparisons were relaxed to use `np.testing.assert_allclose`, which are needed for https://github.com/python-graphblas/graphblas-algorithms/pull/11.  I take a narrow approach with this: I only relax the comparisons needed to allow `graphblas-algorithms` to pass tests.

(btw, I found that a few tests fail when `pytest-randomly` is installed, which runs tests randomly)

### DIFF
--- a/networkx/algorithms/tests/test_cluster.py
+++ b/networkx/algorithms/tests/test_cluster.py
@@ -39,6 +39,9 @@ class TestTriangles:
         G.remove_edge(1, 2)
         assert list(nx.triangles(G).values()) == [5, 3, 3, 5, 5]
         assert nx.triangles(G, 1) == 3
+        G.add_edge(3, 3)  # ignore self-edges
+        assert list(nx.triangles(G).values()) == [5, 3, 3, 5, 5]
+        assert nx.triangles(G, 3) == 5
 
 
 class TestDirectedClustering:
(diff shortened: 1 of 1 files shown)


networkx/algorithms/tests/test_cluster.py
```diff
@@ -170,13 +176,15 @@ def test_triangle_and_edge(self):
         G = nx.cycle_graph(3, create_using=nx.DiGraph())
         G.add_edge(0, 4, weight=2)
         assert nx.clustering(G)[0] == 1.0 / 6.0
-        assert nx.clustering(G, weight="weight")[0] == 1.0 / 12.0
+        np.testing.assert_allclose(nx.clustering(G, weight="weight")[0], 1.0 / 12.0)
```
jarrodmillman, line 179: It may be worth adding a comment above these tests to indicate why we aren't doing exact comparisons.  Maybe something like:
```
# Relaxed comparisons to allow graphblas-algorithms to pass tests
```

eriknw: Done.

eriknw:
I dropped unnecessary `.0` *in the whole file* via `s/\.0\>//g`.  This was easier (and cleaner imho)  than changing only my changes (and subsequently makes review more difficult).

I'm happy to address any other nits you want to share @rossbar.  Also, I'm glad I didn't overstep with the changes found by `pytest-randomly` (there's still one problematic test that I know of).

MridulS (approved):
Thanks for the new tests! :)

PR #5677, opened 2022-06-01
Correcting a typo in the references

RATCOinc:
The correct spelling of the author is "Narsingh Deo" not "Marsingh Deo".


dschult (approved):
Thanks for this!
The automatic tests are failing for reasons that have nothing to do with this PR.
This looks good to me!

rossbar (approved):
The CI failure seems to be cropping up elsewhere and is definitely unrelated, so in this goes. Thanks for the fix @RATCOinc !

PR #5683, opened 2022-06-02
Update subgraph views tests to pass with out of order execution

MridulS:
pytest-randomly fails here (https://github.com/networkx/networkx/pull/4553)


dschult (approved):
LGTM!
It is nice to have the randomly test this sort of thing. :}

rossbar (approved):
Thanks @MridulS ! These changes highlight one of the gotchas of using class-based organization of tests: accidental changing/persistence of state between test cases!

PR #5685, opened 2022-06-02
Another catch by pytest-randomly

MridulS:
I guess we will see a bunch of other things when https://github.com/networkx/networkx/pull/4553 is merged and the cron job runs every day.


dschult (approved):
Nice catch of a subtle morphing of self.G.
:}

rossbar (approved):
Thanks @MridulS !

Issue #5692, opened 2022-06-05
networkx/algorithms/tree/tests/test_mst.py::test_random_spanning_tree_{additive,multiplicative}_small not skipped when scipy is not installed


mgorny:
### Current Behavior
The two following tests fail if scipy is not installed:

```
FAILED networkx/algorithms/tree/tests/test_mst.py::test_random_spanning_tree_multiplicative_small - ModuleNotFoundError: No module n...
FAILED networkx/algorithms/tree/tests/test_mst.py::test_random_spanning_tree_additive_small - ModuleNotFoundError: No module named '...
```

<details><summary>Full traceback</summary>

```pytb
============================================================== FAILURES ===============================================================
___________________________________________ test_random_spanning_tree_multiplicative_small ____________________________________________

    def test_random_spanning_tree_multiplicative_small():
        """
        Using a fixed seed, sample one tree for repeatability.
        """
        from math import exp
    
        pytest.importorskip("numpy")
    
        gamma = {
            (0, 1): -0.6383,
            (0, 2): -0.6827,
            (0, 5): 0,
            (1, 2): -1.0781,
            (1, 4): 0,
            (2, 3): 0,
            (5, 3): -0.2820,
            (5, 4): -0.3327,
            (4, 3): -0.9927,
        }
    
        # The undirected support of gamma
        G = nx.Graph()
        for u, v in gamma:
            G.add_edge(u, v, lambda_key=exp(gamma[(u, v)]))
    
        solution_edges = [(2, 3), (3, 4), (0, 5), (5, 4), (4, 1)]
        solution = nx.Graph()
        solution.add_edges_from(solution_edges)
    
>       sampled_tree = nx.random_spanning_tree(G, "lambda_key", seed=42)

G          = <networkx.classes.graph.Graph object at 0x7fda89729810>
exp        = <built-in function exp>
gamma      = {(0, 1): -0.6383,
 (0, 2): -0.6827,
 (0, 5): 0,
 (1, 2): -1.0781,
 (1, 4): 0,
 (2, 3): 0,
 (4, 3): -0.9927,
 (5, 3): -0.282,
 (5, 4): -0.3327}
solution   = <networkx.classes.graph.Graph object at 0x7fda8972a450>
solution_edges = [(2, 3), (3, 4), (0, 5), (5, 4), (4, 1)]
u          = 4
v          = 3

networkx/algorithms/tree/tests/test_mst.py:401: 
[...]

PR #5693, opened 2022-06-05
importorskip scipy instead of numpy for total spanning tree

MridulS:
fixes https://github.com/networkx/networkx/issues/5692

To catch this we would have to run a CI pipeline with numpy but without scipy. I don't think we should be doing that as it takes up too much CI time, so just adding a fix and no test.


mgorny:
Thanks!

dschult (approved):

mjschwenne (approved):
Thanks for catching this!

mgorny:
> Thanks for catching this!

Not exactly my choice ;-).

jarrodmillman (approved):

PR #5701, opened 2022-06-07
Use inline math

szhorvat:
It seems to me that these pieces in the docs should be inline math, not inline code.


dschult (approved):

lucasmccabe:
@dschult @jarrodmillman Not super important, but I've been noticing the behavior above a few times recently - the checks appear to be passing, but the **ci/circleci: image artifact** action gets stuck in limbo. Any idea why this occurs?

rossbar (approved):
Nice catch, thanks @szhorvat !

PR #5702, opened 2022-06-08
Reconfigure Adigraph interface from NetworkX to Latex

dschult:
This builds on PR #5639 and proposes a new interface focusing on 
- to_latex(G, caption, latex_label, as_document=False)
- write_latex(G, "just_a_figure.tex", caption, latex_label, as_document=False)

I have put it in a separate PR because I want feedback on the interface. I like the functions here, but handling the options for node_color/edge_width/etc feels kludgy.
With two PRs we can more easily compare the two interfaces.

Suggestions?

The Adigraph interface is split into two classes: 
- Adigraph handles a single graph and it's options.
    * ADIG.to_latex_raw()
    * ADIG.to_latex_figure()
    * ADIG.to_latex_document()
- AdigraphCollection handles multiple graphs to be put into subfigures within a figure. 
    * ADIGcoll.to_latex_figure()
    * ADIGcoll.to_latex_document()


rossbar (commented):
I've only taken a high-level look so far, focusing mostly on the interface options. I definitely like the functional interface with `to_latex` and `write_latex` and prefer it to the OO interface of the `Adigraph` class. In fact, I'd advocate to keep the `Adigraph` and `AdigraphCollection` classes private, at least initially. IMO it makes more sense for the user to be able to set the node/edge attributes on the networkx graph objects then use the functions to convert to the LaTeX output rather than introduce new classes with more interfaces for users to learn.

dschult:
I've cleaned up the docs and added more tests. I think this is ready to merge
It's not a full implementation of TikZ graph drawing, but it gives a pretty good framework for most basic stuff and options for some more advanced stuff.

I'm torn between `to_latex` and `generate_latex`.  Also whether this belongs in the drawing subpackage or the readwrite subpackage.

dschult:
I've updated the LaTeX interface to allow either Adigraph output, or TikZ output.
The default is currently set to TikZ because that doesn't restrict you to only color, width and label attributes. 

TikZ allows any valid tikz options, with inputs for 
 - options for the tikzpicture environment itself: e.g. [scale=2]
 - options for the \draw command that draws nodes: e.g. [shape=rectangle]
 - options for the scope environment within which the edges are drawn: e.g. [blue,>->,line width=3]
 - options for each node  (can override the default values above)
 - options for each edge  (can override the default values above)
 - label text for nodes
 - label text for edges
 - options for the style of each edge text label.

I'd like feedback on whether to include both of these, or only the TikZ version. The only advantage I can think of for the adigraph interface is that you can store the values in different attributes on each node or edge. You only get 2 style features: color, width. But you can store them as "blue" and 3 rather than having to store them as a combined single attribute for options: "blue,line width=3".

Also, if anyone has suggestions for TikZ features to add/test that would be cool too. 
I'm putting off an implementation that uses the TikZ Graph Library which includes layout routines and is a competitor of GraphViz I guess.

dschult:
I'm going to close this with a redirection to #6238
We can reopen if adigraph is desired/

PR #5710, opened 2022-06-08
Catch ':' explicitly while working with pydot

MridulS:
pydot has it's own mechanism to deal with quoting string and we shouldn't try to change that.

In https://github.com/networkx/networkx/pull/5667 I added a catch all which didn't work with the internal system of pydot as it adds quotes for some strings.

In this PR I am explicitly looking for `:` and if someone provides a node/edge name/attribute without the double quotes it will fail on the them gracefully. But this change should keep on working for people who were depending on pydot to add quotes to their strings.

### DIFF
--- a/networkx/drawing/nx_pydot.py
+++ b/networkx/drawing/nx_pydot.py
@@ -169,6 +169,13 @@ def from_pydot(P):
     return N
 
 
+def _check_colon_quotes(s):
+    # A quick helper function to check if a string has a colon in it
+    # and if it is quoted properly with double quotes.
+    # refer https://github.com/pydot/pydot/issues/258
+    return ":" in s and (s[0] != '"' or s[-1] != '"')
+
+
 def to_pydot(N):
     """Returns a pydot graph from a NetworkX graph N.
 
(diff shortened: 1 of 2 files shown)


rossbar (approved):
At some point I think it's worth considering how much effort NX should put into working around `pydot` parsing issues... there seem to be quite a few. A quick sampling of pydot parsing issues that have been reported to NX:
 - Problems parsing strings with colons (this PR)
 - Problems related to [parsing strings that start with numbers](https://github.com/networkx/networkx/pull/5667#issuecomment-1149870056)
 - Problems parsing long-ish strings with special characters (#3857)

That being said, +1 for this PR since it's an improvement on a previous attempt to work around pydot issues.

dschult (approved):

networkx/drawing/nx_pydot.py
```diff
@@ -213,25 +220,68 @@ def to_pydot(N):
...
-            edge = pydot.Edge(str(u), str(v), key=str(key), **str_edgedata)
+            str_edgedata = {str(k): str(v) for k, v in edgedata.items() if k != "key"}
+            u, v = str(u), str(v)
+            raise_error = (
+                _check_colon_quotes(u)
+                or _check_colon_quotes(v)
+                or (
+                    any(
+                        (_check_colon_quotes(k) or _check_colon_quotes(v))
+                        for k, v in edgedata.items()
```
peterjc, line 251: Calling ``_check_colon_quotes(k)`` assumes ``k`` is a string, but unlike the other variables passed to ``_check_colon_quotes`` like ``u`` and ``v``, neither ``k`` nor ``v`` have been explicitly cast to a string.

Probably you wanted ``for k, v in edgedata.items()`` --> ``for k, v in str_edgedata.items()``

Cross reference https://github.com/peterjc/thapbi-pict/pull/458#issuecomment-1155136449

PR #5719, opened 2022-06-09
Some more changes to make pytest-randomly happy

MridulS:
After this is merged we should be able to merge https://github.com/networkx/networkx/pull/4553, it passed locally on a bunch of runs after these changes.


rossbar (approved):
LGTM, thanks @MridulS . Longer term it'd be great to make these tests less state-ful :)

dschult (approved):

PR #5730, opened 2022-06-13
docs: Add characteristic polynomial example to polynomials docs

lucasmccabe:
Implements the suggestion from #5726 to add an example of the chromatic polynomial to the `polynomials` docstring.


rossbar (approved):
I took the liberty of fixing up the sympy intersphinx link - it was more complicated than I expected. Other than that this LGTM and I think it's a nice example how to use sympy and networkx together and avoid expanding NX's API footprint with functions that really just wrap sympy functionality. Thanks @lucasmccabe !

lucasmccabe:
@rossbar  This makes sense to me.  Thanks for your help!

jarrodmillman (approved):

PR #5759, opened 2022-06-15
Remove deprecated function nx.info

MridulS:
bye bye nx.info, we should really try to get something which is more useful and informative.


jarrodmillman (approved):

rossbar (approved):
Thanks @MridulS !

mboehme:
Thanks Martin! LGTM.

PR #5763, opened 2022-06-15
Remove pyyaml dependency

jarrodmillman:


rossbar (approved):
All the yaml related changes LGTM. Mypy was failing for some reason (could reproduce locally) but it looked unrelated - maybe bad interaction with mypy and lazy importing? Either way, rather than let it be a blocker or try to dig too deeply I just pushed up a trivial change that makes mypy stop complaining.

PR #5764, opened 2022-06-15
Update return type of attrmatrix

MridulS:
It does feel like I'm missing some update to documentation here.


rossbar (approved):
The failing doctests should pass now - they do locally at least. I approve contingent on this passing CI when it catches up!

jarrodmillman (approved):

PR #5791, opened 2022-06-17
add parenthesis

jnecus:
Added parenthesis  see #5750 .


dschult:
Don't worry about the MacOS tests failing. That started recently and is probably due to changes in the testing environment. We're working on those.

But...  These parentheses should not be needed. Without the parens it is a class object and with the parens it is a class instance. But both should work with this function.  The examples work on our testing platform. And on my local machine.

dschult:
I'm going to close this PR. Let's continue the discussion in #5750  Or leave followup comments here and we can reopen if needed.

PR #5793, opened 2022-06-17
Compute `is_strongly_connected` lazily

matusvalo:
The PR computes strongly connected components lazily. This approach avoids the need of computing all strongly connected components and yields performance increase of graphs having more strongly connected components. The following example is based on directed path graph:

image

```python
import timeit
vertices = [1, 10, 20, 50, 100, 200, 500, 1000, 2000, 5000, 10000, 50000]
times = []
for i in vertices:
    times.append(timeit.timeit(stmt='nx.is_strongly_connected(G)', setup=f'import networkx as nx; G = nx.path_graph({i}, create_using=nx.DiGraph)', number=100))
````

### DIFF
--- a/networkx/algorithms/components/strongly_connected.py
+++ b/networkx/algorithms/components/strongly_connected.py
@@ -350,7 +350,7 @@ def is_strongly_connected(G):
             """Connectivity is undefined for the null graph."""
         )
 
-    return len(list(strongly_connected_components(G))[0]) == len(G)
+    return len(next(strongly_connected_components(G))) == len(G)
 
 
 @not_implemented_for("undirected")

### CODE REVIEW

**Comment by matusvalo** [comment]
The failed check is not connected to this PR. It is caused by failed installation:
```
      pygraphviz/graphviz_wrap.c:2711:10: fatal error: 'graphviz/cgraph.h' file not found
      #include "graphviz/cgraph.h"
               ^~~~~~~~~~~~~~~~~~~
      1 error generated.
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure
× Encountered error while trying to install package.
╰─> pygraphviz
```

dschult:
This is great!!   Thanks very much.  And such a simple idea too.  Instead of looking at the first element of the list, just iterate to  the first element of the list. :}

The trouble with MacOS tests is due to a change in the environment of the testing platform and not related to this PR at all.

dschult (approved):

rossbar (approved):
Ditto!

PR #5795, opened 2022-06-17
Compute `is_weakly_connected` lazily

matusvalo:
This PR computes `is_weakly_connected` lazily. This can yield significant performance increase. E.g. for case of discrete graph:

image

```python
import timeit
vertices = [1, 10, 20, 50, 100, 200, 500, 1000, 2000, 5000, 10000, 50000]
times = []
for i in vertices:
    times_patched.append(timeit.timeit(stmt='nx.is_weakly_connected(G)', setup=f'import networkx as nx; G = nx.DiGraph(); G.add_nodes_from(range({i}))', number=100))
```

### DIFF
--- a/networkx/algorithms/components/weakly_connected.py
+++ b/networkx/algorithms/components/weakly_connected.py
@@ -157,7 +157,7 @@ def is_weakly_connected(G):
             """Connectivity is undefined for the null graph."""
         )
 
-    return len(list(weakly_connected_components(G))[0]) == len(G)
+    return len(next(weakly_connected_components(G))) == len(G)
 
 
 def _plain_bfs(G, source):


matusvalo:
The checks are failing due error not connected to the change in this PR:

```
      #include "graphviz/cgraph.h"
               ^~~~~~~~~~~~~~~~~~~
      1 error generated.
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure
× Encountered error while trying to install package.
╰─> pygraphviz
```

dschult (approved):

dschult:
Thanks!

rossbar (approved):
Excellent, thanks @matusvalo !

PR #5800, opened 2022-06-20
Fix typo in bipartite closeness_centrality and thought-o in tests

### LINKED ISSUE #5620: Bipartite Closeness centrality algorithm does not report nodes with zero value.
### Current Behavior
The function saves closeness values of '0.0' always at closeness[len(nodes)] which is closeness[n]. This results a returned dictionary with less elements than number of nodes in the graph.

### Expected Behavior
The values of '0.0' should instead be saved at closeness[node] within the for-loops iterating over all nodes.

### Steps to Reproduce
/

### Environment
Python version: 3.9.12
NetworkX version: 2.8

### Additional context
In both cases within the source code below
`else:
            closeness[n] = 0.
`
needs to be changed to:
`else:
            closeness[node] = 0.
`
See sourcecode:
https://networkx.org/documentation/stable/_modules/networkx/algorithms/bipartite/centrality.html#closeness_centrality

dschult:
Fixes #5620 
with typo in bipartite closeness_centrality and thought-o in tests

### DIFF
--- a/networkx/algorithms/bipartite/centrality.py
+++ b/networkx/algorithms/bipartite/centrality.py
@@ -252,7 +252,7 @@ def closeness_centrality(G, nodes, normalized=True):
                 s = (len(sp) - 1) / (len(G) - 1)
                 closeness[node] *= s
         else:
-            closeness[n] = 0.0
+            closeness[node] = 0.0
     for node in bottom:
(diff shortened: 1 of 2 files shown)


MridulS (approved):

rossbar (approved):
Thanks @dschult !

PR #5809, opened 2022-06-23
Pydot layout fix

shakedbr:
Fixed an issue caused by #5710:
In the `pydot_layout` function, all the node names are converted to strings in the returned dictionary keys.
Therefore, when using numerical node names (e.g. `0, 1`) the return pos dictionary contains `{'0':..., '1':...}` instead of `{0:..., 1:...}`.


MridulS (approved):
Thanks for the catch @shakedbr!

rossbar (approved):
Thanks @shakedbr !

PR #5810, opened 2022-06-23
Add random_spanning_tree to documentation

mjschwenne:
When I created #5656 I forgot to add `random_spanning_tree` to `tree.rst` so it does not appear in the documentation.


MridulS:
Thanks for the catch @mjschwenne !

PR #5837, opened 2022-06-30
DOC: Rst syntax update.

Carreau:
Just came across a weird thing when trying to render with papyri,
and it looks like what was attempted here was a list enumeration, so
converting to a enumerated list instead of a code/verbatim block


rossbar (approved):
Thanks @Carreau , indeed I think that was the original intention. We can bikeshed over whether the list should start with 0 or 1 elsewhere :)

PR #5844, opened 2022-07-05
Speed up unionfind a bit by not adding root node in the path

MridulS:
As suggested by @dschult in https://github.com/networkx/networkx/pull/4810, but FWIW it doesn't have that much of difference in speed. I'm +0 on this.

### DIFF
--- a/networkx/utils/union_find.py
+++ b/networkx/utils/union_find.py
@@ -53,11 +53,12 @@ class UnionFind:
             return object
 
         # find path of objects leading to the root
-        path = [object]
+        path = []
         root = self.parents[object]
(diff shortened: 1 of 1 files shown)


rossbar (approved):
I'm +0.5 for readability reasons rather than performance: I find the `root != object` more idiomatic than `root != path[-1]`, even if it results in another re-assignment in the loop. This is very subjective though, so YMMV. I'm happy either way!

dschult (approved):
Thanks for getting the issue into this PR. I think it is a marginal improvement in readability. And it closes this issue. :)

PR #5847, opened 2022-07-06
Use py_random_state to control randomness of random_triad

MridulS:
Make sure results are reproducible.


rossbar (approved):
Nice catch, thanks @MridulS

dschult:
I guess that also means this should be added to the  `networkx/tests/test_all_random_functions.py` with a line like:
```python
t(nx.random_triad, G, seed=seed)
```

dschult (approved):

PR #5848, opened 2022-07-06
update tests in base class and simple rename in convert.py

dschult:
This splits out the minor refactoring test improvements from #5836 to ease review.

- One test function was not a test method so not being tested in the multidigraph test subclass -- moved it into a class.
- Tests of subclassing MultiDiGraph were copied and adapted to MultiGraph
- Changed the names of the exception variables in convert.py where nested try/except clauses are named the same. This should ease debugging when it is not clear which exception is being raised.


rossbar (approved):
LGTM, though I always have a hard time traversing the (multiple) inheritance tree for OO tests :). AFAICT this inherits the correct test methods from `TestMultiGraph` and runs all the desired test methods as expected.

Also +1 for the shadowed exception names. Thanks @dschult

jarrodmillman (approved):

PR #5850, opened 2022-07-06
Move factory attributes to the class instead of instance.

dschult:
We originally had copied the class factory attributes to the instance for an attempt as speed-up.
It now looks like it is a small slowdown.  But perhaps more importantly given the minor impact of these speedups is that it should be easier to read the `__init__` function without getting hung up in all the factories. And trying to figure out why we would be copying them from the class to the instance anyway.   BTW my measure of the speed up is about 15-20% of the time to create a Graph object... but that is so small to start with that it probably isn't important.  Still, if this way is slightly faster and easier to read, maybe we should do it.

This has been separated from #5836 to ease review.

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -299,13 +299,6 @@ class DiGraph(Graph):
         {'day': 'Friday'}
 
         """
-        self.graph_attr_dict_factory = self.graph_attr_dict_factory
-        self.node_dict_factory = self.node_dict_factory
-        self.node_attr_dict_factory = self.node_attr_dict_factory
-        self.adjlist_outer_dict_factory = self.adjlist_outer_dict_factory
-        self.adjlist_inner_dict_factory = self.adjlist_inner_dict_factory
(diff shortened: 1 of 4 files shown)


rossbar (approved):
This LGTM. As long as the factories aren't ever intended to be used/modified outside of instance creation (why would they be?) I think it's safe to leave these as class attributes rather than copy the factories all down to the specific instance.

I also think this is a nice improvement from the readability standpoint alone. The performance boost is a nice bonus (though I agree it's probably not too impactful in practice). On my system:

#### On main (8522eea39)
```python
%timeit G = nx.Graph()
1.59 µs ± 24 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each
```

#### This PR (f48ab4af45)
```python
%timeit G = nx.Graph()
1.18 µs ± 5.68 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

jarrodmillman (approved):

PR #5864, opened 2022-07-14
Fix multiple inheritance with Graph

### LINKED ISSUE #5865: Mixins incompatible with Graph
Using a Mixin with `Graph` is currently perilous:
```python
class Specialized(Graph, Mixin): ...

s = Specialized()  # Mixin.__init__ is not called!!
```

NeilGirdhar:
* Make Graph.__init__ call super in case it's used in multiple
  inheritance
* Add Graph._create_attributes to factor out common behavior in
  Graph.__init__ and DiGraph.__init__

Fixes #5865

This is a preliminary pull request in an effort to simplify the giant #4014 to make the review simpler.

### DIFF
--- a/networkx/classes/graph.py
+++ b/networkx/classes/graph.py
@@ -315,6 +315,17 @@ class Graph:
         {'day': 'Friday'}
 
         """
+        super().__init__()
+        # Create object attributes.
+        self._create_attributes()
+        # Attempt to load graph with data.
+        if incoming_graph_data is not None:
+            convert.to_networkx_graph(incoming_graph_data, create_using=self)
+        # Load graph attributes (must be after convert).
+        self.graph.update(attr)
+
+    def _create_attributes(self):
+        # Copy class attributes to members.
         self.graph_attr_dict_factory = self.graph_attr_dict_factory
         self.node_dict_factory = self.node_dict_factory
         self.node_attr_dict_factory = self.node_attr_dict_factory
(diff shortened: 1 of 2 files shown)


NeilGirdhar:
This PR repairs problems when someone does something like:
```python
class Specialized(Graph, Mixin): ...

s = Specialized()  # Mixin.__init__ is not called!!
```

dschult:
Thanks for separating these out from the types PR.  It is **much** easier to review these separately, where the idea for the change is clear.

networkx/classes/graph.py
```diff
@@ -315,24 +315,32 @@ def __init__(self, incoming_graph_data=None, **attr):
         {'day': 'Friday'}
 
         """
+        super().__init__()
```
MridulS, line 318: What happens if there are args or kwargs that need to be passed to the `__init__` constructor of `Mixin`?

This should probably be `super().__init__(**attr)`. And also maybe would need to change the signature of `Graph.__init__` 

PS: I'm personally not a fan of doing this. Adding a `super().__init__()` to a `object` subclass doesn't sound too fun  🤷

NeilGirdhar: > What happens if there are args or kwargs that need to be passed to the __init__ constructor of Mixin?

Yes, I thought of that, but there's nothing you can do in that case.  But at least you'll get a failure rather than silently not initializing the base class, which is much worse.

> This should probably be ...

I think it's okay to support base classes that require no arguments (which are the most prevalent class of initializers).  And to fail on the rest with a clear message at initialization time.  That's what this PR would do.

> PS: I'm personally not a fan of doing this. Adding a super().__init__() to a object subclass doesn't sound too fun

All classes are "object subclasses".  And all non-final classes should theoretically call super in their `__init__` (along with `__exit__`, `__enter__`, etc.)  These methods exhibit a pattern of extending functionality.  There are a variety of linting errors that are starting to be added to catch this potential mistake.

In any case, another benefit of this PR is that it reduces the overlapping code between Graph and Digraph, which reduces the amount of code that needs to be tested, and ensures that changes to Graph's `__init__` affects subclasses.

dschult:
I don't think we should move the attribute assignment parts of `__init__` out of the `__init__` function. It's only a few lines of very readable code so repeating it actually makes it much easier to read -- and maintain than a separate function. I know that repeating code is usually a Bad Thing, but in this case (especially now that we are removing the factory functions from the instances) it is actually more readable and easier to maintain.

PR #5875, opened 2022-07-15
voterank algorithm returns vote value alongside with node

maksim-melaxtech:

### DIFF
--- a/networkx/algorithms/centrality/tests/test_voterank.py
+++ b/networkx/algorithms/centrality/tests/test_voterank.py
@@ -26,21 +26,21 @@ class TestVoteRankCentrality:
                 (4, 6),
             ]
         )
-        assert [0, 7, 6] == nx.voterank(G)
+        assert [0, 7, 6] == [node for node, vote in nx.voterank(G)]
 
     # Graph unit test
     def test_voterank_centrality_2(self):
         G = nx.florentine_families_graph()
         d = nx.voterank(G, 4)
         exact = ["Medici", "Strozzi", "Guadagni", "Castellani"]
-        assert exact == d
+        assert exact == [node for node, vote in d]
 
     # DiGraph unit test
     def test_voterank_centrality_3(self):
         G = nx.gnc_graph(10, seed=7)
         d = nx.voterank(G, 4)
         exact = [3, 6, 8]
-        assert exact == d
+        assert exact == [node for node, vote in d]
 
     # MultiGraph unit test
     def test_voterank_centrality_4(self):
@@ -49,7 +49,7 @@ class TestVoteRankCentrality:
             [(0, 1), (0, 1), (1, 2), (2, 5), (2, 5), (5, 6), (5, 6), (2, 4), (4, 3)]
         )
         exact = [2, 1, 5, 4]
-        assert exact == nx.voterank(G)
+        assert exact == [node for node, vote in nx.voterank(G)]
 
     # MultiDiGraph unit test
     def test_voterank_centrality_5(self):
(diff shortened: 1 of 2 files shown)


rossbar (commented):
This seems like a reasonable feature request, but it will require some additional scaffolding since it results in a breaking API change.

One idea might be a keyword argument like `return_voterank` that can be set to true to return the votes as well as the nodes. Another alternative would be to add a new function (not sure what it'd be named) that returned a dict keyed by node with the votes as the values. I think a dict makes sense and fits better with the API used for other centrality functions.

networkx/algorithms/centrality/voterank_alg.py
```diff
@@ -90,3 +90,92 @@ def voterank(G, number_of_nodes=None):
...
+
+
+def vote_rank(G, number_of_nodes=None):
+    """Select a list of influential nodes in a graph using VoteRank algorithm
+
+    VoteRank [1]_ computes a ranking of the nodes in a graph G based on a
+    voting scheme. With VoteRank, all nodes vote for each of its in-neighbours
+    and the node with the highest votes is elected iteratively. The voting
+    ability of out-neighbors of elected nodes is decreased in subsequent turns.
+
```
dschult, line 102: Explain in a second paragraph the difference between the two functions. You could add something to the original function too telling people to use this version if they want the vote values.

networkx/algorithms/centrality/voterank_alg.py
```diff
@@ -90,3 +90,92 @@ def voterank(G, number_of_nodes=None):
...
+    G : graph
+        A NetworkX graph.
+
+    number_of_nodes : integer, optional
+        Number of ranked nodes to extract (default all nodes).
+
+    Returns
+    -------
+    voterank : dict
+        Unordered dictionary with nodes as a key and vote as a value.
```
dschult, line 114: I think this is an ordered dict, right?  For Python 3.6 or later dicts are ordered in the order the keys are added. If it is ordered, the order should probably be tested.

networkx/algorithms/centrality/tests/test_voterank.py
```diff
@@ -59,3 +102,12 @@ def test_voterank_centrality_5(self):
         )
         exact = [2, 0, 5, 4]
         assert exact == nx.voterank(G)
+
+    # MultiDiGraph unit test
+    def test_vote_rank_centrality_5(self):
```
dschult, line 107: What is this testing? Put that into the test name rather than a number. Then when a test says that `test_vote_rank_centrality_for_MultiDiGraph` failed, we easily know what the test is about. (and you can remove the comment).  Tests with numbers in the name are not as helpful.

networkx/algorithms/centrality/tests/test_voterank.py
```diff
@@ -28,20 +28,54 @@ def test_voterank_centrality_1(self):
         )
         assert [0, 7, 6] == nx.voterank(G)
 
+    def test_vote_rank_centrality_1(self):
```
dschult, line 31: It looks like these tests have the same name as the version they mimic (they don't --- but it looks like they do). That's one reason to make the names of the function different.

networkx/algorithms/centrality/voterank_alg.py
```diff
@@ -90,3 +90,92 @@ def voterank(G, number_of_nodes=None):
             vote_rank[nbr][1] -= 1 / avgDegree
             vote_rank[nbr][1] = max(vote_rank[nbr][1], 0)
     return influential_nodes
+
+
+def vote_rank(G, number_of_nodes=None):
```
dschult, line 95: We need a different name!  
How about `vote_rank_with_vote`?  Kinda long, but not that long.
Thoughts?

dschult (commented):
Awesome -- thanks for this.  

Some comments:
Rather than repeat all that code, can you keep the code in the more general (dict returning) version, and then convert the original to simply calling the new version and just give a list of the keys of the output of the new version?  If this doesn't make sense let me know. Something like return `list(nx.vote_rank_with_votes)`

And then we need a better name. :)

MridulS:
@maksim-melaxtech thanks for your contribution! This hasn't been touched in a while, let me know if you still have the bandwidth to address the comments. No worries if not! Thanks again!

PR #5879, opened 2022-07-17
Adds ```nx.bfs_layers``` method

still-n0thing:
Adds ```bfs_layers``` method to [breadth_first_search.py](https://github.com/networkx/networkx/blob/main/networkx/algorithms/traversal/breadth_first_search.py).
```python3
def bfs_layers(G, sources):
    """Returns an iterator of all the layers in breadth-first search traversal.

    Parameters
    ----------
    G : NetworkX graph
        A graph over which to find the layers using breadth-first search.

    sources : node in `G` or list of nodes in `G`
        Specify starting nodes for single source or multiple sources breadth-first search

    Yields
    -------
    layer: list of nodes
        Yields list of nodes at the same distance from sources

    Examples
    --------
    >>> G = nx.path_graph(5)
    >>> dict(enumerate(nx.bfs_layers(G, [0, 4])))
    {0: [0, 4], 1: [1, 3], 2: [2]}
    >>> H = nx.Graph()
    >>> H.add_edges_from([(0, 1), (0, 2), (1, 3), (1, 4), (2, 5), (2, 6)])
    >>> dict(enumerate(nx.bfs_layers(H, [1])))
    {0: [1], 1: [0, 3, 4], 2: [2], 3: [5, 6]}
    >>> dict(enumerate(nx.bfs_layers(H, [1, 6])))
    {0: [1, 6], 1: [0, 3, 4, 2], 2: [5]}
    """
    ...
```
ToDo:
- [x] adds `bfs_layers` method to the docs 
- [x] adds tests for the `bfs_layers` method 
- [x] changes the implementation of `descendents_at_distance`

### DIFF
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -377,9 +377,9 @@ def bfs_layers(G, sources):
     Parameters
     ----------
     G : NetworkX graph
-        A graph over which to find the layers using breadth first search.
+        A graph over which to find the layers using breadth-first search.
 
-    sources : list of nodes in `G`
+    sources : node in `G` or list of nodes in `G`
         Specify starting nodes for single source or multiple sources breadth-first search
 
     Yields
@@ -399,22 +399,25 @@ def bfs_layers(G, sources):
     >>> dict(enumerate(nx.bfs_layers(H, [1, 6])))
     {0: [1, 6], 1: [0, 3, 4, 2], 2: [5]}
     """
-    for source in list(sources):
-        if not G.has_node(source):
-            raise nx.NetworkXError(f"The node {source} is not in the graph.")
+    if sources in G:
+        sources = [sources]
 
     current_layer = list(sources)
     visited = set(sources)
 
+    for source in current_layer:
+        if source not in G:
+            raise nx.NetworkXError(f"The node {source} is not in the graph.")
+
     # this is basically BFS, except that the current layer only stores the nodes at
     while current_layer:
+        yield current_layer
         next_layer = list()
         for node in current_layer:
             for child in G[node]:
                 if child not in visited:
                     visited.add(child)
                     next_layer.append(child)
-        yield current_layer
         current_layer = next_layer
 
 
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -410,7 +410,7 @@ def bfs_layers(G, sources):
             raise nx.NetworkXError(f"The node {source} is not in the graph.")
 
     # this is basically BFS, except that the current layer only stores the nodes at
-    # the same distance from sources at each iteration
+    # same distance from sources at each iteration
     while current_layer:
         yield current_layer
         next_layer = list()
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -9,6 +9,7 @@ __all__ = [
     "bfs_predecessors",
     "bfs_successors",
     "descendants_at_distance",
+    "bfs_layers",
 ]
 
 
@@ -368,6 +369,62 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
         children = [c]
         parent = p
     yield (parent, children)
+
+
+def bfs_layers(G, sources, layer_limit=None):
+    """Returns all the layes of bfs traversal.
+    Parameters
+    ----------
+    G : NetworkX graph
+        A graph
+
+    source : node in `G` or list of nodes in `G`
+
+    layer_limit : int, optional(default=len(G))
+        Specify the maximum layer depth
+
+    Returns
+    -------
+    dict()
+        key is the distance of layers from the sources and value is the set of nodes in that layer
+
+    Examples
+    --------
+    >>> G = nx.path_graph(5)
+    >>> nx.bfs_layers(G, [0, 4])
+    {0: {0, 4}, 1: {1, 3}, 2: {2}}
+    >>> H = nx.Graph()
+    >>> H.add_edges_from([(0, 1), (0, 2), (1, 3), (1, 4), (2, 5), (2, 6)])
+    >>> nx.bfs_layers(H, 1)
+    {0: {1}, 1: {0, 3, 4}, 2: {2}, 3: {5, 6}}
+    >>> nx.bfs_layers(H, [1, 6])
+    {0: {1, 6}, 1: {0, 2, 3, 4}, 2: {5}}
+    """
+    if sources in G:
+        sources = [sources]
+
+    if layer_limit is None:
+        layer_limit = len(G)
+
+    current_distance = 0
+    current_layer = {node for node in sources}
+    visited = {node for node in sources}
+    all_layers = dict()
+
+    # this is basically BFS, except that the current layer only stores the nodes at
+    # current_distance from source at each iteration
+    while current_distance < layer_limit and len(current_layer) != 0:
+        next_layer = set()
+        for node in current_layer:
+            for child in G[node]:
+                if child not in visited:
+                    visited.add(child)
+                    next_layer.add(child)
+        all_layers[current_distance] = current_layer
+        current_layer = next_layer
+        current_distance += 1
+
+    return all_layers
 
 
 def descendants_at_distance(G, source, distance):
--- a/networkx/algorithms/traversal/breadth_first_search.py
+++ b/networkx/algorithms/traversal/breadth_first_search.py
@@ -371,60 +371,51 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
     yield (parent, children)
 
 
-def bfs_layers(G, sources, layer_limit=None):
-    """Returns all the layes of bfs traversal.
+def bfs_layers(G, sources):
+    """Returns an iterator of all the layers in breadth-first search traversal.
+
     Parameters
     ----------
     G : NetworkX graph
-        A graph
+        A graph over which to find the layers using breadth first search.
 
-    source : node in `G` or list of nodes in `G`
+    sources : list of nodes in `G`
+        Specify starting nodes for single source or multiple sources breadth-first search
 
-    layer_limit : int, optional(default=len(G))
-        Specify the maximum layer depth
-
-    Returns
+    Yields
     -------
-    dict()
-        key is the distance of layers from the sources and value is the set of nodes in that layer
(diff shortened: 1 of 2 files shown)


networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -370,6 +371,62 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
     yield (parent, children)
 
 
+def bfs_layers(G, sources, layer_limit=None):
+    """Returns all the layes of bfs traversal.
+    Parameters
```
dschult, line 376: The documentation building system (Sphinx) needs a blank line after the first "short line" description. The first line is used to describe this function wherever it appears in the docs as a link to this doc_string. So it's nice to have it short and clear (which it is here).  We just need a blank line after it.

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -370,6 +371,62 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
...
+    ----------
+    G : NetworkX graph
+        A graph
+
+    source : node in `G` or list of nodes in `G`
+
```
dschult, line 382: Add a sentence describing what this `source` will be used for. Also, it should be named `sources` instead of `source`.

Similarly, for the incoming graph G, you should add a more informative sentence describing G. "A graph" doesn't say how it will be used in the function. Maybe "The graph over which to find the layers using breadth first search."

Thirdly:  Let's make this function a generator.  Instead of returning all the layers at once in a dict, let's have it `yield` the layers one at a time. Then we don't need the depth limit -- the user can stop when they want to. And they can choose whether they want to store it as a list, dict, tuple, or process each layer when it is computed and throw that away before getting the next layer.  Changes needed: remove the `all_layers` dict and replace the setting of that dict with `yield current_layer`. Remove the line at the end `return all_layers`.  Remove the computation of `current_distance` and simplify the loop spec to `while current_layer:`. (Python evaluates a list as "False" when it is empty.) Change the doc_string to replace the "Returns" section with a "Yields" section. The items yielded should be lists of nodes with a description that they are the nodes in each layer of the BFS. In the doc_string examples, you will need to change the calling code to add `list(...)` around the call (or `dict(enumerate(...))` which will give the output you current provide. If you want the slightly simpler output list for these examples, you'll need to change the expected output accordingly.

Style choice:  `visited = {node for node in sources}` can be `visited = set(sources)`... that's up to you. Similar for `current_layer = list(sources)` if you want that.

Finally, the comments in the issue indicate that we should store the layers as lists instead of sets.

dschult (commented):
That was a quick turnaround. :)  I have some comments below.

Still needed:  
1) We should rewrite `descendents_at_distance` to use this layers generator. It should stop iterating after "distance" iterations and return the final layer.  I also think the check at the beginning of that function should read: `if source not in G:` But that is a minor thing. Let me know if you want help with the loop.
2) Add tests...  They should be added to `tests/test_breadth_first_search.py`.  To start, copy the tests for `descendant_at_distance` and change them so the output is tested to match what this `bfs_layers` should match.  You can test locally with:
`pytest --doctest-modules breadth_first_search,py tests/test_breadth_first_search.py` which tests the doc_strings and the test files. You can just run the test file tests with 
`pytest tests/test_breadth_first_search.py`

still-n0thing:
> ```if source not in G:```
I don't quite understand this part.

still-n0thing:
@dschult I made the suggested changes in case of any issue please inform me.

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -370,6 +371,53 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
...
+    {0: [0, 4], 1: [1, 3], 2: [2]}
+    >>> H = nx.Graph()
+    >>> H.add_edges_from([(0, 1), (0, 2), (1, 3), (1, 4), (2, 5), (2, 6)])
+    >>> dict(enumerate(nx.bfs_layers(H, [1])))
+    {0: [1], 1: [0, 3, 4], 2: [2], 3: [5, 6]}
+    >>> dict(enumerate(nx.bfs_layers(H, [1, 6])))
+    {0: [1, 6], 1: [0, 3, 4, 2], 2: [5]}
+    """
+    for source in list(sources):
+        if not G.has_node(source):
```
dschult, line 403: A good way to check if a node is in a graph is "node in G".  (this doesn't work with edges... only nodes).
A list (or dict or other unhashable) object cannot be a node. And a tuple can be but could also be a container of nodes. So a good type-agnostic way of checking whether the input is a list or a single node is to check if the thing is in G.  So, we check `if sources in G:` and if it is, we consider it to be a single node. We want an iterable of nodes so we wrap it in a list.  Note that this works pretty well for strings too. If input "abcd" is a node in G, it is wrapped in a list -- which seems reasonable. If it is not a node, then it is treated as a container of nodes `['a', 'b', 'c', 'd']`.  That is also reasonable.

After we set up the sources object, we iterate over the sources and check if each is in G, raising an exception if they are.  Finally, we use `sources` to fill up the current_layer, and visited. 

Idea:  This doesn't allow us to handle an iterator.  It would be exhausted after the first loop and not be able to fill the current_layer.  So, let's make this work for either a node, iterable or iterator.  We'll fill the current_layer before we iterate over it to check that the nodes are in G.   What do you think:
```
    # check for single node
    if sources in G:
        sources = [sources]

    current_layer = list(sources)
    visited = set(current_layer)

    for source in current_layer:
        if source not in G:
            raise nx.NetworkXError(...)

```

networkx/algorithms/traversal/tests/test_bfs.py
```diff
@@ -51,6 +51,18 @@ def test_bfs_tree_isolates(self):
         assert sorted(T.nodes()) == [1]
         assert sorted(T.edges()) == []
 
+    def test_bfs_layers(self):
+        assert dict(enumerate(nx.bfs_layers(self.G, [0]))) == {
+            0: [0],
+            1: [1],
+            2: [2, 3],
+            3: [4],
+        }
```
dschult, line 60: You should set this up to test both `sources=[0]` and `sources=0`.
That makes it worthwhile to create a variable to hold the expected values and the formatting is nicer:
```python
    expected = {0: [0], 1: [1], .....}
    assert dict(enumerate(nx.bfs_layers(self.G, [0]))) == expected
    assert dict(enumerate(nx.bfs_layers(self.G, 0))) == expected
```

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -370,6 +371,53 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
...
+
+    # this is basically BFS, except that the current layer only stores the nodes at
+    while current_layer:
+        next_layer = list()
+        for node in current_layer:
+            for child in G[node]:
+                if child not in visited:
+                    visited.add(child)
+                    next_layer.append(child)
+        yield current_layer
```
dschult, line 417: We should `yield` before the loop.   ...   Why? because we know the value of `current_layer` so we can yield. And we might not be asked to visit the next layer (the user might stop the loop by not calling "next" or stopping their loop over the layers).  In that case, we will have saved the time spent generating `next_layer`.

So, I would move the "yield" statement to immediately after the `while` statement.

networkx/algorithms/traversal/tests/test_bfs.py
```diff
@@ -51,6 +51,18 @@ def test_bfs_tree_isolates(self):
...
+
+    def test_bfs_layers_missing_source(self):
+        with pytest.raises(nx.NetworkXError):
+            next(nx.bfs_layers(self.G, ["abc"]))
```
dschult, line 64: Duplicate this test and the `pytest.raises` clause so you can test input `["abc"]` and then `"abc"`.

dschult (commented):
Looking good.  

Another task (if you've had enough we can stop and merge and do this next one in a later PR)  would be to rewrite the `descendents_at_distance` to make it use this function. The advanatge of that is that we have the BFS loop code in one place and can optimize it without having to change it in multiple places.

The idea would be to pull out most of the code for `de..._at_dist...` and replace it with a loop over `distance` layers (or is it "distance + 1"? -- check the current docks for whether we return the starting layer as distance 0 or 1). 
We only need the last layer in this function and we need to stop the loop at the right iteration.  Something like:  (untested)
```python
bfs_generator = nx.bfs_layers(G, source)
for i in range(distance + 1):
    layer = next(bfs_generator)
return layer
```
There are other ways to do this too -- some using `enumerate`.  This should not change the result of the function... so the test and the doc_strings don't need to be changed (unless you see something that could be improved).

still-n0thing:
@dschult I have finished up the suggested changes. 

I would like to work on ```descendents_at_distance``` next. Should I create an issue or you will be creating the issue?

dschult:
Thinking about it more, I think it would be easiest just to put it into this PR.
They are so related, it doesn't really require a separate issue. :}

still-n0thing:
should ```descendents_at_distance``` return list or set? (I might need to change something in test if it returns list)

still-n0thing:
@dschult I have rewritten ```descendents_at_distance``` with ```bfs_layers``` which returns a set of nodes.

dschult:
I think it should return whatever it returned before for backward compatibility.

still-n0thing:
> I think it should return whatever it returned before for backward compatibility.

It use to return set before and rewritten function also returns set.

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -370,6 +371,56 @@ def bfs_successors(G, source, depth_limit=None, sort_neighbors=None):
...
+        sources = [sources]
+
+    current_layer = list(sources)
+    visited = set(sources)
+
+    for source in current_layer:
+        if source not in G:
+            raise nx.NetworkXError(f"The node {source} is not in the graph.")
+
+    # this is basically BFS, except that the current layer only stores the nodes at
```
dschult, line 412: Did this comment get cutoff?

dschult (approved):
This looks good to me!
Thanks for all your work on this. 
I have one optional comment, but I approve this PR in any case.

networkx/algorithms/traversal/breadth_first_search.py
still-n0thing: Yeah, let me fix it

still-n0thing: I have fixed it

MridulS (approved):
This looks good @still-n0thing! Thanks.

PR #5882, opened 2022-07-18
Adds `nx.draw_networkx_multigraph_edges` method

dtekinoglu:
This PR adds `nx.draw_networkx_multigraph_edges` method to the codebase.


networkx/drawing/nx_pylab.py
```diff
@@ -890,6 +891,207 @@ def _connectionstyle(posA, posB, *args, **kwargs):
...
+            arrowstyle = "-|>"
+        else:
+            arrowstyle = "-"
+
+    multiEdgeDict = {}
+    for edge in edgelist:
+        if edge not in multiEdgeDict:
+            multiEdgeDict[edge] = 1
+        else:
+            multiEdgeDict[edge] += 1
```
dschult, line 1038: I suggest using the method [G.number_of_edges(u, v)](https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.number_of_edges.html) to count. Then you can build the dict more easily and also not have to check the opposite direction of the edge.
```python
multiedge_count = {(u, v): G.number_of_edges(u, v) for u, v in G.edges()}
```
Hmmm...  The way I wrote this example computes the number of edges for **each** multiedge -- which is a little silly since they will all be the same value -- and overwrite the previous (identical) entry in the dict. Maybe there is a better way -- but if its not clear don't worry about the extra computes.

Also, just so you know, there is a subtlety that `G.edges()` always yields 2-tuples, while `G.edges` yields 2-tuples for Graph/DiGraph and 3-tuples for MultiGraph/MultiDiGraph.  I don't think it affects things here because you use `G.edges()` for the default above. But it could be that people would include a 3-tuple edge in `edgelist`.  That might be where your use of `e[0]` and `e[1]` will work better than `for u, v in` because it doesn't care if there is a third entry in `e`.

networkx/drawing/nx_pylab.py
```diff
@@ -890,6 +891,207 @@ def _connectionstyle(posA, posB, *args, **kwargs):
...
+
+    for edge in multiEdgeDict:
+        if multiEdgeDict[edge] == 1 and list(G.edges()).count((edge[1], edge[0])) == 0:
+            nx.draw_networkx_edges(
```
dschult, line 1042: To reduce the number of graphical edge containers (and calls to draw_*_edges) maybe it's best to collect all the singular edges in one list/dict and all the multiedge 2-tuples in another list/dict.  Then draw the singular edges with one connectionstyle function call and the multiedges with the arc you compute. I'm not sure that connectionstyle can vary across the edges in a FancyArrowPatch. If not, then you'll have to call the draw function for each multtiedge as you do now. But there would be only 1 LineCollection for the other non-multi edges.

dschult (commented):
Great that this is working!!

I'm hoping you can eventually get it working from within the original edges function.
Meanwhile, you should add the new function to the list of functions in `doc/reference/drawing.rst` so it shows up in the docs. 

I like how it looks!
I've got some comments, questions below. But, in short -- I am impressed. Nice work!

The current code treats selfloops separately from the other edges (line ~870), but still inside the same function. Would it work to do a similar thing with multiedges?

rossbar:
Thinking about this a bit more, I think there are even more dimensions to consider (or: how to make the multiedge drawing at least as complicated as the single-edges!)

For example, we might want to consider the ability to draw multiedges with `LineCollection` for  for performance reasons. It's possible to imagine drawing lines between nodes that are parallel to one another to denote separate multiedges without requiring any arcing (i.e. not necessarily using fancy arrow patches). However; how to implement, test, and switch this behavior seems like another headache: we already have problems switching between LineCollection/FancyArrowPatch edges for the *single* edge case.

Just some more food for thought :) This would definitely be a nice feature though!

dtekinoglu:
I made the new method private so that `nx.draw_networkx_edges` directly calls it if the input graph is a multigraph. However, I have the following problem: I designed `_draw_networkx_multigraph_edges` method such that it calls `nx.draw_networkx_edges` with different `connectionstyle` values for different edges. Now, since the input graph is a multigraph, they keep calling each other 😄 

A solution I could think of was to copy code from `nx.draw_networkx_edges` to `_draw_networkx_multigraph_edges` so that it does not call the previous one back. However, in this case there will be huge amount of code duplication.

What do you suggest? @dschult @rossbar @MridulS

dschult:
Arg -- yes, I see the problem.  Maybe we make two private helper functions: the current edges function and the current multiedges functoin. Then we make a new public version of the edges function which checks for a multigraph and calls either the private multiedge or private edge function.  Finally, change the multiedge private function to call the private edge function when it is looping through the different connection styles. 

This is just thoughts put down in email. Are there better ways to arrange this code?

networkx/drawing/nx_pylab.py
dtekinoglu: If I use `G.number_of_edges` method - instead of counting edges in the edgeList-, would people still be able to use edgelist argument as they expect? For example, a person who prefers to draw exactly one edge between a pair of nodes will not be able to do it if it is a multiedge.

dtekinoglu:
Almost all the test cases are failing for `draw_networkx_edges()` now, and I could not identify the reason. For example, `test_single_edge_color_undirected(edge_color, expected, edgelist)` is giving the following result for the current implementation:
image

My implementation gives the following:
image

There is a gap between the arrows in my output, this is the only difference I could realize. I'm not sure what caused this because the only thing I changed was to make current method private and call it from a public method.

Another failing test is `test_single_edge_color_directed(None-r-red)`. Current implementation gives the following:
image

My implementation's output looks almost the same:

image

Any idea why these tests are failing? @dschult @rossbar @MridulS

networkx/drawing/nx_pylab.py
dschult: That's a good question. Do we want to allow people to choose which edges to draw by listing them in `edgelist`? I think we do.  Do we want to allow them to specify *which* multiedge to draw? Also, probably we do.  So, they should be able to use a 3-tuple in `edgelist` to specify which edges to draw. That means we need to count how many edges in `edgelist` are between the two nodes -- not how many edges in G.edges are between the two nodes...

As for how to count efficiently. I remembered that Python has a collections builtin library with a Counter object. The Counter acts like a dict, only when you construct it or use `count.update(iterable)` it counts how many times each element shows up.  So it is a dict keyed by element to the value that is the number of times that element has been added to the Counter.  I think `from collections import Counter` allows something like: 
```python
edge_counts = Counter(edgelist)
for edge, cnt in edge_counts.items():
    if cnt == 1:
        ...
    else:
        bend = 0.5 / cnt
        ...
```

networkx/drawing/nx_pylab.py
```diff
@@ -638,6 +639,86 @@ def draw_networkx_edges(
...
     """
+
+    if arrowstyle == None:
+        if G.is_directed():
+            arrowstyle = "-|>"
+        else:
+            arrowstyle = "-"
+
+    if type(G) == nx.classes.multigraph.MultiGraph == False:
+        _draw_networkx_edges(
```
dschult, line 650:
```suggestion
        return _draw_networkx_edges(
```
and similarly for the multigraph drawing routine.

dtekinoglu: `draw_networkx_multigraph_edges` calls `_draw_networkx_edges `method multiple times. I am not sure how to return drawn object in that case.

networkx/drawing/nx_pylab.py
dtekinoglu: I used Counter as you suggested. However, I think I should still check for opposite direction of the edge because if the graph is directed, there might be an opposite edge even though the edge count is 1.  Consider the following example:

```
G = nx.MultiDiGraph([(0,1),(1,0),(2,0)])
Counter(list(G.edges()))
```

Output: 

`Counter({(0, 1): 1, (1, 0): 1, (2, 0): 1})`

If I don't check for the opposite side, (0, 1) and (1, 0) will be drawn as straight lines and they will overlap. Is there another way to prevent this without checking as I do?

dschult: Ahh...  hmmm....  I'm not sure how best to handle this case....
Do we bend all the multiedges the same direction (e.g. always to the right?).  

OK...   with multiedges, here's an idea (and maybe there is a better one -- not sure this even works):
- count the number in each direction. Bend them always to the right. Then when the other direction is drawn it will be bent the "other direction" and not overlap. By "other direction" I mean that it still is bent e.g. to the right, but since it is set in the other direction, it will bend away from the edges going the first direction. Does this work?

- In the directed case where we are already drawing FancyArrows, maybe we should bent every edge slightly so that even without multiedges, we still get the opposite direction edges to not overlap. I guess we would want an option to turn that off to get straight line with arrows depicting direction including "both directions".  Hmm.  I guess this can already be done by the user specifying an arc ConnectionStyle for every edge.  maybe??  Not sure...

rossbar:
> This is just thoughts put down in email. Are there better ways to arrange this code?

IMO yes, though the problem is not with the draw_multiedges function proposed here, but the fact that the `draw_networkx_edges` function is extremely difficult to reason about in it's current state.

It would be great to refactor `draw_networkx_edges` so that the two `_draw_networkx_edges_line_collection` and `_draw_networkx_edges_fancy_arrow_patch` functions weren't defined in the module itself rather than internally in the `draw_networkx_edges` function. This should make the logic for handling e.g. self-loops *much* more clear, and should make it more straightforward to plug the multiedge drawing right into the `draw_networkx_edges` function. OTOH this refactor is *probably* a bit tricky. The good news is, the edge drawing is much better tested than it was say 3 releases ago, so developers can be *slightly* more confident that they're not breaking everything in an attempted refactor. It would be a shame to block the new functionality here for these unrelated reasons though, so +1 for pursuing multiedge drawing independent of the `draw_networkx_edges` mess; though I would also strongly advocate to make the multiedge drawing function private (as has been done here).

MridulS:
The issue seems to be coming from https://github.com/networkx/networkx/blob/99a925f695080787d077f620972c6552c4b0b4ba/networkx/drawing/nx_pylab.py#L700 when there are multiple edges between 2 nodes. The local copy of `edge_color` is overwritten by this line and it messes up the subsequent `edge_color`.

networkx/drawing/nx_pylab.py
```diff
@@ -890,6 +965,95 @@ def _connectionstyle(posA, posB, *args, **kwargs):
...
+                    G,
+                    pos,
+                    edgelist=[edge],
+                    width=width,
+                    edge_color=edge_color,
+                    style=style,
+                    alpha=alpha,
+                    arrowstyle=arrowstyle,
+                    arrowsize=arrowsize,
+                    edge_cmap=edge_cmap,
```
MridulS, line 1039: Yeaah, scratch my previous comment. It was the wrong reasoning.

The problem here is that this logic plots one `edge` at a time while the `_draw_networkx_edges` logic for extracting `edge_color` from `edge_cmap` is written under the assumption that all the edges are present while plotting it.

https://github.com/networkx/networkx/blob/main/networkx/drawing/nx_pylab.py#L682-L700 extracts the color mapping from a `cmap` to `edge_color` but this fails as there is only one `edge` being passed here in the `edgelist`. Ideally the `edge_color` should already be extracted here from the `edge_cmap` here before passing it to the `_draw_networkx_edges` to plot that one edge.


I can't think of a not-too-invasive of fixing this right now :/

dschult: Nice!  
I see that it is a tricky issue though.  When `edge_color` is an array, we will need to track the colors assigned to each multiedge and loop over them along with the edge itself. I guess if there are other arguments that allow arrays which match the order of `edgelist` we will need to do the same for them.

The basic idea is the same as what is in the PR: call `_draw_networkx_edges` once for each edge with `edgelist=[edge]`, only now we also need to pull out the colors from `edge_colors` as we loop over `edgelist` and use input arguments `edge_color=[ecolor]`.  I think that also means when multiedges are present that we have to collect not just the counts of how many multiedges there are, but also the colors of those edges.

The collective looping should be done for arguments `width`, `style` and `edge_color` with values collected along with edgelist. Perhaps we can construct a list for each edge with values `edge`, `width`, `style`, `edge_color`, `rad`. Then loop over that container of lists calling the `_draw_networkx_edges` function for each one.
`_draw_networkx_edges(G, pos, edgelist=[edge], width=[ewidth], edge_color=[ecolor], style=[estyle], ...)`.  The hard part I think is computing `rad` because you have to know how many multiedges there are before you assign those values and the values are ordered by edgelist, not necessarily with all the multiedges together.  You could collect the `edge_counts` first and then keep track of how many times each edge has appeared as you construct this container of edge information.

```python
num_edges = len(edgelist)
edge_counts = Counter(edgelist)
edge_so_far = dict.fromkeys(edge_counts, 0)  # make dict keyed by edge to value 0

widthlist = width if hasattr(width, "__iter__") else [width] * num_edges 
colorlist = edge_color if hasattr(edge_color, "__iter__") else [edge_color] * num_edges 
stylelist = style if hasattr(style, "__iter__") else [style] * num_edges 
all_edge_info = list(zip(edgelist, widthlist, colorlist, stylelist)) 
for edge, ewidth, ecolor, estyle in all_edge_info:
    rad = edge_so_far[edge] * (0.5 / edge_counts[edge])
    drawnEdge = _draw_networkx_edges(G, pos, edgelist=[edge], width=ewidth, edge_color=ecolor, style=estyle, ..., connectionstyle="arc3,rad=" + str(rad)[:3], ...)
    edge_so_far[edge] += 1
```
(probably a lot of stuff left out of here... you can come up with something better perhaps.)

dschult:
See also #3813   which has some interesting ideas that might work for placing labels.

MridulS:
Thanks for working on this @dtekinoglu!

I'll close this for now and we can use https://github.com/networkx/networkx/issues/3813 as the tracking issue.

PR #5888, opened 2022-07-21
strategy_saturation_largest_first now accepts partial colorings

gpdwatkins:
Amended functionality of strategy_saturation_largest_first so that the colors input can now be a partial coloring. This is so that the coloring of a partially-colored graph - in which the partial coloring is valid (the new code checks this) - can be completed using DSATUR. This reflects the functionality described in the docstring:
```
``colors`` is a dictionary mapping nodes of ``G`` to colors, for those nodes that have already been colored. 
```
See the discussion of this issue here: https://github.com/networkx/networkx/discussions/5884

There is also a test that checks (on two graphs) that if a partial coloring (of i nodes) matches the color assignments of the first i nodes given by the application of DSATUR to the uncolored graph, then the completed colorings also match.

### DIFF
--- a/networkx/algorithms/coloring/greedy_coloring.py
+++ b/networkx/algorithms/coloring/greedy_coloring.py
@@ -209,29 +209,44 @@ def strategy_saturation_largest_first(G, colors):
 
     """
     distinct_colors = {v: set() for v in G}
-    for i in range(len(G)):
-        # On the first time through, simply choose the node of highest degree.
-        if i == 0:
-            node = max(G, key=G.degree)
-            yield node
-            # Add the color 0 to the distinct colors set for each
-            # neighbors of that node.
-            for v in G[node]:
-                distinct_colors[v].add(0)
-        else:
-            # Compute the maximum saturation and the set of nodes that
-            # achieve that saturation.
-            saturation = {
-                v: len(c) for v, c in distinct_colors.items() if v not in colors
-            }
-            # Yield the node with the highest saturation, and break ties by
-            # degree.
-            node = max(saturation, key=lambda v: (saturation[v], G.degree(v)))
-            yield node
-            # Update the distinct color sets for the neighbors.
-            color = colors[node]
-            for v in G[node]:
-                distinct_colors[v].add(color)
+
+    # Add the node color assignments given in colors to the 
+    # distinct colors set for each neighbor of that node
+    for vertex, color in colors.items():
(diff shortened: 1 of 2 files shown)


networkx/algorithms/coloring/tests/test_coloring.py
```diff
@@ -429,6 +431,54 @@ def test_hardest_prob(self):
         )
         check_state(**params)
 
+    def test_strategy_saturation_largest_first(self):
+        def color_remaining_nodes(G, colored_vertices):
+            color_assignments = []
+            aux_colored_vertices = {
+                key: value for key, value in colored_vertices.items()
+            }
```
dschult, line 439: Is this different than `colored_vertices.copy()`? If not, the copy version is more readable.

networkx/algorithms/coloring/greedy_coloring.py
```diff
@@ -209,29 +209,44 @@ def strategy_saturation_largest_first(G, colors):
...
+    # i.e. no neighboring nodes have the same color
+    if len(colors) >= 2:
+        for vertex, color in colors.items():
+            if color in distinct_colors[vertex]:
+                raise nx.NetworkXError(
+                    "Neighboring vertices must have different colors"
+                )
+
+    # If 0 nodes have been colored, simply choose the node of highest degree.
+    if len(colors) == 0:
```
dschult, line 229: I think PEP8 standards say `if len(colors) == 0:` should be written `if not colors:`

dschult (commented):
This looks good to me!  Thanks!

I have two very minor comments below. 
Thanks for making the review so easy.

networkx/algorithms/coloring/tests/test_coloring.py
gpdwatkins: Good point, I will amend to use copy()

networkx/algorithms/coloring/greedy_coloring.py
gpdwatkins: Ah I didn't know that convention. Will update - thanks!

dschult (approved):
LGTM!  Thanks again...

rossbar (approved):
Thanks @gpdwatkins . It took me a while to grok the test, but I think that's because I find the scaffolding to set up the tests in that file generally confusing. You've managed to fit the new test into that framework very well, LGTM!

PR #5890, opened 2022-07-23
use plain inheritance instead of class variables

danieleades:
use plain inheritance instead of class variables for Graph factory methods.

from discussion in #4014 it seems these factory methods were implemented as class methods for performance reasons, that no longer hold.

Using plain methods and inheritance instead of patch class methods is far less 'magic', and therefore easier to reason about. It uses plain old python inheritance instead.

there's another benefit too- when this library starts using type annotations, then an IDE or linter will ensure that consumers correctly subclass these factory methods, including enforcing return types that subclass `dict`. The use of class methods in the previous implementation confounds this static analysis


danieleades:
FYI @NeilGirdhar

NeilGirdhar:
I love this and the networkx team will as well.

>  enforcing return types that subclass dict

FYI I think this (and the various comments in the code) should say `Mapping` instead of `dict`.

danieleades:
> FYI I think this (and the various comments in the code) should say `Mapping` instead of `dict`.  The phrase "dict-like object" should probably be "mapping" or `collections.abc.Mapping`.

Yeah I imagine that's probably the case

danieleades:
@MridulS this has been open for coming up on on 7 months now. How can I progress this?

danieleades:
> @MridulS this has been open for coming up on on 7 months now. How can I progress this?
> 
> @jarrodmillman, @dschult ?

ping maintainers?

danieleades:
ping @rossbar

NeilGirdhar:
Any reason this hasn't been reviewed in 3 years?

PR #5898, opened 2022-07-27
[ENH] Find and verify a minimal D-separating set in DAG

adam2392:
## Summary
This addresses: https://github.com/networkx/networkx/discussions/5859

where I raised the issue of enhancing the `nx.algorithms.d_separated` submodule with additional two related algorithms.

1. for finding a minimal separating set, and 
2. the other is to verify that a separating set is indeed minimal in a DAG. 

In order to compute these, there are some operations that can be performed on a DAG:

1. moralizing a directed graph
2. computing the v-structure of a directed graph

## Misc.

Do I have to add a changelog here too?

### DIFF
--- a/networkx/algorithms/d_separation.py
+++ b/networkx/algorithms/d_separation.py
@@ -62,7 +62,7 @@ from collections import deque
 import networkx as nx
 from networkx.utils import UnionFind, not_implemented_for
 
-__all__ = ["d_separated"]
+__all__ = ["d_separated", "compute_minimal_separating_set", "is_separating_set_minimal"]
 
 
 @not_implemented_for("undirected")
@@ -140,3 +140,160 @@ def d_separated(G, x, y, z):
         return False
     else:
         return True
+
+
+@not_implemented_for("undirected")
+def compute_minimal_separating_set(G, x, y):
+    """Compute a minimal separating set between X and Y.
+
+    Uses the algorithm presented in [1]_.
+
+    Parameters
+    ----------
+    G : graph
+        A networkx DAG.
+    x : node
+        Node X in the graph, G.
+    y : node
+        Node Y in the graph, G.
+
+    Raises
+    ------
+    NetworkXError
+        The *d-separation* test is commonly used with directed
+        graphical models which are acyclic.  Accordingly, the algorithm
+        raises a :exc:`NetworkXError` if the input graph is not a DAG.
+
+    NodeNotFound
+        If any of the input nodes are not found in the graph,
+        a :exc:`NodeNotFound` exception is raised.
+
+    References
+    ----------
+    .. [1] Tian, J., & Paz, A. (1998). Finding Minimal D-separators.
+    """
+    if not nx.is_directed_acyclic_graph(G):
+        raise nx.NetworkXError("graph should be directed acyclic")
+
+    union_xy = x.union(y)
+
+    if any(n not in G.nodes for n in union_xy):
+        raise nx.NodeNotFound("one or more specified nodes not found in the graph")
+
+    # first construct the set of ancestors of X and Y
+    x_anc = set(G.ancestors(x))
+    y_anc = set(G.ancestors(y))
+    D_anc_xy = x_anc.union(y_anc)
+    D_anc_xy = D_anc_xy.union((x, y))
+
+    # second, construct the moralization of the subgraph of Anc(X,Y)
+    moral_G = nx.moralize_graph(G.subgraph(D_anc_xy))
+
+    # find a separating set Z' in moral_G
+    Z_prime = set(G.parents(x)).union(set(G.parents(y)))
+
+    # perform BFS on the graph from 'x' to mark
+    Z_dprime = _bfs_with_marks(moral_G, x, Z_prime)
+    Z = _bfs_with_marks(moral_G, y, Z_dprime)
+    return Z
+
+
+def is_separating_set_minimal(G, x, y, z):
+    """Determine if a separating set is minimal.
+
+    Uses the algorithm 2 presented in :footcite:`Tian1998FindingMD`.
+
+    Parameters
+    ----------
+    G : nx.DiGraph
+        The graph.
+    x : node
+        X node.
+    y : node
+        Y node.
+    z : Set
+        The separating set to check is minimal.
+
+    Returns
+    -------
+    bool
+        Whether or not the `z` separating set is minimal.
+
+    References
+    ----------
+    .. footbibliography::
+    """
+    x_anc = G.ancestors(x)
+    y_anc = G.ancestors(y)
+    xy_anc = x_anc.union(y_anc)
+
+    # if Z contains any node which is not in ancestors of X or Y
+    # then it is definitely not minimal
+    if any(node not in xy_anc for node in z):
+        return False
+
+    D_anc_xy = x_anc.union(y_anc)
+    D_anc_xy = D_anc_xy.union((x, y))
+
+    # second, construct the moralization of the subgraph
+    moral_G = nx.moralize_graph(G.subgraph(D_anc_xy))
+
+    # start BFS from X
+    marks = _bfs_with_marks(moral_G, x, z)
+
+    # if not all the Z is marked, then the set is not minimal
+    if any(node not in marks for node in z):
+        return False
+
+    # similarly, start BFS from Y and check the marks
+    marks = _bfs_with_marks(moral_G, y, z)
+    # if not all the Z is marked, then the set is not minimal
+    if any(node not in marks for node in z):
+        return False
+
+    return True
+
+
+def _bfs_with_marks(G, start_node, check_set):
+    """Breadth-first-search with markings.
+
+    Performs BFS starting from ``start_node`` and whenever a node
+    inside ``check_set`` is met, it is "marked". Once a node is marked,
+    BFS does not continue along that path. The resulting marked nodes
+    are returned.
+
+    Parameters
+    ----------
+    G : nx.Graph
+        An undirected graph.
+    start_node : node
+        The start of the BFS.
+    check_set : set
+        The set of nodes to check against.
+
+    Returns
+    -------
+    marked : set
+        A set of nodes that were marked.
+    """
+    visited = dict()
+    marked = dict()
+    queue = []
+
+    visited[start_node] = None
+    queue.append(start_node)
+    while queue:
+        m = queue.pop(0)
+
+        for neighbr in G.neighbors(m):
+            if neighbr not in visited:
+                # memoize where we visited so far
+                visited[neighbr] = None
+
+                # mark the node in Z' and do not continue
+                # along that path
+                if neighbr in check_set:
+                    marked[neighbr] = None
+                else:
+                    queue.append(neighbr)
+    return set(marked.keys())
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -8,7 +8,7 @@ to the user to check for that.
 import heapq
 from collections import deque
 from functools import partial
-from itertools import chain, product, starmap
+from itertools import chain, combinations, product, starmap
 from math import gcd
 
 import networkx as nx
@@ -30,6 +30,8 @@ __all__ = [
     "dag_longest_path",
     "dag_longest_path_length",
     "dag_to_branching",
+    "compute_v_structures",
+    "moralize_graph",
 ]
 
 chaini = chain.from_iterable
@@ -1144,3 +1146,85 @@ def dag_to_branching(G):
     B.remove_node(0)
     B.remove_node(-1)
     return B
+
+
+@not_implemented_for("undirected")
+@not_implemented_for("multigraph")
+def moralize_graph(G, inplace=False):
+    """Moralize a graph.
+
+    The moralization of a graph adds an edge between the parents
+    of common children (unshielded colliders, or v-structures)
+    in a DAG and then converts the graph to an undirected graph.
+
+    Parameters
+    ----------
+    G : graph
+        A networkx DAG.
+    inplace : bool
+        Whether to modify the existing DAG in place.
+
+    Returns
+    -------
+    moral_G : nx.Graph
+        An undirected graph of the moralization of ``G``.
+    """
+    if not is_directed_acyclic_graph(G):
+        raise nx.NetworkXError("graph should be directed acyclic")
+
+    if inplace:
+        moral_G = G
+    else:
+        moral_G = G.copy()
+
+    # find all v-structures
+    v_structs = compute_v_structures(moral_G)
+
+    # add an edge between all parents of common children
+    for p1, _, p2 in v_structs:
+        moral_G.add_edge(p1, p2)
+
+    # convert graph to undirected graph
+    moral_G = moral_G.to_undirected()
+    return moral_G
+
+
+@not_implemented_for("undirected")
+def compute_v_structures(graph):
+    """Iterate through the graph to compute all v-structures.
+
+    V-structures, or colliders are triples in the directed graph where
+    two parent nodes point to the same child and the two parent nodes
+    are not adjacent.
+
+    Parameters
+    ----------
+    graph : graph
+        A directed graph.
+
+    Returns
+    -------
+    vstructs : Set[Tuple]
+        The v structures within the graph. Each set has a 3-tuple with the
+        parent, collider, and other parent.
+    """
+    vstructs = set()
+    for node in graph.nodes:
+        # get the set of parents and spouses
+        parents = set(graph.parents(node))
+        children = set(graph.children(node))
+        spouses = set()
+        for child in children:
+            spouses = spouses.union(set(graph.parents(child)))
+        spouses.discard(node)
+
+        triple_candidates = parents.union(spouses)
+        for p1, p2 in combinations(triple_candidates, 2):
+            if (
+                not p2 in graph.adjacency(p1)  # should be unshielded triple
+                and graph.has_edge(p1, node)  # must be connected to the node
+                and graph.has_edge(p2, node)  # must be connected to the node
+            ):
+                p1_, p2_ = sorted((p1, p2))
+                vstructs.add((p1_, node, p2_))
+    return vstructs
--- a/networkx/algorithms/tests/test_d_separation.py
+++ b/networkx/algorithms/tests/test_d_separation.py
@@ -3,6 +3,7 @@ from itertools import combinations
 import pytest
 
 import networkx as nx
+from networkx.classes.digraph import DiGraph
 
 
 def path_graph():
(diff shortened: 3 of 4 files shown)


networkx/algorithms/tests/test_dag.py
```diff
@@ -708,3 +708,22 @@ def test_ancestors_descendants_undirected():
     undirected graphs."""
     G = nx.path_graph(5)
     nx.ancestors(G, 2) == nx.descendants(G, 2) == {0, 1, 3, 4}
+
+
+def test_compute_v_structures_raise():
+    G = nx.Graph()
```
dschult, line 714: `nx.DiGraph()` inside the test?

dschult:
You do not need to add a changelog to `doc/release/release_dev.rst`.  But you are welcome to! :}

networkx/algorithms/tests/test_dag.py
adam2392: Fixed the error that must be raised.

adam2392:
I wonder if there is a way to leverage the existing BFS code in networkx.traversal... 

The main one would prolly be generic_bfs_edges 

But one needs to pass in additional parameters to do BFS with "checking marks". Perhaps adding "neighbors_func_args" would make it possible? Lmk wyt @dschult

networkx/algorithms/d_separation.py
```diff
@@ -140,3 +140,202 @@ def d_separated(G, x, y, z):
         return False
     else:
         return True
+
+
+@not_implemented_for("undirected")
+def minimal_d_separator(G, x, y):
+    """Compute a minimal d-separating set between X and Y.
```
dschult, line 147: Can we use `u, v` instead of `x, y` to avoid confusion between nodes and sets of nodes?

Also, non of the functions (including the existing function) actually says what a d-separation is. It can't be more than a short paragraph, but it would sure be helpful. It could even go in the doc_string for the module if it is longer. Perhaps Wikipedia has some concise language.

And similarly, how hard would it be to have a one paragraph description of the algorithm. You only give the complexity here, and make the reader go to the paper for an inkling of what the algorithm does. Actually a paragraph describing what a "minimal d-separating set" might go a long way toward making this clear without details of the algorithm. At the moment, someone who hasn't heard this term before won't know what is going on.

networkx/algorithms/dag.py
```diff
@@ -30,6 +30,8 @@
     "dag_longest_path",
     "dag_longest_path_length",
     "dag_to_branching",
+    "compute_v_structures",
+    "moralize_graph",
```
dschult, line 34: Would the `moralize_graph` and `compute_v_structures` be better housed in the `moral.py` module?  That module only has a single function in it. Maybe it should be brought into `dag.py`? 

I bring it up because this is a good time to consider where these should go. In the end it doesn't really matter, but it would be good to make the code easily discoverable. You know more about this than I do -- what is a good organization of this code?

dschult (commented):
This looks good. Just a couple of comments at this stage. I'm still getting used to the code and the functionality being added.

networkx/algorithms/dag.py
adam2392: OH didn't even see the moral.py. Oh yeah, perhaps computing v-structures could go in there! I'll sub out the existing moralize_graph function too.

networkx/algorithms/d_separation.py
```diff
@@ -140,3 +140,202 @@ def d_separated(G, x, y, z):
...
+    check_set : set
+        The set of nodes to check against.
+
+    Returns
+    -------
+    marked : set
+        A set of nodes that were marked.
+    """
+    visited = dict()
+    marked = dict()
```
dschult, line 322: I've used `dict` as a set-like object before too, but is there a reason here? It looks like the straight `set()` will do everything here and not require recasting at the end.  Changed involve this kind of syntax.
`visited.add(neighbr)`
`marked.add(neighbr)`
`return marked`

networkx/algorithms/d_separation.py
```diff
@@ -140,3 +140,202 @@ def d_separated(G, x, y, z):
...
+    # similarly, start BFS from Y and check the marks
+    marks = _bfs_with_marks(moral_G, y, z)
+    # if not all the Z is marked, then the set is not minimal
+    if any(node not in marks for node in z):
+        return False
+
+    return True
+
+
+def _bfs_with_marks(G, start_node, check_set):
```
dschult, line 299: The `generic_bfs_edges` function does indeed do BFS. But there are so many possible nice additions to BFS that it can easily get bogged down if we try to put too much in.  Early on we had a beautiful version of BFS with hooks at every turn to allow the user to build just about any version they wanted. But all the hooks ended up making it slower and more importantly difficult to understand. The basic algorithm itself is not long or hard to read (as this code show). So, I would recommend leaving this private function here with the `marks` feature. I'm willing to bend on this if anyone would prefer moving it to the traversal subpackage.

dschult (commented):
Comments on the BFS part. Looking good....

networkx/algorithms/dag.py
adam2392: Refactored to place them inside `moral.py` now.

networkx/algorithms/d_separation.py
adam2392: I moved `_bfs_with_marks` into `traversal` submodule for now. I think perhaps it could also be turned into a more `generic_bfs_edges_with_marks` like function if in the future, there is any algorithm internally that also needs this type of algorithm?

networkx/algorithms/d_separation.py
adam2392: Done.

networkx/algorithms/d_separation.py
adam2392: Added descriptions.

networkx/algorithms/dag.py
adam2392: Actually it doesn't make sense to have them inside `moral.py`, but I would suspect it possibly makes sense to move `moral_graph()` into `dag.py` because I don't see any reason to compute the moral graph on directed graphs with loops... but I could be wrong.

adam2392:
Addressed your comments. Thanks!

networkx/algorithms/tests/test_d_separation.py
```diff
@@ -156,3 +166,38 @@ def test_invalid_nodes_raise_error(asia_graph):
     """
     with pytest.raises(nx.NodeNotFound):
         nx.d_separated(asia_graph, {0}, {1}, {2})
+    with pytest.raises(nx.NodeNotFound):
+        nx.is_minimal_d_separator(asia_graph, {0}, {1}, {2})
+    with pytest.raises(nx.NodeNotFound):
+        nx.minimal_d_separator(asia_graph, {0}, {1})
+
+
+def test_minimal_sep_set():
```
adam2392, line 175:
```suggestion
def test_minimal_d_separator():
```

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -418,3 +419,49 @@ def descendants_at_distance(G, source, distance):
...
+    queue.append(start_node)
+    while queue:
+        m = queue.pop(0)
+
+        for neighbr in G.neighbors(m):
+            if neighbr not in visited:
+                # memoize where we visited so far
+                visited[neighbr] = None
+
+                # mark the node in Z' and do not continue
```
adam2392, line 461:
```suggestion
                # mark the node in 'check_set' and do not continue
```

networkx/algorithms/d_separation.py
```diff
@@ -140,3 +160,190 @@ def d_separated(G, x, y, z):
...
+    Then BFS is ran again starting from ``v`` and marking nodes if they are
+    present in ``Z''``. The returned set of marked nodes is a minimal
+    d-separating set.
+
+    https://en.wikipedia.org/wiki/Bayesian_network#d-separation
+    """
+    if not nx.is_directed_acyclic_graph(G):
+        raise nx.NetworkXError("graph should be directed acyclic")
+
+    union_xy = set(u).union(set(v))
```
dschult, line 226: Would this be more clear?
`union_xy = {u, v}`

And below: `D_anc_xy = D_anc_xy.union(u, v))` could be `D_anc_xy.update((u, v))`

networkx/algorithms/d_separation.py
```diff
@@ -140,3 +160,190 @@ def d_separated(G, x, y, z):
...
+
+@not_implemented_for("undirected")
+def minimal_d_separator(G, u, v):
+    """Compute a minimal d-separating set between X and Y.
+
+    A d-separating set in a DAG is a set that when conditioned on
+    blocks all paths between the two sets of variables. This function
+    constructs a set that is "minimal", meaning it is the smallest
+    d-separating set between the two variables. This is not necessarily
+    unique. For more details, see Notes.
```
dschult, line 173: Can you define the words for readers who don't know this lingo?  Is a d-separating set in a DAG a set of nodes, or edges, or something else? What does the phrase "conditioned on blocks" mean? What is a variable? and what two sets of variables are you talking about?  This function constructs a (add d-separating) set. And what does "between the two variables" mean? 

Also the first line of the doc_string refers to  X and Y. We should be clear in that first line about whether u & v are nodes or sets of nodes. The rest of the docs need to be updated for u and v also.

Below, it talks about a "d-separation test", but that language is not used elsewhere. Also, "is commonly used with" begs the question as to whether it *can* be used with other types of graphs. Is it only defined for DAGs? Could we extend this function to work with cyclic or even undirected graphs?  Maybe the second sentence (or an edited version) would be sufficient to describe the NetworkXError.

networkx/algorithms/d_separation.py
```diff
@@ -7,6 +7,19 @@
 to the actual distribution parameters.  See [1]_ for a formal
 definition.
 
+To understand d-separation intuitively, consider the following example:
+
+For a pair of two nodes, u and v, all paths are considered open initially, unless
+they contain a collider along the path. d-separation is concerned with blocking all
+paths between these u and v. A path is considered blocked by the
+separating set if all open paths contain some variable in the d-separating
+set that is not a collider, or the descendant of a collider.
```
dschult, line 16: I'm not familiar with this terminology. Can you make it clear that "open" is a label being defined by this operation and not a technical term the reader has to look up?  Also, the term "collider" is not well established yet.  Maybe put the paragraph that defines it, or at least before you use the term here. 

In the paragraph about the collider, maybe make it more clear what "open" means by saying:
`then the path through the collider is called "open".` But maybe that is an incorrect definition of open -- I am only trying to guess at what open refers to here. We have no topology right?

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -418,3 +419,49 @@ def descendants_at_distance(G, source, distance):
         current_distance += 1
 
     return current_layer
+
+
+@not_implemented_for("directed")
+def _bfs_with_marks(G, start_node, check_set):
```
dschult, line 425: I have a weak preference to keep the `_bfs_with_marks` function close to the code that uses it -- until we run into another problem that might want to stop the BFS at specific nodes.  Putting it here removes the context in which it is makes sense. Do other maintainers have thoughts on this? I might be persuaded.

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -418,3 +419,49 @@ def descendants_at_distance(G, source, distance):
...
+    visited = dict()
+    marked = set()
+    queue = []
+
+    visited[start_node] = None
+    queue.append(start_node)
+    while queue:
+        m = queue.pop(0)
+
+        for neighbr in G.neighbors(m):
```
dschult, line 456: This is a very minor nit....
Our standard abbreviated variable name for a neighbor is `nbr`.  Not a big deal -- but what do you think?

dschult (commented):
Thanks very much!  Let me know if anything here is not clear. The goal is clarity of the docs and code. So, it matters less if what I suggest is done, and more that the end result is clear.

networkx/algorithms/traversal/breadth_first_search.py
adam2392: Makes sense. Moved back.

networkx/algorithms/traversal/breadth_first_search.py
adam2392: Agreed! Standardization removes cognitive complexity. These tips help for me towards becoming hopefully a highly efficient and productive contributor to networkx :)

networkx/algorithms/d_separation.py
adam2392: Okay, I tried explaining it a bit more. The definition of d-separation in words is quite involved because you have to define what an "open", or "blocked" path is. WDYT of the new writeup?

networkx/algorithms/d_separation.py
adam2392: I have reworded both function. Lmk wdyt?

networkx/algorithms/d_separation.py
adam2392: Done

networkx/algorithms/d_separation.py
```diff
@@ -100,6 +153,12 @@ def d_separated(G, x, y, z):
         If any of the input nodes are not found in the graph,
         a :exc:`NodeNotFound` exception is raised.
 
+    Notes
+    -----
+    A d-separating set in a DAG is a set that when conditioned on
+    blocks all paths between the two sets of variables.
```
dschult, line 159:
```suggestion
    A d-separating set in a DAG is a set of nodes that
    blocks all paths between the two sets. Nodes in `z`
    block a path if they are part of the path and are not 
    the child in a collider structure or a descendant of a collider.
    A collider structure along a path is ``... -> c <- ...`` where
    ``c`` is the collider node.
```
Is my wording even correct?  I'm adding it to make this doc_string more clear and also a bit more self-contained. But change it as you see fit.

networkx/algorithms/d_separation.py
```diff
@@ -11,6 +11,59 @@
...
+Before we overview, we introduce the following terminology to describe paths:
+
+- "open" path: A path between two nodes that can be traversed
+- "blocked" path: A path between two nodes that cannot be traversed
+
+A **collider** is a triplet of nodes along a path that is like the following:
+``... u -> c <- v ...``), where 'c' is a common successor of 'u' and 'v'. A path
+through a collider is considered "blocked". When
+a node that is a collider, or a descendant of a collider is included in
+the d-separating set, then the path through that collider node is "open". If the
```
dschult, line 29: Does the "descendent of a collider" have to be a collider along the path being considered? Or does the descendent have to be on the path? What is the relationship between the path and the collider and also the path and the descendant of the path?

I'm guessing that any colliders will block a path unless the collider node or any of its descendants are in the d-separation set.  So the d-separation set can include nodes very far from the collider and still "un-clock" paths through that collider if the collider has a directed path to a node in the d-separation set.

networkx/algorithms/d_separation.py
```diff
@@ -11,6 +11,59 @@
...
+path through the collider node is open, then we will call this node an open collider.
+
+If a path between 'u' and 'v' contains a non-collider node that is included in the
+d-separation set, then that path is "blocked". In other words, the d-separation set
```
dschult, line 33: I think the single quotes in this entire section should be double-back-ticks e.g., ``u`` so that the font is changes for those names (they aren't input parameters because this isn't a function's doc_string. input parameters would get single back-ticks).

dschult (commented):
Nice!  Thanks --
I've put some more comments...  The suggestions are only suggestions and you can choose to handle them as you like. The formatting of the variable names with back-ticks vs quotes vs double-back-ticks can be annoying sometimes, but it helps `sphinx` understand how to format the text in the generated webpage documentation.

networkx/algorithms/d_separation.py
adam2392: Changed the wording slightly

networkx/algorithms/d_separation.py
adam2392: Yeah any triple with `u -> w <- v` is unblocked if `w` or any of its descendants are in the d-separation set. If we then have `u -> w <- v -> .... -> a` and the rest of the path is not inside the d-separation set, then `u` and `a` are not d-separated on this path at least.

networkx/algorithms/d_separation.py
```diff
@@ -140,3 +199,233 @@ def d_separated(G, x, y, z):
...
+    is reached, then ``'z'`` is minimal.
+
+    For full details, see [1]_.
+
+    https://en.wikipedia.org/wiki/Bayesian_network#d-separation
+    """
+    if not nx.is_directed_acyclic_graph(G):
+        raise nx.NetworkXError("graph should be directed acyclic")
+
+    union_xy = {u, v}
```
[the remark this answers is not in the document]
adam2392: Done.

adam2392 (commented):
Added some comments.

networkx/algorithms/d_separation.py
dschult: That's an interesting nonlinear behavior for the blocking process.  The nodes in the d-separation set block paths -- except if they are in a particular graph structure. And then they open paths. 

Do you know the application that drove people to consider this idea of a "collider" and "opening the collider"?

adam2392: I think the reasons is that two node that collide at a third node should be "independent" of each other in the statistical sense. However, when you condition on the collider, you create a statistical dependency.

E.g. `x -> z <- y` would have `P(x,y) = P(x)P(y)`, but not `P(x, y | z) = P(x | z) P(y | z)`

Intuitively it means they are independent, but dependent when you condition on seeing the effect of both of them. E.g. water sprinklers and presence of clouds are independent, but if you condition on seeing the floor wet, then they are highly dependent.

dschult: Does this mean that an "open" path means that probabilistic dependence occurs along that path?
And the d-separation set if the set of nodes you are wanting to condition your probabilities on?
Is the "d" of d-separation standing for dependence?

dschult (approved):
This LGTM! Thanks @adam2392

networkx/algorithms/d_separation.py
adam2392: Yep to both the first questions. 

Re "d" in d-separation, yes it is trying to tie the idea of "dependence" with separation in a graph. I think also, if interpreted probabilistically, d-separation is tied to conditional probabilistic independence assuming the Markov condition: https://en.wikipedia.org/wiki/Causal_Markov_condition. However, I guess in general d-separation doesn't have to do with anything probabilistically, but based on what I know, it's almost solely only used in probabilistic graphical models. 

^ If you want, I can add some details regarding the explicit connection with Markov condition of graphs and what it means probabilistically?

adam2392:
Not sure why CI is failing but now it looks like it is from some viz issue

dschult:
The viz problem was solved by a merged PR in the last 12 hours or so. Can you merge with main again? Or you can just leave it and we should be able to review this PR without concern for that failed test.

networkx/algorithms/d_separation.py
adam2392: Okay I've updated the doc string with a summary of the description I just added.

adam2392:
Kay merged w/ main and added the summary of d-separation.

jarrodmillman (approved):

jarrodmillman:
Thanks!!

PR #5900, opened 2022-07-27
add support for normalizing node values with matplotlib normalization…

DanielAnthes:
… functions

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -316,6 +316,7 @@ def draw_networkx_nodes(
     node_shape="o",
     alpha=None,
     cmap=None,
+    norm=None,
     vmin=None,
     vmax=None,
     ax=None,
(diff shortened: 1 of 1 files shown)


dschult:
Thanks for this idea -- but I'm afraid that if we try to enable every possible argument to each matplotlib function we will explode. :)   I would prefer that we make the user convert the node color values to their normed versions first. This can be done in the graph using `nx.get_node_attributes/nx.set_node_attributes` along with e.g. `mpl.colors.Logrithmic`.  Or once you have the `node_color` values, you can simply use `mpl.colors.Logrithmic` to get the `normed_node_color`. And then call the draw function. Is there functionality that this approach does not enable?  (untested)

```python
norm = mpl.colors.Logrithmic()
normed_node_colors = norm(node_colors)
```

To change the values stored in the graph attributes:
```python
norm = mpl.colors.Logrithmic()  # or any other normalization-function constructor
wts = nx.get_node_attributes(G, "weight")  # as a dict
norm_wts = norm(list(wts.values()))  # as an ndarray
normed_weights = dict(zip(G.nodes, norm_wts))
nx.set_node_attributes(G, normed_weights, "normed_weights")
```

MridulS:
I agree with @dschult we shouldn't really expose every little knob of matplotlib, it will be hard to manage that. The above code example should achieve the same end output. I'll close this PR. Thanks!

PR #5902, opened 2022-07-29
Fix Louvain_partitions by yielding a copy of the sets in the partition gh-5901

### LINKED ISSUE #5901: louvain_partitions iteration difference
louvain_partitions leads to different clusterings depending on how one iterates over the returned generator

### Current Behavior
If I iterate over the generator returned from louvain_partitions by a for loop I get for each iteration the expected clustering levels. 
However, if I first create a list (or tuple, deque...) and then iterate over the list I get different results. In fact, the levels have certain clusters which are empty if I iterate over the list. 

### Expected Behavior
I would expect that no matter how I iterate over the resulting generator I would get the same results.

### Steps to Reproduce
```
import networkx as nx
import networkx.algorithms.community as nx_comm

G = nx.path_graph(200)

def evaluate(part):
    print(len(part))
    count = 0
    for nodes in part:
        if len(nodes) == 0:
            count += 1
    print(f"\tEmpty Clusters: {count}")

louvian_clustering = list(nx_comm.louvain_partitions(G, resolution=1, seed=20))
for part in louvian_clustering:
    evaluate(part)

print("----")

for part in nx_comm.louvain_partitions(G, resolution=1, seed=20):
    evaluate(part)
```
Output
```
90
    Empty Clusters: 76 
41
    Empty Clusters: 27 
17 
    Empty Clusters: 3
 14
    Empty Clusters: 0
 ----
 90
    Empty Clusters: 0
 41 
    Empty Clusters: 0 
17
    Empty Clusters: 0 
14 
    Empty Clusters: 0
```

### Environment

Python version: 3.10.5
NetworkX version: 2.8.5

dschult:
Fixes #5901 by yielding a list of copies-of-sets in the partition instead of the partition-list-of-sets that will be used for computing the next partition.

### DIFF
--- a/networkx/algorithms/community/louvain.py
+++ b/networkx/algorithms/community/louvain.py
@@ -179,7 +179,8 @@ def louvain_partitions(
     )
     improvement = True
     while improvement:
-        yield partition
+        # gh-5901 protect the sets in the yielded list from further manipulation here
+        yield [s.copy() for s in partition]
         new_mod = modularity(
             graph, inner_partition, resolution=resolution, weight="weight"
         )
--- a/networkx/algorithms/community/tests/test_louvain.py
+++ b/networkx/algorithms/community/tests/test_louvain.py
@@ -2,6 +2,7 @@ import networkx as nx
 from networkx.algorithms.community import (
     is_partition,
     louvain_communities,
+    louvain_partitions,
     modularity,
     partition_quality,
 )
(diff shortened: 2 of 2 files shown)


rossbar (approved):
Thanks @dschult , this looks like a very nice solution. Just wanted to ping @z3y50n to give him a chance to weight in but LGTM!

z3y50n (approved):
LGTM thanks for fixing this!

jarrodmillman (approved):

Issue #5889, opened 2022-07-21
Kou et al. Steiner tree approximation inefficiency


GuyAglionby:
The existing Steiner tree approximation algorithm (from Kou et al.) finds the metric closure of the whole graph `G`, which requires finding all-pairs shortest paths. However, all that is actually needed are the shortest paths between terminal nodes. For even moderately sized graphs, and in cases where there are few terminal nodes relative to total nodes, this therefore unnecessarily increases runtime. 

A solution could be to use `single_source_dijkstra` with a `source` and `target` for each pair, noting that we're restricted to undirected graphs so can avoid doubling up. It would be nice if we could supply multiple targets to this function to avoid throwing away work, but I can't see a function that lets us do this. In any case simply replacing with multiple calls to this method will improve things (see implementation of method that is currently used `all_pairs_dijkstra`).

nishant42491:
hello I have tried to implement an alternative single source shortest path implementation linked with #5905.

PR #5905, opened 2022-07-29
Steiner tree approximation

nishant42491:
with regards to issue #5889 I have taken a stab at implementing multiple calls to the shortest path algorithm.
I have removed the metric closure function from the previous implementation too.

### DIFF
--- a/networkx/algorithms/approximation/steinertree.py
+++ b/networkx/algorithms/approximation/steinertree.py
@@ -55,7 +55,7 @@ def steiner_tree(G, terminal_nodes, weight="weight"):
     for u_index in range(len(terminal_nodes)):
         u = terminal_nodes[u_index]
 
-        for v_index in range(u_index+1, len(terminal_nodes)):
+        for v_index in range(u_index + 1, len(terminal_nodes)):
             v = terminal_nodes[v_index]
             dist_u_v = nx.shortest_path_length(G=G, source=u, target=v, weight=weight)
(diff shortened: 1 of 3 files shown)


networkx/algorithms/approximation/steinertree.py
```diff
@@ -88,11 +49,20 @@ def steiner_tree(G, terminal_nodes, weight="weight"):
...
+    # M is the subgraph induced by terminal_nodes with shortest corresponding pairwise paths in G.
+
+    M = nx.Graph()
+    for u_index in range(len(terminal_nodes)):
+        u = terminal_nodes[u_index]
+
+        for v_index in range(u_index + 1, len(terminal_nodes)):
+            v = terminal_nodes[v_index]
+            dist_u_v = nx.shortest_path_length(G=G, source=u, target=v, weight=weight)
+            path_u_v = nx.shortest_path(G=G, source=u, target=v, weight=weight)
```
GuyAglionby, line 61: I will leave it to the maintainers for a full review, but just to note that here there are two independent calculations of shortest path. It might be better to calculate only `path_u_v` and use that to construct `dist_u_v`, avoiding a potentially expensive re-calculation.

rossbar (commented):
This one's been stale for a while and has conflicts + function removals that are backward incompatible and would have to be reverted anyways, so I will go ahead and close it.

#5889 certainly seems like a worthwhile line of investigation, but if interested I'd encourage opening a new PR that focuses specifically on the optimization without the removal of public functions!

PR #5908, opened 2022-08-01
Graph walks implementation by jfinkels & dtekinoglu

dtekinoglu:
I created test_walks.py for the new methods,  when I tried to run them locally by `pytest test_walks.py` it gives the following output:

Adsız

What should I do to succesfully run the tests? @dschult @rossbar @MridulS

### DIFF
--- a/networkx/algorithms/tests/test_walks.py
+++ b/networkx/algorithms/tests/test_walks.py
@@ -28,7 +28,7 @@ def test_non_integer_nodes():
     expected = {
         "A": {"A": 0, "B": 0, "C": 1},
         "B": {"A": 1, "B": 0, "C": 0},
-        "C": {"A": 0, "B": 1, "C": 0}
+        "C": {"A": 0, "B": 1, "C": 0},
     }
     assert num_walks == expected
 
--- a/networkx/algorithms/tests/test_walks.py
+++ b/networkx/algorithms/tests/test_walks.py
@@ -1,0 +1,103 @@
+"""Unit tests for the :mod:`networkx.algorithms.walks` module."""
+
+from functools import partial
+
+from networkx import (
+    DiGraph,
+    all_pairs_number_of_walks,
+    cycle_graph,
+    empty_graph,
+    number_of_walks,
+    relabel_nodes,
+    single_source_number_of_walks,
+)
+from networkx.utils import pairwise
+
+
+class NumberOfWalks:
+    """Unit tests for the
+    :func:`networkx.algorithms.walks.number_of_walks` function.
+    This is an abstract base class; concrete subclasses must set the
+    :attr:`method` class attribute to be one of the possible values for
+    the `method` keyword argument to the function under test.
+    """
+
+    @classmethod
+    def setUp(self):
+        self.G = DiGraph(pairwise(range(3), cyclic=True))
+        self.number_of_walks = partial(number_of_walks)
+
+    def test_neither_source_nor_target(self):
+        num_walks = self.number_of_walks(self.G, 3)
+        expected = {0: {0: 1, 1: 0, 2: 0}, 1: {0: 0, 1: 1, 2: 0}, 2: {0: 0, 1: 0, 2: 1}}
+        assert num_walks == expected
+
+    def test_source_only(self):
+        num_walks = self.number_of_walks(self.G, 2, source=2)
+        expected = {0: 0, 1: 1, 2: 0}
+        assert num_walks == expected
+
+    def test_target_only(self):
+        num_walks = self.number_of_walks(self.G, 2, target=2)
+        expected = {0: 1, 1: 0, 2: 0}
+        assert num_walks == expected
+
+    def test_source_and_target(self):
+        num_walks = self.number_of_walks(self.G, 2, source=0, target=2)
+        assert num_walks == 1
+
+
+class SingleSourceNumberOfWalksTestBase:
+    """Unit tests for the
+    :func:`networkx.algorithms.walks.single_source_number_of_walks`
+    function.
+    This is an abstract base class; concrete subclasses must set the
+    :attr:`method` class attribute to be one of the possible values for
+    the `method` keyword argument to the function under test.
+    """
+
+    @classmethod
+    def setUp(self):
+        self.single_source_number_of_walks = partial(single_source_number_of_walks)
+
+    def test_no_target(self):
+        G = cycle_graph(4)
+        source = 0
+        k = 3
+        num_walks = self.single_source_number_of_walks(G, k, source)
+        expected = {0: 0, 1: 4, 2: 0, 3: 4}
+        assert num_walks == expected
+
+    def test_target(self):
+        G = cycle_graph(4)
+        source = 0
+        target = 1
+        k = 3
+        num_walks = self.single_source_number_of_walks(G, k, source, target=target)
+        assert num_walks == 4
+
+
+class AllPairsNumberOfWalksTestBase:
+    """Unit tests for the
+    :func:`networkx.algorithms.walks.all_pairs_number_of_walks`
+    function.
+    This is an abstract base class; concrete subclasses must set the
+    :attr:`method` class attribute to be one of the possible values for
+    the `method` keyword argument to the function under test.
+    """
+
+    @classmethod
+    def setUp(self):
+        self.all_pairs_number_of_walks = partial(all_pairs_number_of_walks)
+
+    def test_basic(self):
+        G = cycle_graph(4)
+        k = 3
+        num_walks = self.all_pairs_number_of_walks(G, k)
+        expected = {
+            0: {0: 0, 1: 4, 2: 0, 3: 4},
+            1: {0: 4, 1: 0, 2: 4, 3: 0},
+            2: {0: 0, 1: 4, 2: 0, 3: 4},
+            3: {0: 4, 1: 0, 2: 4, 3: 0},
+        }
+        assert num_walks == expected
--- a/networkx/algorithms/tests/test_walks.py
+++ b/networkx/algorithms/tests/test_walks.py
@@ -1,6 +1,11 @@
 """Unit tests for the :mod:`networkx.algorithms.walks` module."""
 
+import pytest
+
 import networkx as nx
+
+pytest.importorskip("numpy")
+pytest.importorskip("scipy")
 
 
 class TestNumberOfWalks:
--- a/networkx/algorithms/tests/test_walks.py
+++ b/networkx/algorithms/tests/test_walks.py
@@ -8,57 +8,26 @@ pytest.importorskip("numpy")
 pytest.importorskip("scipy")
 
 
-class TestNumberOfWalks:
-    @classmethod
-    def setup_class(cls):
-        cls.G = nx.DiGraph(nx.utils.pairwise(range(3), cyclic=True))
+def test_neither_source_nor_target():
+    G = nx.DiGraph([(0, 1), (1, 2), (2, 0)])
+    num_walks = nx.number_of_walks(G, 3)
+    expected = {0: {0: 1, 1: 0, 2: 0}, 1: {0: 0, 1: 1, 2: 0}, 2: {0: 0, 1: 0, 2: 1}}
+    assert num_walks == expected
 
-    def test_neither_source_nor_target(self):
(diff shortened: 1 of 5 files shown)


networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,103 @@
...
+
+from networkx import (
+    DiGraph,
+    all_pairs_number_of_walks,
+    cycle_graph,
+    empty_graph,
+    number_of_walks,
+    relabel_nodes,
+    single_source_number_of_walks,
+)
```
dschult, line 13: I would say you should just use e.g. `nx.single_source_number_of_walks` instead of importing each individually.  Same for `pairwise`. But that's really just a style issue.

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,103 @@
...
+from networkx.utils import pairwise
+
+
+class NumberOfWalks:
```
dschult, line 17: I think the class names should start with "Test" for pytest to identify it as a class containing tests to run. Try
```python
class TestNumberOfWalks:
```
and see if that does the trick.

dschult (commented):
This is great!  Thanks for making progress here!
I think that pytest is not finding the tests because of a naming issue. see below.

networkx/algorithms/tests/test_walks.py
dtekinoglu: Thanks, it is working now 👍

dtekinoglu: All tests I wrote are passing locally, but they are all failing here 😞

dschult:
The tests are failing now because we test some configurations that don't have numpy/scipy installed. To make the test system skip any functions that use numpy/scipy we use the function `pytest.importorskip()` which is pronounced "import or skip". That is, if it imports, great!  If it doesn't, then skip these tests. You can use it inside a function/class to skip just that part. But in this case I think we just run it at the top of the module because none of these will work without numpy/scipy to make the adjacency matrix.

```python
sp = pytest.importorskip("scipy")
```
Actually, I think you don't call np or sp explicitly in the tests. So you probably don't need the "np = " or "sp = " part.  That just provides the name for the imported library if the import is successful.

rossbar (commented):
This is looking great, thanks @dtekinoglu !

I will follow up with a more detailed review, but at this stage I just wanted to discuss what we wanted to do with the API.

The way I see it, we essentially have one function that does all the work - `all_pairs_number_of_walks` - and two additional functions that essentially just call `all_pairs_number_of_walks` under-the-hood (though `single_source_number_of_walks` actually *duplicates* `all_pairs`) and then munge the output dictionary to present the results in the form requested as specified by the various combinations of keyword arguments.

The main question that jumped out to me is: since we're computing the full dictionary in every case, why add *only* the `all_pairs_number_of_walks` function then show how to get the various results (i.e from a single source to a single target, from a single source to all targets, etc.) in the docstring.

Alternatively, could we consider making `all_pairs_number_of_walks` a generator? Potentially this would mean that `all_pairs_number_of_walks` could be run on larger graphs without eating up a ton of memory, and in which case it probably makes more sense to have the other function(s) to consume the generator and return the desired, specific subset of results.

Just some general thoughts - once a decision is reached on this point I think everything else can move very quickly as the code/docstring/tests look great!

dschult:
I like this new interface. It is quite clear and easy to follow. 

Iit looks like the documentation rst file needs to be updated.

As for whether to implement a generator version, I think the generator would still need to compute the whole matrix just to yield the first part of the result. So we will have memory issues even if we yield parts of the dict-of-dict format. The memory issues stem from using the matrix approach to this problem — but I still think it is the best approach to finding the number of walks between pairs of nodes. So I’m leaning toward using the current PR and not a generator, but I might be missing something.

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,64 @@
...
+        assert num_walks == expected
+
+    def test_source_only(self):
+        num_walks = nx.number_of_walks(self.G, 2)
+        expected = {0: 0, 1: 1, 2: 0}
+        assert num_walks[2] == expected
+
+    def test_source_and_target(self):
+        num_walks = nx.number_of_walks(self.G, 2)
+        assert num_walks[0][2] == 1
```
rossbar, line 28: Now that we're down to only one function `nx.number_of_walks`, these test cases are redundant as the `test_neither_source_nor_target` covers them. The only difference is how the resultant dictionary is accessed, which doesn't really test the behavior of `nx.number_of_walks`.

This goes for some of the tests below as well, which are covered by `test_basic`.

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,64 @@
...
+import networkx as nx
+
+pytest.importorskip("numpy")
+pytest.importorskip("scipy")
+
+
+class TestNumberOfWalks:
+    @classmethod
+    def setup_class(cls):
+        cls.G = nx.DiGraph(nx.utils.pairwise(range(3), cyclic=True))
```
rossbar, line 14: Just a nit, but I find this incantation more difficult to grok than simply constructing the cycle graph manually:

```suggestion
        cls.G = nx.DiGraph([(0, 1), (1, 2), (2, 0)])
```

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,64 @@
...
+
+import networkx as nx
+
+pytest.importorskip("numpy")
+pytest.importorskip("scipy")
+
+
+class TestNumberOfWalks:
+    @classmethod
+    def setup_class(cls):
```
rossbar, line 13: I'd also advocate for getting rid of the test class here, especially since we're down to a single function. I've always found OO tests harder to reason about and more fraught with corner cases.

In this case it should be as simple as:
 1. Removing the class
 2. Removing the `self` arguments from the test methods so that they become test functions instead
 3. Replacing the `setup_class` with either a) a pytest fixture that creates the graph or b) simply creating the graph in each test case (since it's really not that expensive in this case).

This isn't a blocker though - the tests could go in with the current organization (once the other comments have been addressed) and we can re-organize them in a follow-up PR!

rossbar (commented):
Just a few comments from a quick once-over on the tests!

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,33 @@
...
+
+import pytest
+
+import networkx as nx
+
+pytest.importorskip("numpy")
+pytest.importorskip("scipy")
+
+
+def test_neither_source_nor_target():
```
rossbar, line 11:
```suggestion
def test_directed():
```

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,33 @@
...
+    G = nx.DiGraph([(0, 1), (1, 2), (2, 0)])
+    num_walks = nx.number_of_walks(G, 3)
+    expected = {0: {0: 1, 1: 0, 2: 0}, 1: {0: 0, 1: 1, 2: 0}, 2: {0: 0, 1: 0, 2: 1}}
+    assert num_walks == expected
+
+
+def test_no_target():
```
rossbar, line 18:
```suggestion
def test_undirected():
```

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,33 @@
...
+    G = nx.cycle_graph(4)
+    source = 0
+    k = 3
+    num_walks = nx.number_of_walks(G, k)
+    expected = {0: 0, 1: 4, 2: 0, 3: 4}
+    assert num_walks[source] == expected
```
rossbar, line 24: This is very much a nit, but we can save a few lines here by getting rid of some of the local variables:

```suggestion
    G = nx.cycle_graph(4)
    num_walks = nx.number_of_walks(G, 3)
    expected = {0: 0, 1: 4, 2: 0, 3: 4}
    assert num_walks[0] == expected
```

Also, in the grander scheme it might be better to do the "complete" test, i.e. set `expected` to be the full dict-of-dicts that maps all nodes to all other nodes, and assert that this is correct for *all* of num_walks, rather than just one instance.

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_source_and_target():
```
rossbar, line 27:
```suggestion
def test_digraph_non_integer_nodes():
```

networkx/algorithms/tests/test_walks.py
```diff
@@ -0,0 +1,33 @@
...
+    G = nx.DiGraph([("A", "B"), ("A", "C"), ("A", "D"), ("B", "D"), ("C", "D")])
+    source = "A"
+    target = "D"
+    k = 2
+    num_walks = nx.number_of_walks(G, k)
+    assert num_walks[source][target] == 2
```
rossbar, line 33: Same type of comment here - let's get rid of local variables ane expand the `expected` dict to include all the src/tgt combinations and compare it to the full output of `number_of_walks`.

```suggestion
    G = nx.DiGraph([("A", "B"), ("A", "C"), ("A", "D"), ("B", "D"), ("C", "D")])
    source = "A"
    target = "D"
    num_walks = nx.number_of_walks(G, 2)
    assert num_walks[source][target] == 2
```

rossbar (commented):
Much better thanks! Another quick round of review on the tests.

Now that we've updated the tests themselves, we should make sure that the test names make sense. I think all three tests are valuable, but we could rename them to give a better sense of what they actually focus on. See individual comments for name suggestions.

IMO it would also be an improvement to *always* check against the full dict-of-dicts that maps all "sources" to all "targets" as is done in the first test. That's a more complete test and we don't really gain a lot by testing that the dictionary lookup works as expected (this is already extensively tested in Python itself!)

dschult:
Hopefully we will get a `matrix_power` function in scipy.sparse soon. The expression A**n computes the element-wise power. Because A is a matrix of 0 and 1, it might be surprising, but A**n is just A.  :) 
So, some the tests' expected values need to be updated.

The variable walk_length is described as if it is the number of nodes in the sequence. But the length of a walk/path/trail is the number of edges. So I think we should describe walk_length as the number of edges and point out that this is one fewer than the number of nodes in the sequence. So, now the walk length can go to zero and it makes sense.

I changed this PR to use dense arrays and numpy's matrix_power, so we can merge this. To use SciPy, we will need to wait until we stop supporting versions of scipy.sparse that don't have matrix_power -- and that could be a year or more.
I made the suggested test fixes and added to and changed the wording of the doc_string.  I also added the check for a positive walk_length.

networkx/algorithms/walks.py
```diff
@@ -0,0 +1,80 @@
+"""Function for computing walks in a graph.
+"""
+from itertools import starmap
+from operator import eq
+
```
MridulS, line 5:
```suggestion
```

networkx/algorithms/walks.py
```diff
@@ -0,0 +1,80 @@
...
+    1
+
+    """
+    import numpy as np
+
+    if walk_length < 0:
+        raise ValueError(f"`walk_length` cannot be negative: {walk_length}")
+
+    A = nx.adjacency_matrix(G)
+    # power = sp.sparse.linalg.matrix_power(A, walk_length)
```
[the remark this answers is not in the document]
rossbar: I think we can uncomment this - matrix_power is available in scipy.sparse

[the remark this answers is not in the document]
dschult: I can't find matrix_power in scipy.sparse for arrays.  The spmatrix.__pow__ does a version for integer values, but the sparray class only does elementwise power as far as I can tell. Am I missing it?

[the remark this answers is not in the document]
rossbar: Ah sorry, my mistake - I got confused between matrices and arrays. Sparse matrices implement [matrix power](https://github.com/scipy/scipy/blob/e38ddf1422389f6fd8a0eead4f7e186122310ae2/scipy/sparse/_matrix.py#L53) but arrays do element wise. The comment is correct!

MridulS (approved):

rossbar (approved):
Awesome, thanks all for the hard work here!

PR #5910, opened 2022-08-02
Arf layout

cvanelteren:
Implements arf layout (Geipel 2006) which provides improvements over the traditional spring layout.
Short description

    The attractive and repulsive forces (arf) layout (Geipel
    2006) improves  the spring layout in  three ways. First,
    it prevents congestion of  highly connected nodes due to
    strong forcing  between nodes.  Second, it  utilizes the
    layout space  more effectively by preventing  large gaps
    that  spring layout  tends  to create.  Lastly, the  arf
    layout represents symmmetries in  the layout better than
    the default spring layout.

Example:
<details>
<summary>Plotting code </summary>

``` python
import networkx as nx, numpy as np, matplotlib.pyplot as plt
from functools import partial
graphs = [nx.krackhardt_kite_graph(),
          nx.florentine_families_graph(),
          nx.grid_graph((5,5)),
          nx.erdos_renyi_graph(100, 0.02, seed = 0)
          ]
fig, ax = plt.subplots(ncols = 2, nrows = 2, share = 0)

layouts = [partial(nx.arf_layout, alpha = 10, b = 5, seed = 0),
           partial(nx.spring_layout, seed = 0)]
centers = [np.array([-1.5, 0]), np.array([1.5, 0])]
for axi, g in zip(ax, graphs):
    for layout, center in zip(layouts, centers):
        pos = layout(g)
        pos = nx.rescale_layout_dict(pos)
        pos = {x: y + center for x, y in pos.items()}
        nx.draw(g, pos = pos,
                ax = axi, node_size = 10)
    axi.annotate("Spring", xy = [0.75, 1.02], xycoords = axi.transAxes,
        va = "center", ha = "center")
    axi.annotate("Arf", xy = [0.25, 1.02], xycoords = axi.transAxes,
        va = "center", ha = "center")
    axi.axis("equal")
fig.show()
```
</details>

image

### DIFF
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -1190,7 +1190,7 @@ def arf_layout(
     node_order = {node: i for i, node in enumerate(G)}
     for x, y in G.edges():
         if x != y:
-            idx, jdx = [node_order[i] for i in (x, y)]
+            idx, jdx = (node_order[i] for i in (x, y))
             K[idx, jdx] = a
 
     # vectorize values
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -1112,47 +1112,47 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
 
 
 def arf_layout(
-    g, pos=None, b=1, alpha=1.1, etol=1e-6, dt=1e-3, max_iter=1000, seed=None
+    G,
+    pos=None,
+    scaling=1,
+    a=1.1,
+    etol=1e-6,
+    dt=1e-3,
+    max_iter=1000,
 ):
     """Arf layout for networkx
 
-    The  attractive   and  repulsive  forces   (arf)  layout (Geipel 2006)
-    improves  the spring  layout  in three  ways. First,  it
-    prevents  congestion of  highly connected  nodes due  to
-    strong forcing  between nodes.  Second, it  utilizes the
-    layout space  more effectively by preventing  large gaps
-    that  spring layout  tends  to create.  Lastly, the  arf
-    layout represents symmmetries in  the layout better than
+    The attractive and repulsive forces (arf) layout [1]
+    improves the spring layout in three ways. First, it
+    prevents congestion of highly connected nodes due to
+    strong forcing between nodes. Second, it utilizes the
+    layout space more effectively by preventing large gaps
+    that spring layout tends to create. Lastly, the arf
+    layout represents symmmetries in the layout better than
     the default spring layout.
 
     Parameters
     ----------
-    g : nx.Graph or nx.DiGraph
+    G : nx.Graph or nx.DiGraph
         Networkx graph.
     pos : dict
         Initial  position of  the nodes.  If set  to None  a
-        random layout will be used using seed as input.
-    b : float
-        Controls the radius of  the circular space the graph
-        layout will be plotted in.
-    alpha : float
-        Attraction scalar between edges. Should be larger than 1.
+        random layout will be used.
+    scaling : float
+        Scales the radius of the circular layout space.
+    a : float
+        Strength of springs between connected nodes. Should be larger than 1. The greater a, the clearer the separation ofunconnected sub clusters.
     etol : float
-        Termination tolerance
+        Graduent sum of spring forces must be larger than `etol` before succesful termination.
     dt : float
-        Integration factor
+        Time step for force differential equation simulations.
     max_iter : int
         Max iterations before termination of the algorithm.
-    seed : int, RandomState instance or None  optional (default=None)
-        Set the random state for deterministic node layouts.
-        If int, `seed` is the seed used by the random number generator,
-        if numpy.random.RandomState instance, `seed` is the random
-        number generator,
-        if None, the random number generator is the RandomState instance used
-        by numpy.random.
 
     References
-    - https://www.worldscientific.com/doi/abs/10.1142/S0129183107011558
+    .. [1] "Self-Organization Applied to Dynamic Network Layout", M. Geipel,
+            International Jounral of Modern Physics C, 2007, Vol 18, No 10, pp. 1537-1549.
+            https://doi.org/10.1142/S0129183107011558 https://arxiv.org/abs/0704.1748
 
     Returns
     -------
--- a/networkx/drawing/layout.py
+++ b/networkx/drawing/layout.py
@@ -1111,8 +1111,9 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
     return pos
 
 
-def arf_layout(g, pos = None, b = 1,  alpha = 1.1, etol = 1e-6, dt = 1e-3,
-        max_iter = 1000, seed = None):
+def arf_layout(
+    g, pos=None, b=1, alpha=1.1, etol=1e-6, dt=1e-3, max_iter=1000, seed=None
+):
     """Arf layout for networkx
 
     The  attractive   and  repulsive  forces   (arf)  layout (Geipel 2006)
@@ -1167,7 +1168,7 @@ def arf_layout(g, pos = None, b = 1,  alpha = 1.1, etol = 1e-6, dt = 1e-3,
     import numpy as np
 
     assert alpha > 1, "Alpha should be larger than 1"
-    pos_tmp = nx.random_layout(g, seed = seed)
+    pos_tmp = nx.random_layout(g, seed=seed)
     if pos is None:
         pos = pos_tmp.copy()
     else:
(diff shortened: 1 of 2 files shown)


cvanelteren:
Small remark. Tests seem to be failing on test not relating to this PR. Local pytests pass for the changed files.

networkx/drawing/tests/test_layout.py
```diff
@@ -402,7 +411,7 @@ def test_rescale_layout_dict(self):
         for k, v in expectation.items():
             assert (s_vpos[k] == v).all()
         s_vpos = nx.rescale_layout_dict(vpos, scale=2)
-
+        assert s_vpos == {0: (-2, -2), 1: (2, 2), 2: (0, 0)}
```
dschult, line 414: I think this line is causing the test failures. Does it fail locally?

cvanelteren: @dschult I rebased the code. I think I pulled the wrong main which was not up-to-date. Should be ok now.

networkx/drawing/layout.py
```diff
@@ -1110,6 +1111,108 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
...
+    bV = np.sqrt(N) * b  # scaling constant for circular space
+
+    # looping variables
+    error = etol + 1
+    n_iter = 0
+    while error > etol:
+        diff = p[:, None] - p[None]
+        A = np.linalg.norm(diff, axis=-1)[..., None]
+        # attraction_force - repulsions force
+        change = K[..., None] * diff - bV / A * diff
```
dschult, line 1204: I'm not sure what best practices are these days. But at one point it was deemed more readable to use `np.newaxis` instead of `None`. Thoughts?         Similarly, the ellipsis might be easier to read if it is written out as `K[:, :, None]`. But these are just questions, not suggestions.

In the last line, why not factor out the `diff` multiplication. I think the paper does it that way, and it is fewer multiplications, but is there another reason to keep the multiplications separate?

networkx/drawing/layout.py
```diff
@@ -1110,6 +1111,108 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
...
+        change = np.nansum(change, axis=0)
+        p += change * dt
+
+        error = np.linalg.norm(change, axis=-1).sum()
+        # error = abs(change).sum()
```
dschult, line 1209: I think `abs(change).sum()` is the same as `np.linalg.norm(change, ord=1)`. That is, it is the L_1 norm. But line 1208 is summing the norms. So, is the comment correct? Or am I just being confused by it.

networkx/drawing/tests/test_layout.py
```diff
@@ -443,3 +450,12 @@ def test_multipartite_layout_layer_order():
     G.nodes["a"]["subset"] = "layer_0"  # Can't sort mixed strs/ints
     pos_nosort = nx.multipartite_layout(G)  # smoke test: this should not raise
     assert pos_nosort.keys() == pos.keys()
+
+    def test_arf_layout(self):
```
dschult, line 454:
```suggestion
    def test_arf_layout_partial_pos_input(self):
```

networkx/drawing/layout.py
```diff
@@ -1110,6 +1111,108 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
...
+    seed : int, RandomState instance or None  optional (default=None)
+        Set the random state for deterministic node layouts.
+        If int, `seed` is the seed used by the random number generator,
+        if numpy.random.RandomState instance, `seed` is the random
+        number generator,
+        if None, the random number generator is the RandomState instance used
+        by numpy.random.
+
+    References
+    - https://www.worldscientific.com/doi/abs/10.1142/S0129183107011558
```
dschult, line 1155:
```suggestion
    .. [1] "Self-Organization Applied to Dynamic Network Layout", M. Geipel,
            International Jounral of Modern Physics C, 2007, Vol 18, No 10, pp. 1537-1549.
            https://doi.org/10.1142/S0129183107011558 https://arxiv.org/abs/0704.1748
```

dschult (commented):
I hope these comments are helpful. I went through the code this time. And I found the numpy indexing and multiplying out arrays using newaxis to be somewhat difficult to grok... But hopefully you can help me some.

The tests don't contain any test that the values produced are specific values. 
Do you have any cases where you know what the values should be?

networkx/drawing/layout.py
cvanelteren: I commonly use `None` due to it being shorter than `np.newaxis` but for clarity and interface compatibility I think it is better to use `np.newaxis` (although in practice I still end up using `None` a lot, behind the scenes it is the same anyway). I prefer the ellipses as it is clear that the last axis is used.

networkx/drawing/layout.py
```diff
@@ -1110,6 +1111,108 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
...
+        return pos
+    K = np.zeros((N, N))
+    for idx, x in enumerate(g.nodes()):
+        for jdx, y in enumerate(g.nodes()):
+            if idx == jdx:
+                K[idx, jdx] = 0
+            elif g.has_edge(x, y) or g.has_edge(y, x):
+                K[idx, jdx] = alpha
+            else:
+                K[idx, jdx] = 1
```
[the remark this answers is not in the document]
cvanelteren: Good suggestions and should have done this in the first place. Fixed!

networkx/drawing/layout.py
```diff
@@ -1110,6 +1111,108 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
...
+    import numpy as np
+
+    assert alpha > 1, "Alpha should be larger than 1"
+    pos_tmp = nx.random_layout(g, seed=seed)
+    if pos is None:
+        pos = pos_tmp.copy()
+    else:
+        for node in g.nodes():
+            if node not in pos:
+                pos[node] = pos_tmp[node].copy()
```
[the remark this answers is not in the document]
cvanelteren: Numpy leaves a shallow copy. I intended here to ensure that the array being copied  is modified in the new `pos` only. Not sure if this is a bad habit of mine. Here, `pos_tmp` and `pos` don't interact, i.e. `pos_tmp` is not modified after this point in the code. As such the results are not affected by this and the copy is not necessary. Could be a c++ tick that I don't want the pointer to reference something that may be deallocated.

networkx/drawing/layout.py
```diff
@@ -1110,6 +1111,108 @@ def multipartite_layout(G, subset_key="subset", align="vertical", scale=1, cente
...
+        by numpy.random.
+
+    References
+    - https://www.worldscientific.com/doi/abs/10.1142/S0129183107011558
+
+    Returns
+    -------
+    pos : dict
+        A dictionary of positions keyed by node.
+
```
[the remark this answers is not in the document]
cvanelteren: I misrerad `a` to be an `alpha` this is changed in the code now. Additionally, the checks are added for `a`. The `NetworkXError` may be a good idea, but in practice arf does not converge but still produces very usable layouts. I therefore think a warning would be more opt, what do you think?

networkx/drawing/layout.py
cvanelteren: It is indeed the same, I forgot to take this comment out and opted for the norm to make it congruent with the math in the paper. I am in favor for the `norm` notation, but open to change it to 1209 with the proper axis added.

networkx/drawing/tests/test_layout.py
cvanelteren: Fixed

networkx/drawing/layout.py
cvanelteren: Thanks and fixed!

cvanelteren:
> I hope these comments are helpful. I went through the code this time. And I found the numpy indexing and multiplying out arrays using newaxis to be somewhat difficult to grok... But hopefully you can help me some.

The comments were helpful. I incorporated many changes and fixed and or added where necessary. I shared your feeling about numpy's broadcast system. It allows for neatly writing complex operations while obfuscating the operations. Using `np.einsum` could be an alternative but I sometimes find that equally difficult to understand properly. Anyway, the math checks out for this PR, but it could provide an interesting discussion point for future PRs (broadcasting vs einsum).

> The tests don't contain any test that the values produced are specific values. Do you have any cases where you know what the values should be?

I can't recall that the paper provides bounds or proofs of convergence. I believe the author argued that there are no objective markers for what makes a good layout and as such I can't provide an absolute test case other than a known heuristic that converges. The paper uses a few measures to determine the "success" of different test graphs. Are you looking for a particular case which has a known solution? One trivial example may be an empty graph that is approximated by a circle. Something like

<details>  <summary> example </summary>

```
G = nx.empty_graph(10)
pos = nx.arf_layout(G, max_iter = 10000)
nx.draw(G, pos)
```


</details>

image

dschult:
I was thinking more simple than the circle and hub layout obtained from an empty graph (which is nice for testing repulsion but not for attraction).  I was thinking more like 3 nodes, 1 edge.   Or even 2 nodes, 1 edge.
I'm guessing one could construct a steady state of the equations in this kind of special case by hand. Then using those positions as initial conditions, you could test that the results are indeed a solution (though not the only one).
 
But I'm leaning toward worrying about testing of all the spring-based layouts in another PR. This introduces new layout methods to the same level of testing that we are providing for the other layouts. 

It looks like the only test that is failing now is the `isort` linter which wants `numpy` to be imported after `warnings` (because it is not a builtin library) and with a blank line between the builtins and non-builtins as well as after the non-builtins. This is all based on the "Details" link of the failing style test.

networkx/drawing/layout.py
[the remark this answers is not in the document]
dschult: Looks like `spring_layout` also doesn't raise an exception when max_iter i reached...
So let's leave out the warning and the exception. They can be handled in another PR and implemented with whatever is decided for all spring-like layouts.

cvanelteren:
Should be ok now. I would also be in favor of having a separate test suite part as a PR that would test the currently present force layouts. If you want I can have a poke at it and draft up the PR for testing the force layouts along the lines we discussed here and in #5311.

dschult (approved):
That would be great if you could put together a PR to take on the task of testing the force layouts.
Thanks!

jarrodmillman (approved):

PR #5914, opened 2022-08-04
Update lattice.py

Geometrein:
Fixed a mistake in the doctoring for hexagonal_lattice_graph (odd -> even).


dschult (approved):
Thanks very much for this!
I approve this PR.

PR #5924, opened 2022-08-10
Remove old Appveyor cruft

jarrodmillman:

### DIFF
--- a/networkx/conftest.py
+++ b/networkx/conftest.py
@@ -264,7 +264,3 @@ if not has_pydot:
     collect_ignore += needs_pydot
 if not has_sympy:
     collect_ignore += needs_sympy
-
-# FIXME:  This is to avoid errors on AppVeyor
-if sys.platform.startswith("win"):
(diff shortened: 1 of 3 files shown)


jarrodmillman:
According to https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
```
Whether the name can be used to open the file a second time, while the named temporary file
is still open, varies across platforms (it can be so used on Unix; it cannot on Windows).
If delete is true (the default), the file is deleted as soon as it is closed.
The returned object is always a file-like object whose file attribute is the
underlying true file object. This file-like object can be used in a
[with](https://docs.python.org/3/reference/compound_stmts.html#with) statement,
just like a normal file.
```

rossbar (approved):
LGTM, thanks @jarrodmillman 

I'd advocate to replace the uses of `tempfile` in the docstring examples with e.g. `io.StringIO` objects, but that's a subject for another issue/PR!

Issue #5871, opened 2022-07-15
generic_bfs_edges has apparently duplicated input arguments


dschult:
The `generic_bfs_edges()` function has input arguments `neighbors` and `sort_neighbors` and they seems to do the same thing -- or more precisely, their functionality duplicates. 

`neighbors` is "A function that takes a newly visited node of the graph as input and returns an *iterator* (not just a list) of nodes that are neighbors of that node"

while `sort_neighbors` is "A function that takes the list of neighbors of given node as input, and returns an *iterator* over these neighbors but with custom ordering."

In the code, the neighbors of a node are found using:
```python
lambda node: iter(sort_neighbors(neighbors(node)))
```
But that just means that `neighbors` should have been set by the user to this composition of functions.

We need to at least make the doc_string more clear, but it seems that we should eliminate the `sort_neighbors` input and have users just provide `neighbors`. And the doc_strings should correctly reflect whether these function actually need to return iterators or not.

still-n0thing:
- If I get this right ```neighbors``` is a callable object (or function) which decides how the neighbors of visited node should be added to the ```queue``` ? 
- Do we need to remove ```sort_neighbors```(is a callable object or function) and only work with ```neighbors```(is a callable object or function)? since that might cause break some things (I am not really sure about it)

still-n0thing:
@dschult ?

dschult:
You are correct that `neighbors` is a callable function that returns the neighbors of a visited node so they can be added to the queue. And `sort_neighbors` seems to do the same thing -- only it's name emphasizes the function of sorting the results. The idea of this issue is to combine these two input arguments into one and better document what that function does.

I'm not sure exactly how this would/should work. But I'm sure it will require a deprecation where one of the inputs still works but raises a DeprecationWarning if used.

still-n0thing:
https://github.com/networkx/networkx/pull/5925 @dschult

PR #5925, opened 2022-08-10
modifies ```bfs_edges``` and adds warning to ```generic_bfs_edges```

still-n0thing:
- [x] adds the deprecated warning for ```sort_neighbors``` argument in ```generic_bfs_edges```
- [x] modifies the implementation of ```bfs_edges``` where it calls ```generic_bfs_edges``` with only ```neighbors``` parameters.
[#5871](https://github.com/networkx/networkx/issues/5871)


still-n0thing:
[#5871](https://github.com/networkx/networkx/issues/5871)

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -68,6 +64,16 @@ def generic_bfs_edges(G, source, neighbors=None, depth_limit=None, sort_neighbor
     .. _Depth-limited-search: https://en.wikipedia.org/wiki/Depth-limited_search
     """
     if callable(sort_neighbors):
+        import warnings
+
+        warnings.warn(
+            (
+                "sort_neighbors parameter is deprecated and will be removed "
+                "in NetworkX 3.0, use neighbors parameter instead."
```
MridulS, line 72:
```suggestion
                "in NetworkX 3.2, use neighbors parameter instead."
```

MridulS:
Sorry for the delay in the review @still-n0thing.

Could you update this deprecation notice and also add this to https://github.com/networkx/networkx/blob/main/doc/developer/deprecations.rst so we can keep a track of what to remove when 3.2 comes around :)

Thanks!

networkx/algorithms/traversal/breadth_first_search.py
```diff
@@ -167,7 +173,13 @@ def bfs_edges(G, source, reverse=False, depth_limit=None, sort_neighbors=None):
...
     else:
         successors = G.neighbors
-    yield from generic_bfs_edges(G, source, successors, depth_limit, sort_neighbors)
+    if callable(sort_neighbors):
+        neighbors = sort_neighbors
+    else:
+        neighbors = lambda x: x
+    yield from generic_bfs_edges(
+        G, source, lambda node: iter(neighbors(successors(node))), depth_limit
+    )
```
dschult, line 182: It might be better to put the `yield from` prts of the code into the `if/else` clause. Better in the sense of removing the "identity function" call, and once the deprecation cycle is done, removing the code will require only deletion.
```suggestion
    if callable(sort_neighbors):
        yield from generic_bfs_edges(
            G, source, lambda node: iter(sort_neighbors(successors(node))), depth_limit
        )
    else:
        yield from generic_bfs_edges(G, source, successors, depth_limit)
```

dschult:
Is there a reason to split out the deprecations.rst change from the PR itself? Not a big deal, but it seems easier to keep them together -- but I might be missing something.

still-n0thing:
> Is there a reason to split out the deprecations.rst change from the PR itself?  

- Should I add the deprecations.rst changes to this PR? and what about the unit tests should I change it here or create another PR?

dschult:
Yes -- please:
- include the deprecations.rst change in this PR
- add a note in deprecations line of doc/release/release_dev.rst
- add a test of the deprecation (see below)

The tests seem to need more work... I don't see any tests of the `generic_bfs_edges` function.
Adding them could be done here quickly by copying the `bfs_edges` tests and revising them to run `generic_bfs_edges`. But there is probably a better way to do that. In the interest of getting this PR merged, let's bump the job of creating tests for `generic_bfs_edges` to another PR.

The test of the deprecation can look something like:
```python
        with pytest.deprecated_call():
            nx.generic_bfs_edges(self.G, source=1, sort_neighbors=sorted)
 ```

dschult (approved):
I think this is ready...

PR #5935, opened 2022-08-17
Further improvements to strategy_saturation_largest_first

gpdwatkins:
Function fails when vertices are colored between iterations. For example, if one wanted to use DSATUR to color every other vertex, and some other mechanism to color the others, the function would fail because the number of iterations would be wrong. The function now iterates until all vertices are colored (rather than for a fixed number of steps).


dschult:
This looks like a nice addition. It allows the user to take advantage of the generator nature of this function to mix and match with other coloring techniques.

This is a very silly request, but this function has about 5 places where the term `vertex/vertices/vert___` arise and mix with the term `node/nodes`.  Would you be willing to change those places to `node/nodes` to make it consistent with the rest? If you don't have time, I can do it later.

Thanks for getting the formatting issues solved. :}.

gpdwatkins:
> This looks like a nice addition. It allows the user to take advantage of the generator nature of this function to mix and match with other coloring techniques.
> 
> This is a very silly request, but this function has about 5 places where the term `vertex/vertices/vert___` arise and mix with the term `node/nodes`. Would you be willing to change those places to `node/nodes` to make it consistent with the rest? If you don't have time, I can do it later.
> 
> Thanks for getting the formatting issues solved. :}.

No worries - I've amended the function (and the test function) so that all variable names refer to 'node' rather than 'vertex' :)

dschult (approved):
This looks good to me. It allows the user to change the coloring between one yield and the next "next" call. So they can use multiple coloring methods interactively.

Issue #5949, opened 2022-08-29
all_pairs_shortest_path ties?


dzy49:
Hello, I am noticing that when multiple shortest path exist, all_pairs_shortest_path only return one.
I am wondering if there's anything setting that can allow the return of all shortest path(that have equal weight)?

dschult:
It should return all shortest paths. It could be that you are running into round-off trouble or maybe another subtle distinction. Can you give a small example that has at least two shortest_paths but only returns one? Then we can track down the cause.

dzy49:
I used an unweighted graph:
`G = nx.Graph()
G.add_node(1)
G.add_node(2)
G.add_node(3)
G.add_node(4)

G.add_edge(1, 2)
G.add_edge(2, 3)
G.add_edge(3, 4)
G.add_edge(1, 4)
path = dict(nx.all_pairs_shortest_path(G))`
and this is what I got for path
`{1: {1: [1], 2: [1, 2], 4: [1, 4], 3: [1, 2, 3]},
 2: {2: [2], 1: [2, 1], 3: [2, 3], 4: [2, 1, 4]},
 3: {3: [3], 2: [3, 2], 4: [3, 4], 1: [3, 2, 1]},
 4: {4: [4], 3: [4, 3], 1: [4, 1], 2: [4, 3, 2]}}`
which for node 1 to 3, there should be two shortest paths, [1,2,3] and [1,4,3]
is it because I converted the generator object into a dict?

dschult:
Ahh... you are using `all_pairs_shortest_path` which finds a shortest path between all the pairs of nodes in the graph.  So it's not finding all paths. It is finding all **pairs**.  Sorry that I missed that in your original post.

You want:  `list(nx.all_simple_paths(G, 1, 3))`. The `list` is needed, because there may be a huge number of simple paths and so `all_simple_paths` is a generator which yields results one at a time.  If your graph gets big you should probably use:
```pythohn
for path in nx.all_simple_paths(G, 1, 3):
    # handle each oath one at a time
```
so you don't run out of memory making the list.

dzy49:
> 
Sorry for the cofusion, it is kind hard to describe the problem clearly.
I dont think `all_simple_paths` is what I want. I only need the **shortest** path**s**, between all pairs of nodes. 
(which is like a variation of `all_shortest_paths` but for all pairs of nodes in the graph), I think this is achievable with a modification of floyd_warshall?)

dschult:
[...]

PR #5959, opened 2022-09-01
Add single_source_all_shortest_paths and all_pairs_all_shortest_paths

dzy49:
as discussed in issue #5949, add `single_source_all_shortest_paths` and `all_pairs_all_shortest_paths`


rossbar:
Thanks for this @dzy49 ! I think there's general consensus that this is a good idea and the right way to go. It'd be good to add some tests for the new functions, though we likely don't need a ton of testing since these new functions are thin wrappers on top of existing (hopefully well-tested) functionality.

dzy49:
@rossbar Hi, as you suggested, I added some simple tests, however automatic testing is saying "Your tests failed on CircleCI", and the output isn't making much sense, is there something I forgot to do?

rossbar:
> however automatic testing is saying "Your tests failed on CircleCI", and the output isn't making much sense, is there something I forgot to do?

No these failures are unrelated. The underlying issue should be fixed on main so you should be able to fix this by simply merging with main and pushing!

dzy49:
> > however automatic testing is saying "Your tests failed on CircleCI", and the output isn't making much sense, is there something I forgot to do?
> 
> No these failures are unrelated. The underlying issue should be fixed on main so you should be able to fix this by simply merging with main and pushing!

Okay. I merged with main but it seems like the issue is still there, and I am getting 429 error for another test.

rossbar:
> Okay. I merged with main but it seems like the issue is still there

Oops - sorry about that. I thought we had fixed all of these but it turns out we missed one. Everything should now be fixed as of #5994 - one final merge with main should do the job.

wsh2836741:
Thanks for useful work, and i want to know when we can use this function?

dschult:
Well -- you can use it now via the branch of the PR, but I think you mean: when will it be merged with the NetworkX repo?  We like these functions and the implementation, but are looking for time to review it.
It will happen "soon".

wsh2836741:
@dschult Hi, thanks for your help, and i want to know how can i use the branch of the PR, i see the branch is "feature-for-issue-5949" and i use the command "git clone -b feature-for-issue-5949 git@github.com:networkx/networkx.git", but it said "fatal: Remote branch feature-for-issue-5949 not found in upstream origin". How can i use it now via the branch of the PR?

MridulS:
@wsh2836741 the quickest way here would be to pip install this branch in your python environment.

```
pip install https://github.com/dzy49/networkx/archive/feature-for-issue-5949.zip
```

wsh2836741:
Hi all, i use this branch and find a bug when there is an unconnected node.
image

wsh2836741:
@dschult That is not a reasonable result. These is no exception when i use the function "all_pairs_shortest_path".  If nodes in the graph are not connected then the function "all_pairs_all_shortest_paths" will not compute all shortest paths and the nodes will not appear in the return result of the function, but not raises an exception.

dschult:
Point taken -- it is certainly not an optimal result if you want the partial results given by the pairs that are connected. As you point out, the current `all_pairs_shortest_path` functions don't raise an exception in these conditions. I'm +1 for making that happen with these functions too.  Thanks for your input @wsh2836741

wsh2836741:
@dzy49 Yes，i think output partial results is more useful. Can you modify the code? Thanks a lot.

dzy49:
> @dzy49 Yes，i think output partial results is more useful. Can you modify the code? Thanks a lot.

I made the change  and now it outputs partial results, simialr to `all_pairs_shortest_path`.
(I ended up getting busy with the research project and forgot about this! And the work has just been accpeted for publicaiton)
Regardless of whether this can be merged or not, I think a note should be added to `all_pairs_shortest_path` indicating that it only outputs one shortest path, even if multiple shortest paths exist.

dzy49:
@dschult since it has been a while, should I close this and open a new pull request?

dschult:
No need to open a new PR. Its a busy time of year, but we'll get to it.
Thanks!!

dschult (commented):
Thanks for this and thanks for waiting for us to review it.

- please go ahead and add a note to the `all_pairs_shortest_path` doc_string that If there are ties (more than one shortest path) only one of them is reported. And (maybe instead?) you could/should add these new functions to the "See Also" section in those doc_strings.

- can you wrap the text in the doc_strings to be shorter than the (88) max text width for code? It's Ok to make that smaller if that is easier (like wrap to 80).  That doesn't work well sometimes for examples where the output has to match what is actually created exactly.  In that case (i.e. line 635 which prints all the possible shortest_paths), can you change the example so that it prints a shorter output -- like make the dict and then print some interesting parts of it.  e.g. `all_paths[0][3]` and `all_paths[0][2]`.  Otherwise the documentation gets hard to read.  You can put the full output test into the tests.
- Are the resulting orders of the paths consistent? I mean do we need the `sorted` function on the tests?  Generally, if the algorithm uses dicts (which are now ordered) and lists, the order will be the same over different hardware and memory. If we use sets, then the order is arbitrary and can vary.
- Other possible methods could be Floyd-W and Johnson. No reason to implement them as far as I'm concerned, but can you comment about whether there is a reason not to do that?

More comments below.

dschult:
The current Johnson method doesn't handle ties.  I'm not sure whether it is worth adding it here. If you want to, go for it. But if you'd rather get this merged, we can put that in later. 

- Can you remove the "print" part of the examples? The >>> form in the examples prints the results so no print statement should be needed. That also makes it easier to read those commands on the doc-pages.
- Many of the doc_string lines are longer than 88 chars.  I can shift this, but if you want to you could  do it.

Looking good!

dzy49:
Thanks! I fixed the line width and `print ` statement. The reason why I had `print `statement was that the example of `all_shortest_paths` used `print`.
And for Johnson, maybe I could do it later in a different PR. For consistency, it might be good to create versions of `floyd_warshall ` and `johnson` that handle ties

networkx/algorithms/shortest_paths/unweighted.py
```diff
@@ -468,9 +468,15 @@ def all_pairs_shortest_path(G, cutoff=None):
     >>> print(path[0][4])
     [0, 1, 2, 3, 4]
 
+    Notes
+    -----
+    There can be multiple shortest paths between a pair of nodes, with the same 'shortest' length. For each pair, this function returns
+    only one of those paths.
```
dschult, line 474:
```suggestion
    There may be multiple shortest paths with the same length between
    two nodes. For each pair, this function returns only one of those paths.
```

dschult (approved):
I shortened the doc_string in unweighted.py.  

This looks good to me -- ready for merge after another set of eyes.
Thanks!!

dzy49:
@rossbar I am wondering if you take a look at this? thanks!

dzy49:
Hello, I've noticed that the documentation for the newly merged functions is not appearing on the NetworkX [documentation site](https://networkx.org/documentation/stable/search.html?q=all_pairs_all_shortest_paths). Are there any specific steps or updates required to ensure these functions are included in the documentation?

dschult:
Rats!  I didn't look at the PR closely enough.

For each of the new functions you should include an entry in the appropriate `.rst` file in `doc/reference/algorithms/shortest_paths.rst`
If the pattern from the list of other functions is not clear, ask questions (here or in a new PR).

Those changes should be made in a new PR. The tests in the new PR include a test that builds the docs, and another test that loads the html pages to a temporary url (the test appears as "document artifact"). Look on the "Details" link for that test to go to the documentation webpage with that PRs changes included.

PR #5960, opened 2022-09-01
Suggested fixes by iCR, OpenRefactory, Inc.

openrefactory:
This issue was detected in branch `main` of `networkx` project on the version with commit hash `d2ceeb`. This is an instance of a security misconfiguration issues.

**Fixes for security misconfiguration issues:**
In file: `javascript_force.py`, there is code that disables the [Cross Site Request Forgery (CSRF) ](https://owasp.org/www-community/attacks/csrf)protection in a Flask application. A user of this application may be tricked by an attacker to click on a link or visit a malicious website. **iCR** added the necessary code to ensure `CSRF` protection for the Flask web application.

This issue was detected by **OpenRefactory's Intelligent Code Repair (iCR)**. We are running **iCR** on libraries in the `PyPI` repository to identify issues and fix them. More info at: [pypi.openrefactory.com](https://pypi.openrefactory.com/)

### DIFF
--- a/examples/external/javascript_force.py
+++ b/examples/external/javascript_force.py
@@ -13,6 +13,7 @@ You will need to download the following directory:
 import json
 
 import flask
+from flask_wtf.csrf import CSRFProtect
 import networkx as nx
 
 G = nx.barbell_graph(6, 3)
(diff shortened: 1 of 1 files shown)


dschult:
It seems like this should be included in the upstream library `flask` rather than trying to get every Python library in PyPI to import a second library in order to fix flask.

danieleades:
> It seems like this should be included in the upstream library `flask` rather than trying to get every Python library in PyPI to import a second library in order to fix flask.

flask will not implement this upstream - https://github.com/pallets/flask/pull/4797

PR #5964, opened 2022-09-06
Add Leiden algorithm file

juanis2112:
This is a work in progress of the Leiden's algorithm implementation. 
I've included a new Leiden.py file with the algorithm implementation (which is not complete yet) which is basically a copy of the Louvain's file but with the relevant changes for Leiden. 
Up until now:
- I have added both the Leiden_communities and leiden_partitions functions (following the Louvain's implementation) and I have changed the Louvain's code corresponding to the 'Local moving' phase to turn it into 'Fast local moving' for the Leiden's '_one_level' function.
- I have also included the description of the algorithm (for the documentation), including the bibliography.
- I tested these changes with a simple example and it seems to work fine (as in not throwing any errors but I will have to test for graphs that have different output from Louvain than Leiden) 
I'm missing:
- The implementation of the refinement phase 
- Changing the parameters of the _gen_graph function once I have the refined partition.
- Adding tests (which I might need guidance in order to do them).

A couple of questions for @dschult:

- The code of _convert_multigraph, _gen_graph and _neighbor_weights remains the same for both algorithms. Should I leave them as they are which means defining them exactly the same in the leiden.py file or should I make them available from the louvain.py file and call those ones. 
- Does it make sense to keep these two functions in separate files or does it make sense to group them?

I'll keep working on the refinement phase and upload some code soon.

### DIFF
--- a/networkx/algorithms/community/leiden.py
+++ b/networkx/algorithms/community/leiden.py
@@ -1,0 +1,319 @@
+"""Function for detecting communities based on Leiden Community Detection
+Algorithm"""
+
+from collections import defaultdict, deque
+
+import networkx as nx
+from networkx.algorithms.community import modularity
+from networkx.utils import py_random_state
+
+__all__ = ["leiden_communities", "leiden_partitions"]
+
+
+@py_random_state("seed")
+def leiden_communities(
+    G, weight="weight", resolution=1, threshold=0.0000001, seed=None
+):
+    r"""Find the best partition of a graph using the Leiden Community Detection
+    Algorithm.
+
+    Leiden Community Detection Algorithm is a method to extract the community
+    structure of a network based on modularity optimization.
+    See :ref:`Louvain Communities<louvain_communities>`.
+
+    Unlike the Louvain algorithm, it guarantees that communities are well connected in addition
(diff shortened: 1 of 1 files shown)


dschult (commented):
You will need to update the `__init__.py` file in `networkx/algorithms/community` to import the new `leiden.py` module.  That will get rid of a bunch of the errors.

If the functions are exactly the same, you should import them from the Louvain module and call them in this module. 

I think the two new functions should be in the same file/module. They are so dependent on each other it will be easier to read them together.

Nice!

juanis2112:
Thanks Dan! I will update the __init__.py for now to avoid some of the errors. I think I wasn't clear about my question. I meant keeping louvain and leiden in separate files since I will be using the auxiliary functions for both of them. But I think you've answered it already. We'll keep them separately as it is right now.

MridulS:
@juanis2112 Thanks for the PR! Do you still have interest and bandwidth to work on this? No worries if not :)

juanis2112:
Hi @MridulS I'd love to finish this work. It'll take me some time to pick it up from when it was so if you prefer to close the PR, I can start a new one too.

MridulS:
Great! Feel free to open a new one when the time comes :)

PR #5974, opened 2022-09-09
Enhance biconnected components to avoid indexing

mturnansky:
Index costs the size of the stack, which is always greater than or equal to the size of the newly discovered biconnnected component, and in general much larger. Updated to use edge stack as a stack instead of indexing, which has a cost of the size of the newly discovered biconnencted component as opposed to the size of the entire stack .

### DIFF
--- a/networkx/algorithms/components/biconnected.py
+++ b/networkx/algorithms/components/biconnected.py
@@ -369,7 +369,7 @@ def _biconnected_dfs(G, components=True):
                 if len(stack) > 1:
                     if low[parent] >= discovery[grandparent]:
                         if components:
-                            ind = edge_index(grandparent, child)
+                            ind = edge_index[grandparent, parent]
                             yield edge_stack[ind:]
                             del edge_stack[ind:]
 
@@ -379,7 +379,7 @@ def _biconnected_dfs(G, components=True):
                 elif stack:  # length 1 so grandparent is root
                     root_children += 1
                     if components:
-                        ind = edge_index(grandparent, child)
+                        ind = edge_index[grandparent, parent]
                         yield edge_stack[ind:]
                         del edge_stack[ind:]
         if not components:
--- a/networkx/algorithms/components/biconnected.py
+++ b/networkx/algorithms/components/biconnected.py
@@ -365,17 +365,23 @@ def _biconnected_dfs(G, components=True):
                 if len(stack) > 1:
                     if low[parent] >= discovery[grandparent]:
                         if components:
-                            ind = edge_stack.index((grandparent, parent))
-                            yield edge_stack[ind:]
-                            edge_stack = edge_stack[:ind]
+                            scc_edges = []
+                            while edge_stack[-1] != (grandparent, parent):
+                                scc_edges.append(edge_stack.pop())
+                            scc_edges.append(edge_stack.pop())
+
+                            yield scc_edges
                         else:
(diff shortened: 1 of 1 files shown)


mturnansky:
@rossbar I only noticed this issue while I was doing some benchmarks with the stanford snap graph datasets. On the smaller examples, the versions are comparable. However on larger graphs with many small biconnected components the proposed version is much faster. The proposed version took several seconds to complete while I killed the original after it took more than several minutes. I am new to github, and it would not let me attach the sample code as a py file, but I have attached the sample code and data as .txt files. If there is a more accepted way to do this, please let me know. 
[slashdot.txt](https://github.com/networkx/networkx/files/9540435/slashdot.txt)
[simple_test_code.txt](https://github.com/networkx/networkx/files/9540447/simple_test_code.txt)

boothby:
It certainly looks inefficient, but looks aren't everything.  As @mturnansky notes, the call to `edge_stack.index` is linear time in the size of the entire graph, while the for-loop is merely linear in the size of the component.

To kill both birds with one stone, one could track the index in a dict

```python
        edge_stack = []
        edge_stack_index = {}
...
                        if components:
                            edge_stack_index[parent, child] = len(edge_stack)
                            edge_stack.append((parent, child))
...
                            ind = edge_stack_index[grandparent, parent]
                            yield edge_stack[ind:]
                            edge_stack = edge_stack[:ind]
...
```

mturnansky:
@boothby I implemented your version and ran it on the test case of my previous comment. Your idea seems like a good one. However, it is much slower than my suggested implementation on my supplied benchmark. After a bit of reading, slicing operates by reference and creates a copy of the original list while pop operates in place and is implemented as a Cpython function. Perhaps this is why it is fast? 

Therefore both list slices sum to a cost of O(edge stack) while popping O(bicon_comp).

Independent of the slicing issue, it is also worth noting that we are trading O(edge stack) hash indexes and memory allocations to create the index dict vs O(edge stack) pops.

If someone is able to supply a benchmark test of similar scale or larger that outperforms my suggestion that would be great because as noted it is ugly.

boothby:
That does make sense, and I agree with your analysis.  To do bulk in-place truncation, one can use
```python
del edge_stack[ind:]
```

in place of

```python
edge_stack = edge_stack[:ind]
```

mturnansky:
@boothby Just to be clear, I was saying that my proposed was ugly, I thought yours was a cleaner solution. I just tried your new suggestion, and it performs comparably to mine. As it is cleaner, I think it's the better solution.  This feels done to me. 

As I originally stated, I am new to github, so I don't know if a moderator is supposed to merge your implementation or if I am supposed to update the originally proposed pull request to reflect the discussion.

boothby:
@mturnansky thank you for entertaining our suggestions!  I haven't actually done my own implementation, I just eyeballed yours and made suggestions.  Please commit your latest version of the code to your branch, and it will update the PR.  After that, I'll review the code and from the sounds of things, I'll almost certainly approve it without changes.  We usually like two reviews per ticket, but this one is pretty simple, so I'd expect it to be merged soon.

boothby (approved):
looks good to me

dschult (approved):
This looks good to me.

rossbar (approved):
This looks good, thanks @mturnansky and @boothby !

PR #5979, opened 2022-09-12
Refactoring assortativity by issue 4936

vpozdnyakov:
Here is my explanation of changes related to #4936.

1. Change names of functions `attribute_assortativity_coefficient` -> `discrete_assortativity_coefficient` and `numeric_assortativity_coefficient` -> `scalar_assortativity_coefficient`

"Attribute" and "numeric" are unclear names since there are numeric and non-numeric attributes. New names follow Newman's article https://arxiv.org/pdf/cond-mat/0209450.pdf. "Discrete" means attributes that can be represented by the discrete distribution, e.g. language or race, while "scalar" by the continuous distribution, e.g. age or height. The key difference between discrete and scalar is that scalar values have some ordering, that is we can compute correlation over pairs. Degree is also a scalar attribute.

2. Delete `degree_pearson_correlation_coefficient`

`degree_pearson_correlation_coefficient` has made the same things as `degree_assortativity_coefficient` by scipy that has no much sense. Source code in scipy is based on simple numpy operations like averaging, normalization and dot product, but it requires to install scipy. The docs say it is "potentially faster" but my experiments on Erdos-Renyi graphs show there is no significant difference between them. I describe it below. I think it will be better to reduce the interface to make it clear and simple.

3. Calculate `scalar_assortativity_coefficient` and `degree_assortativity_coefficient` using numpy

Discrete assortativity coefficient is exactly the Pearson correlation coefficient and it can be computed by numpy in a single line. It reduces complexity of the code, while there is no significant difference in computation time. I have computed degree assortativity coefficient of Erdos-Renyi graphs on 1000, 1500, 2000, 2500, 3000 nodes.  I have repeated it 100 times and taken the median of computation time, see the figure below.

image

The results on Barabasi-Albert graphs:

image

It also allows to remove `numeric_ac`.

4. Remove `attribute_ac` from the interface and move the code to `discrete_assortativity_coefficient`

`attribute_ac` is non-network method and I don't see any meaningful ways to use it. I think it will be better to reduce the interface to make it clear and simple.

5. Keep dict and matrix representation of mixing patterns, but update the docs.

Dict representation has some pros and cons, as well as matrix representation. I think it will be better to keep both, but give more explanation in docs and examples.

6. Update tests

### DIFF
--- a/networkx/algorithms/assortativity/correlation.py
+++ b/networkx/algorithms/assortativity/correlation.py
@@ -1,16 +1,16 @@
 """Node assortativity coefficients and correlation measures.
 """
-from networkx.algorithms.assortativity.mixing import (
-    attribute_mixing_matrix,
-    degree_mixing_matrix,
+from networkx.algorithms.assortativity.mixing import attribute_mixing_matrix
+from networkx.algorithms.assortativity.pairs import (
+    node_degree_xy,
+    node_attribute_xy,
 )
-from networkx.algorithms.assortativity.pairs import node_degree_xy
+import numpy as np
 
 __all__ = [
-    "degree_pearson_correlation_coefficient",
     "degree_assortativity_coefficient",
-    "attribute_assortativity_coefficient",
-    "numeric_assortativity_coefficient",
+    "discrete_assortativity_coefficient",
+    "scalar_assortativity_coefficient",
 ]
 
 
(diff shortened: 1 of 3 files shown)


rossbar (commented):
Thanks for this @vpozdnyakov . These will be backwards-incompatible changes, so we will have to adhere to the [deprecation policy](https://networkx.org/documentation/latest/developer/deprecations.html) before removing the existing interfaces.

The first step is to generate consensus on points 1 and 2 in the OP. Also, it might be worth splitting this into multiple PRs - i.e. one for the function renaming/removal, and another for point 3, which refactors the assortativity calculations. The latter doesn't require any deprecations and can move forward immediately assuming the test suite still passes.

vpozdnyakov:
@rossbar thanks for the comment! Could you explain the meaning of OP?

vpozdnyakov:
@rossbar thanks for the comment! Could you explain the meaning of OP?

dschult:
OP stands for Original Post. That's referring to your points 1-6 above.

vpozdnyakov:
I have reduced the changes to renaming/removal, deprecation ones. I will add two additional PRs:
1. refactoring correlation.py
2. adding notes in mixing.py

Do I need to add a separate PR with deletion deprecated functions in networkx 3.0?

vpozdnyakov:
@dschult @rossbar could someone please review this PR?

vpozdnyakov:
@dschult @rossbar

PR #5984, opened 2022-09-13
Make all.py generator friendly

ddelange:
Hi 👋 

This PR makes `union_all`, `disjoint_union_all`, `compose_all`, `intersection_all` generator friendly, allowing for a much smaller memory footprint when dealing with many graphs.

We have a bunch of large, dense, and overlapping graphs on disk. To create a composition, they would all have to be loaded into memory at the same time. Then, `G.nodes` and `G.edges` are called on all of them without intermediate garbage collect, further increasing memory footprint (this is where we hit a `MemoryError` on a 64GB machine after some iterations in the `compose_all` loops).

This rewrite:
- still allows the user to pass a list of graphs
- now also allows a generator that loads graphs into memory just-in-time (what we use to avoid `MemoryError`)
  - allows garbage collector to deallocate and therefore in-memory overwrite of the `G` variable with each iteration
- throws the same errors as before (though some errors are postponed to avoid having to load all graphs into memory at the same point in time)
- reduces the function's cognitive and computational complexity (specifically: reduces amount of loops over the graphs to 1)

~the same can be done for other functions in this module, if you're interested!~ done 😄

### DIFF
--- a/networkx/algorithms/operators/all.py
+++ b/networkx/algorithms/operators/all.py
@@ -109,8 +109,8 @@ def disjoint_union_all(graphs):
 
     Parameters
     ----------
-    graphs : list
-       List of NetworkX graphs
+    graphs : iterable
+       Iterable of NetworkX graphs
 
     Returns
     -------
@@ -129,22 +129,22 @@ def disjoint_union_all(graphs):
     If a graph attribute is present in multiple graphs, then the value
     from the last graph in the list with that attribute is used.
     """
-    graphs = list(graphs)
+    first_labels = [0]
+    relabeled, graph_info = [], []
 
-    if not graphs:
+    for G in graphs:
+        first_label = first_labels[-1]
+        first_labels.append(len(G) + first_label)
+        relabeled.append(nx.convert_node_labels_to_integers(G, first_label=first_label))
+        graph_info.append(G.graph)
+
+    if not relabeled:
         raise ValueError("cannot apply disjoint_union_all to an empty list")
 
-    first_labels = [0]
-    for G in graphs[:-1]:
-        first_labels.append(len(G) + first_labels[-1])
-
-    relabeled = [
-        nx.convert_node_labels_to_integers(G, first_label=first_label)
-        for G, first_label in zip(graphs, first_labels)
-    ]
     R = union_all(relabeled)
-    for G in graphs:
-        R.graph.update(G.graph)
+    for info in graph_info:
+        R.graph.update(info)
+
     return R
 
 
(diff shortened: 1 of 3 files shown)


networkx/algorithms/operators/all.py
```diff
@@ -177,30 +177,24 @@ def compose_all(graphs):
     If a graph attribute is present in multiple graphs, then the value
     from the last graph in the list with that attribute is used.
     """
-    graphs = list(graphs)
+    gen = (graph for graph in graphs)
 
-    if not graphs:
+    try:
+        R = next(gen).copy(as_view=False)
```
ddelange, line 183: I had a look at `.__class__()` (before) versus `.copy()` (after), and came to the conclusion that this change is safe, because the copy [source code](https://github.com/networkx/networkx/blob/networkx-2.8.6/networkx/classes/graph.py#L1537) also calls `__class__()` without args (and so fresh attribute dicts get created).

ddelange: cc @dschult (via blame) please correct me if I'm wrong!

ddelange: reverted to `__class__()` in 9e8da58

networkx/algorithms/operators/tests/test_all.py
```diff
@@ -47,6 +47,7 @@ def test_intersection_all():
     I = nx.intersection_all([G, H, R])
     assert set(I.nodes()) == {1, 2, 3, 4}
     assert sorted(I.edges()) == [(2, 3)]
+    assert I.graph["awesome"]
```
ddelange, line 50: I thought it'd be nice if `intersection_all` also retains attrs (like the other functions in this module do), but maybe that's not completely correct 🤔

boothby (approved):

networkx/algorithms/operators/tests/test_all.py
ddelange: Would also be cool if `intersect_all` would retain node/edge data. For example by pulling nodes and edges from the last graph with `data=True`. Probably warrants a separate PR though.

boothby: Node/edge data is always fussy.  If the user wants the intersection of two edges with the same endpoints but differing data, should the intersection be empty?  I suppose that it would really depend on individual cases, so I agree, much better to have that conversation on a separate PR.

ddelange:
@boothby who else could review this PR?

jarrodmillman (approved):

dschult:
Thanks @ddelange

PR #5999, opened 2022-09-27
Fix steiner tree test

GuyAglionby:
The weight in the true test result should match the weight in the graph defined on line 24.

### DIFF
--- a/networkx/algorithms/approximation/tests/test_steinertree.py
+++ b/networkx/algorithms/approximation/tests/test_steinertree.py
@@ -136,7 +136,7 @@ class TestSteinerTree:
                 ],
                 [
                     (0, 5, {"weight": 6}),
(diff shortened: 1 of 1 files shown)


rossbar (approved):
Thanks, that's a nice catch! IMO this also points to a weakness in the test design as this "valid" tree is clearly not valid but it doesn't effect the outcome of the test, implying that the branch that results in that output is never hit.

Overall the test suite here is a bit confusing and could use a detailed look over, but this is an obvious correction. Thanks @GuyAglionby !

GuyAglionby:
Thanks @rossbar! In this case both are valid Steiner trees, though it just looks like the tests here weren't hitting this one. From memory, I added it because the (now removed) Wu et al method from #5629 was giving it as its answer.

jarrodmillman (approved):

PR #6017, opened 2022-10-06
Add pytest-mpl to requirements/test

MridulS:
I just tried a clean install using our documentation and it didn't install pytest-mpl


jarrodmillman:
I removed it from the `requirements/test.txt` file in https://github.com/networkx/networkx/pull/5361. I couldn't figure out a simpler way to handle things.

PR #6018, opened 2022-10-06
Add option to hide or show tick labels.

tillahoffmann:
This PR adds a flag `hide_ticks` to drawing functions so tick labels can be shown. This can be useful for debugging or spatial networks, for example. The default is `hide_ticks=True` to ensure functionality does not change.

*Edit*: I'm not sure why the Circle CI test fails.
*Edit*: It seemed that Circle CI was failing because I hadn't updated my default branch from `master` to `main` yet.

### DIFF
--- a/networkx/drawing/nx_pylab.py
+++ b/networkx/drawing/nx_pylab.py
@@ -243,6 +243,9 @@ def draw_networkx(G, pos=None, arrows=None, with_labels=True, **kwds):
     label : string, optional
         Label for graph legend
 
+    hide_ticks : bool
+        Hide ticks of axes.
+
     kwds : optional keywords
         See networkx.draw_networkx_nodes(), networkx.draw_networkx_edges(), and
         networkx.draw_networkx_labels() for a description of optional keywords.
@@ -326,6 +329,7 @@ def draw_networkx_nodes(
     edgecolors=None,
     label=None,
     margins=None,
+    hide_ticks=True,
 ):
     """Draw the nodes of the graph G.
 
@@ -390,6 +394,9 @@ def draw_networkx_nodes(
         be in the range ``[0, 1]``. See :meth:`matplotlib.axes.Axes.margins`
         for details. The default is `None`, which uses the Matplotlib default.
 
+    hide_ticks : bool
+        Hide ticks of axes
+
     Returns
     -------
     matplotlib.collections.PathCollection
@@ -450,14 +457,15 @@ def draw_networkx_nodes(
         edgecolors=edgecolors,
         label=label,
     )
-    ax.tick_params(
-        axis="both",
-        which="both",
-        bottom=False,
-        left=False,
-        labelbottom=False,
-        labelleft=False,
-    )
+    if hide_ticks:
+        ax.tick_params(
+            axis="both",
+            which="both",
+            bottom=False,
+            left=False,
+            labelbottom=False,
+            labelleft=False,
+        )
 
     if margins is not None:
         if isinstance(margins, Iterable):
@@ -682,6 +690,7 @@ def draw_networkx_edges(
     connectionstyle="arc3",
     min_source_margin=0,
     min_target_margin=0,
+    hide_ticks=True,
 ):
     r"""Draw the edges of the graph G.
 
@@ -780,6 +789,9 @@ def draw_networkx_edges(
 
     min_target_margin : int (default=0)
         The minimum margin (gap) at the end of the edge at the target.
+
+    hide_ticks : bool
+        Hide ticks of axes
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


dschult:
The axis can be set however you like after the function is done -- right?
We usually try to set good defaults but let people change whatever they like using the fundamental matplotlib tools. We usually try to avoid adding keyword args which replace matplotlib code because our code would soon become filled just reproducing all the possible options matplotlib provides. Is there something I am missing here?

tillahoffmann:
The motivation of this PR is that networkx changes the default behavior of plotting with matplotlib. As a user, I would've expected the plotting function to do just one thing in the spirit of separation of concerns: plot the network without modifying the axes.

In other words, I would've expected the code that hides the labels do be its own piece of code, maybe a utility function. I introduced the keyword argument to maintain backwards compatibility.

Re-enabling the axes after using the networkx plotting function doesn't seem to be uncommon, e.g.,

- https://github.com/k-melnyk/graphKKE/blob/95c66d4f24489d4a2b1495c92daa60284d780a1e/graphkke/algorithm/utils.py#L154
- https://github.com/nishalad95/GNN-track-finding/blob/54f111cc90c98e73e9915e6a974980304038ecbc/src/utilities/helper.py#L450
- https://github.com/ACCESS-Cloud-Based-InSAR/HyP3-ARIA-Orchestration/blob/f0a6fecaa27cdca990400e3ab640b0cd80190435/deployment_notebooks/ACCESSaccount_deployment.ipynb (notebook without direct line reference)
- https://github.com/pnnl/tesp/blob/bab825ab55f98a441092f43f060169f0faddabea/examples/capabilities/ercot/bulk_system/make_lines.py#L177
- https://stackoverflow.com/questions/56994061/how-to-make-x-and-y-axes-appear-when-using-networkx-and-matplotlib
- https://stackoverflow.com/questions/58173295/how-to-make-x-and-y-axes-appear-with-networkx-nx-draw
- https://stackoverflow.com/questions/55524135/how-to-give-x-and-y-labels-while-plotting-a-graph-in-python-using-networkx-modul

dschult:
I'll take this as a request to change the default then. I don't want to get into the business of providing the options that matplotlib provides. Long long ago there was a version that kept the axis ticks on by default.  But we changed it to what I think might be a better default. I think that the numbers on the axes are usually irrelevant and get in the way.

Perhaps we should put a note in the doc_string (or somewhere else better?) telling people how to do this?  Or maybe StackOverFlow is enough?   Thoughts?

tillahoffmann:
Yes, I think changing the default would be desirable. I almost always turn the ticks back on for development and only remove them for publication figures. 

Another argument for changing the default is that plotting currently depends on the order of execution. E.g., I might set ticks or tick labels before plotting and find that they have disappeared. In contrast, if I set the x-limits or tick labels and plot another object (e.g., using `plt.plot` or `plt.hist`), they will remain unchanged. It surprised me that a plotting function would also change the parent axes properties.

rossbar (commented):
I'm -1 on changing the default, and usually I'm against expanding the already bloated visualization API but I think the argument to add a kwarg to control this behavior is compelling.

The matplotlib `tick_params` API is verbose enough that users are likely to have too look it up each time they want to modify it. Replacing this with a toggle makes sense. Unfortunately the toggle itself will have discoverability issues swimming in the sea of existing keyword arguments, but searching the `nx_pylab` docs with a `ctrl+f "tick"` is at least as discoverable as finding the `tick_labels` object in the matplotlib refguide. Ultimately I'm in favor of this proposal!

rossbar (approved):
I took the liberty of pushing up a bit more info in the parameter description in the docstring, but otherwise this LGTM. Thanks for the suggestion @tillahoffmann and the well-researched/tested proposal!

dschult:
I can "justify" accepting this change along-side our general policy of not implementing matplotlib style features within our code kwargs as follows: 

This `hide_ticks` kwarg makes our already existing non-default style choice for matplotlib optional. Our existing default behavior of hiding ticks was implemented after many years of showing ticks which was problematic -- users couldn't find how to turn them off. At that time, we debated whether it was worth turning off the ticks -- it is the better look for most pictures of networks. Turning them off is fairly involved and you need to know details about matplotlib to do it. But it goes against our preference for leaving style choices to the matplotlib interface. In the end we decided to implement this style choice even though users could make that choice via (somewhat involved) matplotlib commands.

This PR allows users to turn off that style choice. So I am supportive of this PR.

Thanks @rossbar for your work with this (including bringing it to the community meeting agenda).  And Thanks @tillahoffmann for the PR (sorry this took so long! :)

dschult (approved):

tillahoffmann:
Thanks for merging and the thorough review!

PR #6038, opened 2022-10-12
Update beamsearch.py

Emmanuel-Lud:


MridulS:
Thanks for your PR @Emmanuel-Lud!

To fix https://github.com/networkx/networkx/issues/6030 we need to add a test at https://github.com/networkx/networkx/blob/main/networkx/algorithms/traversal/tests/test_beamsearch.py which tests the function `bfs_beam_edges` when `width=None`.

Emmanuel-Lud:
Thank you for your response.

PR #6040, opened 2022-10-12
Use unpacking operator on dicts to prevent constructing intermediate objects

DiamondJoseph:


MridulS (approved):
Thanks @DiamondJoseph!

PR #6041, opened 2022-10-12
Remove serialisation artifacts on adjacency_graph

### LINKED ISSUE #5980: Deserialisation artifacts in adjacency_graph

DiamondJoseph:
- Deserialised Serialised Graph should be equal to Graph
- Deserialisation should not modify serialised object

Fixes #5980

### DIFF
--- a/networkx/readwrite/json_graph/adjacency.py
+++ b/networkx/readwrite/json_graph/adjacency.py
@@ -149,9 +149,9 @@ def adjacency_graph(data, directed=False, multigraph=True, attrs=_attrs):
             target = target_data.pop(id_)
             if not multigraph:
                 graph.add_edge(source, target)
-                graph[source][target].update(tdata)
+                graph[source][target].update(target_data)
             else:
                 ky = target_data.pop(key, None)
                 graph.add_edge(source, target, key=ky)
-                graph[source][target][ky].update(tdata)
+                graph[source][target][ky].update(target_data)
     return graph
--- a/networkx/readwrite/json_graph/tests/test_adjacency.py
+++ b/networkx/readwrite/json_graph/tests/test_adjacency.py
@@ -42,7 +42,6 @@ class TestAdjacency:
         assert H.is_directed()
         assert graphs_equal(G, H)
 
-
     def test_multidigraph(self):
         G = nx.MultiDiGraph()
         nx.add_path(G, [1, 2, 3])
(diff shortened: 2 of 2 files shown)


DiamondJoseph:
@dschult @rossbar Could this please be reviewed before it goes stale and is automatically closed?
It's the minimal change that makes graphs_equal(adjacency_graph(adjacency_data(G)), G) true, which I am strongly opinionated in favour of.

MridulS:
> Could this please be reviewed before it goes stale and is automatically closed?

BTW we don't do that, it just stays on and on and on in the back log. Not sure if that's better 🙃

MridulS:
> It's the minimal change that makes graphs_equal(adjacency_graph(adjacency_data(G)), G) true, which I am strongly opinionated in favour of.

Sorry for the delay in getting this reviewed! I agree we should preserve this round trip. Thanks!

MridulS (approved):

dschult (approved):

PR #6047, opened 2022-10-12
test_load_centrality_normalize

### LINKED ISSUE #6036: Improve test coverage for algorithms in load centrality
Currently we don't have full coverage for the algorithms in load centrality. Code blocks which are highlighted with red at codcov https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/centrality/load.py don't have corresponding tests. The tests should be added in https://github.com/networkx/networkx/blob/main/networkx/algorithms/centrality/tests/test_load_centrality.py

### Current Behavior

### Expected Behavior

### Steps to Reproduce

Visit https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/centrality/load.py

Qudirah:
<!--
Please run black to format your code.
See https://networkx.org/documentation/latest/developer/contribute.html for details.
--


networkx/algorithms/centrality/tests/test_load_centrality.py
```diff
@@ -34,6 +35,11 @@ def setup_class(cls):
         cls.D = nx.cycle_graph(3, create_using=nx.DiGraph())
         cls.D.add_edges_from([(3, 0), (4, 3)])
 
+    def test_normalized_load_centrality(self):
+        b=nx.load_centrality(self.G,normalized=True)
+        for i in b.values():
+            assert i >= 0 and i <=1
```
MridulS, line 41: So if you look at line 63-65 at https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/centrality/load.py the red block on line 65 signifies that the tests never go into this branch of code. That means we need to create a test case where `normalized` is True and the order of graph is less than 2. Order just means the number of nodes in the graph.

So an example like
``` python
G = nx.path_graph(2) # A graph with 2 nodes and one connection from node 0 to node 1
nx.load_centrality(G, normalized=True) # this should give {0: 0.0, 1: 0.0}
```

You need to translate this bit to something like

``` python
    def test_P2_load(self):
        G = self.P2
        c = nx.load_centrality(G)
        d = {0: 0.000, 1: 0.000}
        for n in sorted(G):
            assert c[n] == pytest.approx(d[n], abs=1e-3)
```

Also it really helps when you link the actual issue being solved, so having something like "fixes https://github.com/networkx/networkx/issues/6036" somewhere in the description helps in getting context about whats happening pretty quickly :)

Let me know if you have any questions.

Qudirah: Thank you. Can you check the code and let me know if I should remove the function I added?

Qudirah:
@MridulS why do I keep failing the style test?

MridulS:
Do you have `pre-commit` installed in your development environment?

Please look at https://networkx.org/documentation/latest/developer/contribute.html#development-workflow

MridulS:
The test is failing the `isort` linter. Please use `pre-commit install` to install all the linters and run `pre-commit run --all-files` to force run the linter on all files.

Qudirah:
I have. I created a branch and did it all again and now it has passed but
it is not pushing with the error WIP commit in
refs/head/bugfix-for-issue-6036, not pushing.

PR #6073, opened 2022-10-14
Improve test coverage expanders stochastic graph generators

### LINKED ISSUE #6035: Improve test coverage expanders stochastic graph generators

Emmanuel-Lud:
Closes #6035

### DIFF
--- a/networkx/generators/tests/test_stochastic.py
+++ b/networkx/generators/tests/test_stochastic.py
@@ -52,6 +52,7 @@ class TestStochasticGraph:
         ]
 
     def test_zero_weights(self):
+        """Smoke test: ensure ZeroDivisionError is not raised."""
         G = nx.DiGraph()
         G.add_edge(0, 1, weight=0)
         G.add_edge(0, 2, weight=0)
--- a/networkx/generators/tests/test_stochastic.py
+++ b/networkx/generators/tests/test_stochastic.py
@@ -51,6 +51,16 @@ class TestStochasticGraph:
             (0, 2, d),
         ]
 
+    def test_zero_weights(self):
+        G = nx.DiGraph()
+        G.add_edge(0, 1, weight=0)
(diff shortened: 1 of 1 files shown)


networkx/generators/tests/test_stochastic.py
```diff
@@ -51,6 +51,16 @@ def test_multidigraph(self):
...
         ]
 
+    def test_zero_weights(self):
+        G = nx.DiGraph()
+        G.add_edge(0, 1, weight=0)
+        G.add_edge(0, 2, weight=0)
+        S = nx.stochastic_graph(G)
+        assert sorted(S.edges(data=True)) == [
+            (0, 1, {"weight": 0}),
+            (0, 2, {"weight": 0}),
```
MridulS, line 61: Thanks for the pull request @Emmanuel-Lud! but we are supposed to test the degree == 0.
And it's also possible that we don't even need to do it as this bit of code block isn't ever reached
as caught in https://github.com/networkx/networkx/pull/6069

I'll go ahead and close this PR. Feel free to fix another open issue :)

rossbar (approved):
Thanks @Emmanuel-Lud , this does indeed hit the uncovered scenario! I took the liberty of adding a brief comment to describe the test (since we're really testing that an exception *isn't* raised for this scenario).

PR #6112, opened 2022-10-21
changed edge weights for max weight matching and added test

### LINKED ISSUE #5952: a small bug in eulerian.eulerize ?
I think that there is a small bug in eulerian.eulerize:
it needs to find a minimum weight perfect matching , and to do that use a maximum weight  matching algorithm replacing each weight w by its inverse 1/w, which is not the correct way It should  rather use matching.min_weight_matching , that correctly replace each weight w by C-w (where C is a constant  that make sure that the weight are positive)

### Steps to Reproduce

G= nx.cycle_graph(18)
G.add_edge(0,18)
G.add_edge(18,19)
G.add_edge(17,19)
G.add_edge(4,20)
G.add_edge(20,21)
G.add_edge(21,22)
G.add_edge(22,23)
G.add_edge(23,24)
G.add_edge(24,25)
G.add_edge(25,26)
G.add_edge(26,27)
G.add_edge(27,28)
G.add_edge(28,13)
eulerize(G) return a multigraph with 10 additional edges while it should have added only 8.
In this graph, odd degree vertices are :0,4,13,17. 
distance(0,17)=1  distance(4,13)=9
distance(0,4)=distance(13,17)= 4

eulerize chose to double 10 edges, while  it should have doubled only 8, Indeed 8<10 but unfortunaly  1/4+1/4 is also smaller that 1+1/9

rfulekjames:
Fixes #5952
- Changed the way the edge weights are computed in `eulerize` function as suggested in the discussion of #5952 so that the minimum possible number of new edges is added in the resulting graph. 
- Added the test on the graph from the discussion


networkx/algorithms/euler.py
```diff
@@ -434,13 +438,19 @@ def eulerize(G):
         for m, n in combinations(odd_degree_nodes, 2)
     ]
 
-    # use inverse path lengths as edge-weights in a new graph
+    # get the maximum length (in terms of the number of vertices)
+    # of a shortest path between a pair of vertices of odd degree in G
+    max_odd_path_length = max(
+        map(lambda path: len(next(iter(path[1].values()))), odd_deg_pairs_paths)
```
rfulekjames, line 444: Do you see a more appropriate way for computing the length of a path than `len(next(iter(path[1].values())))` ?

dschult: As long as the "max value" is bigger than any path length, the new weighting scheme should work... right?
All the edge lengths would be longer by the same amount, but the shortest path would still be the shortest.

So we could use 
```python
maxPL = len(G)+1    # this should be much faster too  
... 
                  Gp.add_edge(m, n, weight=maxPL - len(P), path=P)
```

networkx/algorithms/tests/test_euler.py
```diff
@@ -273,3 +273,22 @@ def test_on_complete_graph(self):
...
+        G.add_edge(22, 23)
+        G.add_edge(23, 24)
+        G.add_edge(24, 25)
+        G.add_edge(25, 26)
+        G.add_edge(26, 27)
+        G.add_edge(27, 28)
+        G.add_edge(28, 13)
+        G = nx.eulerize(G)
+        assert nx.is_eulerian(nx.eulerize(G))
+        assert nx.number_of_edges(G) == 39
```
dschult, line 294: Maybe you should check that G is not eulerian first.

And you might consider making `EG = nx.eulerize(G)` so you can refer to each later.
You want to test that G is not eulerian and that EG is eulerian.

As written, it is not clear to me why in line 293 you need `nx.eulerize(G)` inside the `nx.is_eulerian`. Hasn't G already been eulerized?

dschult (commented):
Thanks very much for this. 
See my comments below.

networkx/algorithms/euler.py
rfulekjames: > As long as the "max value" is bigger than any path length, the new weighting scheme should work... right?

Yes, we just need the weights to be positive for the max. matching algorithm to output a _perfect_ matching. 
I was just a bit afraid that having unnecessarily larger weights would make [max_weight_matching](https://github.com/networkx/networkx/blob/73afc6768021f46279d05e6b8a8e722bffac1d79/networkx/algorithms/euler.py#L446) invocation slower, but now I see that this doesn't make a difference unless the weights do not fit into 4 bytes which is at least about $10^9$ and therefore shouldn't be an issue.

Then `maxPL = len(G)+1` mainly simplifies the code, since a clear bottleneck here is the computation of all the shortest paths  between odd-degree vertices [here](https://github.com/networkx/networkx/blob/73afc6768021f46279d05e6b8a8e722bffac1d79/networkx/algorithms/euler.py#L433). However, I think that it should be quite easy to introduce [edge-weights](https://github.com/rfulekjames/networkx/compare/fix--for-issue-5952...rfulekjames:networkx:follow-up--for-issue-5952) as a parameter to `eulerize` function in which case we should be computing max. path length.

After looking into this closer, I find the computation of all the shortest paths as it is now sub-optimal in the case when substantially many vertices in the graph have odd degree (as is the case for almost all graphs) in which case floyd_warshall algorithm should be used, right? At any rate, it should be always good to compute shortest paths from one source to multiple targets with a single call of Dijkstra algorithm.
For example, on a 30 x 30 hexagonal lattice `eulerize` runs for more than half an hour on my laptop mainly due to the computation of all the shortest paths.

dschult: I agree that it would be better/faster most of the time to compute all the targets for any starting node rather than computing the shortest paths between each possible pair of odd degree nodes. There might be cases where this is slower than the current setup (e.g. when the number of even degree nodes is much much greater than the number of odd degree nodes). But a single dijkstra pass should be better in most all cases.  

I notice that we compute all those paths, but then mostly only use their lengths.  In fact the only place we need the paths themselves is to store the path as an attribute on the edge of the new graph.  Can you check the time of a version that computes path_length instead of path?  I suspect the path_length routines are faster than the path routines by enough that we should consider an option to not compute the paths here. But the default should be to have the paths I suspect.  But this whole paragraph is probably not as important as the previous paragraph. If we can reduce the number of graph traversals we need to do that will have the biggest benefit (I think).

rfulekjames: > Can you check the time of a version that computes path_length instead of path? I suspect the path_length routines are faster than the path routines by enough that we should consider an option to not compute the paths here

On `hexagonal_lattice_graph(20, 20)` and `hexagonal_lattice_graph(30, 30)` the running time improvement if not computing all the shortest paths is about 10% on my laptop (10 vs 9 sec. and 77 vs 70 sec, resp.), but the main benefit of not computing them should be in saving memory.


> There might be cases where this is slower than the current setup (e.g. when the number of even degree nodes is much much greater than the number of odd degree nodes). But a single dijkstra pass should be better in most all cases.

On `hexagonal_lattice_graph(30, 30)`  the running time went down considerably, to roughly 1/30 (if using a single  dijkstra pass from each vertex) of what it is now on my laptop. If odd degree nodes are _localized_ to a small part of graph the way it is implemented now is can be, of course, much better. 

To solve both cases (_localized_ and _non-localized_) at once a function computing shortest path lengths from a single source to multiple targets (without need to always search the whole graph) would be desirable, but I coudn't find such a function in the library. Have I overlooked anything? If there is no such function, perhaps, adding it would be beneficial also for different use cases.

Alternatively one can introduce a parameter in the `eulerize` function that would switch between the strategies as for how to compute shortest path lengths.

networkx/algorithms/tests/test_euler.py
rfulekjames: > As written, it is not clear to me why in line 293 you need nx.eulerize(G) inside the nx.is_eulerian. Hasn't G already been eulerized?

Yes. Fixed.

rossbar (commented):
Thanks for this @rfulekjames . One quick thought I had while scanning the discussion here and in the related issue - is there any chance you could split this up into two PRs? One for the bug fix and a second to encapsulate the `localized` proposal? As it stands, it's hard for me to quickly parse out what is related to the bug fix and what is related to the new feature proposal. I'm certainly not against either, but it's useful to have that context when reviewing!

If it's too much work to split up into multiple PRs that's totally fine - don't stress about it; it will likely just be stuck at the review stage for a bit longer :)

rfulekjames:
@rossbar Thanks for the feedback!

> is there any chance you could split this up into two PRs? One for the bug fix and a second to encapsulate the localized proposal? 

Sure, will do it shortly.

networkx/algorithms/euler.py
```diff
@@ -423,30 +433,51 @@ def eulerize(G):
         raise nx.NetworkXPointlessConcept("Cannot Eulerize null graph")
     if not nx.is_connected(G):
         raise nx.NetworkXError("G is not connected")
-    odd_degree_nodes = [n for n, d in G.degree() if d % 2 == 1]
+    odd_degree_nodes = {n for n, d in G.degree() if d % 2 == 1}
```
rfulekjames, line 436: To speed up line 455

networkx/algorithms/euler.py
```diff
@@ -423,30 +432,51 @@ def eulerize(G):
...
 
-    # get all shortest paths between vertices of odd degree
-    odd_deg_pairs_paths = [
-        (m, {n: nx.shortest_path(G, source=m, target=n)})
-        for m, n in combinations(odd_degree_nodes, 2)
-    ]
-
-    # use inverse path lengths as edge-weights in a new graph
-    # store the paths in the graph for easy indexing later
+    # get all shortest paths lengths between vertices of odd degree
```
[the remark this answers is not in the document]
rfulekjames: > I'll reword my question. Have you seen a case when it is faster to use the localized option? Can you find one?

Yes, for example, we can take a large toroidal grid, let's say 500 x 500, `G = nx.grid_2d_graph(500, 500, periodic=True)`
which is Eulerian and add locally some forward diagonals to make it non Eulerian:
```
for i in range(50):
    G.add_edge(2*(i % 10, i // 10), (2 * (i % 10) + 1 , 2 * (i // 10) + 1))
```
The example is a bit extreme, since the odd degree vertices are very close to each other, but it nicely illustrates the difference.
The running time of  `nx.eulerize(G, localized=False)`  was 49.65 sec and for  `nx.eulerize(G, localized=True)`  only 4.76 sec on my machine.

As I mentioned earlier, if there was an option to use a function computing shortest path lengths from a single source to multiple targets then  both localized and non-localized case could be handled more efficiently without this additional parameter.

rfulekjames:
Thanks @rossbar 

Resolved the conflicts. There was an issue with diffs w.r.t. `main`  which should be now showing correctly.

PR #6113, opened 2022-10-21
Fix doc inconsistencies related to cutoff in connectivity.py and disjoint_paths.py

paulitapb:
In `local_edge_connectivity` and `local_node_connectivity` from **connectivity.py** the cutoff description implies that only `edmonds_karp()` and `shortest_augmenting_path()` support cutoff. That is not true. Now those algorithms are given as examples of flow functions that allow cutoff. I tried to keep the description as general as possible and reference to [Flow](https://networkx.org/documentation/stable/reference/algorithms/flow.html) as the source of truth so in the future, if new flow algorithms are added there's no need to change this. 
Also changed the cutoff description in `edge_connectivity` from **connectivity.py** to match `local_edge_connectivity` and `local_node_connectivity` cutoff description. 
Finally in **disjoint_paths.py** added a reference to [Flow](https://networkx.org/documentation/stable/reference/algorithms/flow.html) in the cutoff description of `node_disjoint_paths `and `edge_disjoint_paths`.

### DIFF
--- a/networkx/algorithms/connectivity/connectivity.py
+++ b/networkx/algorithms/connectivity/connectivity.py
@@ -80,7 +80,7 @@ def local_node_connectivity(
         flow value reaches or exceeds the cutoff. This is only for the
         algorithms that support the cutoff parameter: e.g., :meth:`edmonds_karp`,
         and :meth:`shortest_augmenting_path`. Other algorithms will ignore this
-        parameter (see :meth:`maximum_flow` for details). Default value: None.
+        parameter (see :algorithms:`flow` for details). Default value: None.
 
     Returns
     -------
@@ -534,7 +534,7 @@ def local_edge_connectivity(
         flow value reaches or exceeds the cutoff. This is only for the
         algorithms that support the cutoff parameter: e.g., :meth:`edmonds_karp`,
         and :meth:`shortest_augmenting_path`. Other algorithms will ignore this
-        parameter (see :meth:`maximum_flow` for details). Default value: None.
+        parameter (see :algorithms:`flow` for details). Default value: None.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


networkx/algorithms/connectivity/connectivity.py
```diff
@@ -78,9 +78,9 @@ def local_node_connectivity(
     cutoff : integer, float
         If specified, the maximum flow algorithm will terminate when the
         flow value reaches or exceeds the cutoff. This is only for the
-        algorithms that support the cutoff parameter: :meth:`edmonds_karp`
-        and :meth:`shortest_augmenting_path`. Other algorithms will ignore
-        this parameter. Default value: None.
+        algorithms that support the cutoff parameter: e.g., :meth:`edmonds_karp`,
```
dschult, line 81: Didn't we decide that all the flow functions except `prepush_flow` supported `cutoff`?
In that case, maybe we can simplify these with something like:

``` 
    cutoff : integer, float, or None (default: None)
        If specified, the maximum flow algorithm will terminate when the
        flow value reaches or exceeds the cutoff. This doesn't work for the
        algorithm :meth:`prepush_flow` in which case `cutoff` is ignored.
```

dschult, line 81: Didn't we decide that all the flow functions except `prepush_flow` supported `cutoff`?
In that case, maybe we can simplify these with something like:

``` 
    cutoff : integer, float, or None (default: None)
        If specified, the maximum flow algorithm will terminate when the
        flow value reaches or exceeds the cutoff. This doesn't work for the
        algorithm :meth:`prepush_flow` in which case `cutoff` is ignored.
```

paulitapb: That was my first idea but then I thought what happens if there's some new flow function added that doesn't support cutoff.  Also maybe cutoff could be added to preflow push (I couldn't find anything about that online so I wanted to read the Cormen to fully understand the algorithm and decide if it could be possible to add without changing the algorithm itself). I thought this could be easier to maintain over time.  But I would be happy to change it if you think that's better. @dschult

dschult: Yes -- I see the problem....      How about:

```
cutoff : integer, float, or None (default: None)
        If specified, the maximum flow algorithm will terminate when the
        flow value reaches or exceeds the cutoff. This only works for flows
        that support the cutoff parameter (most do) and is ignored otherwise.
```

paulitapb: Perfect! I love that!

rossbar (approved):
LGTM, thanks @paulitapb !

PR #6127, opened 2022-10-24
DOC: Specifically branch off main, instead of current branch

MridulS:
Update the new contributor documentation to branch off from main. This will make sure people are creating branches from their local copy of main instead of the current working branch.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -94,7 +94,7 @@ Development Workflow
      branch name will appear in the merge message, use a sensible name
      such as 'bugfix-for-issue-1480'::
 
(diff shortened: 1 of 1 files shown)


rossbar (approved):
In principle this should be taken care of by the step above, which has the added benefit that `main` is up-to-date when the branching occurs. OTOH we've been seeing a lot of submissions where the branching is clearly a problem...

I'm willing to give it a try, if only to reduce the number of PRs that have branching problems. In a perfect world, we could admonish users to *really* pay attention to the contributor guide, but there's only so much you can do without going into too much detail about the git branching model.

dschult (approved):

PR #6130, opened 2022-10-25
Improve test coverage for digraph class

### LINKED ISSUE #6109: Improve test coverage for digraph class
I discovered there are still some red spots in this link https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/classes/digraph.py
I will like to work on it.

The tests will be added in https://github.com/networkx/networkx/blob/main/networkx/classes/tests/test_digraph.py

Current Behavior

Expected Behavior

Steps to Reproduce
Visit https://app.codecov.io/gh/networkx/networkx/blob/main//networkx/classes/digraph.py

Qudirah:
fixes #6109


MridulS (approved):
Thanks @Qudirah!

PR #6148, opened 2022-10-29
Bug fix in swap.py

### LINKED ISSUE #6144: directed_edge_swap: ZeroDivisionError when passing a digraph without edges
I was working on improving test coverage for `swap.py` and tested `directed_edge_swap` with a digraph without edges. That resulted in a ZeroDivisionError from `nx.utils.cumulative_distribution`. 

### Steps to Reproduce

```python
from networkx as nx

G = nx.DiGraph()
G.add_nodes_from([0, 2, 4, 3])

G = nx.directed_edge_swap(G, nswap=1, max_tries=500, seed=1)
```
### Current Behavior

ZeroDivisionError is raised because in `nx.utils.cumulative_distribution` there's a division by the sum of the input distribution that is zero. The distribution that `nx.utils.cumulative_distribution` is called with in this example is the degrees of a digraph without edges (all degrees are zero). This is the error log. 

```
  File "test.py", line 12, in <module>
    G = nx.directed_edge_swap(G, nswap=1, max_tries=500, seed=1)
  File "/home/paula/Outreachy/networkX/networkx/networkx/utils/decorators.py", line 766, in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
  File "<class 'networkx.utils.decorators.argmap'> compilation 21", line 5, in argmap_directed_edge_swap_17
  File "/home/paula/Outreachy/networkX/networkx/networkx/algorithms/swap.py", line 81, in directed_edge_swap
    cdf = nx.utils.cumulative_distribution(degrees)  # cdf of degree
  File "/home/paula/Outreachy/networkX/networkx/networkx/utils/random_sequence.py", line 103, in cumulative_distribution
    cdf.append(cdf[i] + distribution[i] / psum)
```
### Expected Behavior

I think the best is to raise an exception in `directed_edge_swap` with a message that exposes this situation.  Also, this should be added to the [doc entry](https://output.circle-artifacts.com/output/job/ab098af6-f76c-4bf9-a732-5c243aca96da/artifacts/0/doc/build/html/reference/algorithms/generated/networkx.algorithms.swap.directed_edge_swap.html#directed-edge-swap). 

### Environment

Python version: 3.8
NetworkX version: 2.8.7

### Further Discoveries

Something else that should be added is that if the digraph has less than 3 edges then is impossible to swap edges. In that case, the current behavior is that the max number of tries is reached. In my opinion, this case should raise an exception with an appropriate message. Also, this could be mentioned in the  [doc entry](https://output.circle-artifacts.com/output/job/ab098af6-f76c-4bf9-a732-5c243aca96da/artifacts/0/doc/build/html/reference/algorithms/generated/networkx.algorithms.swap.directed_edge_swap.html#directed-edge-swap). 

I will work on these changes and will look closely at the other functions in `swap.py`. 
Also, I think it will be good to review the functions that use `nx.utils.cumulative_distribution` to check for this error. I can do that too.

paulitapb:
Closes #6144 
I changed   `directed_edge_swap`, now an exception is raised if the number of edges is less than 3. Also updated the documentation.

### DIFF
--- a/networkx/algorithms/swap.py
+++ b/networkx/algorithms/swap.py
@@ -46,7 +46,7 @@ def directed_edge_swap(G, *, nswap=1, max_tries=100, seed=None):
     NetworkXError
         If `G` is not directed, or
         If nswap > max_tries, or
-        If there are fewer than 4 nodes in `G`
+        If there are fewer than 4 nodes or 3 edges in `G`.
     NetworkXAlgorithmError
         If the number of swap attempts exceeds `max_tries` before `nswap` swaps are made
 
@@ -70,9 +70,9 @@ def directed_edge_swap(G, *, nswap=1, max_tries=100, seed=None):
     if nswap > max_tries:
         raise nx.NetworkXError("Number of swaps > number of tries allowed.")
     if len(G) < 4:
-        raise nx.NetworkXError("Graph has less than four nodes.")
-    if len(G.edges) == 0:
-        raise nx.NetworkXError("Graph has no edges.")
+        raise nx.NetworkXError("DiGraph has less than four nodes.")
+    elif len(G.edges) < 3:
+        raise nx.NetworkXError("DiGraph has less than 3 edges")
 
     # Instead of choosing uniformly at random from a generated edge list,
     # this algorithm chooses nonuniformly from the set of nodes with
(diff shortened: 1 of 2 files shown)


paulitapb:
Added a more general PR for this changes

PR #6160, opened 2022-11-01
Add ThinGraph example to Multi*Graph doc_strings

nsengiyumva-wilberforce:
This PR is for https://github.com/networkx/networkx/issues/5854


rossbar (approved):
This at least gets us to the point where the docstrings between the 4 graph classes all have an example included. In principle it'd be great to modify the subclassing example for each case (i.e. inherit from MultiGraph for the MultiGraph example) but that gets into the question of whether/how docstrings should be kept in sync with one another.

IMO this is sufficient to close #5854. Thanks @nsengiyumva-wilberforce !

PR #6167, opened 2022-11-01
Sampled recursive largest first coloring

JonahBreslow:
Implemented a coloring algorithm not yet included. Performs a bit better than the greedy algorithms. I think I will need help with updating `doc/release/release_dev.rst`. Any code review feedback will be super helpful :)

### DIFF
--- a/networkx/algorithms/coloring/sampled_recursive_largest_first.py
+++ b/networkx/algorithms/coloring/sampled_recursive_largest_first.py
@@ -1,13 +1,14 @@
 from copy import copy
 from typing import Dict
 
+import networkx as nx
 from networkx.algorithms.mis import maximal_independent_set
 from networkx.classes import Graph
 
 __all__ = ["sampled_rlf"]
 
 
-def sampled_rlf(G: Graph, n_searches: int = 1e4) -> Dict:
+def sampled_rlf(G: Graph, n_searches: int = 1_000) -> Dict:
     """This is a non-deterministic approach to the Recursive
     Largest First algorithm that leverages the
     `networkx.maximal_independent_set` method, which randomly
@@ -24,6 +25,10 @@ def sampled_rlf(G: Graph, n_searches: int = 1e4) -> Dict:
     Returns:
         Dict: A dictionary where the key is the node and the value is the color
     """
+    if not isinstance(n_searches, int):
+        raise nx.NetworkXError(
+            "n_searches must be an integer. " f"{n_searches} not valid."
+        )
     color = {}
 
     temp_graph = copy(G)
--- a/networkx/algorithms/coloring/sampled_recursive_largest_first.py
+++ b/networkx/algorithms/coloring/sampled_recursive_largest_first.py
@@ -5,29 +5,73 @@ import networkx as nx
 from networkx.algorithms.mis import maximal_independent_set
 from networkx.classes import Graph
 
-__all__ = ["sampled_rlf"]
+__all__ = ["sampled_rlf_color"]
 
 
-def sampled_rlf(G: Graph, n_searches: int = 1_000) -> Dict:
-    """This is a non-deterministic approach to the Recursive
-    Largest First algorithm that leverages the
-    `networkx.maximal_independent_set` method, which randomly
-    selects a subset of nodes in graph G such that no edges
-    can be drawn between any of the selected nodes.
-    This implementation will execute this method `n_searches`
-    number of times and will always select the subset of nodes
-    that is the largest.
+def sampled_rlf_color(G: Graph, n_searches: int = 1_000) -> Dict:
+    """Color a graph using variant of Recursive Largest First algorithm.
 
-    Args:
-        G (Graph): an undirected networkx graph
-        n_searches (int): the number of samples to take
+    Attempts to color a graph using as few colors as possible, where no
+    neighbours of a node can have same color as the node itself. The
+    number `n_searches` is the number of maximal independent sets
+    sampled. Of all the maximal independent sets, the set with the
+    greatest number of nodes is selected to be colored. This is a
+    non-deterministic approach. The pseudo-code for this algorithm
+    is as follows:
 
-    Returns:
-        Dict: A dictionary where the key is the node and the value is the color
+    1. Let `G` be an undirected graph with nodes `n_1 ... n_n`
+    2. Select the node from `n_1 ... n_n` with the highest degree.
+    If there are multiple, just select the first one.
+    3. Sample `n_searches` maximal idependent sets using
+    `nx.maximal_independent_set()` and choose the set with
+    the greatest number of nodes, called `n_max`.
+    4. Assign `n_max` to the same color.
+    5. Remove `n_max` from the graph.
+    6. Repeat steps 2 through 5 until there are no more nodes
+    in the graph.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+
+    n_searches : int
+       the number of maximal independent set samples to take
+       from graph G
+
+    Returns
+    -------
+    A dictionary with keys representing nodes and values representing
+    corresponding coloring.
+
+    Examples
+    --------
+    >>> G = nx.Graph([(0, 1), (1, 2), (1, 3)])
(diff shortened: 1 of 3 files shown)


networkx/algorithms/coloring/sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,105 @@
...
+            "n_searches must be an integer. " f"{n_searches} not valid."
+        )
+
+    if not isinstance(G, Graph):
+        raise nx.NetworkXNotImplemented(
+            "G must be an undirected networkx graph. " f"{G} not valid."
+        )
+    color = {}
+
+    temp_graph = deepcopy(G)
```
dschult, line 78: Why does this need a deepcopy? This code doesn't change the attributes -- it only remove nodes...
So I would like `temp_graph = G.copy()` would be sufficient.

dschult (commented):
I haven't worked through all of this yet, but in a first pass I had some small questions...
Thanks for this!

networkx/algorithms/coloring/sampled_recursive_largest_first.py
JonahBreslow: Yeah, this is right. I'll switch it to a shallow copy.

networkx/algorithms/coloring/tests/test_sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,123 @@
...
+        for neighbor in graph.neighbors(node):
+            if coloring[neighbor] == color:
+                return False
+
+    return True
+
+
+def verify_length(coloring, expected):
+    coloring = dict_to_sets(coloring)
+    return len(coloring) == expected
```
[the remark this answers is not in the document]
JonahBreslow: Yes, this is a much more readable test. I'll make this change.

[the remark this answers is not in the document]
JonahBreslow: Ok, well the change is actually slightly different since it needs to account for the empty graph test. but the `dicts_to_sets` conversion can be removed for increased readability still.

networkx/algorithms/coloring/sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,105 @@
...
+            "G must be an undirected networkx graph. " f"{G} not valid."
+        )
+    color = {}
+
+    temp_graph = copy(G)
+    color_iter = 0
+
+    while len(temp_graph.nodes) > 0:
+        # finding node with the highest degree
+        init_node = sorted(G.degree, key=lambda x: x[1], reverse=True)[0][0]
```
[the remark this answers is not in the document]
JonahBreslow: updated

networkx/algorithms/coloring/sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,103 @@
...
+    References
+    ----------
+    .. [1] 'Recursive Largest First Algorithm.'
+       Wikipedia. Wikimedia Foundation, January 15, 2022.
+       https://en.wikipedia.org/wiki/Recursive_largest_first_algorithm#cite_ref-:1_2-0.
+    """
+    if not isinstance(n_searches, int):
+        raise nx.NetworkXError(
+            "n_searches must be an integer. " f"{n_searches} not valid."
+        )
```
dschult, line 69: I think we should remove this error checking because it will raise an exception later `while iteration < n_searches` and so long as it can be compared to `iteration`, the code will run just fine. For example, a Numpy.uint8 or similar might be used.  

Similarly the check for G being a networkx graph can be done by `maximum_independent_set`. If we test every function for G being a networkx graph we'll spend a lot of time checking. This is probably a bias or style of coding, but the networkx style is to not check the type of an input advertised as a networkx graph.  We do check if it is directed or multigraph when that matters, but otherwise assume that an input indicated to be a networkx graph is such a thing.

networkx/algorithms/coloring/sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,103 @@
+from typing import Dict
+
+import networkx as nx
+from networkx.algorithms.mis import maximal_independent_set
+from networkx.classes import Graph
+
+__all__ = ["sampled_rlf_color"]
+
+
+def sampled_rlf_color(G: Graph, n_searches: int = 1_000) -> Dict:
```
dschult, line 10: We are considering putting type annotations into stub files for NetworkX, but we don't include type annotations in the source code. We do indicate the type of the arguments in the doc_string under the parameters section.

networkx/algorithms/coloring/tests/test_sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,108 @@
...
+# Basic tests for all strategies
+# For each basic graph function, specify the number of expected colors.
+BASIC_TEST_CASES = {
+    empty_graph: 0,
+    one_node_graph: 1,
+    two_node_graph: 2,
+    disconnected: 2,
+    three_node_clique: 3,
+}
+
```
dschult, line 60: I think the BASIC_TEST_CASES could be specified more clearly and compactly by creating objects rather than functions which create the objects. The test code doesn't modify the graphs so even if we add tests which reuse this list of basic tests, we should be OK using objects.

I'll also suggest that you consider changing the name of `n_nodes` to `n_colors` because the value for e.g. `disconnected` is definitely not the number of nodes.

Here's how I'm suggesting we implement objects rather than functions:
```python
BASIC_TEST_CASES = [
    nx.Graph(),  # empty graph
    nx.path_graph([1]),  # one_node_graph
    nx.Graph([(1, 2)]),  # single_edge_graph
    nx.cycle_graph(3),   # triangle (three_node_clique)
    nx.Graph([(1, 2), (2, 3), (4, 5), (5, 6)]),  # disconnected graph
]
```
Then in the test for directed graphs, hardcode in the directed graph `graph = nx.DiGraph([(1, 2), (2, 3), (4, 5), (5, 6)])`

I'm not saying it has to be this way, but its my suggestion.

networkx/algorithms/coloring/sampled_recursive_largest_first.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    if not isinstance(n_searches, int):
+        raise nx.NetworkXError(
+            "n_searches must be an integer. " f"{n_searches} not valid."
+        )
+
+    if not isinstance(G, Graph):
+        raise nx.NetworkXNotImplemented(
+            "G must be an undirected networkx graph. " f"{G} not valid."
+        )
```
dschult, line 74: If G is a directed graph, **this code** will not raise an exception. (Because isinstance check for a Graph will pass for any subclass of a Graph as well.). The code in `maximal_independent_set` will raise an exception though.

So, I'd prefer that this code be removed or replaced with a check using `G.is_directed()` or by using the NetworkX decorator `@nx.utils.not_implemented_for('directed')` when starting the function. I think simply removing it here is the better choice, but I don't feel strongly about it.

dschult (commented):
I've got a few more suggestions and requests. The code and tests are solid as fr as I can tell. The comments are mostly about style, conventions and aiming for simplicity. Feel free to push back on any of them.

networkx/algorithms/coloring/sampled_recursive_largest_first.py
JonahBreslow: Right, that makes sense since a directed graph is a Graph subclass. I will just remove this since `maximal_independent_set` already handles this.

dschult (commented):
This looks good!  Thanks!
I have some suggestions for rewording the doc_string.

The main ideas I'm pushing are to 1) remove the node names given in the algorithm pseudocode and 2) to adapt the algorithm description to focus more on the overview
of the method used -- leaving definitions of parameters to other sections of the docs.

See what you think and make changes to these suggestions as you see fit.

dschult (commented):
The code looks good to me. 

I wonder if the module name should be the abbreviation and the function should be the full name. The __init__.py file seems to suggest that the full name is to be included in the top namespace. `nx.smampled_recursive_largest_first`. But that is currently the module... not the function.

Issue #6191, opened 2022-11-09
Testing increase in generators/line


Mjh9122:
Missing many branches in line.py. 
See link: https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/generators/line.py
### Expected Behavior
More tests are needed. 
### Steps to Reproduce
Look at the coverage.

MridulS:
Done in https://github.com/networkx/networkx/pull/6215

PR #6192, opened 2022-11-09
Line.py tests 

Mjh9122:
Adds coverage to Line.py. Slight issue with coverage, tests are not consistent, I believe it is due to the use of the arbitrary element. Only changes to non testing code are 1. adding error catching, 2. removed a branch that is redundant. #6191

### DIFF
--- a/networkx/generators/line.py
+++ b/networkx/generators/line.py
@@ -286,9 +286,6 @@ def inverse_line_graph(G):
         for u in p:
             P_count[u] += 1
 
-    if max(P_count.values()) > 2:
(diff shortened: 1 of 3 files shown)


networkx/generators/line.py
```diff
@@ -286,9 +286,6 @@ def inverse_line_graph(G):
         for u in p:
             P_count[u] += 1
 
-    if max(P_count.values()) > 2:
```
Mjh9122, line 289: I do not believe that this branch can be reached due to the fact that later functions force any vertex to be in at most 2 cliques

PR #6195, opened 2022-11-10
Added test in redundancy. Coverage up to 100.

Mjh9122:


networkx/algorithms/bipartite/tests/test_redundancy.py
```diff
@@ -25,6 +25,12 @@ def test_redundant_nodes():
         assert redundancy[v] == 1
 
 
+def test_redundant_nodes_2():
```
MridulS, line 28: Just a nit pick about the name.

Could you rename this to something like `test_redundant_nodes_subset`

PR #6203, opened 2022-11-10
Minor updates in matching.py 

Mjh9122:
Condensed some nested ifs, removed code that did not have any effect, and the comment explaining the code that had no effect.


rossbar (commented):
Same deal here with the newlines. I'll hold off looking at the rest of the PRs - please double check that these newline changes are undone.

PR #6204, opened 2022-11-10
Testing improvement in generators/line.py

Mjh9122:


rossbar (commented):
Can you revert all of the deleted newlines in `networkx/generators/line.py`? Not sure what happened there but we definitely don't want to get rid of those.

rossbar (commented):
Now the newlines are gone but there are extra whitespaces at the start of the lines. If possible, it'd be better to *revert* the commit where the newlines were added instead of manually adding them back to avoid other formatting issues like this.

Also check out the [developer guide](https://networkx.org/documentation/latest/developer/contribute.html), particularly the section on setting up a development environment and running e.g. `black`, which should prevent issues like this.

PR #6209, opened 2022-11-13
Add warning in bipartite module documentation  

paulitapb:
I noticed that on the main page of the bipartite module there's a comment about how to import the algorithms because they are not part of `networkx` base namespace. I made this a warning so it's easier for users to notice.


paulitapb:
As discussed in another PR, this is not a fit for a warning.

PR #6210, opened 2022-11-13
Add warning in community module documentation

paulitapb:
I noticed that on the main page of the community module there's a comment about how to import the functions because they are not part of networkx base namespace. I made this a warning so it's easier for users to notice.


rossbar (commented):
Same comment here, this isn't a good fit for a warning IMO

PR #6211, opened 2022-11-13
Add warning in documentation: import modules

paulitapb:
I noticed that on the main page of some modules there's a comment about how to import the functions because they are not part of the `networkx` base namespace. I checked all the documentation and made this a warning so it's easier for users to notice.
I added this in `node_classification` and  `tournament`.

### DIFF
--- a/networkx/algorithms/community/__init__.py
+++ b/networkx/algorithms/community/__init__.py
@@ -1,9 +1,10 @@
 """Functions for computing and measuring community structure.
 
-The functions in this class are not imported into the top-level
-:mod:`networkx` namespace. You can access these functions by importing
-the :mod:`networkx.algorithms.community` module, then accessing the
-functions as attributes of ``community``. For example::
+.. warning:: The functions in this class are not imported into the top-level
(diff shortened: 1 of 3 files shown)


rossbar (commented):
I'm -1 on adding these as a warning. Warnings are typically more for things that may change or may give unexpected results. In this case the functions are just organized into a sub-namespace, which is a language feature and (arguably) a best-practice (see e.g. PEP 20).

My vote would be to leave the current explanations as is!

paulitapb:
> I'm -1 on adding these as a warning. Warnings are typically more for things that may change or may give unexpected results. In this case the functions are just organized into a sub-namespace, which is a language feature and (arguably) a best-practice (see e.g. PEP 20).
> 
> My vote would be to leave the current explanations as is!

Ok! I saw it in [approximation doc entry](https://output.circle-artifacts.com/output/job/ab098af6-f76c-4bf9-a732-5c243aca96da/artifacts/0/doc/build/html/reference/algorithms/approximation.html) and I thought It was fine. But given this is not a good fit for a warning I'll change it there!

PR #6214, opened 2022-11-13
Improve coverage for current flow betweenness subset

Qudirah:
fixes for #6213


rossbar (commented):
Thanks for the suggestion @Qudirah - it turns out this was already covered by #6143, therefore I will close this!

PR #6224, opened 2022-11-18
Doc update; Newman-watts is almost correct

blwh:
I noted that the Newman-Watts-Strogatz algorithm isn't completely true to the original paper that introduces it. However, it's arguable that for large enough networks, and small `p` and `k`, this should not be a problem. Added this to the docs.


blwh:
I am struggling to find the style error in my commit. I'd appreciate some help finding it (if my pull request is OK, of course).

MridulS (changes requested):
Do you have pre-commit installed? It would catch these issues :)

blwh:
> Do you have pre-commit installed? It would catch these issues :) https://networkx.org/documentation/latest/developer/contribute.html

I am still a bit new to code contributions with git. Thank you!

PR #6228, opened 2022-11-20
fix wording in error message

ikethefifth:
The wording of this error message had the graph and collection switched around, so I fixed it


dschult (approved):
Thanks very much!

PR #6230, opened 2022-11-21
Setup vf2pp files for expansion to subgraph problems

dschult:
This PR sets up the functions to allow easier addition of the subgraph versions of the problem.
The main change are to add the degree dicts to the namedtuple data structure and 
make the `return False` statements in the main generator function more useful.  

Notes:
- Moved the degree dict creation to `_initialize_parameters`. Moved this call to before where the degrees are used.
- added G1_degree and G2_degree degree dicts to the graph parameters namedtuple. 
- Reordered the intro of the main function vf2pp_all_isomorphisms.
  - Replaced `return False` (which doesn't do much inside a generator) with `return "helpful string"`, which is the same as `raise StopIteration("helpful string")`. The string is only seen if you manually use `next` on the generator function. But that can be helpful for debugging why there is e.g. no isomorphism.
  - Inline the function `precheck_label_properties` and make more readable.
- Removed the degree dict input parameter from the `_find_candidates` helper function.
- changed the name of utils.labels_different to utils.labels_many.
- adjusted the tests accordingly.
- Reworded, unified language and corrected the main module doc_strings.

If this is too much to review I can split it up. Most of the changes are due to the change in the namedtuple.


MridulS:
@dschult is this ready for review? I guess this has some conflicts after https://github.com/networkx/networkx/pull/6334 ?

MridulS:
Is this still relevant @dschult ?

dschult:
I'm sprinkling these changes into a (currently) larger version that tries to do the same thing. So I am closing this. Thanks for the nudge!

PR #6236, opened 2022-11-25
Add conversion functions for node/edge_match to Graph based labels

dschult:
**Needed: a way to convert from `node_match` to label-based isomorphism checking.**
This PR provides helper functions for converting from node/edge_match to label-based matching in the isomorphism subpackage `node_match_to_label` and `edge_match_to_label`.

`nx.node_match_to_label(node_match, G, attrname)` checks the nodes in G and creates a label for each group of nodes for which node_match returns True. It update G to include the new label under the node attribute given by `attrname`. There is a similar function for edge_match.

To make sure that multiple graphs all use the same set of integer labels for nodes that match, the functions allow a list of graphs to process.

**Details**
One large API change from VF2 to VF2++ isomorphism algorithms is the way we handle matching nodes and edges. The current `is_isomorphic` uses optional `node_match/edge_match` functions which take 2 datadict inputs and return `True/False` to indicate whether those objects can be mapped to each other. The new `vf2pp_is_isomorphic` uses label names and looks up labels on the graph to determine the node labels. The nodes match when the labels match. We will be updating it to include matching of edges too.

The `node_match` functions are very flexible -- perhaps too flexible. There is no check that the matching is commutative (nodes u and v match exactly when nodes v and u match). And there is no check for transitivity (u ~ v and v ~ w means u ~ w).  There are also many special cases like checking numerical equality versus checking string or category values.  So in some ways the current interface is complicated, especially given that the only information provided to `node_match` are the datadicts for the two nodes.

Any valid matching conditions can be implemented via labels stored on the graph. But in complicated situations, that might require some work. For example, someone who currently has written a `node_match` function may find it hard to reformulate that into node labels stored in the graph. Hopefully this PR provides some help for that.

**Allowing more descriptive labels**
One difficulty with these new functions is that the labels are integers. A user might like to replace them with e.g. more meaningful strings. To help with this, the function returns a dict keyed by label to a representative datadict, from which they can relabel the nodes. An example of how to use this is provided in the doc_strings.

### DIFF
--- a/networkx/algorithms/isomorphism/matchhelpers.py
+++ b/networkx/algorithms/isomorphism/matchhelpers.py
@@ -1,11 +1,15 @@
 """Functions which help end users define customize node_match and
 edge_match functions to use during isomorphism checks.
 """
+import itertools
 import math
 import types
-from itertools import permutations
+
+import networkx as nx
 
 __all__ = [
+    "node_match_to_label",
+    "edge_match_to_label",
     "categorical_node_match",
     "categorical_edge_match",
     "categorical_multiedge_match",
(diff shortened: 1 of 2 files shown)


networkx/algorithms/isomorphism/matchhelpers.py
```diff
@@ -19,6 +23,214 @@
...
+
+    The `node_match` function takes a pair of node datadicts and returns `True`
+    if the nodes are considered matched in the sense of isomorphism. This function
+    calls `node_match` to determine which groups of nodes all match each other.
+    Those nodes are given an integer label, unique across all graphs in Glist.
+    The node attribute `attrname` is created/updated in each graph to store the labels.
+    Node labels are equal iff the corresponding nodes match across all graphs in Glist.
+
+    The `node_match` function must be commutative and transitive across all graphs
+    (if a node matches with 2 other nodes, then the two other nodes must match).
```
amcandio, line 37: Might be clearer to refer to this like if `node_match(u, v)` then..

networkx/algorithms/isomorphism/matchhelpers.py
```diff
@@ -19,6 +23,214 @@
...
+    ddict_by_label = {}
+    for G in Glist:
+        for edge in G.edges:
+            ddn = G.edges[edge]
+            for lbl, dd in ddict_by_label.items():
+                if edge_match(ddn, dd):
+                    break
+            else:
+                lbl = next(new_label)
+                ddict_by_label[lbl] = ddn
```
amcandio, line 229: This part could be deduped between the two. But just a preference

dschult:
I have been wrestling with our node/edge matching API while trying to revive #6236.  I've come to the realization that our current matching interface is more flexible than can be described by data attribute labels. And that we could easily allow more flexible matching beyond what we currently offer.  What follows is a discussion of my thoughts and a proposal that we allow a dict-of-containers for matching each of nodes and edges in addition to the data-attribute-name and matching-function currently offered. We convert the data attribute names and the matching functions to dict-of-sets internally now.

==== Details ====

**Q: Do the matching relationships need to be commutative and transitive?**
What if we had a triangle graph with nodes 0,1,2 and node values 53, 57 and 60. A numerical node_match function with threshold 5 would allow (0, 1) and (1, 2) to match while (0, 2) would not match. That's not transitive. Do we run into a problem? The valid isomorphisms from G to itself, G -> G (automorphisms) are {0:0, 1:1, 2:2}, {0:0, 1:2, 2:1}, and {0:1, 1:0, 2:2}. There doesn't seem to be any trouble with a non-transitive matching function in this example. And we can consider a subgraph morphism problem from this triangle to a graph with nodes a, b, c, d and node values 53, 57, 60, 64. So node matches consist of (0, a), (0, b), (1, a), (1, b), (1, c), (2, b, (2, c), (2, d). There doesn't seem to be any trouble with the isomorphism algorithms if we apply the restriction that mappings must consist of only these matching nodes.

Our current matching function helpers support these kind of non-transitive matching functions for ISMAGS and for VF2. Our vf2pp algorithm requires node labels (categorical node matching) which are always transitive. If u and v have the same label and v and w have the same label then u and w have the same label.  The vf2pp paper only considers this case, but there doesn't seem to be any reason that the algorithm must be restricted to this kind of matching.  We should probably support numerical_node_match functions in vf2pp just as we do for ISMAGS and vf2.

**A: No** we don't need transitive matching relations

**Q: What is a general way to encode and think about matching?**
Many papers on graph and subgraph isomorphism use node data via a node/edge label function `label(u)` and require `label(u) == label(v)` in the mappings. This is a categorical mapping and it is commutative and transitive. Many papers also ignore edge mapping using statements like: imposing restrictions on edge labels is straight-foward but we don't consider it because it complicates the discussion.

I have realized in working on this PR to provide helper functions which convert "match functions" to "labels" that numerical match functions cannot be described using categorical labels. I'm guessing many other matching functions don't map to labels.  In general, if `match(u, v)` computes `f(u) == f(v)` we can use labels. But if the matching involves any interaction between u and v like `NL(u, v) == 0` where `NL` is a nonlinear function we can't use labels. The numerical_match helper functions use the formula `math.isclose(f(u), f(v), rtol, atol)` which is essentially `|f(u) - f(v)| <= atol` and a similar `rtol` condition. This is nonlinear and can't be converted to labels.

I'm pretty sure a general way to encode matching conditions is as a dict keyed by SG object (either node or edge) to a container of FG objects.  The ISMAGS, VF2 and vf2pp algorithms create this kind of data structure as a dict of candidate sets. You might also notice that the relationship is a bipartitie graph with two sets "objects of SG" and "objects or FG" and connections only if the two objects match.  The dict of candidate sets is a directed adjacency representation of this bipartite graph. But it is simpler both conceptually and in terms of memory and speed to consider the matching relationship as a dict keyed by SG node to candidate image FG nodes. For edges, 

I'd like to propose that we allow matching restrictions of 3 types: labels (f(u)), binary match functions (bool f(u, v)), or matching sets (dict-of-containers) and that these each get converted to dict-of-candidate-sets early in the isomorphism process for all isomorphism algorithms.  This is more flexible than what we currently provide and aligns with the algorithm treatments of computing such a data structure during pre-processing.

**A: lets allow dict-of-containers as a matching input**

rossbar:
Thanks for the writeup @dschult , the examples specifically really helped me get a feel for the mechanics of matching in practice!

I'm +1 on expanding the matching restrictions (approaches?) to allow the dict-of-candidates pattern. The fact that the existing matching mechanisms (different matching types and different algorithms) all tend towards creating a mapping-of-candidates under-the-hood is a good indication that it's a nice, general way to frame matching! Furthermore, I suspect that decoupling the candidate-generation from the "inner-loop" of the matching code could provide an avenue for improved understanding of this core component (i.e. comparing candidate matches) of the isomorphism algorithms. I'm having trouble framing this clearly as I'm not comfortable with the terminology - but the point I'm trying to get across is: candidate matching is fundamental for the *morphism problem, and providing a pattern where candidates can be supplied explicitly (and works with most/all the *morphism algorithms) would be a great improvement. What this entails in practice I have no idea, but emphatic +1 for exploring!

PR #6237, opened 2022-11-25
Warn users about duplicate nodes in generator function input

dschult:
As we discussed last summer, some surprising results can occur if you input duplicate nodes in functions that use the nodes_or_number decorator. We decided to leave them in the code (some people might rely on them and they don't seem to be a problem for users).

For example:  nx.complete_graph([1, 1]) returns a graph with a self-loop.  nx.complete_graph([1, 2, 2, 3, 3, 3]) is the usual complete_graph on 3 nodes with selfloops added for node 2 and 3. 

This PR adds a warning to the doc_strings for these functions just so the users have a hint that duplicate nodes in the list of nodes can cause unusual results.

### DIFF
--- a/networkx/generators/classic.py
+++ b/networkx/generators/classic.py
@@ -233,6 +233,8 @@ def complete_graph(n, create_using=None):
     n : int or iterable container of nodes
         If n is an integer, nodes are from range(n).
         If n is a container of nodes, those nodes appear in the graph.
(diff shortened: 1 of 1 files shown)


MridulS (approved):
LGTM

jarrodmillman (approved):

Issue #5246, opened 2022-01-03
Improving the performance of triangles


zephyr111:
Hello everyone,

Based on [this StackOverflow post](https://stackoverflow.com/questions/70554095/counting-triangles-in-a-graph-by-iteratively-removing-high-degree-nodes/70559707), we have found that `triangles` of NetworkX is a bit slow and can be improved (assuming a higher memory footprint is acceptable). 

For non-directed graph, the following code can be used:

```python
def triangles(G):
    nodeNeighbours = {
        # The filtering of the set ensure each triangle is only computed once
        node: set(n for n in edgeInfos.keys() if n > node)
        for node, edgeInfos in G.adjacency()
    }
    
    res = {node: 0 for node in G.nodes()}
    
    for node1, neighbours in nodeNeighbours.items():
        for node2 in neighbours:
            for node3 in neighbours & nodeNeighbours[node2]:
                # Dispatch the counts to each node participating to the triangle found
                res[node1] += 1
                res[node2] += 1
                res[node3] += 1

    return res
```

This code is significantly faster big graphs.

For directed graphs, one need to compute the incoming edges of each nodes. This can be computed using a dictionary and a basic walk on the full graph (similar to what is done with `nodeNeighbours`).

Is this acceptable to use this implementation to improve the existing code?

Thank you.

### Steps to Reproduce

Here is an example to test the performance of `triangles`:

```python
import networkx as nx
# For bigger tests (slow):  G = nx.erdos_renyi_graph(15000, 0.005)
G = nx.erdos_renyi_graph(1000, 0.1)
res1 = nx.triangles(G)
res2 = triangles(G)
assert res1 == res2
```

### Environment

Python version: 3.9.9
NetworkX version: 2.6.3

SultanOrazbayev:
Two tiny suggestions:
- ignore nodes with only one neighbour,
- use `res=defaultdict(int)`

These are relatively harmless in dense graphs, but can give a boost on large, sparse graphs.

```python
from collections import defaultdict

def triangles(G):
    nodeNeighbours = {
        # The filtering of the set ensure each triangle is only computed once
        node: set(n for n in edgeInfos.keys() if n > node)
        for node, edgeInfos in G.adjacency()
        if len(edgeInfos)>1
    }
    
    res = defaultdict(int)
    
    for node1, neighbours in nodeNeighbours.items():
        for node2 in neighbours:
            for node3 in neighbours & nodeNeighbours[node2]:
[...]

PR #6258, opened 2022-12-06
Update calculation of triangles

SultanOrazbayev:
This PR follows up on issue https://github.com/networkx/networkx/issues/5246, the key function is mostly as suggested by @dschult and @zephyr111 (minor variable renaming, hopefully for better clarity).

Note that the existing docs are missing the `int` return when a single node is provided (even though it is one of the examples), so the docs are modified to account for that also.


networkx/algorithms/cluster.py
```diff
@@ -47,17 +49,65 @@ def triangles(G, nodes=None):
...
+
+    # iterate over the nodes in a graph
+    for node, neighbors in G.adjacency():
+        later_neighbors[node] = {
+            n for n in neighbors if n not in later_neighbors and n is not node
+        }
+
+    # instantiate with a zero count for each node
+    # add 1 to the count if a nodes neighbor's neighbor is also a neighbor
+    triangle_counts = {node: 0 for node in G}
```
eriknw, line 102:
```suggestion
    triangle_counts = dict.fromkeys(G, 0)
```

SultanOrazbayev: Somehow the tests are failing now.

SultanOrazbayev (commented):
the code seems to be fine

dschult:
Do we have any evidence yet that this speeds up the calculation? It looks like a fair amount of extra code and that usually ends up slowing things down. My comment in the original issue about looking up one neighbor not speeding things up made me think of that again here. How is this version doing?

SultanOrazbayev:
Yes, there's definitely a speed-up (my original SO question was prompted by very slow triangles calculation).

I didn't do comprehensive benchmarking, but here's a small python script to examine the times:https://gist.github.com/SultanOrazbayev/c4d5be83b5a8dd35ad13f0eff847135d. For an Erdos-Reny graph with 1500 nodes and 0.1 edge probability I get a speed up of about 10x-20x (note I gave up on trying larger graphs, since nx current version is taking rather long.

SultanOrazbayev:
And in the above script, I'm not controlling for seed etc, but the results are large and consistent enough.

SultanOrazbayev:
For the single-node case, though, yes, the current version appears to be much, much faster. This is likely due to the sub-graph creation part. Let me investigate a bit more.

SultanOrazbayev:
After further inspection, it seems that the speedup is observed only when computing triangles for the whole graph (possibly, there is also speed up when `nodes` contains a lot of nodes). For single node or for small subset of nodes the old approach is much, much faster.

As a way to combine the strengths of both approaches, the old approach is preserved if `nodes` is specified, if nodes kwarg is `None`, then the new approach is used.

dschult:
To speedup the subgraph part, you might try making a copy of the subgraph. Subgraph Views are not very performant. When traversing the graph, you still have to look at all edges and reject the ones that are not in the graph. If memory is not a bottleneck, and if you expect to look at an edge more than once on average, it is likely to be faster to add `.copy()` to the end of your subgraph call.  This does take time of course -- because you have to create that new graph to hold the subgraph. So, no guarantees...
That also would mean you should have some logic to handle the case where `nodes is None` so that you don't make a copy/subgraph of the whole graph.  :)

SultanOrazbayev:
Thank you for these details, I didn't think about the overhead of using views vs copy. Right now, the iterative approach of the original code is preserved for the case when nodes are provided. Do you think this will work?

dschult:
Yes -- that is a good way to handle the case when nodes are provided.

networkx/algorithms/cluster.py
```diff
@@ -47,17 +49,42 @@ def triangles(G, nodes=None):
...
+
+    # instantiate with a zero count for each node
+    # add 1 to the count if a nodes neighbor's neighbor is also a neighbor
+    triangle_counts = dict.fromkeys(G, 0)
+    for node1, neighbors in later_neighbors.items():
+        for node2 in neighbors:
+            for node3 in neighbors & later_neighbors[node2]:
+                triangle_counts[node1] += 1
+                triangle_counts[node2] += 1
+                triangle_counts[node3] += 1
```
dschult, line 85: This can be sped up by moving two increments out of the inner loop and using a `collections.Counter` object to push the third increment into compiled `Counter.update` code.
```suggestion
    triangle_counts = collections.Counter()
    for node1, neighbors in later_neighbors.items():
        for node2 in neighbors:
            third_nodes = neighbors & later_neighbors[node2]
            m = len(third_nodes)
            triangle_counts[node1] += m
            triangle_counts[node2] += m
            triangle_counts.update(third_nodes)
```

dschult (commented):
This look good! 
The later_neighbors approach creates an extra data structure (more RAM) but avoids the repeated lookup of neighbors in the previous `set(G[w]) - {w}`. I think the extra RAM is at most a dict of sets with N pointers to sets with at most N elements (N being the number of nodes). That's less than a copy of the graph so I view this as a good enhancement.

For performance, I checked a complete_graph on 1000 nodes and the old/current code is 28s while the PR code is 34s. So, the new version is slower for denser graphs. This makes sense because the bulk of the time savings with the new approach is pre-computation of `set(G[w]) - {w}` and the cost of the new method is looping over nodes in `neighbors & later_neighbors[node2]` instead of using it's length. When those "third node" sets are large the extra inner loop is costly. And when they are small, using lookups makes up for the extra inner loop cost. Ordering the nodes to avoid counting twice helps too, but the main savings is replacing set creation time with lookup time.

I then checked random graphs with 1500 nodes and p=0.3 and 0.5. When p=0.5, the two methods both took about 20.7s.  So, I think that is the density where the methods take essentially the same time.

I was able to speed up the PR method by a factor of 2-4 by moving the update of `triangle_counts` outside the inner loop and increasing it by `m` instead of `1` where `m` is the number of nodes to be looped over in the inner loop. Using a `Counter` to handle the inner loop also helps speed things up.

I tried adding pre-computation to the current networkx version and it helped a lot -- about 3-4 times faster. But that is still slower than the current PR for sparse graphs and slower than the inner-loop-speedup version for complete graphs. So, overall we should go with the PR with the inner-loop speedup with the inner loop pushed into the compiled code that is within `Counter.update()`.

We're looking at a speedup of 10+ times for sparse graphs and about 7 for density p=0.5 and about 4 for the complete_graph...  The complete graph on 1500 nodes takes 1min 56sec for the existing code, and 28sec for the PR with inner loop speedup.
See the comments below...

networkx/algorithms/cluster.py
```diff
@@ -27,13 +27,15 @@ def triangles(G, nodes=None):
     ----------
     G : graph
        A networkx graph
-    nodes : container of nodes, optional (default= all nodes in G)
+
+    nodes : container of nodes, optional (default = compute for all nodes in G)
```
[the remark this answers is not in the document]
SultanOrazbayev: For me the potential inconsistency was that default value of `nodes` kwarg is `None`, but on a second read, the original phrasing is better.

networkx/algorithms/cluster.py
```diff
@@ -27,13 +27,15 @@ def triangles(G, nodes=None):
...
        Compute triangles for nodes in this container.
 
     Returns
     -------
-    out : dictionary
-       Number of triangles keyed by node label.
+    out : dictionary or int
```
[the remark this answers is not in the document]
SultanOrazbayev: Thank you, incorporated these comments!

networkx/algorithms/cluster.py
```diff
@@ -47,17 +49,42 @@ def triangles(G, nodes=None):
...
     # If `nodes` represents a single node in the graph, return only its number
     # of triangles.
     if nodes in G:
         return next(_triangles_and_degree_iter(G, nodes))[2] // 2
-    # Otherwise, `nodes` represents an iterable of nodes, so return a
+
+    # if `nodes` is a container of nodes, then return a
     # dictionary mapping node to number of triangles.
-    return {v: t // 2 for v, d, t, _ in _triangles_and_degree_iter(G, nodes)}
+    if nodes:
```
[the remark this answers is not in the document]
SultanOrazbayev: Ah, interesting, didn't consider this! 

If `nodes=0` and G contains this node, then we would get back result from line 58. If `nodes=0` and G does not contain this node, without implementing `if nodes is not None`, the rest of the code would execute, triggering unnecessary compute. 

The tricky part is that if we change line 62 to `if nodes is not None`, then the returned value is going to be a `dict`, while the user would probably expect an `int` (since `nodes` is a single node `0`).

Another tricky example is if nodes is set to `False`, which could also be a valid node, but for which `if nodes` is not a valid solution.

So, I'm not sure if this is a good way out of this, but how about checking whether adding a check whether `nodes` could refer to a single node (by checking if it is hashable), if it is, then return the triangle count (or 0 if the node is not in the graph). Kindly check the updated version.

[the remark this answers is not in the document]
dschult: This idiom appears often in our code, and we sometimes refer to it as a `nbunch`. It could be a node. If could be a collection of nodes, and it could be None which means all the nodes. Because nodes can be containers, like the grid coordinates: (2, 3), we need to handle them first. If (2, 3) is in G, then we will treat it as a single node. And if (2, 3) is not in G then we will treat it as a container. This same logic works for non-containers too. We first check if it is in G. If not, then we check if it is None and handle the resulting container appropriately. The old `nx.triangle` code does this too. So we can leave some of that from before if that's easier.

```python
if nodes in G:
    return singleton  # nodes is a node of the graph G

if nodes is not None:
    #  nodes must be an iterator of nodes, so handle that case
    return dict(...)

# build the dict for all nodes in G
```

networkx/algorithms/cluster.py
dschult: In github you can Edit a message you have written by clicking the three dots next to the emoji smiley face in the upper right. :)

Also, Nice Catch on starting the Counter with a dict keyed by node to the value 0! :} Otherwise we would lose the nodes without any triangles.

And another style issue we often wrestle with is whether to spell out the full word `neighbors` or abbreviate it `nbrs`.  The long way is "better" because nobody has to guess what it is. The short way helps keep the code from being split over many lines. We often use `nbrs` when it helps reduce line-splitting. And we often use `neighbors` when it doesn't.  We can get away with that because `nbrs` is a pretty unmistakable abbreviation. 
:)

networkx/algorithms/cluster.py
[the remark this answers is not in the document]
dschult: Yes -- I agree, it would be better to change the wording to be even more clear:
(adjust as you think is better)
```
    nodes : node, iterable of nodes, or None (default=None)
        If a singleton node, return the number of triangles for that node. 
        If an iterable, compute the number of triangles for each of those nodes.
        If `None` (the default) compute the number of triangles for all nodes in `G`.
```

networkx/algorithms/cluster.py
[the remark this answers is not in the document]
SultanOrazbayev: Maybe I'm thinking of very unrealistic edge cases, but consider the snippet below:

```python
from networkx import path_graph, triangles

G = path_graph(5)
target_node = (3,4)

# compute triangles for the target node if it's not part of a graph (expect 0)
print(triangles(G, nodes=target_node))
# {3: 0, 4: 0}

# compute triangles for the target node if it's part of a graph (expect 0)
G.add_node(target_node)
print(triangles(G, nodes=(3, 4)))
# 0
```

For me the behaviour is potentially ambiguous. Specifically, in the first case, if there's an explicit check on the hashability of the object passed to `nodes`, then we would have received 0, since the code would check whether `target_node` is part of the graph.

I think this is a larger question about flexibility/ambiguity of `nbunch`, perhaps there should be a separate class for `nbunch`... For now, I think having an explicit check on hashability is going to resolve this edge case.

networkx/algorithms/cluster.py
```diff
@@ -42,22 +44,49 @@ def triangles(G, nodes=None):
...
-    if nodes in G:
-        return next(_triangles_and_degree_iter(G, nodes))[2] // 2
-    # Otherwise, `nodes` represents an iterable of nodes, so return a
-    # dictionary mapping node to number of triangles.
-    return {v: t // 2 for v, d, t, _ in _triangles_and_degree_iter(G, nodes)}
+    if nodes is not None:
+        # this checks if `nodes` could potentially refer to a single node
+        # If `nodes` represents a single node, return only its number of triangles
+        # (if it's in the graph) or zero (if the node is not in the graph).
+        if callable(getattr(nodes, "__hash__")):
```
rossbar, line 59: Is there a way to avoid diving into the object model? That might help with readability. For example:

```suggestion
        # If `nodes` represents a single node, return only its number of triangles
        if nodes in G:
```

Note also that `triangles` currently raises when `nodes` represents a single node that is not in the Graph, so I think this might've changed behavior for that corner case. A test should be added just to be sure!

networkx/algorithms/cluster.py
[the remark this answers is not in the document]
dschult: Checking hashability doesn't solve the issue because strings and tuples are iteraables that are hashable (and frozensets, frozen networkx.Graphs, etc).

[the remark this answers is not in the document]
SultanOrazbayev: Hmm, from the [tutorial](https://networkx.org/documentation/stable/tutorial.html): "a node can be any hashable object (except `None`)". Hence, if we check that an object is hashable, it could potentially be a node and so could be tested with `some_object in G`. So, strings/tuples/etc can be a node and when passed as an individual object to `nodes` they will be interpreted as a node, rather than as a container of nodes. I will elaborate in the issue on `nbunch`.

networkx/algorithms/cluster.py
SultanOrazbayev: Interesting, so suppose `nodes` is a list, then `if nodes in G` will be False, but we still don't know if it's because `nodes` is a container or because the single node in `nodes` is not in the graph... that's why it seems better to first check whether `nodes` is a container vs single node.

re: raising for a single node not in the graph, do you mean this (raising a KeyError or NetworkxError) is the desired behavior (to test against)?

rossbar: > Interesting, so suppose nodes is a list, then if nodes in G will be False, but we still don't know if it's because nodes is a container or because the single node in nodes is not in the graph... that's why it seems better to first check whether nodes is a container vs single node.

Unfortunately I think this is a fundamental ambiguity, since it's perfectly valid to have nodes that are iterable that meet this criteria.

In fact, the docstring itself is ambiguous: the parameter description says that `nodes` is supposed to be a container of nodes, but the examples section shows an example of a single node :shrug: .

Either way this is a clear corner case. For example, consider the following:

```python
>>> G = nx.Graph()
>>> G.add_node((0, 0))  # Nodes are 2-tuples, perhaps representing 2D coordinates for example
>>> G.add_node((2, 3))
# Let's assume the user intends (0, 1) to represent a single node, that happens not to be in G.
# What happens in this case?
>>> nx.triangles(G, (0, 1))  
```
On main:
```python
>>> nx.triangles(G, (0, 1))
{}
```
This PR:
```python
>>> nx.triangles(G, (0, 1))
   ...
StopIteration
```

Now as to which behavior is "correct" - I'd argue neither! But either way this is also what I was referring to with this being an untested corner case.

SultanOrazbayev: Oops, I fixed the `StopIteration` bug. Thank you for pointing this out! I also added a test, for now assuming that the correct return is 0 triangles for a node not in the graph.

SultanOrazbayev: Also, re: ambiguity: isn't this resolved if we check whether `nodes` could be a key in a python dict (so checking if it's hashable)? This way even if the object is iterable, we know whether it could potentially be a key in the dict (and hence a node in the graph, ignoring special case of `None`)

rossbar: > isn't this resolved if we check whether nodes could be a key in a python dict (so checking if it's hashable)?

The ambiguity arises from the fact that `nodes` can be interpreted as either a single node or a collection of nodes. When the user specifies `nodes=(0, 1)`, this can either be interpreted as: a collection containing nodes `0` and `1`; or a single node, the 2-tuple `(0, 1)`.

This ambiguity *wouldn't* exist if there were stronger enforcement of the parameter description (i.e. `nodes : container of nodes, optional (default = all nodes in G)`), but unfortunately [that ship has sailed](https://github.com/networkx/networkx/blob/0f91550007fd3a95261d858b1a6a623ef8bda38a/networkx/algorithms/cluster.py#L41).

The question then becomes what to do about it. Should we drop support for treating `nodes` as a single node? That would resolve the ambiguity but would change behavior and likely affect users. Should we add a bunch of input validation boilerplate?

Generally I think this is a larger discussion/separate issue - I only raised it here because I noticed that the proposed changes would change behavior for this (and potentially other) untested corner case(s).

dschult: If possible, can we leave the behavior the same for the following cases:
- a single node in the graph (if nodes in G: return a single integer value)
- a container that happens to hold at least one node of the graph (and return a dict keyed by node to the number)
- None as the default and leads to a dict keyed by each node of G to the number for that node.

It'd be great if the ambiguous corner cases could either be treated better or the same as they are currently treated.

SultanOrazbayev:
TLDR; I simplified the code and made the behaviour consistent with the original version. This is ready for a review now.

<details>

The ambiguity of `nodes` kwarg makes it difficult to have a nice implementation that would return 0 if a node is not in the graph. As a result, this test that I wanted to add will not be added:

```python
    def test_node_not_in_graph(self):
        G = nx.Graph()
        G.add_node((0, 0))
        G.add_node((2, 3))
        assert nx.triangles(G, (0, 1)) == 0
```

The tricky part here is that if one passes an iterable containing nodes that are not in the graph, one would expect a dictionary with keys corresponding to the nodes in the iterable and values being zero. (somewhat similar behaviour for a list containing some nodes in the graph and some nodes not in the graph). Since a node could potentially also be an iterable itself, this leads to ambiguity (if we should return zero because the iterable is not a node in the graph or try to iterate over the contents of the iterable and return a dict with key/val containing zeros for nodes that are not in the graph and n-triangles for nodes that are in the graph). Implementing this would not lead to a readable code, so for now the most efficient path is to revert to the original behaviour.

</details>

networkx/algorithms/cluster.py
```diff
@@ -42,22 +46,47 @@ def triangles(G, nodes=None):
...
+
+    # iterate over the nodes in a graph
+    for node, neighbors in G.adjacency():
+        later_neighbors[node] = {
+            n for n in neighbors if n not in later_neighbors and n is not node
+        }
+
+    # instantiate with a zero count for each node
+    # add 1 to the count if a nodes neighbor's neighbor is also a neighbor
+    triangle_counts = Counter(dict.fromkeys(G, 0))
```
aaronzo, line 80: Why is instantiating an empty counter necessary? we are guaranteed to have all nodes in `later_neighbours` so an empty `Counter` should be ok I would have thought.

```suggestion
    Counter()
```

If for whatever reason the initialized Counter is necessary, you can avoid instantiating 2 dictionaries by doing
```python
defaultdict(int).fromkeys(G, 0)
```
instead, without much of a performance difference.

networkx/algorithms/cluster.py
```diff
@@ -42,22 +46,47 @@ def triangles(G, nodes=None):
...
+
+    # dict used to avoid visiting the same nodes twice
+    # this allows calculating/counting each triangle only once
+    later_neighbors = {}
+
+    # iterate over the nodes in a graph
+    for node, neighbors in G.adjacency():
+        later_neighbors[node] = {
+            n for n in neighbors if n not in later_neighbors and n is not node
+        }
```
aaronzo, line 76: you can avoid looping over this twice by using a generator:
```suggestion
    def later_neighbors_gen():
        for node, neighbors in G.adjacency():
            nbrs = {
                n for n in neighbors
                if n not in later_neighbors and n is not node
            }
            later_neighbors[node] = nbrs
            yield node, nbrs
```
and then on line 81:
```python
for node1, neighbors in later_neighbors_gen():
    ...
```

dschult: Keeping the loop **out of a generator function** makes it easier to read and the generator doesn't save any memory.  Am I missing an advantage of `later_neighbors_gen`?  If not, I'd prefer the original to this suggestion.

aaronzo: The advantage would be that the nodes are only looped over once, not twice (once on creating `later_neighbors`, once on counting triangles).

But on testing this I've just realised that the whole dictionary of `later_neighbors` is needed before the second loop can begin because it's used on line 82, so this should stay as is.

networkx/algorithms/cluster.py
SultanOrazbayev: It's not much, but it helps with ensuring that nodes without neighbors have zero counts. If we have `triangle_counts` initiated using blank `Counter`, there would have to be a separate loop to fill in zeros for nodes not in `triangle_counts`.

aaronzo: All nodes appear in `G.adjacency()`, so all nodes appear in `later_neighbors` - if they have no neighbors they just appear with the empty set. So, you can avoid initializing the Counter by adding a statement inside the first block, but outside the second:

```python
triangle_counts = Counter(dict.fromkeys(G, 0))
for node1, neighbors in later_neighbors.items():
    if not neighbors:
        triangle_counts[node1] = 0
    for node2 in neighbors:
        ...
```
checking the boolean is optional, I guess.

SultanOrazbayev: The performance change is interesting, but a more robust benchmarking would be needed for maintainers. This is one of the missing pieces in the current PR for it to be considered for merging. 

Also, your suggestion might have a better chance as a standalone PR because in this one I tried to combine the triangle count change with a more explicit handling of `nodes`/`nbunch` (the unresolved conversation above). I reverted that behavior to be consistent with the original code, but I guess it doesn't look very clean. This is not to say that your suggestions are not welcome for this PR :), but to say that after performance benchmarking there might still be an issue with the code readability.

SultanOrazbayev: Looking closer at your suggestion though, I am not actually sure about the source of the performance improvement. 

```diff
- triangle_counts = Counter(dict.fromkeys(G, 0))
+ triangle_counts = Counter()
  for node1, neighbors in later_neighbors.items():
+   acc = 0
    for node2 in neighbors:
       third_nodes = neighbors & later_neighbors[node2]
       m = len(third_nodes)
+      acc += m
-      triangle_counts[node1] += m
       triangle_counts[node2] += m
       triangle_counts.update(third_nodes)
+   triangle_counts[node1] += acc
```

As you note, we are modifying the `node1` entry of `triangle_counts` less frequently, but this seems a bit like a micro-optimisation (see a talk by James Powell that has nice examples: https://youtu.be/Ix04KpZiUA8?t=754). Specifically, this *might* have performance boost for nodes that have a lot of neighbors if updating counter element is more expensive than updating a variable. However, in terms of readability (and this is subjective) the reader has to keep track of three lines (involving `acc`) to understand what's going on, which decreases readability (but again, this is subjective).

aaronzo: Will leave it up to you, but my preference here would be the minor optimization for the very minor increase in code length. My thinking is that triangle counting will be a function used often, so speed will improve the user experience.
> might have performance boost for nodes that have a lot of neighbors

I think we can say more than that, variable updates are faster than dict lookups, and initializing a dictionary vs not is clearly slower. The worst case scenario for this algorithm is highly connected graphs - so improving that case will be beneficial since graphs that are not highly connected will be fast anyway.

But, with all that considered, it's still a minor thing :) If you are happier with your code as it is, it should be merged that way - I just wanted to mention this as it jumped out at me when I first read it.

dschult (approved):
Thanks for this -- and sorry for the delay. I tried to include the new changes in the helper function used by other functions for triangles, degree, generalized degree. But the loops are different enough that it didn't end up helping there.  Let's get this merged. :)
Thanks!

SultanOrazbayev:
Thank you very much, @dschult !

jarrodmillman (approved):

PR #6261, opened 2022-12-07
Changed arguments list of GraphMLWriterLxml.dump()

patrick-nicodemus:
Changed arguments list of GraphMLWriterLxml.dump to be consistent with the type of GraphMLWriter.dump. Otherwise, if we have some GraphMLWriterLxml A, and some user calls A.dump(stream) (maybe in the context of a function which takes a GraphMLWriter) then it will throw an error.
This also makes static analysis tools happier, which reduces the amount of things to look at when debugging.


MridulS (approved):

rossbar (approved):
Thanks @patrick-nicodemus !

Issue #6264, opened 2022-12-08
OutMultiEdgeView.__call__ order of arguments


patrick-nicodemus:
### Discussed in https://github.com/networkx/networkx/discussions/6263

<div type='discussions-op-text'>

<sup>Originally posted by **patrick-nicodemus** December  7, 2022</sup>
The __call__ method of `classes.reportviews.OutMultiEdgeView` is defined as:
```
def __call__(self, nbunch=None, data=False, keys=False default=None, ):
```
but the __call__ method of its parent class `classes.reportviews.OutEdgeView` is
```
def __call__(self, nbunch=None, data=False, default=None):
```
I would suggest that `keys` be moved to the last position. Otherwise if a user expects an OutEdgeView and gets an OutMultiEdgeView they will get a wrong answer.</div>

SultanOrazbayev:
Thank you for flagging this @patrick-nicodemus !

PR #6267, opened 2022-12-09
Patch view signature

SultanOrazbayev:
Resolves https://github.com/networkx/networkx/issues/6264

Ready for review.

### DIFF
--- a/networkx/classes/tests/test_reportviews.py
+++ b/networkx/classes/tests/test_reportviews.py
@@ -872,9 +872,8 @@ class TestMultiEdgeView(TestEdgeView):
             assert len(e) == 3
         elist = sorted([(i, i + 1, 0) for i in range(8)] + [(1, 2, 3)])
         assert sorted(ev) == elist
-        # test order of arguments:graph, nbunch, data, keys, default
-        ev = evr((1, 2), "foo", keys=True, default=1)
         # test that the keyword arguments are passed correctly
+        ev = evr((1, 2), "foo", keys=True, default=1)
         with pytest.raises(TypeError):
             evr((1, 2), "foo", True, 1)
         with pytest.raises(TypeError):
--- a/networkx/classes/tests/test_reportviews.py
+++ b/networkx/classes/tests/test_reportviews.py
@@ -878,7 +878,7 @@ class TestMultiEdgeView(TestEdgeView):
         with pytest.raises(TypeError):
             evr((1, 2), "foo", True, 1)
         with pytest.raises(TypeError):
-            evr((1, 2), "foo", keys=True, 1)
+            evr((1, 2), "foo", True, default=1)
         for e in ev:
             if set(e[:2]) == {1, 2}:
                 assert e[2] in {0, 3}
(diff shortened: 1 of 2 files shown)


rossbar (commented):
I very much like the idea of resolving this issue with keyword-only arguments. This will break user code for any users who are calling all the way out to the `default` or (`keys` in the case of the `Multi` views) using argument positions. [We generally try to warn users about upcoming breaking changes](https://networkx.org/documentation/latest/developer/deprecations.html), so this probably needs some discussion to see how best to move forward.

dschult:
I think for the most part you would still simply want to check whether a value is different from the default. That is, I would favor API design that doesn't distinguish whether someone used keyword or positional syntax to provide the input value.

But -- there are tools that can do it in [Python's `inspect` library](https://docs.python.org/3/library/inspect.html). In particular the `inspect.Parameter` and `inspect.Signature` objects.  But I would not say that these are not as easy way to check for positional/keyword passing.

networkx/classes/tests/test_reportviews.py
```diff
@@ -873,7 +873,7 @@ def test_iterkeys(self):
         elist = sorted([(i, i + 1, 0) for i in range(8)] + [(1, 2, 3)])
         assert sorted(list(ev)) == elist
         # test order of arguments:graph, nbunch, data, keys, default
-        ev = evr((1, 2), "foo", True, 1)
+        ev = evr((1, 2), "foo", keys=True, default=1)
```
dschult, line 876: Should we add a test that it is keyword only?  I mean something like:
```python
with pytest.raises(TypeError):
    evr((1, 2), "foo", True, 1)
```
Or is it overkill to test for whether the arguments are keyword only or not.  I guess to do a thorough job you'd want to test that using a second positional argument doesn't raise a TypeError, while the third positional argument does. If we're going to be eventually using keyword-only arguments in many places in the library the complete tests could become onerous.  But it would have caught the potential bug that this PR fixes...

SultanOrazbayev: Thank you, @dschult ! I implemented this, but I agree that this won't scale well (in general) if keyword-only arguments are more widely used.

networkx/classes/tests/test_reportviews.py
```diff
@@ -873,7 +873,12 @@ def test_iterkeys(self):
...
+        # test that the keyword arguments are passed correctly
```
dschult, line 877: Let's clean up the comments here too. The leading comment line about `test of order arguments` doesn't even seem to be correct anymore. So maybe just move your new comment line up to replace the comment line on line 875.  That also puts line 876 under that description, which I think is accurate. Lots of flexibility here, so do what you think is best -- but I think the old line 875 should be removed/changed.

dschult: Yes -- the `graph` is the part that was confusing before -- and the keys/default are no longer ordered. So probably best to just get rid of the claim to an ordered in a comment.

SultanOrazbayev: Thanks, fixed!

dschult (approved):
Looks good to me... :}  Thanks!

SultanOrazbayev:
Given the discussion at the last meeting (re: breaking changes in API), does it make sense to make a separate PR with an advance warning and keep this PR open until 2 minor releases have been done?

rossbar (approved):
This is still technically a breaking change, but to hit this corner case in practice you'd need to call the view methods with both `nbunch` and `data` specified by position, which is pretty unreadable. Note also there were no instances of this in the test suite (excepting the one test that was explicitly for testing argument order), which, by weak proxy, lends more confidence that this likely doesn't happen often in user-code.

Given all this, I'm game for putting this in without a warning if others are! If it ends up seriously biting users, it can always be reverted and re-introduced with a warning cycle.

Issue #2717, opened 2017-10-20
 remove_edges_from causing RuntimeError on certain view generators


ahrenberg:
In certain cases when remove_edges_from is operating on an view generator it will fail with a RuntimeError due to the yielded objects being the same as those removed.

I have observed it in two cases:

1. On on view returned by nx.selfloop_edges operating on MultiGraph and MultiDiGraph objects.  `G.remove_edges_from(nx.selfloop_edges(G))` is a standard construction for removing self-loops. It works with Graph and DiGraph, but fails for MultiGraph and MultiDiGraph.

2. On the view returned by nx.edges for all graph types. The construction `G.remove_edges_from(nx.edges(G))` might not be very useful, but the error also occurs when using functions from itertools to manipulate the view of nx.edges(G). Examples below.

Of course, in 2 the behaviour might perhaps be expected - the view seems to be an iterator to the same object as being manipulated, which is problematic. (A workaround is of course to create a list from it.) 

However in 1, the selfloop view seems to be handled for Graph and DiGraph, so might it not be expected to work also for the ones provided for MultiGraph and MultiDiGraph?

## Error-producing example
These three examples all fail:

### Removing self-loops from MultiGraph
```
import networkx as nx
G = nx.MultiGraph([(0,0), (1,1), (0,1)])
G.remove_edges_from(nx.selfloop_edges(G)) # Fail here
```
#### Error Message
```
Traceback (most recent call last):                                                                                    
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/networkx/classes/multigraph.py", line 603, in remove_edges_from              
    for e in ebunch:                                                                                                  
  File "/usr/lib/python3.6/site-packages/networkx/classes/function.py", line 1150, in <genexpr>
    if n in nbrs for d in nbrs[n].values())
  File "/usr/lib/python3.6/_collections_abc.py", line 761, in __iter__
    for key in self._mapping:
RuntimeError: dictionary changed size during iteration
```

### Removing all edges from Graph
```
import networkx as nx
G = nx.Graph([(0,1), (1,2), (2,3)])
G.remove_edges_from(nx.edges(G)) # Fail here
```
#### Error Message
```
Traceback (most recent call last):                                                                                    
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/networkx/classes/graph.py", line 1038, in remove_edges_from                  
    for e in ebunch:                                                                                                  
[...]

PR #6268, opened 2022-12-09
doc: update documentation when providing an iterator over current graph to add/remove_edges_from.

SultanOrazbayev:
Resolves https://github.com/networkx/networkx/issues/2717

### DIFF
--- a/networkx/classes/digraph.py
+++ b/networkx/classes/digraph.py
@@ -702,6 +702,16 @@ class DiGraph(Graph):
         Edge attributes specified in an ebunch take precedence over
         attributes specified via keyword arguments.
 
+        Warning
+        -------
+        When providing a container of edges, make sure that it is not an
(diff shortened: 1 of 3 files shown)


dschult (commented):
This looks nice -- and the issue has been around a long time without action... so thanks!

I have wording suggestions -- they are optional, so decide what words you think work best. The start of the notes seems awkward and in the last sentence, the word "variable" isn't quite right since the list often isn't assigned to a variable name.

For the notes sections, perhaps start with:
When adding nodes from an iterator over the graph you are changing, ...

for the last sentence, maybe replace "variable" with "object" in two places in the last sentence.

Thanks!!

dschult (approved):
Thanks!  This looks good!

MridulS (approved):
Thanks @SultanOrazbayev!

Issue #3912, opened 2020-04-12
Lack of library-wide ordered set hurts readability of algorithms


xpe:
**Updated as of 2020-04-12 1:41pm PDT**

**Problem**

There is at least one place in NetworkX that uses [collections.OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) as a substitute for an insertion-ordered set. This is an undesirable workaround: it makes code harder than necessary to read.

Consider this example:

```py
visited[child] = None
```

What do you think it does? I wouldn't fault anyone for (on a first pass) thinking that it marks `child` as not visited. Well, in reality, **adds** an item to an collections.OrderedDict (used as an insertion-ordered set).

In my view, the above code doesn't just "read badly" it is also misleading. In context of a particular algorithm, [simple_paths.py][1], it makes it harder to follow.

[1]: https://github.com/networkx/networkx/blob/b0ca6ce081cc6e66d509c5f7fed60728da821c25/networkx/algorithms/simple_paths.py#L241

**Solution**

There are options. I understand that Python's collections library does not provide an ordered set container. I recommend either:

* using an existing ordered set library, or
* add a library-wide wrapper on top of OrderedDict.

For example, please consider [IndexedSet from the boltons package](https://boltons.readthedocs.io/en/latest/setutils.html#boltons.setutils.IndexedSet):

> IndexedSet is a collections.MutableSet that maintains insertion order and uniqueness of inserted elements. It’s a hybrid type, mostly like an OrderedSet, but also list-like, in that it supports indexing and slicing.

SultanOrazbayev:
My understanding is this issue can be closed. The function referenced (`_all_simple_paths_graph`) is now using ordinary `dict`, which since Python 3.7 is guaranteed to preserve the insertion order (https://docs.python.org/3/library/collections.html#ordereddict-objects).

dschult:
I think we could close this.  
Another option is to change `visited[child] = None` to `visited[child] = True`. Setting to `True` is perhaps easier to understand as adding child to the dict while setting to `None` might mean turning off a flag held in the dict.  If this is "fixed", it should be changed in any other places in that module where the same idiom is used: probably a search for `= None` will quickly give the places where it occurs.

PR #6273, opened 2022-12-10
Update simple_paths.py to improve readability of the BFS.

SultanOrazbayev:
Improve readability of the BFS.

Resolves https://github.com/networkx/networkx/issues/3912.

### DIFF
--- a/networkx/algorithms/simple_paths.py
+++ b/networkx/algorithms/simple_paths.py
@@ -257,7 +257,7 @@ def _empty_generator():
 
 
 def _all_simple_paths_graph(G, source, targets, cutoff):
-    visited = dict.fromkeys([source])
+    visited = dict.fromkeys([source], value=True)
     stack = [iter(G[source])]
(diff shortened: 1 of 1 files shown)


networkx/algorithms/simple_paths.py
```diff
@@ -257,7 +257,7 @@ def _empty_generator():
 
 
 def _all_simple_paths_graph(G, source, targets, cutoff):
-    visited = dict.fromkeys([source])
+    visited = dict.fromkeys([source], True)
```
dschult, line 260: Would this be more readable? 
```suggestion
    visited = {source: True}
```

SultanOrazbayev:
Agree with the change, I hesitated with proposing it initially, but agree that it's more readable that way.

dschult (approved):
Thanks!   LGTM

MridulS (approved):
Thanks @SultanOrazbayev!

PR #6286, opened 2022-12-16
[DOC] Follow numpydoc standard in barbell_graph documentation

SultanOrazbayev:
Resolves https://github.com/networkx/networkx/issues/6285

### DIFF
--- a/networkx/generators/classic.py
+++ b/networkx/generators/classic.py
@@ -165,7 +165,6 @@ def barbell_graph(m1, m2, create_using=None):
     Notes
     -----
 
-    For $m1 > 2$ and $m2 >= 0$.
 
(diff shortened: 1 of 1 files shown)


networkx/generators/classic.py
```diff
@@ -145,7 +145,27 @@ def balanced_tree(r, h, create_using=None):
...
+
+    Returns
+    -------
+    G : NetworkX graph
+        A barbell graph.
+
+    Notes
+    -----
+
+    For $m1 > 2$ and $m2 >= 0$.
```
MridulS, line 168:
```suggestion
```
I think we can remove this bit, the information is captured in the Parameters section.

MridulS (approved):

dschult (approved):

PR #2209, opened 2016-07-31
Test ordering

hagberg:
Addresses some of #2203

PR #6291, opened 2022-12-20
Fix ineffective `nodes` args of graph6 writers

e10s:
Expecting `G.nodes()` to report in order of `G.add_nodes_from(nodes)`.

### DIFF
--- a/networkx/readwrite/graph6.py
+++ b/networkx/readwrite/graph6.py
@@ -176,7 +176,10 @@ def to_graph6_bytes(G, nodes=None, header=True):
 
     """
     if nodes is not None:
-        G = G.subgraph(nodes)
+        H = G.__class__()
+        H.add_nodes_from(nodes)
+        H.add_edges_from((u, v) for (u, v) in G.edges() if u in H if v in H)
+        G = H
     H = nx.convert_node_labels_to_integers(G)
     nodes = sorted(H.nodes())
     return b"".join(_generate_graph6_bytes(H, nodes, header))
@@ -364,7 +367,10 @@ def write_graph6_file(G, f, nodes=None, header=True):
 
     """
     if nodes is not None:
-        G = G.subgraph(nodes)
+        H = G.__class__()
+        H.add_nodes_from(nodes)
+        H.add_edges_from((u, v) for (u, v) in G.edges() if u in H if v in H)
+        G = H
     H = nx.convert_node_labels_to_integers(G)
     nodes = sorted(H.nodes())
     for b in _generate_graph6_bytes(H, nodes, header):
--- a/networkx/readwrite/tests/test_graph6.py
+++ b/networkx/readwrite/tests/test_graph6.py
@@ -118,6 +118,16 @@ class TestWriteGraph6:
         f.seek(0)
         assert f.read() == b">>graph6<<A_\n"
 
+    def test_ordering_nodes(self):
+        G = nx.Graph()
+        G.add_nodes_from(range(4))
+        G.add_edges_from([(1, 2), (1, 3)])
+        nodes = [1, 2, 3, 0]
+        f = BytesIO()
+        nx.write_graph6(G, f, nodes=nodes, header=False)
+        f.seek(0)
+        assert f.read() == b"Co\n"
+
 
 class TestToGraph6Bytes:
     def test_null_graph(self):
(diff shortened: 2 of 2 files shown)


dschult:
Thanks for the suggestion. But I think that will change the existing behavior is unexpected ways for others. It seems like this could be done using the `_generate_graph6_bytes` function (which is a private function so a little harder to get to, but you seem pretty flexible with that sort of thing.  Calling the private function directly bypasses the setup code in `write_graph6` so you customize it yourself. I'm suggesting something like:

```python
G = nx.Graph()
G.add_nodes_from(range(4))
G.add_edges_from([(1, 2), (2, 3)])
nodes = [1, 2, 3 ,0]

f = BytesIO()
lines = nx.readwrite.graph6._generate_graph6_bytes(G, nodes, False)
f.writelines(lines)
f.seek(0)
newG=nx.read_graph6(f)

print(G.nodes, G.edges)   # output: (NodeView((0, 1, 2, 3)), EdgeView([(1, 2), (2, 3)]))
print(newG.nodes, newG.edges)  # output: (NodeView((0, 1, 2, 3)), EdgeView([(0, 1), (1, 2)]))
```
Which seems like what you want...

e10s:
Thank you for your feedback.

> But I think that will change the existing behavior is unexpected ways for others.

Please note that due to some modifications between NetworkX 1.11 and 2.0, the behavior also changed accidentally or not, therefore it differs from the documents left unchanged.

I made a test case shown below.
```python
import networkx as nx

print(f"NetworkX {nx.__version__}")

G = nx.Graph()
G.add_nodes_from(range(4))
G.add_edges_from([(1, 2), (2, 3)])
nodes = [1, 2, 3, 0]
print(nx.generate_graph6(G, header=False))
print(nx.generate_graph6(G, nodes=nodes, header=False))
```
I tested it with two versions of NetworkX on Python 3.6.15.
```
NetworkX 1.11
CH
Cg
```
```
NetworkX 2.0
CH
CH
```
I expect the former.

Adding a note, afterward, `generate_graph6` became `to_graph6_bytes` that returns in `bytes` with `\n`.

dschult:
OK.. If I understand you correctly, the problem is that the documentation says:
```
Nodes are labeled 0...n-1 in the order provided.  If None the ordering given by `G.nodes()` is used.
```
when in fact the nodes are ordered the same as they are in `list(G.nodes)`.  The behavior and the documentation don't agree.

NetworkX 1.11 and 2.0 are old enough that dict's did not maintain order. So the node order in NetworkX graphs was arbitrary and implementation specific. I believe the change you are referring to is inside #2209 when changes were made to ensure that dict order was treated as arbitrary (so that Jython could work). But I don't think tracking that down helps us decide what the code should be now. I'm open to changing it if there is a reason to do so. But there is a cost, so there should be some justification for the change.

Either the documentation should be changed or the behavior should be changed. Let's talk about which should be changed.  Why do you need this behavior?

e10s:
Yes, the inconsistency is the problem. I'm sorry for my lack of explanation. It bothered me that the change in #2209 had been never covered by any unit tests, so I judged the old behavior before the change was what it should be and I thought `nodes` lost its raison d'être except to create a subgraph.

Now that it seems your first suggestion will make me happy, I don't think my proposal is the best solution for this problem.

By the way, I have another idea. That is, both the documentation and the behavior should be changed. For example:
```python
def to_graph6_bytes(G, nodes=None, header=True, consider_order=False):
    if nodes is not None:
        if consider_order:
            H = G.__class__()
            H.add_nodes_from(nodes)
            H.add_edges_from((u, v) for (u, v) in G.edges() if u in H if v in H)
            G = H
        else:
            H = G.subgraph(nodes)
            G = H
    # snip
```
This looks a little ugly but will keep backward compatibility. Now, `nodes` has a only one purpose if `consider_order` is `False`. It's for just making an induced subgraph by itself and its nodes will be enumerated in order of _subgraph_`.nodes()` as a subsequence of its _supergraph_`.nodes()`, in the same manner as the current codes. Or, if `True`, `nodes` will have two purposes. A subgraph is made in a different way so that enumeration will be performed by  _subgraph_`.nodes()` in order of `nodes`, which will be regarded as a new feature rather than a breaking change by users who are familiar with the existing behavior. `consider_order` is a bad name, though.

Issue #6279, opened 2022-12-14
Error in documentation of katz_centrality? 


erikkemperman:
### Current Behavior

The docs for katz centrality state:
https://github.com/networkx/networkx/blob/f20ba5f2f6462af8f1ac9b1772ffa5b4bcca4d58/networkx/algorithms/centrality/katz.py#L131-L132

The value for `beta` here was changed from 1 to 0 a long time ago, in this commit:
https://github.com/networkx/networkx/commit/b33ebd46a32558b1f33bd3caa6030e21502e17bf#diff-fd4d6088095ce7547e9e07adf7326d408cb55f6ef43d47aeba7f53d405b44b8dL259

Which I think may have been an incorrect response to issue https://github.com/networkx/networkx/issues/1247.

### Expected Behavior

I believe that the previous note with `beta=1` was actually correct. For what it's worth, here are the relevant parts of the implementation in current master:

https://github.com/networkx/networkx/blob/6971d845e25aa4b76879831bb85ac2fa23ae0c9e/networkx/algorithms/centrality/katz.py#L153-L166

If we assume that `nstart` is left to the default of `None`, we see that `x` starts out being all zeroes. But then, unless `beta` is non-zero for at least some of the nodes, it is impossible for any element of `x` to ever become non-zero:

https://github.com/networkx/networkx/blob/6971d845e25aa4b76879831bb85ac2fa23ae0c9e/networkx/algorithms/centrality/katz.py#L168-L177

In other words, it looks to me like if `nstart=None` and `beta=0` then Katz centrality will always produce an array of zeroes. And in particular, even if `alpha` is (ever so slightly less than) `1/lambda_max` it will *not* resemble the eigenvector centrality.

### Steps to Reproduce

(EDITED) Figured I should add an example, adapted from the example in the docs:

```python3
import math
G = nx.path_graph(4)
phi = (1 + math.sqrt(5)) / 2.0  # largest eigenvalue of adj matrix

print(f"For reference: eigenvector centrality")
eig_centrality = nx.eigenvector_centrality(G)
for n, c in sorted(eig_centrality.items()):
    print(f"{n} {c:.2f}")
# 0 0.37
# 1 0.60
# 2 0.60
# 3 0.37

print(f"Katz with beta=1.0")
katz_centrality_0 = nx.katz_centrality(G, alpha=(1 / phi - 0.01), beta=1.0)
for n, c in sorted(katz_centrality_0.items()):
    print(f"{n} {c:.2f}")
[...]

PR #6294, opened 2022-12-21
Fixed an error in the documentation of the katz centrality

BrunoBaldissera:
Issue  #6279 reported a sentence in the documentation of the method stating

>  When $\alpha = 1/\lambda_{\max}$ and $\beta=0$, Katz centrality is the same 
>      as eigenvector centrality. 

I replaced $\beta=0$ with $\beta=1$, according to the issue's suggestion (and fixed a small typo along the way).

### DIFF
--- a/networkx/algorithms/centrality/katz.py
+++ b/networkx/algorithms/centrality/katz.py
@@ -59,7 +59,7 @@ def katz_centrality(
 
     beta : scalar or dictionary, optional (default=1.0)
       Weight attributed to the immediate neighborhood. If not a scalar, the
-      dictionary must have an value for every node.
+      dictionary must have a value for every node.
 
     max_iter : integer, optional (default=1000)
       Maximum number of iterations in power method.
@@ -128,7 +128,7 @@ def katz_centrality(
     The iteration will stop after ``max_iter`` iterations or an error tolerance of
     ``number_of_nodes(G) * tol`` has been reached.
 
-    When $\alpha = 1/\lambda_{\max}$ and $\beta=0$, Katz centrality is the same
+    When $\alpha = 1/\lambda_{\max}$ and $\beta=1$, Katz centrality is the same
     as eigenvector centrality.
 
     For directed graphs this finds "left" eigenvectors which corresponds
--- a/networkx/algorithms/centrality/katz.py
+++ b/networkx/algorithms/centrality/katz.py
@@ -128,11 +128,11 @@ def katz_centrality(
     The iteration will stop after ``max_iter`` iterations or an error tolerance of
     ``number_of_nodes(G) * tol`` has been reached.
 
-    When $\alpha = 1/\lambda_{\max}$ and $\beta=1$, Katz centrality is the same
-    as eigenvector centrality.
+    For strongly connected graphs, as $\alpha \to 1/\lambda_{\max}$, and $\beta > 0$,
+    Katz centrality approaches the results for eigenvector centrality.
 
     For directed graphs this finds "left" eigenvectors which corresponds
-    to the in-edges in the graph. For out-edges Katz centrality
+    to the in-edges in the graph. For out-edges Katz centrality,
     first reverse the graph with ``G.reverse()``.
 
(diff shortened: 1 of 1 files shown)


dschult:
There is another typo in a comment on line 172:   It has the formula as "-" Beta when the code clearly uses "+".

Thanks for this!

Qudirah:
@BrunoBaldissera are you done with this issue?

vigna:
For $\alpha = 1/\lambda_\mathrm{max}$ the summation does not converge. Convergence is possible only when $\alpha < 1/\lambda_\mathrm{max}$.

In the limit case, for $\alpha \to 1/\lambda_\mathrm{max}$ there is some theory that will tell you where Katz index is going, and in particular for strongly connected graphs (of at least for a semisimple $\lambda_\mathrm{max}$) it is the dominant eigenvector, but when the graph is not strongly connected there are many such vectors and the situation is a bit complicated. I know because I [proved the result in full generality](https://www.cambridge.org/core/journals/network-science/article/spectral-ranking/99ACDCD0CC1B774AB0041FB16AB43D1B). What you get is the projection of the preference vector onto the eigenspace.

dschult:
Excellent @vigna 
Can you suggest some wording here relating it to the eigenvector centrality? Or is that difficult....

Maybe something like:  For strongly connected graphs, as $\alpha \to 1/\lambda_\mathrm{max}$ and $beta=1$ this approaches the result for eigenvector centrality.

vigna:
OK, I just read the definition and actually it's a bit weird but the original statement was correct. If $\beta=0$, this is not Katz centrality. It is eigenvector centrality, period. And it will have a solution only if $\alpha$ is the reciprocal of an eigenvalue. Simply, in the definition, it should be $\beta > 0$.

What is true is when $\beta > 0$ if you let $\alpha$ go to the limit value you will tend in the direction of the dominant eigenvector, or more generally in the direction of the projection of the preference vector onto the dominant eigenspace. (You'll need to renormalize tho as the vector norm will go to infinity).

BTW, $\beta$ can be different for every node, _and the code allows that_, so probably it would be better to have a definition matching the code, with a parameter $\beta_i$ instead of the same $\beta$ for all nodes.

Said that, I've always seen Katz centrality defined as $\beta \sum_{n\geq 0}A^nx^n$ (here $\beta$ is a vector but I can't bold it), and while this definition is equivalent, I don't think it is really idiomatic. The original formulation by Katz is in them of an attenuated path summation (voters of voters of voters...). The linear recurrence in the paper is derived much later.

vigna:
> Excellent @vigna Can you suggest some wording here relating it to the eigenvector centrality? Or is that difficult....
> 
> Maybe something like: For strongly connected graphs, as α→1/λmax and beta=1 this approaches the result for eigenvector centrality.

Yes, this is correct for every $\beta\neq 0$. If you want a precise explanation for the non-strongly-connected case just put a reference to my paper, it is definitely _not_ something you want to explain in general documentation...

dschult:
It can be merged once it is clear what the good resolution of the issue is. Can you suggest some wording for the docs to make this all clear? The discussion above makes me believe that the change to the documentation should not be `beta=1`. But there's a lot of other information you added that is not included in the doc_string. If you can write out what you think this change ought to be then it would help move this PR closer to merging.

vigna:
Oops. I didn't even realize this is not my branch. I guess I should fork from this branch or something like that?

dschult:
You could load the branch by making a git "remote" to it, but you still couldn't push to it. You could make a PR to the PR-branch, but that still relies on the original author responding.  Another approach would be to pull the branch from a git remote and then add some commits, push it to your fork of networkx and make a new PR. But maybe that's more than needed. 

How complex are the changes you are thinking about? Can you make a suggestion in the comments?
I can make changes to the PR if you want to make a suggestion.

dschult (approved):
Apologies for taking so long to get to this documentation change.
I think it is all ready to merge.

jarrodmillman (approved):

PR #6295, opened 2022-12-24
Isomorphism improve documentation

paulitapb:
I added some more details to the functions `could_be_isomorphic` and `fast_could_be_isomorphic`.


networkx/algorithms/community/__init__.py
```diff
@@ -1,9 +1,10 @@
...
 
-The functions in this class are not imported into the top-level
-:mod:`networkx` namespace. You can access these functions by importing
-the :mod:`networkx.algorithms.community` module, then accessing the
-functions as attributes of ``community``. For example::
+.. warning:: The functions in this class are not imported into the top-level
+:mod:`networkx` namespace. 
+
+They can be imported using the :mod:`networkx.algorithms.community` module,
+then accessing the functions as attributes of ``community``. For example::
```
dschult, line 7: Users can now access these function via e.g. `nx.community.girvan_newman` without a new import statement.
Should we advertise that more in places like this? 

We should go through the codebase and update claims like these now that we are using lazy_loading. But that (and maybe this comment) belong in a different PR.

networkx/algorithms/isomorphism/isomorph.py
```diff
@@ -24,6 +24,10 @@ def could_be_isomorphic(G1, G2):
     Notes
     -----
     Checks for matching degree, triangle, and number of cliques sequences.
+    The triangle sequence contains the number of triangles each node is part of.
+    The clique sequence contains for each node the size of the maximal clique
+    involving that node.
```
dschult, line 29:
```suggestion
    The clique sequence contains for each node the number of maximal cliques
    involving that node.
```
Usually, a node will be part of more than one maximal clique. A maximal clique cannot add any nodes while remaining a clique. That is different from the maximum clique for a node which is the largest of all the maximal cliques.  This function checks the number of maximal cliques in which this node participates.

dschult (commented):
Thanks for this!  Looks good!  I have two comments below.

networkx/algorithms/community/__init__.py
paulitapb: This change does belongs to a different PR. I'll go ahead and reverse it

dschult (approved):
This looks good to me!

MridulS (approved):

PR #6296, opened 2022-12-24
Broken link in isomorphism documentation

paulitapb:
As I was reading the documentation I found that the link in the reference section of `is_isomorphic` is broken. I replaced the current link with a new version but I'm not sure if it's the right one.  The paper seems to be the same. Maybe the citation format is not right.


MridulS (commented):
Looks like this is change from a wrong branch?

The changes aren't for the isomorphism docs.

paulitapb:
> Looks like this is change from a wrong branch?
> 
> The changes aren't for the isomorphism docs.



> Yes, It seems this is on my main branch. I will fix that.

rossbar (approved):
The new link works and links to the correct doc so LGTM, thanks @paulitapb !

PR #6299, opened 2022-12-25
Improve _single_shortest_path_length function

Tortar:
Hi! I have modified the function _single_shortest_path_length making it more performant. I was inspired by the first implementation in #3647, because seemed better to me not only in performance but also in the fact that it yielded the nodes in a reproducible fashion because passing nodes to set was avoided. I improved that version and recalculated the times of execution agains the current main with the following code:```

```
graphs = [  "nx.random_regular_graph(d=1, n=10, seed=4241)",
            "nx.random_regular_graph(d=9, n=10, seed=4241)",
            "nx.random_regular_graph(d=10, n=100, seed=4241)",
            "nx.random_regular_graph(d=99, n=100, seed=4241)",
            "nx.random_regular_graph(d=1, n=1000, seed=4241)",
            "nx.random_regular_graph(d=3, n=1000, seed=4241)",
            "nx.random_regular_graph(d=10, n=1000, seed=4241)",
            "nx.random_regular_graph(d=50, n=1000, seed=4241)",
            "nx.random_regular_graph(d=100, n=1000, seed=4241)",
            "nx.random_regular_graph(d=200, n=1000, seed=4241)",
            "nx.random_regular_graph(d=500, n=1000, seed=4241)",
            "nx.random_regular_graph(d=700, n=1000, seed=4241)",
            "nx.erdos_renyi_graph(n=100, p=0.9, seed=4241)",
            "nx.watts_strogatz_graph(n=100, k=10, p=0.3, seed=4241)",
            "nx.connected_caveman_graph(100,10)",  
            "nx.caveman_graph(100,10)",  "nx.windmill_graph(10, 10)",       
            "nx.full_rary_tree(3, 1000)",  "nx.complete_graph(100)",
            "nx.complete_graph(10)",  "nx.cycle_graph(100)",
            "nx.cycle_graph(10)" ]

stmt_1 = "list(_single_shortest_path_length(G.adj, {random.randrange(len(G)-1): 1}, float('inf')))"
stmt_2 = "list(_single_shortest_path_length_2(G.adj, {random.randrange(len(G)-1): 1}, float('inf')))"
values = {}
for graph in graphs:
    setup_1 = "import networkx as nx; import collections, random; from __main__ import _single_shortest_path_length; random.seed(4241); G = {}".format(graph)
    setup_2 = "import networkx as nx; import collections, random; from __main__ import _single_shortest_path_length_2; random.seed(4241); G = {}".format(graph)
    main_time = timeit.timeit(stmt_1, setup_1, number=20000)
    pr_time = timeit.timeit(stmt_2, setup_2, number=20000)
    values[main_time/pr_time] = "{} : {}, {}, {}".format(graph, round(main_time,2), round(pr_time,2), round(main_time/pr_time, 2))
for x in values.keys(): print(values[x])
```

giving the following result:

```
+--------------------------------------------------------+--------------------+---------------+-------------+
|                         graph                          | time: current main | time: this pr | times ratio |
+--------------------------------------------------------+--------------------+---------------+-------------+
|     nx.random_regular_graph(d=1, n=10, seed=4241)      |        0.05        |      0.04     |     1.17    |
|     nx.random_regular_graph(d=9, n=10, seed=4241)      |        0.06        |      0.06     |     1.05    |
|    nx.random_regular_graph(d=10, n=100, seed=4241)     |        0.87        |      0.69     |     1.26    |
|    nx.random_regular_graph(d=99, n=100, seed=4241)     |        0.28        |      0.27     |     1.05    |
|    nx.random_regular_graph(d=1, n=1000, seed=4241)     |        0.06        |      0.05     |     1.22    |
|    nx.random_regular_graph(d=3, n=1000, seed=4241)     |       10.86        |     10.32     |     1.05    |
|    nx.random_regular_graph(d=10, n=1000, seed=4241)    |        8.74        |      9.12     |     0.96    |
|    nx.random_regular_graph(d=50, n=1000, seed=4241)    |       16.12        |      6.74     |     2.39    |
|   nx.random_regular_graph(d=100, n=1000, seed=4241)    |        5.53        |      6.24     |     0.89    |
|   nx.random_regular_graph(d=200, n=1000, seed=4241)    |        10.5        |      5.76     |     1.82    |
|   nx.random_regular_graph(d=500, n=1000, seed=4241)    |       44.36        |      4.86     |     9.13    |
|   nx.random_regular_graph(d=700, n=1000, seed=4241)    |       82.79        |      4.36     |    18.97    |
|     nx.erdos_renyi_graph(n=100, p=0.9, seed=4241)      |        1.91        |      0.32     |     5.93    |
| nx.watts_strogatz_graph(n=100, k=10, p=0.3, seed=4241) |        0.92        |      0.82     |     1.13    |
|           nx.connected_caveman_graph(100,10)           |       11.66        |     13.37     |     0.87    |
|                nx.caveman_graph(100,10)                |        0.14        |      0.16     |     0.85    |
|               nx.windmill_graph(10, 10)                |        0.34        |      0.33     |     1.01    |
|               nx.full_rary_tree(3, 1000)               |        6.41        |      6.81     |     0.94    |
|                 nx.complete_graph(100)                 |        0.28        |      0.27     |     1.04    |
|                 nx.complete_graph(10)                  |        0.06        |      0.05     |     1.07    |
|                  nx.cycle_graph(100)                   |        1.18        |      0.96     |     1.23    |
|                   nx.cycle_graph(10)                   |        0.13        |      0.1      |     1.23    |
+--------------------------------------------------------+--------------------+---------------+-------------+
```

As you see the new implementation is marginally better for many cases, but there are also some cases where the time of execution is more than an order of magnitude less than the current one. I want also to stress the point about reproducibility which can be very useful in some cases (I actually tried to optimize this algorithm for this reason!).


Tortar:
I now realized that the reason why this algorithm works much better in some cases is that the while loop works on a node at a time, instead of a level at a time as was done before, so it can happen that the last level is explored much less than in the current algorithm, which requires to build the `nextlevel` set in its entirety.

dschult:
Thanks for this!

Do you have some ideas about why this is faster than what we compared 3 years ago? Some things I see:
- You've made it so that only nodes not in `seen` get added to the list for the next level. This avoids the need for `nextlevel` to be a `set` because the same node is never added to the queue a second time. That's what allows you to get the preserved node order.  
- I think eliminating nodes in seen from going into the queue may also lead to the large speed up for dense graphs. You have essentially pushed the collection of the next level nodes to the next iteration of the loop. That allows processing the new nodes as soon as you get them from `adj[node]`. The current code collects them one level earlier, but still processes them later, so they must get stored in "nextlevel" in the meantime. Basically, the current code adds the same node to the queue many times and that increases processing. 
- It also looks like you have eliminated the set/list creation needed in the current version for "nextlevel" and "found". But I can't think that takes much time since that is only done once per level.
- In the first level, you avoid all the checking of nodes in seen and adding them to the queue by preallocating them.  Again I don't think that saves a huge amount of time. 

I wonder if checking the speed of the current code with a change of `nextlevel.update(adj[v])` to `nextlevel.update(adj[v].keys() - seen.keys())` would be comparable.  But I doubt it. The new code has a "just-in-time" feel to it.  In that sense, could you try in your new code changing:
```python
           for v in adj[node]:
                 if v not in seen:  
```
to
```python
            for v in adj[node].keys() - seen:
```

Nice work!  Thanks!

Tortar:
Your second reasoning strikes me as the correct cause of the large speed-up for dense graphs! Your other points are nonetheless important for keeping the performance comparable/a bit better for sparse graphs.

Tried to change in `for v in adj[node].keys() - seen:`, it degredes the performance using the Erdős-Rényi graph of 500 nodes in relation to the probability of edge creation : now the speed-up  is from 0.4 to 17 while before it was from 0.95 to 29.5. I think that the reason is that the reason is that it's more costly to make a difference between two sets and than to iterate over the result than the current way because it is a sort of double iteration.

Tortar:
Ops, sorry maybe I misunderstood what you asked about changing the code, I think what you meant was

- Change the current code with `nextlevel.update(adj[v].keys() - seen.keys())`
- Change the new code with `for v in adj[node].keys() - seen:` 

I did this and made the comparison: the new code is from 1.25 to 107 times faster (also consider that the new code with this modification is actually worse than the one without it as I previously calculated)

dschult:
Interesting!  Thanks...  
So, I think that means the new code with `for v in adj[node].keys() - seen` is the fastest of these tests -- but I'm losing track of which base is being used for the 1.25 to 107 times faster. :}  Whichever is the best, lets go with that. :)

Tortar:
The fastest is actually the current new one,  the one with `for v in adj[node].keys() - seen` is actually 2x slower than the current new one as I tried to explain in the comment where I misunderstood what you were saying, if it wasn't clear it's probably because I'm not a native english speaker so I didn't explain myself rightly :-)

So let's go with this one!

dschult (approved):
Your english is fine!  :}   I just didn't know if the 1.25 to 107 times faster was faster than the original code, or the PR or the tweaked version of the original code.  (faster than what?)...

Thanks for checking the relative speeds of these different approaches. And thanks for all the work timing and trying the different versions.

I approve this PR.      But -- now I wonder if the same sort of changes would help the `_single_source_shortest_path` function.   :)  

Thanks again!

Tortar:
it was that faster in relation to the current code modified with `nextlevel.update(adj[v].keys() - seen.keys())`, so this slowed down that code more than changing the new one with `for v in adj[node].keys() - seen:`.

Yes, I think that you are totally right: I also realized it when looking at the code and I think I will try to optimize also those parts in the near future!

Tortar:
ok, found a way with the more primitive data structures (no deque) which improve a bit the speed , so I change with this version!

dschult:
This looks even better than the previous version. It is easy to read with no figuring out whether the deque is getting emptied correctly.  The main part of the speedup in this PR is making sure not to add previously seen nodes to the queue. This version also removes some extra data structures by refactoring the loop. Should be easier for people to review too.

I reaffirm my approval of this PR. :}

dschult:
@mriduls and @rossbar  Might there be a chance to get this reviewed and into v3.0?  It could speed up many algorithms that use shortest_path_length. Also I think the changes are short enough to review fairly easily.  If not that's OK.  We will expand these ideas to e.g. shortest_path and maybe others and can make it a plus for the whole module in v3.1.
:)

Tortar:
Hi @dschult, thank for the support for this PR. I actually tried the same approach also to shortest_path here, but some of the changes applied here were already present in shortest_path, I'd like to ask you what do you think of this one https://github.com/networkx/networkx/compare/main...Tortar:networkx:patch-4 ?

dschult:
I agree that it is less worthy. But what do the timings say? 
It does cut the final level once all nodes have been seen.
But rather than converting the input to a list, we should change all the calling functions so they create a list rather than a dict. But I'm not sure that is worth doing either.  How can we tell if it makes a difference?

But that patch should be a different PR anyway.

Tortar:
Created a PR with the benchmark!

dschult:
I'd be interested in seeing a comparison between using passing `G.adj` into this function and `G._adj`. See #4895 for an issue discussing this.  Because `G.adj` is a view on `G._adj` it is slower, and since this function never updates G there is no risk of exposing the internal data structure to corruption. Also, this code might really benefit from the direct access to the dict-of-dict storage.

networkx/algorithms/shortest_paths/unweighted.py
```diff
@@ -72,25 +72,24 @@ def _single_shortest_path_length(adj, firstlevel, cutoff):
...
+        thislevel = nextlevel
+        nextlevel = []
+        for v in thislevel:
+            for w in adj[v]:
+                if w not in seen:
+                    seen.add(w)
+                    nextlevel.append(w)
+                    yield (w, level)
+            if len(seen) == n:
+                return
```
dschult, line 92: I was thinking about the `del seen` code that has been removed in this PR.  I think we need it to avoid large memory hoarding by the (potentially many) exhausted iterator/generator.  Granted this is only a problem if the user code is calling this many times and not releasing the references to the generator. But I can imagine that happening, and that was the concern driving having `del seen` in the current code.

The potential problem is that a large set `seen` is created in the generator. If we don't delete that set before `return`, the generator stops yielding results, but it can avoid "garbage collection" if any references to the generator in the user code aren't removed. Usually keeping exhausted iterators is not a memory problem because they are small. But in this can `seen` could be fairly large.  

What do you think?  The fix would be:
```suggestion
            if len(seen) == n:
                del seen
                return
    del seen
```

Tortar: mmh I don't think it's useful to have that because when you hit return in a generator it raises StopIteration (see https://stackoverflow.com/questions/16780002/return-in-generator-together-with-yield) which starts the garbage collection.

As an example I made:

```
def gen():
    a = [i for i in range(10**6)]
    for x in a:
        if x != 1:
            yield x
        else:
            return

k = gen()
next(k) # memory usage increases
next(k) # memory is freed
```

Am I misunderstanding your point?

Tortar:
wow! `G._adj` increases the speed by a good amount (1.5-2x on the current implementation). Will change the code to call this!

networkx/algorithms/shortest_paths/unweighted.py
dschult: Unless something has changed in python since I last looked at this (a long time ago), the memory is not freed when an iterator is exhausted. In your example, the memory would not be freed until the variable `k` releases its reference to the iterator object. For example if you did `del k` then it would release the memory used by the list `a`. Certainly the variable `k` still exists after the StopIteration and you can call `next(k)` again even after the StopIteration has been raised.  But maybe Python is smart enough now to delete the local namespace of the instance when the StopIteration is raised. But I don't think so. That could cause problems with generators that get restarted.

Tortar: Or are there any subtlety I don't see? Can you point me out a concrete example if this is the case? :-)

dschult: You are right!   I think this must be an out-of-date understanding that I had about generators. It looks like StopIteration releases all the refcounts within the generator.

Thanks!  I learned something ---   makes me wonder if there are other `del something` in the code base. I'll search for them... :)

Tortar: glad to know it! :-)

dschult:
This is ready for another set of eyes. I think it is ready for merging, with perhaps discussion/thought about
1) whether to include the changes of `G.adj -> G._adj` in this PR or to make them more pervasively in the codebase as a separate PR.
2) whether  to include the other BFS improvement PRs together with this one, or to keep them separate.

I am in favor of including the `G.adj -> G._adj` change here as a proof of concept and experiment for how and whether it impacts people.  I am in favor of keeping the various BFS improvements in separate PRs because it makes review of each one easier. But I admit there are positives for doing it the other way on both of these issues.

Tortar:
Waiting for it! I'm favor of keeping all as it is now to make the review process faster as you said :-)

rossbar (approved):
The changes to code LGTM. I didn't try to reproduce the benchmark results, though it'd be nice to do so. As noted in the discussion, most of the changes seem like they should result in speedups.

@MridulS I'd be interested to see benchmarking results from your setup if it's not difficult to do. Otherwise LGTM, thanks @Tortar !

Tortar:
Any update on how to finalize with the optimizations I included in the PRs I opened @rossbar @dschult @MridulS ? I'd like to see them included (or rejected) since they took me a bit of time to work them out :-) Let me know if squashing some of them can be easier to make them more "reviewable"

PR #6306, opened 2022-12-28
feat: prototype of higher-level API and helper functions

dPys:
* Addresses issue #4064 by providing a much-needed parallelization API (`NxParallel`) for Networkx that aspires to be backend-agnostic by leveraging a joblib framework.
* Supports default multiprocessing, loky, and threading backends, and by extension ipyparallel, dask, and ray backends with associated optional imports as well.
* The API avoids the need to write redundant _parallel variants of each parallelizable algorithm. Rather, to minimize footprint and avoid breaking changes across the repo, capitalizing on the API to parallelize existing or future NetworkX algorithms would simply involve adding an additional kwarg to the algorithm of interest, where the kwarg consumes a single callable instance generated by `NxParallel`.
* Harmonizes the efforts in several PR's:
  - https://github.com/networkx/networkx/pull/2265
  - https://github.com/networkx/networkx/pull/3440
  - https://github.com/networkx/networkx/pull/3581
  - https://github.com/networkx/networkx/pull/3587
  - https://github.com/networkx/networkx/pull/4007
 * Builds upon the work done in `https://github.com/networkx/networkx/blob/master/examples/advanced/plot_parallel_betweenness.py`
 * Includes unit testing.
 * Something like `pip install networkx[parallel]` might be a nice bottom-line for ultimately addressing the added joblib dependency, but the current dependency handling in this PR should be considered a stub, and will likely still need to undergo strategic planning by the NetworkX core team.
 * Pending any other necessary revisions, this PR will still need to add a brief developer guide + example of usage before it can be merged.
 
Happy new year,
 @dPys


MridulS:
Thanks for working on this @dPys and getting the conversation started again! One recurrent theme with discussions about adding parallelization in networkx is not adding any dependencies which will make our codebase less readable. This approach really does help with that but in this case we will still need to maintain these parallel implementations inside networkx (like the one presented for `betweenness_centrality`).

I am not sure if you have seen this yet but we have added an experimental plugin backend system, where users can swap in a totally different backend for computations but keep using the networkx API. Some documentation is available at https://networkx.org/documentation/latest/reference/classes/index.html#backends, this was added in #6000. This is currently used with https://github.com/python-graphblas/graphblas-algorithms to dispatch a certain set of algorithms to GraphBLAS.

If we use this plugin infrastructure we can write a new package, let's call it `nx_parallel`. Now this package can have the implementations for parallel networkx code, and we don't even need to worry about the dictionary graph data structure of networkx, backends can have different data structures. They just need to follow certain guidelines like should be able to import and export from networkx graph object and have a `__networkx_plugin__` attribute. This opens up an opportunity to write a much more performant parallel implementation (if anyone fancies that :D )

Just an example of how this will work for a user:
``` python
In [1]: import networkx as nx; import nx_parallel

In [2]: G = nx.erdos_renyi_graph(10, 0.5)

# You need to convert your networkx graph object specifically to a parallel graph object
In [3]: H = nx_parallel.ParallelGraph(G) 

# networkx can take a ParallelGraph object and dispatch it to the dask implementation of betweenness_centrality
In [4]: nx.betweenness_centrality(H) 

Out[4]: 
{0: 0.0,
 1: 0.0,
 2: 0.0,
 3: 0.0,
 4: 0.0,
 5: 0.0,
 6: 0.0,
 7: 0.0,
 8: 0.0,
 9: 0.0}
```
I have made a rough sketch of this package here https://github.com/MridulS/nx_parallel, let me know if this makes sense!

MridulS:
Closing this as the discussion has moved to https://github.com/networkx/nx_parallel 

Thanks @dPys!

PR #6309, opened 2022-12-29
Improving the Time and Space Complexity of the Tree Isomorphism Algorithm.

maucarrui:
Hello. I've recently finished my Bachelor of Science Thesis on Computer Science were I studied, implemented and proved the correctness of several algorithms with some heavy applications on graph theory. One of the algorithms that I worked with was the Tree Isomorphism Algorithm first presented by A. V. Aho, J. E. Hopcroft and J. D. Ullman. 

The current implementation of this algorithm takes O(n log n) time to return an isomorphism. In this pull request I propose an implementation which takes O(n) time and space. 

# Algorithm for Rooted Trees

The main idea of the algorithm to tell whether two rooted trees are isomorphic is explained in the book "Algorithms on Trees and Graphs" by G. Valiente on pages 151-166. The idea somewhat similar to the one you have implemented. In this case, we check that each level of the rooted trees have the same "structure". Informally, a structure of a vertex v is the "form" the sub-tree rooted on v has.

In the implementation we assign every vertex a value that corresponds to its structure it has on the level. In the beginning, the leaves of the tree are the only ones with a value, this value is defined as 0. The structure of a vertex v is then defined as the multiset that contains the values the values of v's children. Then, for each structure present in a level, we assign a value to each unique structure present.

Thus, we check in each level that all the structures present in the level of one tree are also present on the other tree. If for some level we notice that the levels are different, then we stop the execution and return false and an empty isomorphism. If all levels coincide then we can safely say that both trees are isomorphic.

To determine if two levels have the same structure we have to compare all of its structures. The best way to obtain this is by "sorting" these structures and then comparing that they're equal. However, we must first define a linear way to sort this structures. The algorithm to sort the previous structures taking O(n) time and space is presented on "The Design And Analysis of Computer Algorithms" by A. V. Aho, J. E. Hopcroft and J. D. Ullman on pages 78-84.

This way we now have a linear algorithm to tell if two rooted trees are isomorphic. 

# Algorithm for an Arbitrary Tree Isomorphism

We now have to obtain an algorithm to tell if two non-rooted trees are isomorphic. To do the previous we have to consider a "good" root for our trees, as trees can only have 1 or 2 centers then its a good idea to root the centers and then tell if these rooted trees are isomorphic. I couldn't find in the library a function to find the centers of a tree in linear time. So I implemented a function in which in two BFS traversals can find the centers of a tree, this function takes O(n) time and space.

Now that we have our centers, we can root our trees and use our previous algorithm to determine if there exists an isomorphism for the trees. In total this operations take O(n) for each center, thus the algorithm in total takes O(2n) = O(n).

# Implementation notes

As I've mentioned before, the structure of a vertex is represented as a multiset, thus the PR contains a class `NaturalMultiset` which contains only the necessary operations for the main algorithm to work. After the multiset class we add some functions that mimic the algorithms presented on "The Design And Analysis of Computer Algorithms". The algorithms related to multisets and the ordering of these structures are:

1. `NaturalMultiset`

- `add`
- `contains`
- `to_list`
- `from_list`
- `__eq__`
- `__len__`
- `__hash__`
- `__str__`

2. `categorize_entries`
3. `categorize_lists`
4. `sort_lists_of_naturals`
5. `sort_natural_multisets`

We now have the functions that mimic the algorithm presented on "Algorithms on Trees and Graphs".

1. `get_levels`
2. `get_initial_values`
3. `get_initial_children`
4. `assign_structure`
5. `get_multisets_list_of_level`
6. `update_values`
7. `build_isomorphism`
8. `levels_verification`
9. `get_height`
10. `get_parenthood`
11. `rooted_tree_isomorphism_n`
12. `get_centers_of_tree`
13. `tree_isomorphism_n`

# API changes

The proposed main algorithms returns two values. The first one is a boolean to indicate is the trees were indeed isomorphic or not. And the second one, an isomorphism **mapping**. 

The reason I return a map is because I consider them to be more accurate to the mathematical definition of an isomorphism: A mapping of the vertices of T1 onto the vertices of T2. And the reason I return two values is because one may only be interested in considering when two trees are isomorphic. Additionally, if the passed trees are empty then the returned isomorphism would be empty, which could be considered as that they're not isomorphic; a false negative.
However, all of these are my opinions on the matter and I have no problem with it behaving the same way as the original.

If the previous changes are accepted, then I would request your help to update the file `doc/release/release_dev.rst` because I didn't understand fully how to indicate the changes proposed here.

# Unit tests

Most of the previous functions have non-trivial unit tests assigned to them. The ones that don't have one are tested in another more complex unit test.

# Documentation and Format

All of the previously mentioned functions are documented following the Numpy Style Guide. With the only exception that it didn't specify how to write dictionaries, in those cases I used the Google Docstring Guide. The unit tests also have a brief comment on what they do and what function are they testing. All of them have been formatted with the Python formatter, black.

# References

I added 2 additional references for the algorithms implemented. The first one is the Book by G. Valiente "Algorithms on Trees and Graphs". And the second one is my thesis, were I explain the algorithms with more detail, prove the time and space complexity, and provide a low level implementation for each one; the implementations proposed here are based on the ones I have on my thesis. I hope there is no problem with adding my thesis as a reference here. Unfortunately, due to some bureaucratic timings, my thesis is yet to be archived on the Internet. In case you want to read or verify that I've actually studied these algorithms you can send an email to my account maucarrui@ciencias.unam.mx.

# Performance

Now, the results you've been waiting for, how much of a difference is there between the current algorithm and the one proposed? Beforehand, sorry for the bad formatting, but I hope it is clear enough.

## Positive Tests

The following test takes all the non-isomorphic trees of size k, let T be one of these trees. Then create a "copy" of T, modify it in some way (i.e. change the nodes order or edges order) and then verify that both trees are isomorphic. As the created tree is a "copy" of the original, then it should always yield an isomorphism. The leftmost number is the number of vertices in the trees, the second number is the amount of trees for which the previous process was applied, and the last number is how long it took to get an isomorphism for all of these trees.

The following times are with the current algorithm.
```
2 1 0.0004029273986816406
3 1 0.00027370452880859375
4 2 0.0005514621734619141
5 3 0.0015108585357666016
6 6 0.002101898193359375
7 11 0.00490117073059082
8 23 0.014166593551635742
9 47 0.030213117599487305
10 106 0.07018899917602539
11 235 0.16405081748962402
12 551 0.40637993812561035
13 1301 1.086066484451294
14 3159 2.883446455001831
15 7741 7.693917274475098
16 19320 20.695265531539917
17 48629 57.28187799453735
```

With the proposed algorithm.
```
2 1 0.00026869773864746094
3 1 0.0002827644348144531
4 2 0.0006053447723388672
5 3 0.0017850399017333984
6 6 0.0030426979064941406
7 11 0.0047910213470458984
8 23 0.014776468276977539
9 47 0.02821493148803711
10 106 0.06437063217163086
11 235 0.14986014366149902
12 551 0.38997745513916016
13 1301 0.9029171466827393
14 3159 2.2917373180389404
15 7741 6.050358295440674
16 19320 16.294824361801147
17 48629 44.54627561569214
```

For k's greater than 14 there starts to be a noticeable difference between the algorithms.

## Negative tests.
The following test takes all the non-isomorphic trees of size k, let T be one of these trees. Takes every possible pair of these trees and checks that the algorithm effectively returns that they're different.

The following times are with the current algorithm.
```
4 1 0.00017452239990234375
5 3 0.00015354156494140625
6 15 0.0010170936584472656
7 55 0.0031976699829101562
8 253 0.018256664276123047
9 1081 0.07431364059448242
10 5565 0.3633444309234619
11 27495 1.8973579406738281
12 151525 9.942706108093262
13 845650 57.08123278617859
14 4988061 346.85636019706726
```

The following times are with the proposed algorithm.
```
4 1 8.296966552734375e-05
5 3 0.00012087821960449219
6 15 0.0005776882171630859
7 55 0.0028438568115234375
8 253 0.012853145599365234
9 1081 0.057640790939331055
10 5565 0.28786420822143555
11 27495 1.4543352127075195
12 151525 8.296680212020874
13 845650 45.35695219039917
14 4988061 277.9806983470917
```

For k's greater than 11 there starts to be a noticeable difference.

## Reproduce these tests

In the PR I left all of the previous functions, this way you can run these tests and verify by yourself the performance of both algorithms.

# Further questions

Please, if you have any question about how the algorithm works, don't doubt in writing a comment about it. I'll try my best to explain it.

### DIFF
--- a/networkx/algorithms/isomorphism/tree_isomorphism.py
+++ b/networkx/algorithms/isomorphism/tree_isomorphism.py
@@ -242,6 +242,80 @@ def categorize_lists(S):
             
     return LENGTH
 
+def sort_lists_of_naturals(S):
+    """
+    Given a list of lists whose entries are numbers between 0 and m. The
+    function sort_lists_of_naturals performs an algorithm similar to bucket sort
+    to return a sorted list with the elements of S. The expected order is a
+    lexicographical order.
+    
+    Parameters
+    ----------
+    S : list
+        The list of lists of natural numbers.
+    
+    Returns
+    -------
+    A sorted list with the elements of S.
+    
+    References
+    ----------
+        .. [1] A. V. Aho, J. E. Hopcroft y J. D. Ullman, The Desing And Analysis 
+               of Computer Algorithms, Addison Wesley Publishing Company, 1974.
+               Pages 80-82.
+
+        .. [4] M. Carrasco-Ruiz, Herramientas para el Manejo Computacional de
+               Gráficas, Bachelor of Science Thesis, UNAM, 2021.
+    
+    Notes
+    -----
+    This algorithm runs in O(l_total + m) time and O(l_total) space, where 
+    l_total is the sum of the lengths of the lists in S and m is the max value
+    of all the lists in S.
+    """
+    # Get the maximum value contained in S.
+    m = max([max(s) for s in S])
+
+    # Define a list that will behave like a queue.
+    Q = []
+
+    # Define an empty array of size m whose entries are empty queues.
+    A = [[] for i in range(m+1)]
+    
+    print(A)
+    
+    # The max length found in S.
+    l_max = max([len(s) for s in S])
+    
+    # Obtain the NONEMPTY and LENGTH functions.
+    NONEMPTY = categorize_entries(S)
+    LENGTH = categorize_lists(S)
+    
+    l = l_max
+    while (l >= 1):
+        # All of the elements in Q have length greater than l. For each element
+        # s in Q, append q to the queue A[s[l]].
+        while (len(Q) > 0):
+            s = Q.pop()
+            A[s[l-1]].append(s)
+
+        # Add the lists s of size l to the queue A[s[l]] if there are any.
+        if l in LENGTH:
+            for s in LENGTH[l]:
+                A[s[l-1]].append(s)
+            
+        # Sort the elements of A according to its l-th entry.
+        for i in NONEMPTY[l-1]:
+            while (len(A[i]) > 0):
+                s = A[i].pop()
+                Q.append(s)
+        
+        # Proceed with the lists of size l-1.
+        l = l - 1
+    
+    # Q now contains all of the elements of S sorted lexicographically.
+    return Q
+
 def get_levels(T, root, height):
     """
     Returns a function (map or dict) M: N -> P(V) such that, for each i in
--- a/networkx/algorithms/isomorphism/tree_isomorphism.py
+++ b/networkx/algorithms/isomorphism/tree_isomorphism.py
@@ -164,9 +164,12 @@ def categorize_entries(S):
     
     References
     ----------
-        .. [1] A. V. Aho, J. E. Hopcroft y J. D. Ullman, The Desing And Analysis 
+        .. [1] A. V. Aho, J. E. Hopcroft y J. D. Ullman, The Desing And Analysis
                of Computer Algorithms, Addison Wesley Publishing Company, 1974.
                Pages 80-82.
+
+        .. [4] M. Carrasco-Ruiz, Herramientas para el Manejo Computacional de
+               Gráficas, Bachelor of Science Thesis, UNAM, 2021.
     
     Notes
     -----
@@ -186,7 +189,6 @@ def categorize_entries(S):
         M = NaturalMultiset()
         
         for s in S:
-            print(s, i)
             # Only consider lists of length greater or equal than i, and whose
             # i-th entry is not already contained in the multiset.
             if len(s) < (i+1) or M.contains(s[i]):
@@ -220,6 +222,9 @@ def categorize_lists(S):
         .. [1] A. V. Aho, J. E. Hopcroft y J. D. Ullman, The Desing And Analysis 
                of Computer Algorithms, Addison Wesley Publishing Company, 1974.
                Pages 80-82.
+
+        .. [4] M. Carrasco-Ruiz, Herramientas para el Manejo Computacional de
+               Gráficas, Bachelor of Science Thesis, UNAM, 2021.
     
     Notes
     -----
--- a/networkx/algorithms/isomorphism/tree_isomorphism.py
+++ b/networkx/algorithms/isomorphism/tree_isomorphism.py
@@ -644,6 +644,204 @@ def update_values(S, MS, values):
         # For the next unique structure, assign a new value.
         current_val += 1
 
+def levels_verification(values_T1, values_T2, levels_T1, levels_T2, parenthood_T1, parenthood_T2, height):
+    """
+    For each level i in T1, check that all the structures present in the
+    i-th level are also present in i-th level of T2. If all the levels are the
+    same for both trees, return true.
+    
+    Parameters
+    ----------
+    values_T1 : dictionary
+        The value function for the vertices of the tree T1.
+
+    values_T2 : dictionary
+        The value function for the vertices of the tree T2.
+    
+    levels_T1 : dictionary
+        The level function for the tree T1.
+
+    levels_T2 : dictionary
+        The level function for the tree T2.
+    
+    parenthood_T1 : dictionary
+        The parenthood function defined for T1.
+
+    parenthood_T2 : dictionary
+        The parenthood function defined for T2.
+    
+    height : int
+        The height of the trees T1 and T2.
+    
+    Note
+    ----
+    This algorithm runs in O(n) time and space.
+    """
+    # Start at level 1 as all the vertices at level 0 are leaves.
+    current_level = 1
+    
+    # check that all the structures present in the i-th level are also present
+    # in i-th level of T2.
+    while current_level <= height:
+        # Assign the structures for the vertices of the current level for T1 and
+        # T2.
+        struct_T1 = assign_structure(parenthood_T1, levels_T1, 
+                                     values_T1, current_level)
+
+        struct_T2 = assign_structure(parenthood_T2, levels_T2, 
+                                     values_T2, current_level)
+        
+        # Build the list of structures and the mapping.
+        S_T1, MS_T1 = get_multisets_list_of_level(levels_T1, values_T1, 
+                                                  struct_T1, current_level)
+
+        S_T2, MS_T2 = get_multisets_list_of_level(levels_T2, values_T2, 
+                                                  struct_T2, current_level)
+        
+        # Sort the list of structures.
+        sorted_S_T1 = sort_natural_multisets(S_T1)
+        sorted_S_T2 = sort_natural_multisets(S_T2)
+
+        # If the sorted lists are different, return false.
+        if sorted_S_T1 != sorted_S_T2:
+            return False
+        
+        # Update the values.
+        update_values(sorted_S_T1, MS_T1, values_T1)
+        update_values(sorted_S_T2, MS_T2, values_T2)
+        
+        # Move to the next level.
+        current_level += 1
+        
+    return True
+
+def get_height(T, root):
+    """
+    Returns the height of a rooted tree.
+    
+    Parameters
+    ----------
+    T : NetworkX Graph
+        A rooted tree.
+    
+    root : node
+        The root of the tree.
+    
+    Note
+    ----
+    This algorithm runs in O(n) time and space.
+    """
+    # Define a map for the distance to the root.
+    distance_to_root = {}
+    
+    # The max distance between a root and another vertex.
+    max_distance = 0
+    
+    distance_to_root[root] = 0
+    
+    # Perform a BFS traversal to determine the distances.
+    for (parent, child) in nx.bfs_edges(T, root):
+        distance_to_root[child] = distance_to_root[parent] + 1
+        
+        if max_distance < distance_to_root[child]:
+            max_distance = distance_to_root[child]
+            
+    return max_distance
+
+def get_parenthood(T, root):
+    """
+    Returns the parenthood function of a rooted tree.
+    
+    Parameters
+    ----------
+    T : NetworkX Graph
+        A rooted tree.
+    
+    root : node
+        The root of the tree.
+    
+    Note
+    ----
+    This algorithm runs in O(n) time and space.
+    """
+    # Define an empty parenthood function.
+    parenthood = {}
+
+    # Perform a BFS traversal to determine the parenthood function.
+    for (parent, child) in nx.bfs_edges(T, root):    
+        parenthood[child] = parent
+        
+    return parenthood
+        
+
+def rooted_tree_isomorphism_n(T1, root_T1, T2, root_T2):
+    """
+    Returns if two rooted trees are isomorph.
+    
+    Parameters
+    ----------
+    T1 : NetworkX Graph
+        A rooted tree.
+    
+    root_T1 : node
+        The root of T1.
+
+    T2 : NetworkX Graph
+        A rooted tree.
+    
+    root_t2 : node
+        The root of T2.
+    
+    Note
+    ----
+    This algorithm runs in O(n) time and space.
+    """
+    # If both trees have different amount of vertices, return false.
+    if T1.order() != T2.order():
+        return False
+    
+    # If both trees are empty, return true.
+    if T1.order() == 0:
+        return True
+    
+    # Get the height's of the trees.
+    height_T1 = get_height(T1, root_T1)
+    height_T2 = get_height(T2, root_T2)
+    
+    # If the trees differ in height, return false.
+    if height_T1 != height_T2:
+        return False
+    elif height_T1 == 0:
+        # If both trees have height 0, then there are only conformed by the
+        # roots, they're isomorph.
+        return True
+    else:
+        # If they have the same amount of levels, check that all levels coincide
+        # in structure.
+        levels_T1 = get_levels(T1, root_T1, height_T1)
+        levels_T2 = get_levels(T2, root_T2, height_T2)
+        
+        # For every level i, check that both trees have the same amount of
+        # vertices on the i-th level. If they differ for some level, return
+        # false.
+        for i in range(height_T1):
+            if len(levels_T1[i]) != len(levels_T2[i]):
+                return False
+        
+        # Set the initial values for T1 and T2.
+        values_T1 = get_initial_values(T1, root_T1)
+        values_T2 = get_initial_values(T2, root_T2)
+        
+        # Set the parenthood function for T1 and T2.
+        parenthood_T1 = get_parenthood(T1, root_T1)
+        parenthood_T2 = get_parenthood(T2, root_T2)
+        
+        return levels_verification(values_T1, values_T2,
+                                   levels_T1, levels_T2,
+                                   parenthood_T1, parenthood_T2,
+                                   height_T1)
+    
+
 def root_trees(t1, root1, t2, root2):
     """Create a single digraph dT of free trees t1 and t2
     #   with roots root1 and root2 respectively
--- a/networkx/algorithms/isomorphism/tree_isomorphism.py
+++ b/networkx/algorithms/isomorphism/tree_isomorphism.py
@@ -24,6 +24,60 @@ from networkx.utils.decorators import not_implemented_for
 
 __all__ = ["rooted_tree_isomorphism", "tree_isomorphism"]
 
+def get_levels(T, root, height):
+    """
+    Returns a function (map or dict) M: N -> P(V) such that, for each i in
+    {0,...,h}, M(i) is the set of vertices that are found in the i-th level
+    of T.
+    
+    Let h be the height of the tree. The amount of levels a tree has is equal to
+    its height. If a vertex v is in the i-th level, then its childs are in the
+    (i-1)-th level. Thus, by definition, the only vertex in the h-th level is
+    the root of the tree and the only vertices in the level 0 are leaves.
+    
+    Parameters
+    ----------
+    T : NetworkX Graph
+        The rooted tree.
+    
+    root : node
+        The root of the tree.
+    
+    height : int
+        The height of the tree.
+    
+    Returns
+    -------
+    M : dictionary
+        A map M: N -> P(V) such that, for each i in {0,...,h}, M(i) is the set
+        of vertices that are found in the i-th level of T.
+    
+    Notes
+    -----
+    The algorithm runs in O(n) time and space.
+    """
+    # Map M such that M(i) is the set of vertices that are in the i-th level.
+    lvl = {}
+    
+    # Map in_lvl such that in_lvl(v) is the level where v is found.
+    in_lvl = {}
+    
+    # Define the root height.
+    in_lvl[root] = height
+    
+    # Perform a BFS traversal to fill the in_lvl function.
+    for (parent, child) in nx.bfs_edges(T, root):
+        in_lvl[child] = in_lvl[parent] - 1
+        
+    # Traverse all vertices and group them by levels.
+    for v in in_lvl.keys():
+        l = in_lvl[v]
+        if l in lvl:
+            lvl[l].add(v)
+        else:
+            lvl[l] = {v}
+            
+    return lvl
 
 def root_trees(t1, root1, t2, root2):
     """Create a single digraph dT of free trees t1 and t2
--- a/networkx/algorithms/isomorphism/tree_isomorphism.py
+++ b/networkx/algorithms/isomorphism/tree_isomorphism.py
@@ -79,6 +79,60 @@ def get_levels(T, root, height):
             
     return lvl
 
+def get_initial_values(T, root):
+    """
+    Returns a function (map or dict) M: N -> P(V) such that, for every
+    vertex v in the tree T, if v is a leave then v's initial value is 0,
+    otherwise its defined as -1.
+    
+    Parameters
+    ----------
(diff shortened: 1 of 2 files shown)


Aufinal:
Aside from the extensive tests (that I haven't checked), it seems to me there is only one big difference between your algorithm and the one before : when you get a list of lists of `int`s corresponding to the structure of the previous level, you use a bucket sort instead of the naive python sort. I agree this is smart, since you know there are at most `n` different such lists at each step, but the code seems unnecessarily complicated for such a change.

Apart from the small improvements (e.g. not checking the degree sequence, or the O(n) algorithm to find centers), can you explain what is the difference between what you are doing and imply implementing a `bucket_sort` function for list of lists and  replacing the `sorted` calls of the previous algorithm by such a function ?

dschult:
Looks like New Year's Day is a good day to look at this PR. :}  Thanks for that nice comment @Aufinal 

This PR has nice formatting, tests and clear reporting of theoretical time dependence for the functions. I can verify that this new code is faster than the previous -- as advertised. I extended the timing up to 1000 nodes. The new code is consistently scaling at O(n) where n is the number of nodes (also the number of edges since this is a tree). I checked for agreement with the older code making sure the isomorphisms agree.  This is tricky because if there are any symmetries in the trees, the number of isomorphisms become large really quickly.  So I limited myself to a pinwheel shaped Graph:
```python
def pinwheel_graph(n):
    G = nx.Graph()
    G.add_node(0)
    for i in range(n):
        nx.add_path(G, [0] + list(range(len(G), 2*len(G))))
    return G
```
which doubles the number of nodes along each spoke out from the center node. This means it has n-spokes and 2^n nodes for n>=1. There are unique isomorphisms for n>=3. And it is not slow to produce graphs with thousands of nodes.

With each double of the number of nodes, the time doubled very nicely. The older code is obviously slower since it has dependence O(n ln(n)). For 512 nodes, I get 2.62 s for the older code and 0.095 s for the new code.  This shows the impact of using a better algorithm! :}

I agree with @Aufinal that many parts of the code seem like too much for this task. We don't want to try to maintain our own Data Structures within NetworkX unless they are really needed. And at least some of the functions are the same between the two versions.

For the data structure `NaturalMultiSet` there is a Python data structure `collections.Counter` which provides almost all features. The only method not easy to port over is `MS.to_list()` which can be constructed with (an admittedly somewhat long) list comprehension. 

The main idea is that a Python dict holding the mapping `NaturalMultiSet.m` contains the set `NaturalMultiSet.X` as its keys. So, you don't need to maintain them separately. 
```python
from collections import Counter

nlist = [1, 3, 2, 4, 2, 3, 4, 3, 2, 2, 3, 12, 1]
MS = Counter()     # can also use Counter(nlist)
MS.update(nlist)   # use update() instead of from_list() 
# MS.add(3) --> MS[3] += 1    or   MS.update([3])
# length   --> sum(MS.values())  And you could subclass Counter to keep track of the sum incrementally if desired
# contains, __eq__, __len__ are same

# the MS.to_list() method is a little more tricky
[x for x in range(max(MS)+1) for i in range(MS.get(x, 0))]
# if you need the result to be hashable you can convert to a string, or just use a tuple instead of a list.
tuple(x for x in range(max(MS)+1) for i in range(MS.get(x, 0)))
```

To summarize then, 
```python
nlist = [1, 3, 2, 4, 2, 3, 4, 3, 2, 2, 3, 12, 1]
MS = Counter(nlist)
MS[3] += 1      # same as MS.add(3)
MS.update([2, 3, 1, 4, 3, 2, 3])    # same as MS.from_list([2, 3, 1, 4, 3, 2, 3]) 
MS[4] += 1      # same as MS.add(4)
MSlist = [x for x in range(max(MS) + 1) for _ in range(MS.get(x, 0))]
# output:  [1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 12]
```

Sorry this is so long, but I'll end with restating what @aufinal requested and adding a request to it:
1) can you replace only the sorting parts of the older code with a bucket_sort function and get the same speedup? (What precisely is the difference between the new method and the old?)
2) can you reduce the future maintenance potential of the new data structure by using `collections.Counter`?

maucarrui:
Thanks for the feedback @Aufinal and @dschult!

Now that I think about it, I agree that most of the functions implemented are a little bit of an overkill. To answer your question, besides from the sorting algorithm, I think that other difference is on how we label the vertices of the rooted trees. 

In the current algorithm we build a new tree that contains the two rooted trees by adding an artificial vertex as the new root. Then we traverse each level and assign levels and children until the very last level. Finally, we check that both of the roots have the same label (in my case I call these labels values); if both of these roots have the same label then they're isomorphic and we build the isomorphism. A drawback of this method is that if both trees differ on a level, then there is no way to know this until the very end when we have already traversed all of the rest of the levels. To explain a little bit better the previous problem. When joining both trees as one, we also group all of the labels. Thus when sorting all of these labels there is no direct way of knowing if both levels have the same structure. This could be solved by traversing each label and building a list which contains the sorted labels of the first tree and another one for the other tree, and then checking if they're the same. In terms of complexity, the algorithm would still be O(n), but the difference with the proposed algorithm is that as soon as some levels differ we stop the algorithm and return false (or an empty isomorphism); when working with very big graphs this can save up a lot of time as we don't have to traverse (and sort) all of the vertices.

The proposed algorithms keeps both trees separated and traverses each level of both trees at the same time, thus when building the sorted labels we have the labels of the first and second separated, allowing us to check if they're equal and if not so, end the execution and return false. In my opinion, the proposed method stays more loyal to the original algorithm proposed by Aho, Hopcroft and Ullman, and perhaps its easier to understand how the level verification works and understand why if all levels share the same structure then both trees are isomorphic. There is also a space benefit with this method (even if its not that big of an impact), for we only have to work with the given trees and not build a tree of size 2n + 1 (all of the vertices of the first and second tree and the artificial root). All of the dictionaries used in the proposed algorithm are also used, in some way, in the current algorithm, so in that sense I don't think there is much of a difference.

What do you think? Do you agree with some of my points?

Nonetheless, I'll work on an alternative that uses the collection `Counter` and some of the improvements I mentioned earlier while trying to stay loyal to the current implementation and not use so many functions.

Aufinal:
The second version looks much better ! However, i have a question left : do you *really* need to sort the multisets ?

First, consider [this part](https://github.com/networkx/networkx/pull/6309/files#diff-2ac63d95565391abdff221258d89bbde241cc18665df9e64c9cc9433b0723e49R701) of your code: you sort the two lists `S_T1`and `S_T2` to compare them, but you could just as well compare them as multisets ! 

You also use this ordering in the function [`update_values`](https://github.com/networkx/networkx/pull/6309/files#diff-2ac63d95565391abdff221258d89bbde241cc18665df9e64c9cc9433b0723e49R485) to make sure that both labels are the same, but for that you can just use any ordering (e.g. `list(S_T1.keys()`) as long as you make sure to use the same ordering for both sides...

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+    This algorithm runs in O(l_total + m) time and O(l_total) space, where
+    l_total is the sum of the lengths of the lists in S and m is the max value
+    of all the lists in S.
+
+    """
+    # Define the an empty NONEMPTY map.
+    NONEMPTY = {}
+
+    # The max length found in S.
+    l_max = max([len(s) for s in S])
```
Aufinal, line 61: Minor nitpick : you don't need the square brackets in this line (since `max` works on iterators)

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+    -----
+    This algorithm runs in O(|S|) time and space.
+
+    """
+    # Define an empty LENGTH map.
+    LENGTH = {}
+
+    # Traverse the lists in S and add them to the LENGTH.
+    for s in S:
+        if len(s) in LENGTH:
```
Aufinal, line 113: This kind of code (that you also use in other places) can be simplified with a `collections.defaultdict(list)`, if you want.

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+
+    # The max length found in S.
+    l_max = max([len(s) for s in S])
+
+    # Obtain the NONEMPTY and LENGTH maps.
+    NONEMPTY = categorize_entries(S)
+    LENGTH = categorize_lists(S)
+
+    l = l_max
+    while l >= 1:
```
Aufinal, line 162: Why not `for l in range(l_max, 0, -1)` ?

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+                Q.append(s)
+
+        # Proceed with the lists of size l-1.
+        l = l - 1
+
+    # Q now contains all of the elements of S sorted lexicographically.
+    return Q
+
+
+def get_levels(T, root, height):
```
Aufinal, line 187: Each of these `get_` functions (there are around 5 of them) run its own BFS for the task; wouldn't it be faster to run one BFS and compute everything at once ?

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+    # Get the initial children found on the 0-th level.
+    children_T1 = get_initial_children(parenthood_T1, levels_T1)
+    children_T2 = get_initial_children(parenthood_T2, levels_T2)
+
+    # Start at level 1 as all the vertices at level 0 are leaves.
+    current_level = 1
+
+    # check that all the structures present on the i-th level are also present
+    # in i-th level of T2.
+    while current_level <= height:
```
Aufinal, line 681: Why not a `for` loop ?

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+        for v in centers_T2:
+            is_iso, iso = rooted_tree_isomorphism_n(T1, u, T2, v)
+            if is_iso:
+                return is_iso, iso
+
+        # If none of the previous trees were isomorphic, then return false.
+        return False, {}
+
+
 def root_trees(t1, root1, t2, root2):
```
Aufinal, line 1019: Do you need to keep this code ?

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+            values_T2,
+            levels_T1,
+            levels_T2,
+            parenthood_T1,
+            parenthood_T2,
+            height_T1,
+        )
+
+
+def get_centers_of_tree(T):
```
Aufinal, line 888: This is a nice function that's faster than `nx.center` for trees; it might be useful to move it next to `nx.center` and make it available to the API (as, idk, `nx.tree_center` ?) 🙂

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+    # If both trees have different amount of vertices, return false.
+    if T1.order() != T2.order():
+        return False, {}
+
+    # If both trees are empty, return true.
+    if T1.order() == 0:
+        return True, {}
+
+    # If trees have more than one node, verify that they're indeed trees.
+    assert nx.is_tree(T1)
```
Aufinal, line 839: I know it was already the case in the previous code, but a more explicit error message could be useful here.

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+
+    # Traverse each non-leave vertex on the i-th level and append its structure
+    # to the list.
+    for u in levels[i]:
+        if values[u] != 0:
+            # Get the structure (counter) associated with u.
+            c = struct[u]
+
+            # Get the list representation of the counter.
+            lc = tuple(x for x in range(max(c) + 1) for _ in range(c.get(x, 0)))
```
Aufinal, line 471: IIRC, you don't need `c.get`: `Counter`s return 0 for missing items already

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,29 +1,1021 @@
...
+
+            # Append the list representation to the list.
+            S.append(lc)
+
+            # Indicate that c is a structure u has.
+            if lc in MS:
```
Aufinal, line 477: Idem, `defaultdict(set)` can help here.

Aufinal (approved):
Did a small code review as well, mostly for minor nitpicks and improvements.

maucarrui:
In the current PR is a more polished version of the whole algorithm. There are some big improvements taking into consideration @Aufinal's great feedback.

The first big change is that most of the `get` functions were compiled into a single one, thus the following functions:

- `get_levels`
- `get_initial_values`
- `get_initial_children`
- `get_parenthood`
- `get_height`

Are now contained in the new function `get_initial_maps_and_height`. The main advantage, besides the code being smaller, is that we only traverse two times the nodes in the tree to get maps. The first time we can determine the maps `VALUES`, `PARENTHOOD` and also get the tree's height. The second time we can obtain the rest of the maps: `LEVELS` and `CHILDREN`.

And the second big change. @Aufinal made a very good observation. There really is no need to sort multisets to determine if the rooted trees share the same structures on the same level. We can do the previous by using a `Counter` that contains the list representation of these structures and comparing them directly to determine if both trees share the same structures the same amount of times. Thus, we only have to modify the previous `update_values` function so that vertices with the same structure share the same value on both trees. This way, the previous function is only called one time on each level, an each call updates the values of the vertices by the order specified by the `Counter`'s iterator; I guess the only drawback would be that there is no way of knowing which value a structure is going to get as these are not sorted anymore, but it really isn't that important.

Furthermore, these changes reduce the amount of time the algorithm takes as the amount of BFS traversals have been reduced and we don't have to sort multisets! The algorithm still is O(n), but the times improved by quite a lot. The times I obtained while executing the algorithms on my laptop were the following:

### Positive tests
```
Positive test with new function

2 1 0.00023293495178222656
3 1 0.00023436546325683594
4 2 0.0004878044128417969
5 3 0.0007808208465576172
6 6 0.0019073486328125
7 11 0.003464221954345703
8 23 0.008425474166870117
9 47 0.019379854202270508
10 106 0.04652571678161621
11 235 0.1063690185546875
12 551 0.26069021224975586
13 1301 0.6475820541381836
14 3159 1.7945880889892578
15 7741 4.563429832458496
16 19320 11.326676368713379
17 48629 29.660635948181152
```

### Negative tests
```
Negative test with new function

4 1 9.250640869140625e-05
5 3 0.0001385211944580078
6 15 0.0006356239318847656
7 55 0.002710103988647461
8 253 0.010787725448608398
9 1081 0.05134773254394531
10 5565 0.2679872512817383
11 27495 1.3304953575134277
12 151525 7.602773666381836
13 845650 43.96673941612244
14 4988061 262.4102749824524
```

For example, in the case of the positives tests with `k=17` we almost saved up to 30 seconds!

A final observation. If I'm not wrong, we traverse all the nodes of both trees five times in total to determine if they're isomorphic and if so, return an isomorphism. The first two traversals are to determine the initial maps. Then, in the levels verification, for each level we traverse the vertices in the current level and the ones in the level below (to assign structures). And a final traversal to build the isomorphism.

I took into consideration the recommendations about `defaultdict` and some of the `while`s. Also, the previous functions were deleted as they're no longer used. The unit tests left test the proposed functions for the algorithm, it takes around 2 seconds for all them to complete when `k=13`and `k=11` for the positive and negative tests respectively.

I hope the code and functions are now much more clearer and not so much of an overkill. What do you think?

Aufinal:
Great job ! I think you're having the right idea with the `CHILDREN` map, but you should take even one step further. Specifically, you can build this map for the whole tree instead of only the first level, in the BFS itself (probably with a `defaultdict(set)` or something).

Then your function `assign_structure` simply becomes `Counter(values[u] for u in children[v])`. This assigns a value of `Counter()`(the empty counter) to the leaves, which is meaningful and different from any other value, so you wouldn't need to treat them separately ! I think this also drops the requirement for the `PARENTHOOD` map as well.

maucarrui:
You're right @Aufinal! Let's say we are on the `i`-th level, thus we updated the values of the `(i-1)`-th level and added all of the vertices of this level to its corresponding father's list, this way  we have the `CHILDREN` values for all the vertices on the `i`-th level. We can build the structures and leaves directly using only dictionaries and set comprehensions, which removes the function `assign_values` and the usage of the `PARENTHOOD` map in this case.

However, I don't think we can fill the `CHILDREN` map at the beginning. This map doesn't only represent the children of a vertex in a rooted tree, its actually a list of the ordered children of a vertex. The way I'm filling this map is from bot to top, thus vertices with same value are appended in the same order to its parent list on both trees, that's also the reason why I need the `PARENTHOOD` map. And finally, we perform a DFS from to build the isomorphism taking into consideration the defined order.

Right now I'm not sure if there is a shortcut for building the isomorphism with this method that avoids using the `PARENTHOOD` map. The reason is, let's say we are on the `i`-th level, its true that if "v" maps to "w" in an isomorphism then "v" and "w" have the same value, but the inverse is not necessarily true (i.e. if "v" and "w" have the same value then "v" maps to "w"). The previous can be seen with a small example, lets say we have the following trees:
```
v0
├── v1
│   ├── v4
│   └── v5
├── v2
│   └── v6
└── v3

w0
├── w1
│   └── w4
├── w2
│   ├── w5
│   └── w6
└── w3
```

A possible isomorphism is the following `{v0: w0, v1: w2, v2: w1, v3: w3, v4: w6, v5: w5, v6: w4}`. So, let's say we are on the `0`-th level where we have only the leaves `v4, v5, v6, w4, w5, w6`. If we're not careful, we could assign the following isomorphism `{v4: w4, v5: w5, v6: w6}`, but the previous is not valid as the parent of `w4` only has one child. That's why we can't build the isomorphism on our first from bot to top traversal, as we need more information about the parents on the following level. As this applies to all levels, then the only way we can be sure is when we reach the top (in this case the roots) and follow the order specified by the children in order to build an isomorphism.

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,279 +1,686 @@
...
+    # Define an empty map for LEVELS, VALUES, CHILDREN and PARENTHOOD.
+    LEVELS = defaultdict(set)
+    CHILDREN = defaultdict(list)
+    VALUES = {}
+    PARENTHOOD = {}
 
-    dT = nx.DiGraph()
+    # Auxiliary variables.
+    from_distance = defaultdict(set)
+    distance_to_root = {}
```
Aufinal, line 98: You don't need this variable (or the 3 below it) ; at the end of the BFS, the height is simply `len(from_distance) - 1` !

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,279 +1,686 @@
...
-        namemap[new] = old
+        # If the previous distance is greater than the max, update it.
+        if max_distance_to_root < distance_to_root[child]:
+            max_distance_to_root = distance_to_root[child]
 
-    return (dT, namemap, newroot1, newroot2)
+    # The max distance to the root is the rooted tree's height.
+    height = max_distance_to_root
 
+    # Build the LEVELS map.
```
Aufinal, line 129: maybe slightly more pythonic : `LEVELS =  {height-d:s for (d, s) in from_distance.items() }` .

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,279 +1,686 @@
...
+            children_T1,
+            children_T2,
+            parenthood_T1,
+            parenthood_T2,
+            leaves_T1,
+            leaves_T2,
+        )
+
+        # Move to the next level.
+        current_level += 1
```
Aufinal, line 501: Isn't this a mistake since you already use a `for` loop ?

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -1,279 +1,686 @@
...
+            return False, {}
+
+    # Start at level 1 as all the vertices at level 0 are leaves. Check that
+    # all the structures present on the i-th level are also present in i-th
+    # level of T2.
+    for current_level in range(1, height_T1 + 1):
+        # Assign the structures for the vertices of the current level for T1 and
+        # T2.
+        struct_T1 = {
+            v: Counter([values_T1[u] for u in children_T1[v]])
```
Aufinal, line 461: You don't need a list here since `Counter` accepts iterators as arguments (I think)

Aufinal:
There is still something that confuses me : why do you treat the leaves as a special case ? If `v` is a leaf, then `children[v] = []` and thus the value of `v` is `Counter([])`, which is a perfectly defined structure...

Aufinal:
Looks great ! I have no more comments, I leave it to the maintainers 😊

maucarrui:
Thanks so much for the feedback and time @Aufinal! You're right, there was no need to consider the leaves as a special case.

PR #6311, opened 2023-01-02
New algorithm for calculating the longest path between two nodes

matheusdiogenesandrade:
A new algorithm for finding the longest path between two nodes in a DAG.

### DIFF
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -1087,9 +1087,9 @@ def dag_longest_path_between(
     >>> DG = nx.DiGraph([(0, 1, {'cost':1}), (1, 2, {'cost':1}), (0, 2, {'cost':42})])
     >>> list(nx.all_simple_paths(DG, 0, 2))
     [[0, 1, 2], [0, 2]]
-    >>> nx.dag_longest_path_between(DG, "cost", 1, source=0, target=2)
+    >>> nx.dag_longest_path_between(DG, 0, 2, "cost", 1)
     [0, 1, 2]
-    >>> nx.dag_longest_path_between(DG, weight="cost", source=0)
+    >>> nx.dag_longest_path_between(DG, 0, None, weight="cost")
     [0, 2]
 
     In the case where multiple valid topological orderings exist, `topo_order`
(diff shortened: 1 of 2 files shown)


matheusdiogenesandrade:
Implementation of an algorithm for calculating the longest path between two nodes.

rossbar (commented):
This seems like a nice addition to me! I haven't had a chance to review in detail but this seems like useful functionality.

I realize it's just matching the pattern established by `dag_longest_path_length`, but I wonder if we couldn't get away with *not* adding the `dag_longest_path_length_between` function - the reason being that this is really something that users might be better constructing themselves. It boils down to a one-liner (`sum(G[u][v].get(weight, default_weight) for u, v in pairwise(path))`) for every case *except* multigraphs, where `max` is chosen arbitrarily as the multi-edge reduction function. One could envision scenarios where users would want a *different* reduction function (e.g. `sum`), in which case the API would have to be expanded to accept a callable. IMO it's much more straightforward to have a single function and derive the path length separately.

Sorry, that's all tangential to this PR! I'll open a separate issue. I'm generally +1 for the `dag_longest_path_between` function!

matheusdiogenesandrade:
Sorry, closed the pull request accidentally. Opening this again. @rossbar thanks for the feedback, I will be watching the issues list in order to know the next steps.

dschult:
I wonder if the functionality for these two functions could be embedded into `dag_longest_path` and `dag_longest_path_length`. We would need to add two new input arguments. And I haven't looked closely at the code to see how much duplication this would avoid. The new code handles the case when either `source` or `target` are provided, and the older code handles the case when both are `None`.

I suggest this to build on the api developed for shortest_paths where we have functions to return the path that change their output depending on whether `source` or `target` are provided in the function call.

matheusdiogenesandrade:
Sorry, since I was having some problems in merging the pull request with the newest commits, I thought I could relink the pull repository to a new one. But it seems it is not possible. Anyways, after walking through the code, I noticed that the feature of this pull request can be achieved by a "simple" graph manipulation. I have written the algorithm in this [page](https://matheusdio.wordpress.com/2023/08/01/longest-path-with-mandatory-components-in-dags/), case you want to go straight to the point just go the section "Fixed start- or end-ing nodes". 

So, we have three options here:

1.  Ignore this feature, and instead just write some code snippet tutoring the graph manipulation procedure;
2.  Code the graph manipulation as the feature; Or
3.  Ignore the graph manipulation technique and go back to the previous code.

These are the options I could list for now, case you have other in mind, please let me know. And, if you identified any issue with regards to the manipulation scheme, let me know as well.

Thanks and regards.

PR #6315, opened 2023-01-03
use ruff to lint and fix E713

danieleades:
see the [flake8-docs](https://www.flake8rules.com/rules/E713.html)

> Tests for membership should use the form `x not in the_list` rather than `not x in the_list`. The former example is simply more readable.
> 
> ### Anti-pattern
> 
> ```python
> my_list = [1, 2, 3]
> if not num in my_list:
>     print(num)
> ```
> 
> ### Best practice
> 
> ```python
> my_list = [1, 2, 3]
> if num not in my_list:
>     print(num)
> ```

[ruff](https://github.com/charliermarsh/ruff) is a linting tool similar to flake8, but which is a couple of orders of magnitude faster, and is capable of auto-fixing most of its lints. This PR only enables a single lint (E713).


danieleades:
the mypy errors are not related to this PR

MridulS (approved):

danieleades:
merged into #6371

PR #6316, opened 2023-01-03
use black '--preview' to normalise strings

danieleades:
by default, Black doesn't attempt to normalise strings. Enabling the '--preview' feature enables this feature. It also reformats some tuples.

the '--preview' feature isn't stabilised since there are edge cases it doesn't handle well. Therefore I propose to run it as a one-time deal for now, rather than leaving it enabled.

### DIFF
--- a/networkx/generators/tests/test_internet_as_graphs.py
+++ b/networkx/generators/tests/test_internet_as_graphs.py
@@ -28,8 +28,8 @@ class TestInternetASTopology:
                 cls.CP.append(i)
             else:
                 raise ValueError(
-                    "Inconsistent data in the graph\
-                        node attributes"
+                    "Inconsistent data in the graph                        node"
+                    " attributes"
                 )
             cls.set_customers(i)
             cls.set_providers(i)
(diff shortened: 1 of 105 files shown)


networkx/generators/tests/test_internet_as_graphs.py
```diff
@@ -66,8 +66,8 @@ def set_providers(cls, i):
                         cls.providers[i].add(j)
                     elif j != customer:
                         raise ValueError(
-                            "Inconsistent data in the graph\
```
danieleades, line 69: i'm guessing this is a bug in the original?

MridulS:
This is an experimental feature in black itself https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#preview-style. We follow the black official releases in our pre-commit config https://github.com/networkx/networkx/blob/71ad516a90c89c7294e32bf28e1f05c63c5f17e4/.pre-commit-config.yaml#L6 if and when this is incorporated in a proper black release we can add this to networkx. The whole point for me to use `black` is to not think about these things :)

networkx/generators/tests/test_internet_as_graphs.py
MridulS: Yeah it's a whitespace issue.

PR #6317, opened 2023-01-03
use ruff to lint and fix C4 (comprehensions)

danieleades:
this is equivalent to introducing the 'flake8-comprehenions' plugin, however using [ruff](https://github.com/charliermarsh/ruff) is a couple of orders of magnitude faster, and provides the ability to auto-fix the lints.

using list/dict/set literals provides a small performance boost, as well as improving readability.


networkx/algorithms/isomorphism/tests/test_ismags.py
```diff
@@ -13,19 +13,19 @@ def _matches_to_sets(matches):
     Helper function to facilitate comparing collections of dictionaries in
     which order does not matter.
     """
-    return set(map(lambda m: frozenset(m.items()), matches))
+    return {frozenset(m.items()) for m in matches}
```
MridulS, line 16: well that's definitely better

MridulS (approved):

PR #6319, opened 2023-01-03
forbid blanket 'type: ignore' comments

danieleades:
error suppression should use the smallest possible scope. Additionally, if specific error codes are used, then mypy will return an error if that error is no longer triggered (with a blanket statement this behaviour does not occur)

This PR uses ruff to forbid blanket ignore statements and requires users to set a specific error code.


MridulS (approved):

danieleades:
merged into #6371

PR #6322, opened 2023-01-03
Update copyright years to 2023

MridulS:
Time for the new-year-update-copyright PR. Happy new year!

There is a debate about this if we even need to keep the years in the notice[1][2] but until that is resolved let's update them.

[1] https://hynek.me/til/copyright-years/
[2] https://github.com/scipy/scipy/issues/15282#issuecomment-1003762552

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -2,7 +2,7 @@ NetworkX is distributed with the 3-clause BSD license.
 
 ::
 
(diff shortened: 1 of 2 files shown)


rossbar (approved):
I grepped through everything and these were the only two instances I found too. Thanks @MridulS !

dschult (approved):

PR #6323, opened 2023-01-04
fix warnings for make doctest

dschult:
Fix the warnings that arise from `make doctest`


rossbar (approved):
LGTM, thanks @dschult !

MridulS (approved):

PR #6324, opened 2023-01-04
fix whitespace issue in 'networkx.generators.tests.test_internet_as_g…

danieleades:
the original strings are incorrectly split. I suspect rather than

```python
                raise ValueError(
                    "Inconsistent data in the graph\
                        node attributes"
                )
```

the intention was probably to have

```python
                raise ValueError(
                    "Inconsistent data in the graph"
                    " node attributes"
                )
```

but it all fits on one line anyway


MridulS (approved):
Thanks @danieleades !

dschult (approved):

PR #6333, opened 2023-01-05
doc:improve doc of possible values of nodes and expected behaviour

### LINKED ISSUE #6274: doc: improve doc of possible values of `nodes` and expected behaviour
This is a note to update the documentation of functions that accept `nodes` kwarg, so that it's more explicit what the behaviour of the function is depending on the passed object.

Potentially, the docstrings would be adjusted with:
```
    nodes : node, iterable of nodes, or None (default=None)
        If a singleton node, return the number of triangles for that node. 
        If an iterable, compute the number of triangles for each of those nodes.
        If `None` (the default) compute the number of triangles for all nodes in `G`.
```

_Originally posted by @dschult in https://github.com/networkx/networkx/pull/6258#discussion_r1045180012_

tinaoberoi:
Closes #6274

### DIFF
--- a/networkx/algorithms/cluster.py
+++ b/networkx/algorithms/cluster.py
@@ -329,8 +329,10 @@ def clustering(G, nodes=None, weight=None):
     ----------
     G : graph
 
-    nodes : container of nodes, optional (default=all nodes in G)
-       Compute clustering for nodes in this container.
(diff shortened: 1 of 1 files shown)


MridulS (approved):
Thanks @tinaoberoi!

rossbar (approved):
LGTM, thanks @tinaoberoi !

PR #6344, opened 2023-01-07
Request to add algorithms for finding envy-free matchings in bipartite-graphs we've implemented

danigil:
Hello we're students from Ariel University under Erel Segal (https://github.com/erelsgl)
and as part of a research algorithms course we implemented algorithms from an article that find envy-free matching inside bipartite graphs in polynomial time (based on https://arxiv.org/abs/1901.09527).
We've confirmed that the pytests passed successfully after the changes we've made and added our own tests.


danigil:
style / format (3.8) (pull_request)  - test failed partly not for us to blame

networkx/algorithms/bipartite/EFM_algorithms.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Y_S = set()
+    for subset in Y_subsets:
+        Y_S.update(subset)
+
+    return set(X) - X_S, X_S, set(Y) - Y_S, Y_S
+
+
+@not_implemented_for("directed", "multigraph")
+def envy_free_matching(G, top_nodes=None):
```
erelsgl, line 195: The two main functions should be the first functions in the file.

erelsgl, line 195: The name of this function should indicate that it returns an envy-free matching of maximum cardinality. Perhaps "largest_envy_free_matching" or "max_size_envy_free_matching".

networkx/algorithms/bipartite/EFM_algorithms.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    M = nx.bipartite.maximum_matching(G, top_nodes=top_nodes)
+    EFM_PARTITION = _EFM_partition(G, M, top_nodes)
+    un = EFM_PARTITION[1].union(EFM_PARTITION[3])
+    M = {node: M[node] for node in M if node not in un and M[node] not in un}
+    return M
+
+
+@not_implemented_for("directed", "multigraph")
+def minimum_weight_envy_free_matching(G, top_nodes=None):
```
erelsgl, line 249: The name of this function should indicate that it returns an envy-free matching of maximum cardinality. Perhaps "min_weight_largest_envy_free_matching" or "min_weight_max_size_envy_free_matching".

erelsgl (changes requested):

BenjaminSaldman (commented):
Thanks for the review!

danigil:
The PR was created back in january 2023 and it seems to have stagnated, is there anything to do
to progress its integration? I'm very determined to make it work :)

Thanks in any case!

erelsgl:
Hi @dschult . I have already reviewed this PR and it looks fine, as far as I can tell.

erelsgl (approved):
Hi @dschult . I have already reviewed this PR and wrote several comments, and the authors have handled all of them. Now the PR looks fine, as far as I can tell.

dschult:
Thanks @erelsgl 
Can you "resolve" the comments/requests for changes you made?

@danigil this is a little unusual because there has already been so much back and forth here in this PR. Perhaps it would have been better for you and @erelsgl to have developed this code on a fork so that the maintainers here would be notified when you pushed the resulting work to the main repo. But this is fine to work with.  Could you create a comment like it is a first comment in a PR? Explain what this is and why it might be useful and any comments about the algorithm and process?  I think @erelsgl could be helpful in providing larger context for that. For example, this might be a one-off contribution for you due to this class, but might be part of a larger focus of work collectively over the years of students taking this class. I don't mean for this to be longer or involved.  I just want some sort of introduction...  I've never worked with envy-free bipartite matchings. And I don't know or understand how it relates to other work that this class developed. It'd be nice to hear from @erelsgl whether there is a larger plan or are these PRs largely independent of each other. Pointing to other discussions/issues and PRs could be helpful too, but maybe not. I don't really know. :)

Some things to do in the meantime:
- the doc_strings should start with a single-line short description of what the function does, followed by a blank line and then followed by a paragraph that describes/defines in more detail what is going on. You should not assume that the reader knows what any of this is about -- define terms and point to general references.  Then the parameters, returns/yields sections and examples/notes/references sections can appear. Indentation matters for documentation too... the sphinx rendering system uses that to format the docs. 
- look at the docs "artifact" that is created in the CI tests.  Make sure your doc_strings are being parsed reasonably and look good on the documentation webpages. You probably need to add some blank lines to allow the sphinx system to identify paragraphs/sections/etc.  The artifact is found by clicking on the "Show all checks" link and then scrolling to near the bottom where there is a artifact document "test". Clicking on the link to get the details for that test shows the new webpages for this PR.
- you should probably change all functions that start with double underscores to single underscores (single means "private function" while double is typically reserved for Python language hidden functions. That's not a hard rule. So if there 's a reason to use __ push back and we'll figure it out.
- why are some of your functions in a separate module and others in the main matching.py module? Would it make sense to put all the envy-free stuff into a single module? Or are you thinking it is better to put matching stuff with the other matching stuff? My first thought is to keep similar algorithms together -- and it is likely that this PR is all dealing with similar algorithms. But that's only a first thought.

dschult:
For me, it is a large button in the lower left of the "conversation" just after the place to Reply. It is not in the drop-down menu.  The button says "resolve conversation".  If that isn't there, should I just resolve all discussion?

erelsgl:
Here is what I see at the lower-left of the conversation: 
image
I do not see a "resolve conversation" button. Could you do it please?

dschult:
That's too low -- its the lower left corner of just that one "conversation" in the whole thread...
I'll resolve them.
Here's what I see:
Screenshot 2023-06-13 at 9 28 20 PM

erelsgl:
@dschult Thanks! Strangely, I did not see this button here, and I do not see it in other PRs on which I commented (e.g. here: https://github.com/networkx/networkx/pull/6372 , I see this:  
image
).

dschult:
Hmmm...  Maybe there is a permission that restricts who can resolve conversations.  That is very strange. The person starting the conversation should be able to mark it as resolved... at least that makes sense to me...

Anyway, it is done now.  At least in my display of the conversation the comments on code are shrunk and marked as "resolved".  :)

danigil:
Hi Dan, thanks for the reply
I tried looking at the generated docs as you mentioned but it says "the logs for this run have expired and are no longer available"
So i'm guessing I need to retrigger the CI pipeline.
Regarding the fact there's separate files for our algorithms I agree they need to be in the matching.py file so I'll remove them, I'm not too keen as for how to update the PR, do I create more commits and push them into the branch and then theres an option to append them to the PR?

As to what you said in the start of our conversation I got a little bit confused, the code is indeed created on a fork of the original networkx repo, if it weren't then maybe a PR is impossible as to my understanding, but maybe i'm wrong.

I will implement the changes you suggested.
Thanks a lot!

dschult:
You should push any additional changes to your branch on your github repo (the fork you made of the networkx repo). Any push to that branch is automatically picked up by the PR. It should be the same as what you were doing for all the other commits. (you can see all the previous commits you made on the PR too.)

It might also be good to merge your changes with the main networkx branch so you get the latest CI tooling that was installed in April/May/June. Using git, add a remote (often called upstream) to the networkx repo. Then `git merge upstream main`. There may be "conflicts" where things you changed have been changed in a different way on the main branch. If so, follow the instructions git gives you, update the files with conflicts -- looking for indicators >>>> ==== <<<< to see what the conflicts are. The code between >>>> and ==== is on one branch while the code between ==== and <<<< is from the other branch. You have to choose what the resulting code should be and then git add those files and `git merge --continue` to go to the next conflict.  Hopefully you won't have many conflicts because you are mostly adding new  code and it isn't likely others have added new code in the same files that you did.  If the merge goes badly or you get stuck, use `git merge --abort` to cancel the merge.  All of this will be on your local machine. So you aren't hurting your github repo or fork in this process.  In the worst case you can reclone your fork to your local machine and start over that way.  But again, hopefully none of that is needed and the merge goes smoothly.  Once the merge completes successfully, makes sure it looks OK and then push to the branch on your fork -- the PR will automatically get those same changes.

danigil:
Hi!
I merged the main branch of networkx into out PR
changed the function with double underscore to single underscore
and deleted the redundant files as you mentioned in the comment above.

I tried looking at the docs but the algorithms we added don't appear in them, I also looked the CI log and it doesn't generate docs for our algorithms which reside in algorithms/bipartite/matching.py
can you clue me in as to why this is happening?

dschult (commented):
To have the documentation pick up this new file, you need to add the name of the file to the `doc/reference/algorithms/matching.rst` file.  Hopefully it is clear from the other functions there.

You will also need blank lines between each section of the doc_strings to get them to be parsed correctly. That also makes them easier to read in their ASCII form, so it's a Good Thing.

Also, since this is a new function, can you please make the keyword argumnts "keyword only" by adding a `*,` between the positional arguments and the keyword arguments?

danigil:
Hey @dschult!
I've added the changes you suggested regrading forcing the keyword arguments to be keyword only.
I've also reformatted the doc_strings to be more like the other doc_strings present in the bipartite/matching.py file. It looks pretty good to me in the doc artifact, can you take a look and share your opinion?
Thank you very much!

danigil:
> Hey @dschult! I've added the changes you suggested regrading forcing the keyword arguments to be keyword only. I've also reformatted the doc_strings to be more like the other doc_strings present in the bipartite/matching.py file. It looks pretty good to me in the doc artifact, can you take a look and share your opinion? Thank you very much!

ping @dschult

networkx/algorithms/bipartite/matching.py
```diff
@@ -582,3 +585,310 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+    {3, 4}
+
+    >>> _neighbours_of_set(G, {4})
+    {0, 1, 2}
+    """
+    ret_set = {}
+    for node in node_set:
+        ret_set.update(G[node])
+
+    return set(ret_set)
```
MridulS, line 620: This code will just work fine, but you can speed this bit up with something like :)
```suggestion
    # add import at the top of the file
    from itertools import chain
    neighbors_iter = chain.from_iterable(G.neighbors(n) for n in node_set)
    return set(neighbors_iter)
```

networkx/algorithms/bipartite/matching.py
```diff
@@ -582,3 +585,310 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+    >>> G = nx.Graph([(0, 4), (1, 5), (2, 6)])
+    >>> _neighbours_of_set(G, {0, 1})
+    {4, 5}
+
+    >>> G=nx.Graph([(0,3),(3,0),(0,4),(4,0),(1,4),(4,1),(2,4),(4,2)])
+    >>> _neighbours_of_set(G, {0, 1})
+    {3, 4}
+
+    >>> _neighbours_of_set(G, {4})
+    {0, 1, 2}
```
MridulS, line 614: This bit of documentation is never really built or served as anything starting with `_` will be ignored. We can keep the docs minimal here :)
```suggestion
    r"""Returns a set of the neighbours of a given set of nodes `node_set` in the bipartite graph `G`.

    >>> G = nx.complete_bipartite_graph(3,3)
    >>> _neighbours_of_set(G, {})
    set()
    >>> _neighbours_of_set(G, {1, 2})
    {3, 4, 5}
```

networkx/algorithms/bipartite/matching.py
```diff
@@ -582,3 +585,310 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+    The iterative pattern stops when either Y_k or X_k are an empty set for a certain k, forever staying an empty set from k onward
+    the alternating sequece is proven to always stop in the article.
+
+    >>> G = nx.complete_bipartite_graph(3,3)
+    >>> M = nx.bipartite.hopcroft_karp_matching(G)
+    >>> _M_alternating_sequence(G, M)
+    ((), ())
+
+    >>> G = nx.Graph([(0,3),(3,0),(0,4),(4,0),(1,4),(4,1),(2,4),(4,2)])
+    >>> M = {0:3, 3:0, 1:4, 4:1}
```
MridulS, line 646: Why exactly is the example bit written like this? For an undirected graph, adding an edge `0 -> 3` will also add `3 -> 0` so something like `nx.Graph([(0,3),(0,4),(1,4),(2,4)])` is equivalent to the above example.

networkx/algorithms/bipartite/matching.py
```diff
@@ -582,3 +585,310 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+        )
+        X_current = _neighbours_of_set(G_M, Y_current)
+        if X_current == set():
+            break
+
+        Y_subsets.append(Y_current)
+        X_subsets.append(X_current)
+
+        Y_subgroups_accumulate.update(Y_current)
+        index += 1
```
MridulS, line 699: `index` doesn't seem to be used anywhere.

networkx/algorithms/bipartite/matching.py
```diff
@@ -582,3 +585,310 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+    >>> G = nx.Graph([(0,6),(6,0),(1,6),(6,1),(1,7),(7,1),(2,6),(6,2),(2,8),(8,2),(3,9),(9,3),(3,6),(6,3),(4,8),(8,4),(4,7),(7,4),(5,9),(9,5)])
+    >>> M = {0:6,6:0,1:7,7:1,2:8,8:2,3:9,9:3}
+    >>> _M_alternating_sequence(G, M)
+    (({4, 5}, {1, 2, 3}, {0}), ({8, 9, 7}, {6}))
+
+    """
+    X, Y = nx.bipartite.sets(G, top_nodes=top_nodes)
+    X_0 = set(X) - set(M.keys())
+    if X_0 == set():
+        return (), ()
```
MridulS, line 669: I have question about the algorithms (sorry I have not read the paper yet!)
If I switch one node from one set to another bipartite set then the this would not that work right?
``` python
In [97]: G = nx.Graph([(0,3), (0,4), (1,4), (2,0)])

In [98]: M = nx.bipartite.maximum_matching(G)

In [99]: _M_alternating_sequence(G, M)
Out[99]: ((), ())

In [100]: G = nx.Graph([(0,3), (0,4), (1,4), (2,4)])

In [101]: M = nx.bipartite.maximum_matching(G)

In [102]: _M_alternating_sequence(G, M)
Out[102]: (({2}, {1}), ({4},))
``` 
should the output of 99 and 102 be the same?

I moved node `2` to the other side, should I be getting something like `(({1}, ), ({4}, {2}))` ?

MridulS (commented):
Thanks for your work on this @danigil
I did start doing some review here, but I think I need to come back to this a bit later.

I do have some comments till then :)

networkx/algorithms/bipartite/matching.py
danigil: I see! Nice use of itertools, I'll add it as a separate commit.
BTW, maybe we can add this function not inside bipartite/matching.py but instead make it available to the whole package, I'm pretty sure someone else will find usage for this in the future.
To take it a step further maybe we can modify Graph.neighbors to take in an iterable effectively extending the functionality of it. I'm guessing if you think this is a good idea maybe I can create a separate PR for this, as this isn't a function only used for the calculation of envy-free bipartite matchings, and in general the neighbors set is something that's very common in graph theory and so if this can be of help to someone in the future, why not!

networkx/algorithms/bipartite/matching.py
danigil: Makes sense!

networkx/algorithms/bipartite/matching.py
danigil: You're right, it seems to serve no purpose, This is probably an artifact from the past :P

danigil: I added a commit that removes this index variable.

networkx/algorithms/bipartite/matching.py
danigil: Correct. I removed these redundancies in all places where I noticed them in the docstrings.
I'm guessing the matching should stay as it is, I noticed the existing networkx matching algorithms do return a matching of this form.

networkx/algorithms/bipartite/matching.py
danigil: I went ahead and added this in a commit.
I noticed when looking at the imports that there is a 'import itertools' line so I didn't import chain specifically as you noted, instead, I called itertools.chain -  I thought it is a good practice to avoid as much as possible editing these common grounds in the code to avoid merging issues, correct me if I'm wrong.

networkx/algorithms/bipartite/matching.py
danigil: When you say "would not work" I'm not sure what exactly you mean.
when you took 2 and moved it to the other bipartite set and connected it to 4 you just created a different graph, the resulting M-alternating-sequence may be different.
The output in this algorithm is a partition of the two bipartite sets in the graph into disjoint sets.
This is similar to an M-alternating path.
All the algorithms we implemented are implementations of the ones written in pseudo-code or in words in the article.
When you get the output '((), ())' this happens because the M-alternating-sequence is a sequence starting from vertices unmatched by M, so if M matches all the vertices in X (where X,Y are the bipartite sets of G) the algorithm terminates immediately and the M-alternating-sequence is empty.
'I moved node 2 to the other side, should I be getting something like (({1}, ), ({4}, {2})) ?' this depends, the output of the algorithm depends on the matching returned (keep in mind there may be multiple different maximum matchings for a given bipartite graph) as well as the edges in the graph.
In general I can say that it may not be of utmost importance for you to understand the calculation of M-alternating-sequences since this is only a helper function used for calculating the envy-free bipartite matching which is the subject of this PR, and the contribution in this PR is mainly calculating these envy-free bipartite matching, I didn't mean for these helper functions to be made public in the library (but I'm not opposed to it).

Anyway, I'd be glad to offer any help in further clarifying how these algorithms work if you have any questions (as well as further explaining about what you asked above).
If you have time/want to you can read the article, but it's really not mandatory in order to obtain a higher-level understanding of these algorithms.

I'm not sure if this is a conventional custom, but if you want we can maybe schedule a virtual meeting so we can chat and I can explain the algorithms thoroughly.

dschult (commented):
I took the liberty of pushing a commit with some changes (so you will need to pull that from your fork to your local machine before making more changes):
- spelling `neighbours`, cardinallity, to american spelling and vertex -> node
- import of not_implemented_for and usage of not_implemented_for
      (putting both directed and multigraph in the same line rules out MultiDiGraph only. It's an "and". We need two calls to rules out the "or" version (rules out directed. then rules out multigraph. so both types are rules out and only Graph is left).
- reduced and cleaned up code for `Y_current` and `X_current`. I hope it is more readable, and should also speed it up. 

Still to do:
- should it be called "minimum cost" or "minimum weight"? What does the literature (in but also outside of the paper you are using) call this? Then we can be consistent with one name for it. If both are used, we can say that somewhere, but otherwise be consistent with the name.  Similar for "cardinality" and "size". Let's choose one of those.
- Add a description of each function after the first full paragraph that describes what "envy free matching" means. We need to know about what "maximum cardinality" means and what "min_weight" means. And it would be nice to know why someone might be interested in computing these quantities.
- there are no tests for `envy_free_matching_partition`
- need more description of the four sets in the partition. What properties the nodes have in each, what "matchings are contained in one partition" means (??"matchings involve only nodes in X_L and Y_L."??)
- there are many places where the difference between "a matching" and "the matching" are important here. "a matching" implies there could be others. "the matching" implies it is the only one.
- only use "good nodes" and "bad nodes" if you explain what you mean by that somewhere.

danigil:
> I took the liberty of pushing a commit with some changes (so you will need to pull that from your fork to your local machine before making more changes):
> 
> * spelling `neighbours`, cardinallity, to american spelling and vertex -> node
> * import of not_implemented_for and usage of not_implemented_for
>   (putting both directed and multigraph in the same line rules out MultiDiGraph only. It's an "and". We need two calls to rules out the "or" version (rules out directed. then rules out multigraph. so both types are rules out and only Graph is left).
> * reduced and cleaned up code for `Y_current` and `X_current`. I hope it is more readable, and should also speed it up.
> 
> Still to do:
> 
> * should it be called "minimum cost" or "minimum weight"? What does the literature (in but also outside of the paper you are using) call this? Then we can be consistent with one name for it. If both are used, we can say that somewhere, but otherwise be consistent with the name.  Similar for "cardinality" and "size". Let's choose one of those.
> * Add a description of each function after the first full paragraph that describes what "envy free matching" means. We need to know about what "maximum cardinality" means and what "min_weight" means. And it would be nice to know why someone might be interested in computing these quantities.
> * there are no tests for `envy_free_matching_partition`
> * need more description of the four sets in the partition. What properties the nodes have in each, what "matchings are contained in one partition" means (??"matchings involve only nodes in X_L and Y_L."??)
> * there are many places where the difference between "a matching" and "the matching" are important here. "a matching" implies there could be others. "the matching" implies it is the only one.
> * only use "good nodes" and "bad nodes" if you explain what you mean by that somewhere.

Hi Dan!
Thanks a lot for the review.

I saw the changes you applied and they look good to me, thanks for the help.

I will address the problems you indicated above one by one.

- "minimum cost" and "minimum weight" are terms used interchangeably in the scientific literature that mean the same thing (at least in this context).
- The same goes for "cardinality" and "size"

In terms of picking one of the terms to use throughout the documentation, I would go with the simple, more intuitive terms - so I would pick "minimum weight" and "size". I think this choice of words has a better chance of being recognizable by people with a lesser background in math and related research.

That being said, looking at the rest of the bipartite/matching.py module, it seems the terms "cardinality" (for example, in "hopcroft_karp_matching") and "minimum weight" (for example, in "minimum_weight_full_matching") are used by the functions that were there previously ("size" and "minimum cost" aren't used at all), so I will use the same terms to make my documentation uniform with the existing code as I can, to make my contributions fit in.

I will go ahead and add a commit that unifies this choice of words.

danigil:
Hi @dschult, I addressed all the points you mentioned, please take a look!

networkx/algorithms/bipartite/matching.py
```diff
@@ -378,7 +379,8 @@ def _alternating_dfs(u, along_matched=True):
     # check for alternating paths starting with edges not in the
     # matching.
     return _alternating_dfs(v, along_matched=True) or _alternating_dfs(
-        v, along_matched=False
+        v,
+        along_matched=False,
```
dschult, line 383: This passes the linters in its original form...  Is there a reason to change it?

networkx/algorithms/bipartite/matching.py
```diff
@@ -415,7 +417,11 @@ def _connected_by_alternating_paths(G, matching, targets):
         for v in G
         if v in targets
         or _is_connected_by_alternating_path(
-            G, v, matched_edges, unmatched_edges, targets
+            G,
+            v,
+            matched_edges,
+            unmatched_edges,
+            targets,
```
dschult, line 424: similarly here -- this should pass the linters as one line of arguments...

networkx/algorithms/bipartite/matching.py
```diff
@@ -577,7 +582,11 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
     # where edges are missing (as opposed to zeros, which is what one would
     # get by using toarray on the sparse matrix).
     weights_sparse = biadjacency_matrix(
-        G, row_order=U, column_order=V, weight=weight, format="coo"
+        G,
+        row_order=U,
+        column_order=V,
+        weight=weight,
+        format="coo",
```
dschult, line 589: same here.

networkx/algorithms/bipartite/matching.py
```diff
@@ -587,3 +596,363 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+    r"""Return a unique EFM (Envy Free Matching) partition of bipartite graph `G`.
+
+    An EFM partition of a bipartite graph with bipartite sets X,Y is a unique partition of X - X_L, X_S and Y - X_L, X_S.
+    This unique partition is proven to have the following properties:
+    1. Every matching that matches all the nodes in X_L and is contained in G[X_L,Y_L] (all edges are between X_L and Y_L)
+    is an envy-free matching.
+    2. Every envy-free matching in `G` is contained in G[X_L,Y_L].
+
+    Due to these properties we can say that X_L and Y_L are the "*good*" nodes and X_S and Y_S are the "*bad*" nodes,
+    because - in the context of finding envy-free matchings, we will only find them in X_L and Y_L.
```
dschult, line 701: There is a bunch of notation here that isn't introduced or explained.
Why are the subscripts `L` and `S`? That must stand for something.
What is the notation `X - X_L, X_S` representing? It doesn't look like a partition of X unless X_L is X_S. Or maybe it is a partial partition? or subtraction of sets is not what I'm expecting.

What is G[X_L, Y_L]?  

A good place to start would probably be to explain very briefly what an envy free matching is.
Make the first sentence shorter (it is used for links to the documentation for this function. Qualify terms like `G` being bipartite in the first sentence of the paragraph rather than the first line of the doc_string. 
Also, you can use latex in the paragraph -- but keep it simple because that's what looks ok in the docs.
 
Here's a start, but I don't know this material well, so please correct/change.
```
    r"""Returns an envy free matching partition of `G`.

    An Envy Free Matching (EFM) partition of a bipartite graph `G`
    with bipartite sets $(X, Y)$ is a unique partition into $(X - X_L, X_S)$
    and $(Y - X_L, X_S)$ where subscripts `L` and `S` denote....
```

networkx/algorithms/bipartite/matching.py
```diff
@@ -587,3 +596,363 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+def maximum_envy_free_matching(G, *, top_nodes=None):
+    r"""Return an envy-free matching of maximum cardinality of the bipartite graph `G`
+
+    A matching in a bipartite graph with bipartite sets X and Y is called envy-free,
+    if no unmatched node in X is adjacent to a matched node in Y.
+
+    An envy-free bipartite matching has *maximum cardinality* if it is the largest (has the most edges) possible matching
+    that is still envy-free.
+    There can be multiple different envy-free matching of maximum cardinality, this function returns *one* of them.
+
```
dschult, line 794: Remember to wrap the doc_strings to about 80 chars (like code, but not enforced by a linter).

dschult (commented):
Thanks for this and sorry for the delay in review.
This is mostly about doc_strings and getting a handle on changes outside the scope of the new functions. But the code seems very compact so should be easy to review once I understand the whole ecosystem of functions better.

So, theer's more for me to do, but here's some things to work on.
Thanks!

danigil:
Hi @dschult @MridulS 
A while back we created an interactive site for inputting graph representations and running the algorithms we implemented and getting the output as a plot with the various components and an explanation, it just occurred to me that you can use this site to understand how the algorithm works - I can't believe I haven't thought of this until now... 

Anyways, here's the link: https://efm.csariel.xyz/algo?type=non_weighted 
the weighted version is broken right now, maybe I'll fix it in the future.

networkx/algorithms/bipartite/matching.py
danigil: My code didn't pass the black linter, so I reformatted it using the black linter and maybe it caused these changes...
So I guess there isn't any reason to change it.

networkx/algorithms/bipartite/matching.py
danigil: Ditto ^

networkx/algorithms/bipartite/matching.py
danigil: Ditto ^

danigil: I pip installed the latest black linter package into my env and ran black linter on the code and it caused these changes, I'll reverse these changes.

networkx/algorithms/bipartite/matching.py
dschult: Can you give me some feedback on this change of docs? Or make the changes in the PR directly?
Thanks!

networkx/algorithms/bipartite/matching.py
danigil: Okay!

networkx/algorithms/bipartite/matching.py
danigil: Do you prefer that I'll explain the unclear things you mentioned here in the conversation and we can write docs together? I can also just edit the documentation in the code if you prefer that

danigil: I'll comment here for now, I think it can help quicken this process
the sets `X_L`, `X_S` and `Y_L`, `Y_S` are disjoint sets and  are defined as a partition of `X` and `Y` respectively, i.e `X = X_L union X_S` and `Y = X_L union X_S`
Furthermore, by definition (see article) `X_L = X \ X_S` and `Y_L = Y \ Y_S`
the algorithm works by constructing `X_S` and `Y_S` following a M-alternating sequence, hence `S` means **sequence**,
`X_L` and `Y_L` are the leftover nodes which didn't make it into `X_S` and `Y_S`, hence they're called **leftover**

I wrote " An EFM partition of a bipartite graph with bipartite sets X,Y is a unique partition of X - X_L, X_S and Y - X_L, X_S."
There is a typo - it should be X - X_L, X_S and Y - Y_L, Y_S. - sorry for this mistake!
perhaps the use of the `-` character is confusing as well. maybe it can be changed to "X := X_L U X_S and Y := Y_L U Y_S".
What do you think?

`G[X_L, Y_L]` means a subgraph of G induced by the nodes in `X_L` and `Y_L` - it means you take all the nodes and edges that connect nodes from `X_L` and `Y_L`, I'll add some brief explanation of this.

dschult: Thanks -- 

I think we can start with the definition of an Envy-Free-Matching partition:
```
An EFM partition of a bipartite graph with bipartite sets X,Y is a unique partition of X - X_L, X_S and Y - X_L, X_S.
```
That really doesn't define the EFM partition.  It says that it is a unique partition. But which one? What property separates it from all the other partitions?

Wikipedia says an [envy-free matching](https://en.wikipedia.org/wiki/Envy-free_matching) is defined as (my words):
```
Suppose G is a bipartite graph G = (X+Y, E). An envy-free matching is a
matching in which no unmatched node in X is adjacent to a matched node in Y.
That is, the matched nodes are not connected to unmatched nodes.
```

Notation:  Let's use latex in the doc_strings to write the symbols. Latex can be used in Sphinx processed doc_strings just by including the code between dollar signs.  And while the [relative complement](https://en.wikipedia.org/wiki/Complement_(set_theory)) of B within a set A is written A\B, it is also called the "set difference" and written A - B. The latter is close to Python code, so let's use that notation. Also, we might be able to avoid X_L completely by talking about "partitioning the nodes of X into those that are in the alternating sequence $X_S$ and those that are not." But I suspect you want to use $X_L$ in some places because those are the nodes that will be in the matching. So, in that sense $X_L$ is the more important set.  Maybe it should be called $X_M$ since those nodes are in the maximal matching.  But you can choose.

Perhaps the properties you say are "proven" are the properties that define the partition??? 
How about this:
```
An envy-free matching (EFM) partition of a bipartite graph $G(X+Y, E)$ separates nodes of G into those in the matching $X_M$ and $Y_M$, and those not in the matching, call them $X_S$ and $Y_S$. It is proven that $X_S$ and $Y_S$ form a M-alternating sequence for the envy-free matching $M$. The partition also represents a partition or X and of Y. It has been proven that the partition is unique for $G$, that every envy-free matching in $G$ is contained in the induced subgraph $G[X_M \union Y_M]$, and that every matching that matches all the nodes in $X_M$ and is contained in $G[X_M \union Y_M]$ is an envy-free matching.
```

I think a paragraph like this one would need to have the definition of an envy-free matching before defining the partition.  Sorry -- this is getting long. I hope I am on the right track.

danigil: Thanks for the quick response!

I think it's best to stick to the original terms that were used in the article since we're basing everything on it. So $X_L$ and not $X_M$

The whole doc of the EFM function says: 

    Return a unique EFM (Envy Free Matching) partition of bipartite graph `G`.

    An EFM partition of a bipartite graph with bipartite sets X,Y is a unique partition of X - X_L, X_S and Y - Y_L, Y_S.
    This unique partition is proven to have the following properties:
    1. Every matching that matches all the nodes in X_L and is contained in G[X_L,Y_L] (all edges are between X_L and Y_L)
    is an envy-free matching.
    2. Every envy-free matching in `G` is contained in G[X_L,Y_L].

    Due to these properties we can say that X_L and Y_L are the "*good*" nodes and X_S and Y_S are the "*bad*" nodes,
    because - in the context of finding envy-free matchings, we will only find them in X_L and Y_L.

I thought the first sentence that appears when you mouse over should be short and people who want a better explanation can look in the docs to see the rest of it.

Do you think it's better that the whole description fits in the first description sentence?
I can also make some fixes in this description above, something like:

    Return a unique EFM (Envy Free Matching) partition of bipartite graph `G`.

    An EFM partition of a bipartite graph with bipartite sets X,Y is a unique partition of $ X := X_L U X_S $ and $ Y := Y_L U Y_S$.
    This unique partition is proven to have the following properties:
    1. Every matching that matches all the nodes in X_L and is contained in the induced subgraph G[X_L,Y_L] (all edges are between X_L and Y_L)
    is an envy-free matching.
    2. Every envy-free matching in `G` is contained in the induced subgraph G[X_L,Y_L].

    Due to these properties we can say that X_L and Y_L are the "*good*" nodes and X_S and Y_S are the "*bad*" nodes,
    because - in the context of finding envy-free matchings, we will only find them in X_L and Y_L.

dschult: The first line should be short and can be less precise. It is used to describe a link to the docs for this function. That said, I would not use EFM or parens in it, just Envy Free Matching. We should define EFM in the paragraph.

I'm having trouble with the paragraph. It says that an EFM is a unique partition. And then it lists two sets that seem to be partitioned (so is it one partition or two partitions, or what?). Then it says that this partition is proven to have some properties -- but we haven't defined what makes the partition yet. How are we jumping to proving things about the partition when we haven't defined it yet.  Perhaps you mean that these properties **define** the partition and that it can be proven that this partition is unique??? but I'm guessing here.

dschult:
@erelsgl can you go through the documentation for this PR and make sure the definitions and statements of things that are proven are correct and precise? 

Also, just like writing a paper for publication, can you make sure the terminology is consistent and matches with terminology of similar code within NetworkX?  

It would be quite helpful to have the PRs prepared for review by the professor before having the relatively naive networkx maintainers try to work through the details. Think of it as a publication being prepared for the journal's editors and reviewers. Thanks!

erelsgl:
@dschult  I will be happy to write the documentation, but I am not sure where to put it in the file, as there are some terms that should be defined, that are related to several functions.

Maybe, instead of having all functions in the `matching.py` file, it is better to move them to a different file `envy_free_matching.py`? Then, the terms common to all functions can be clearly explained at the top of the file.

dschult:
Yes, let's put the envy-free matching code into a separate module with that name. That'll help with the documentation.
Thanks!

erelsgl:
@danigil  yes

danigil:
> @danigil yes

@erelsgl 
It's done.
I've output the envy-free matching code to `algorithms/bipartite/envy_free_matching.py` and `algorithms/bipartite/tests/test_envy_free_matching.py`

networkx/algorithms/bipartite/envy_free_matching.py
```diff
@@ -0,0 +1,375 @@
+import itertools as it
```
erelsgl, line 1: """
An *envy-free matching* (EFM) in an unweighted bipartite graph G = (X+Y, E) is matching in which no unmatched vertex in X is adjacent to a matched vertex in Y. 
The concept is motivated by problems of fair allocation. Suppose the vertices of X represent people, the vertices of Y represent houses, and an edge between a person x and a house y represents the fact that x is willing to live in y. Then, an EFM is a partial allocation of houses to people such that each house-less person does not envy any person with a house, since he/she does not like any allocated house anyway.

An empty matching is trivially envy-free. A perfect matching, in which all vertices of X are matched, is also envy-free. But there are graphs that do not have a perfect matching, but do have a non-empty envy-free matching.

This module provides the following functions, related to EFM:

* `maximum_envy_free_matching` - receives an unweighted bipartite graph, and computes an EFM of maximum cardinality.
* `minimum_weight_envy_free_matching`  - receives a weighted bipartite graph, and computes an EFM of minimum weight, among all EFM of maximum cardinality.
* `envy_free_matching_partition` - receives a bipartite graph G = (X+Y, E), and computes a partition of its vertices into two subsets: XL+YL and XM+YM, that satisfies some unique features related to EFM (details below)
"""

dschult: It'd be good to use `node` instead of `vertex` to match networkx. 
Thanks for this!

networkx/conftest.py
```diff
@@ -209,13 +209,14 @@ def add_nx(doctest_namespace):
...
 
-
 # List of files that pytest should ignore
 
 collect_ignore = []
 
 needs_numpy = [
     "algorithms/approximation/traveling_salesman.py",
+    "algorithms/bipartite/matching.py",
+    "algorithms/bipartite/tests/test_matching.py",
```
dschult, line 219: These lines in networkx/conftest.py also have to be updated to match the new module name for the envy-free-matching stuff. Since the now smaller `matching.py` doesn't use `numpy` (I think), you should remove that name and add the envy-free module name.  Similarly for `needs_scipy` below.

networkx/algorithms/bipartite/envy_free_matching.py
erelsgl: > It'd be good to use `node` instead of `vertex` to match networkx. Thanks for this!

Done

networkx/algorithms/bipartite/envy_free_matching.py
```diff
@@ -0,0 +1,375 @@
...
+        Y_subsets.append(Y_current)
+        X_subsets.append(X_current)
+
+        Y_subgroups_accumulate.update(Y_current)
+
+    return tuple(X_subsets), tuple(Y_subsets)
+
+
+def envy_free_matching_partition(G, *, M=None, top_nodes=None):
+    r"""Return a unique EFM (Envy Free Matching) partition of bipartite graph `G`.
```
erelsgl, line 109: """
Returns the envy-free-matching partition (EFM partition) of the given bipartite graph `G`.

Given a bipartite graph G = (X+Y, E), an EFM partition is a partition of the nodes into two disjoint parts: XS+YS and XL+YL, satisfying the following conditions:

(a) There are no edges between XS and YL ;
(b) The subgraph G[XL,YL] is X-saturated; this means that it contains a matching in which all vertices of XL are matched.
(c) The subgraph G[XS,YS] is Y-path-saturated; this means that it contains a structure similar to an odd path, which guarantees that the only EFM it contains is the empty matching.

Intuitively, G[XL,YL] is the "good" part of G, that contains the largest possible EFM; G[XS,YS] is the "bad" part of G, which contains only an empty EFM. 

In the paper [1], it is proved that there is a *unique* partition satisfying properties (a), (b), (c). Moreover, this this unique partition has the following additional properties:
(d) Every maximum matching in G[XL,YL] is an EFM in G.
(e) Every EFM in G is contained in G[XL,YL].

networkx/algorithms/bipartite/envy_free_matching.py
```diff
@@ -0,0 +1,375 @@
...
+)
+
+__all__ = [
+    "envy_free_matching_partition",
+    "maximum_envy_free_matching",
+    "minimum_weight_envy_free_matching",
+]
+
+
+def _M_alternating_sequence(G, M, *, top_nodes=None):
```
erelsgl, line 16: Maybe move this function to the bottom, as it is an internal function.

networkx/algorithms/bipartite/envy_free_matching.py
```diff
@@ -0,0 +1,375 @@
...
+    Y_S = set(it.chain.from_iterable(Y_subsets))
+    return set(X) - X_S, X_S, set(Y) - Y_S, Y_S
+
+
+@nx.utils.not_implemented_for("directed")
+@nx.utils.not_implemented_for("multigraph")
+def maximum_envy_free_matching(G, *, top_nodes=None):
+    r"""Return an envy-free matching of maximum cardinality of the bipartite graph `G`
+
+    A matching in a bipartite graph with bipartite sets X and Y is called envy-free,
```
erelsgl, line 205: Since we already defined EFM at the top of the file, I am not sure it is needed to redefine it here (but it may still be good to redefine, for people who read only this specific function).

erelsgl:
@dschult I have suggested new documentations for the top of the file and for the function envy_free_matching_partition.

PR #6351, opened 2023-01-08
add SIM11* lints

danieleades:
use 'any' and 'all' to simplify checking conditions against every element in an iterator


MridulS (approved):

danieleades:
merged into #6371

PR #6353, opened 2023-01-10
Add nx-guide link in the tutorial

paulitapb:
I thought that it would be useful for new users to have an explicit mention of the nx-guides in the tutorial. Also added a link to nx-guide using the intersphinx format.


rossbar (approved):
Thanks @paulitapb , this seems like a nice improvement to increase discoverability of nx-guides.

I took the liberty of capitalizing the 'X' in the section title to match the styling in the logo. My ulterior motive was to push something up so that CI would trigger since this was submitted when GH actions were down :). FWIW I built this locally and checked the link, which was indeed correct. Thanks again!

PR #6366, opened 2023-01-15
simplify stack in dfs

Tortar:
The stack in dfs can have only parent and children, the depth can be controlled from the outside of it, making the function slightly better in terms of time and space, but also readibility in my opinion :-)

### DIFF
--- a/networkx/algorithms/traversal/depth_first_search.py
+++ b/networkx/algorithms/traversal/depth_first_search.py
@@ -435,7 +435,7 @@ def dfs_labeled_edges(G, source=None, depth_limit=None):
         stack = [(start, iter(G[start]))]
         depth_now = 1
         while stack:
-            parent, depth_now, children = stack[-1]
+            parent, children = stack[-1]
             try:
                 child = next(children)
                 if child in visited:
--- a/networkx/algorithms/traversal/depth_first_search.py
+++ b/networkx/algorithms/traversal/depth_first_search.py
@@ -436,7 +436,7 @@ def dfs_labeled_edges(G, source=None, depth_limit=None):
         depth_now = 1
         while stack:
             parent, children = stack[-1]
-           for child in children:
+            for child in children:
                 if child in visited:
                     yield parent, child, "nontree"
                 else:
--- a/networkx/algorithms/traversal/depth_first_search.py
+++ b/networkx/algorithms/traversal/depth_first_search.py
@@ -74,25 +74,29 @@ def dfs_edges(G, source=None, depth_limit=None):
     else:
         # edges for components with source
         nodes = [source]
-    visited = set()
     if depth_limit is None:
         depth_limit = len(G)
+
+    visited = set()
     for start in nodes:
         if start in visited:
             continue
         visited.add(start)
-        stack = [(start, depth_limit, iter(G[start]))]
+        stack = [(start, iter(G[start]))]
+        depth_now = 1
         while stack:
-            parent, depth_now, children = stack[-1]
+            parent, children = stack[-1]
             try:
                 child = next(children)
                 if child not in visited:
                     yield parent, child
                     visited.add(child)
-                    if depth_now > 1:
-                        stack.append((child, depth_now - 1, iter(G[child])))
+                    if depth_now < depth_limit:
+                        stack.append((child, iter(G[child])))
+                        depth_now += 1
             except StopIteration:
                 stack.pop()
+                depth_now -= 1
 
 
 def dfs_tree(G, source=None, depth_limit=None):
@@ -427,7 +431,8 @@ def dfs_labeled_edges(G, source=None, depth_limit=None):
             continue
         yield start, start, "forward"
         visited.add(start)
-        stack = [(start, depth_limit, iter(G[start]))]
+        stack = [(start, iter(G[start]))]
+        depth_now = 1
         while stack:
             parent, depth_now, children = stack[-1]
             try:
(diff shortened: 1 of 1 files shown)


rossbar (commented):
I personally find it harder to reason about with the depth tracked separately outside of the stack, but that's admittedly pretty subjective. Out of curiosity what's the primary motivation for the proposed change?

Tortar:
From the speed point of view I measure from a tiny 3% better to a 15% better in the best case so not that much. I think that memory-wise the stack this way consumes probably 1/3 less space, which is good given that it is the second data structure in this method in terms of memory occupied, all in all, it is not much of a difference, but I found it interesting to think about this optimization.

rossbar:
> I think that memory-wise the stack this way consumes probably 1/3 less space

Well, at least 1/3 the *number of objects*, which is not the same as memory since `depth_now` is just an integer. I was curious what this would look like in practice I tested with `path_graph(10_000_000)`[^1] and used `memray` for memory profiling:

#### On main

mem_main

#### This PR

mem_patch8

So that's a 400 MB reduction in total usage and 300 fewer allocations for this call (which should help with performance), but accounting for the other changes it ends up being about 300 MB saving for `dfs_edges` (out of 6.8 GB, so ~4%).

Sorry for the bad screengrab quality - to reproduce (though you may want to use fewer nodes for systems with <16GB ram):

1. Create a script, e.g. `dfs_memtest.py`

   ```python
   import networkx as nx
   G = nx.path_graph(10_000_000)
   e = list(nx.dfs_edges(G, source=0))
   ```

2. Follow the [memray instructions to create the flamegraph](https://github.com/bloomberg/memray#usage)

[^1]: The reasoning here was that this graph would maximize the memory saving, as each node only ever has one child, and the nodes are also integers.

Tortar:
that's interesting, thanks for the memgraph and the explanation on how to reproduce it!

rossbar:
>  it is actually clearer with this edit

This is inherently subjective - there is no "right" way to do it. The reason I find it more readable is that the state (i.e. the current depth) is tracked with the current node, whereas the proposed change tracks the state separately. Though again, this is entirely subjective so we're unlikely to make any progress discussing on this front :)

A minor improvement to memory consumption is a more concrete justification for the proposed change. I'd like to hear what others think!

Tortar:
indeed I was too assertive, I meant to say that it is easier to track when we change levels in the dfs, but anyway the real motivation  for the change is the little improvement in memory/time!

ImHereForTheCookies:
This makes sense to me. 👍

boothby:
FWIW I agree that a local value is preferable to loading the stack down with longer state tuples.  I prefer a slightly different idiom than the one used here:

```python
     while stack:
            parent, children = stack[-1]
            try:
                child = next(children)
                ....
            except StopIteration:
                stack.pop()

# without try/except

    while stack:
            parent, children = stack[-1]
            for child in children:
                ...
                break
            else:
                stack.pop()

```
In avoiding exceptions, the `for,break,else` tends to be a little faster.

boothby:
@Tortar I have implemented my suggestion here: https://github.com/boothby/networkx/tree/tortar-patch-8 -- I would push the changes to your repo but the pre-commit hooks are currently faulty in your branch (not your fault; easily resolved with `git rebase -i main`) -- after rebasing I cannot push to your branch without `--force` which is awfully impolite.

Note that we only need to `break` when we grow the stack -- the `for,break,else` pattern is most effective in `dfs_labeled_edges` where non-tree edges are yielded "for free" within the iterator resulting in fewer touches on the stack:

```python
        while stack:
            parent, children = stack[-1]
            for child in children:
                if child in visited:
                    yield parent, child, "nontree"
                else:
                    yield parent, child, "forward"
                    visited.add(child)
                    if depth_now < depth_limit:
                        stack.append((child, iter(G[child])))
                        depth_now += 1
                        break
                    else:
                        yield parent, child, "reverse-depth_limit"
            else:
                stack.pop()
                depth_now -= 1
                if stack:
                    yield stack[-1][0], parent, "reverse"
```

networkx/algorithms/traversal/depth_first_search.py
```diff
@@ -74,25 +74,29 @@ def dfs_edges(G, source=None, depth_limit=None):
...
                 child = next(children)
                 if child not in visited:
                     yield parent, child
                     visited.add(child)
-                    if depth_now > 1:
-                        stack.append((child, depth_now - 1, iter(G[child])))
+                    if depth_now < depth_limit:
+                        stack.append((child, iter(G[child])))
+                        depth_now += 1
             except StopIteration:
```
boothby, line 97: see comments above

networkx/algorithms/traversal/depth_first_search.py
```diff
@@ -419,30 +423,34 @@ def dfs_labeled_edges(G, source=None, depth_limit=None):
...
                     yield parent, child, "forward"
                     visited.add(child)
-                    if depth_now > 1:
-                        stack.append((child, depth_now - 1, iter(G[child])))
+                    if depth_now < depth_limit:
+                        stack.append((child, iter(G[child])))
+                        depth_now += 1
                     else:
                         yield parent, child, "reverse-depth_limit"
             except StopIteration:
```
boothby, line 451: see comments above

boothby (changes requested):

Tortar:
thanks @boothby for the improvements! will push them in a moment and make you a co-author of the pull :-)

boothby (approved):

dschult (approved):
There is also #6714 which looks to fix #6479...  

I think we can/should go ahead and merge these changes as a first step -- and then address the strange behavior seen with depth-limited search. 

I'm approving this and I'll have @boothby press the green button. There are a lot of PRs related to this and the order may impact merge conflicts. Best to have control in those cases I guess. :}

Thanks!

PR #6369, opened 2023-01-16
Fix ecuation in clustering documentation

### LINKED ISSUE #6365: Small documentation error in clustering coefficient for weighted networks
### Current Behavior

The networkx documenation gives the following formula for the clustering coefficient for unweighted directed networks: 
https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.clustering.html#networkx.algorithms.cluster.clustering

_Picture from networkx documenation_
grafik

Please note the factor 2 in the numerator. 
In the original publication the equation is the following: 

_Picture from original publication_
grafik
Please note the factor 2 in the denominator.

I checked the implementation and there it seems to be correct [here](https://github.com/networkx/networkx/blob/59f5b01c80064b3b88e1d044bab1c4cf61fdbc85/networkx/algorithms/cluster.py#L376) (great work to whoever made it so easy to compare it to the equation from the original publication 🎉 )

```python
t / ((dt * (dt - 1) - 2 * db) * 2)
```
Please note the factor two in the denominator.

### Expected Behavior
I think the equation in the documentation should be updated so that the 2 is in the denominator.

paulitapb:
Closes #6365


MridulS (approved):
Thanks @paulitapb

rossbar (approved):
Agreed - this now matches the implementation. Thanks @paulitapb !

PR #6370, opened 2023-01-16
Maximum Priority Matching

RoiMeshulam:
We added the Maximum Priority Algorithm directory.
This directory is implementation of two algorithms that displayed in the articles 'Maximum Priority Matching' and 'Faster Maximium Priority Matchings in Bipartite Graphs' that were written by Jonathan Turner.

In addition we used cython functions in order to improve the run time of the algorithms and there is an option to use them or run the algorithm without them.


MridulS:
Just FYI, we do not accept cython/C code in networkx. NetworkX is a pure python library and it isn't changing anytime soon (I think)

PR #6373, opened 2023-01-16
Add reference to paper in vf2pp

paulitapb:
I noticed there's no reference in the vf2++ documentation. So I added a reference section with the paper where the algorithm is described.


rossbar (approved):
Nice, catch! I took liberty of slightly adjusting the spacing to conform to the numpydoc standard and also added the reference to the top of the module docstring. Thanks @paulitapb !

PR #6376, opened 2023-01-18
New functions for efficient subgraph isomorphism using genetic algorithm in networkx

Lioo7:
Hello,

We are excited to contribute to your library by adding two new (main) algorithms: "find_subgraph_isomorphism_with_genetic_algorithm" and "find_subgraph_isomorphism_with_genetic_algorithm_cython".

These functions use a genetic algorithm to improve the subgraph isomorphism, which is an NP-hard problem. The algorithm is based on the paper "Malware Detection based on Dependency Graph using Hybrid Genetic Algorithm" by Keehyung Kim and Byung-Ro Moon.

We have also included a Cython implementation of the function, "find_subgraph_isomorphism_with_genetic_algorithm_cython" for improved performance.

We hope that these functions will be a valuable addition to the networkx library and we look forward to your feedback.

Thank you,
Bar and Lioz


MridulS:
Thanks for the contribution! But networkx is a pure python library. We don't have cython code, or planning on adding it anytime soon.

dschult:
Could Cython code be provided by a separate library that provides a backend for NetworkX? I'm thinking that the data storage is the same, but the algorithm is written in cython (or anything else). Would that work?  It still wouldn't be part of the networkx package. But it would be useable by people who install both networkx and the separate library. The separate library would be in charge of keeping itself working with networkx and with the networkx backend interface.  Am I missing anything?  Is this something we want to encourage or discourage or neither?

MridulS:
Cython in a different package that is compatible with NetworkX is something we can definitely do! It can work with the backend plugin system too, so just a single import on top of the script can let users use cython versions. If there are people in interested in taking it up and maintaining the package then we can definitely have a cythonised networkx :)

Lioo7:
Thank you for your feedback. 
We understand that networkx is a pure python library and we apologize for any inconvenience caused by including the Cython implementation in our pull request. 
We will remove the Cython code and only submit the pure python implementation. 
We will also work on fixing the code to pass all the checks. 

We appreciate the suggestion to provide the Cython code in a separate library that is compatible with networkx. 
We are excited to hear that this is something that can be encouraged and we can take on the task of creating and maintaining a separate package with the Cython code. 
We will ensure that this package is compatible with networkx and can be used with the backend plugin system. 

We also will fix the code to pass all the checks. Please keep in mind that we are both students and this is our first attempt to contribute code to an open source package, so we will appreciate any feedback. 

If it's not too much trouble, could you also explain the reason why networkx is a pure python library? And if it's common to use multiple languages in other packages, why not in networkx?

Thanks

Lioo7:
All Cython code has been removed as requested and all tests have passed. Ready for review.

dschult:
I took a quick pass through this code and it looks to me like this code is better suited for a separate library.  The functions are very specific to malware detection. That is, they are not constructed for general subgraph isomorphism problems. I think you've got some good ideas and the cython code is probably really helpful too.  Why don't you make a release of a library for malware detection that uses NetworkX where that is helpful?

erelsgl:
Is `genetic_subgraph_isomorphism` really specific to malware detection? I think it can be used in any situation where we need a heuristic answer to whether one graph is isomorphic to a subgraph of another. It seems like a quite general problem.

Lioo7:
Thanks for your feedback. We appreciate your insights and agree with both of you. We are considering releasing a library for malware detection that uses NetworkX where that is helpful. However, we agree with Erel, in our humble opinion the function we wrote (genetic_subgraph_isomorphism) can be used for much more than just malware detection and can be used for general subgraph isomorphism problems as well.

PR #6380, opened 2023-01-19
Remove an instance of random.sample from a set (deprecated in Python 3.9)

eriknw:
I encountered this DeprecationWarning today. This change was missed in #4602


rossbar (approved):
Nice catch, thanks @eriknw !

The fact this wasn't caught previously means this branch is probably untested :upside_down_face:

MridulS (approved):

dschult (approved):

PR #2265, opened 2016-09-27
Implements all-pairs shortest paths in parallel

jfinkels:
This commit modifies the implementation of the unweighted all-pairs
shortest paths to use the multiprocessing module.

This pull request is intended to experiment with employing parallelism directly within NetworkX. See issue #356.

PR #6382, opened 2023-01-20
all_pairs_shortest_path_length now runs in multiple cores.

miguel-bermudo:
Added multiprocessing capabilities to this function for unwheighted graphs, tested it for Freebase, NELL and WordNet, speed increased significantly.

### DIFF
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -2,6 +2,8 @@
 Shortest path algorithms for unweighted graphs.
 """
 import networkx as nx
+from multiprocessing import Process, Queue, cpu_count
+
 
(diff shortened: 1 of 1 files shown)


rossbar (commented):
`TODO` notwithstanding, I think this would be a better fit for a parallel backend. Introducing parallelism directly complicates both the readability and portability of the function, so I'm generally -1 on parallelizing internally.

Check out this discussion for details about the new possibilities of parallel backends with NX 3.0: 

https://github.com/networkx/networkx/pull/6306#issuecomment-1385383777

miguel-bermudo:
Will keep this for my own purposes and keep an eye on the efforts for multiprocessing as its own separate package.
as #2265 already offered a solution to this problem.

PR #6413, opened 2023-02-06
Add Plausible Analytics to our docs

MridulS:
This gives us an anonymous way of tracking usage of our documentation website. By this PR we will only be able to track usage of the latest dev website, which I think is a good start.

This uses the instances hosted by the Scientific Python project.

Should we also put up some disclaimer or note in our README that we are using this?

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -1,9 +1,8 @@
-{% extends "!layout.html" %} 
-
-{% block extrahead %}
-    <script defer data-domain="networkx.org" src="https://views.scientific-python.org/js/script.js"></script>
-    {{ super() }}
-{% endblock %}
-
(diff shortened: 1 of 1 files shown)


MridulS:
Tried on a couple of random pages, seems to load fine.

Screenshot 2023-02-07 at 12 50 53 AM

dschult:
Looks good!  
I guess prettier pre-commit wants different white-space.
But it's good to see that the code is in place and works. (at least the page loads and the script is in the html)
:)

rossbar:
> Should we also put up some disclaimer or note in our README that we are using this?

Good question - probably a good discussion topic for the community meeting!

MridulS:
I'll go ahead and merge this one, as discussed in the previous meeting. Fixed the lint issue.

Issue #6428, opened 2023-02-13
Reference for inverse_line_graph


barbireau:
There is a spelling mistake in the reference for the documentation of inverse_line_graph (latest version). 

https://networkx.org/documentation/stable/reference/generated/networkx.generators.line.inverse_line_graph.html

The author is "N. D. Roussopoulos" instead of N. Roussopolous (note the location of "ou"). Also, it would be nice to add a link or DOI to the article. The complete bibliographic info is below. 

@article{ROUSSOPOULOS1973108,
title = {A max {m,n} algorithm for determining the graph H from its line graph G},
journal = {Information Processing Letters},
volume = {2},
number = {4},
pages = {108-112},
year = {1973},
issn = {0020-0190},
doi = {https://doi.org/10.1016/0020-0190(73)90029-X},
url = {https://www.sciencedirect.com/science/article/pii/002001907390029X},
author = {Nicholas D. Roussopoulos},
keywords = {graph, maximal subgroup, complete graph, complete bipartite, automorphism}
}

Blueclaus13:
Could you assign to me this issue?

barbireau:
I think I can't do that. Can somebody else do this? Possibly @rossbar?

dschult:
We don't usually use the "assign" function in Github to assign Issues.
Please open a PR with the fix for this issue and we will review it.
Thanks!

Blueclaus13:
great thanks!

Blueclaus13:
I'm sorry @barbireau I have a question with your dio link. It redirects me to a DOI not found when I click on it. I'm doing something wrong or the link is wrong ?

barbireau:
The DOI link should be without the "}" at the end. Github automaticaly makes the doi a clickable link, but adds the "}" by mistake.

Blueclaus13:
> I'm sorry @barbireau I have a question with your dio link. It redirects me to a DOI not found when I click on it. I'm doing something wrong or the link is wrong ?

ooh I got it, it took a square bracket as part of the link

PR #6431, opened 2023-02-15
DocumentationFixed-in-line.py-inverse_line_graph

Blueclaus13:
This is a pull request with the changes in the documentation to resolve "Reference for inverse_line_graph #6428 "


dschult:
It looks like an extra file got included in the `git commit`.  You should be able to do something like:
```
git reset HEAD~1
git restore get-pip.py
git add networkx/generators/line.py
git commit
git push 
```

PR #6432, opened 2023-02-15
Add project desc for visualization and ISMAGs

dschult:
I've added two possible project topics.

I could also add isomorphism VF3 or RI if we want to continue to focus on cleaning up the isomorphism subpackage.

I removed my name from the Notebooks project mentoring because I haven't been doing much review there.
Please add your name as an interested mentor if you are interested. (or leave a comment and I will add it).

Suggestions?

### DIFF
--- a/doc/developer/projects.rst
+++ b/doc/developer/projects.rst
@@ -58,6 +58,8 @@ Visualization API with Matplotlib
 
 - Interested Mentors: `@dschult <https://github.com/dschult/>`__,
 
+- Expected time commitment: This project will be a full time 10 week project (~350 hrs).
+
 Incorporate a Python library for ISMAGs isomorphism calculations
 ----------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


rossbar (approved):
+1 for both of these additions. Thanks @dschult !

doc/developer/projects.rst
```diff
@@ -32,15 +32,50 @@ Pedagogical Interactive Notebooks for Algorithms Implemented in NetworkX
...
+  subpackages of tools, or more likely a combination of the two sets of code, or a
+  third incantation that combines good features from each.
+
+- Recommended Skills: Python, graph algorithms.
+
+- Expected Outcome: A plan for how to best incorporate ISMAGS into NetworkX along
+  with code to do that incorporation.
+
+- Interested Mentors: `@dschult <https://github.com/dschult/>`__,
+
```
MridulS, line 78: Both sound perfect!
Could you also add a section on `Expected time commitment` as then we will be able to use these as GSoC ideas too. There are supposed to be either 350 or 175 hours projects.

dschult: I think both of these projects could be even more than full time for 10 weeks, but that's what I put because both outreachy and GSOC are 10 weeks. It seems like it would be harder to make enough progress in the 175 hrs category.

PR #6441, opened 2023-02-19
Start using ruff for pyupgrade and isort

MridulS:
Use ruff's isort and pyupgrade in the pre-commit pipeline.

All works except there is a ruff-isort change in `networkx/algorithms/centrality/betweenness_subset.py`, not a 100% sure if this is an issue or not. I'm fine either way.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,4 +1,6 @@
 select = [
+    "I",       # isort
+    "UP",      # pyupgrade
     "C4",      # flake8-comprehensions
     "E713",    # use 'key not in list'
     "PIE",     # flake8-pie
@@ -11,3 +13,12 @@ select = [
     "SIM118",  # use 'key in dict'
     "SIM2",    # simplify boolean comparisons
 ]
+
+target-version = "py38"
+
+[per-file-ignores]
+"__init__.py" = ["I"]
+"setup.py" = ["I"]
+"examples/*.py" = ["I"]
+"doc/*.py" = ["I"]
+"tools/*.py" = ["I"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,22 +6,10 @@ repos:
     rev: 23.1.0
     hooks:
       - id: black
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
-    hooks:
-      - id: pyupgrade
-        args: [--py38-plus]
   - repo: https://github.com/asottile/blacken-docs
     rev: 1.13.0
     hooks:
       - id: blacken-docs
-  - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
-    hooks:
-      - id: isort
-        name: isort (python)
-        args: ["--profile", "black", "--filter-files", "--skip", "__init__.py"]
-        files: ^networkx/
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v2.7.1
     hooks:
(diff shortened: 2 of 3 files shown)


dschult:
Looks like there are 3 imports in `betweenness_subset.py` that are from the same module and could be combined into a single import statement. I'm not sure why ruff's isort would spread a line with 69 chars into a 3 line statement, but if it is going to do that we might as well combine it with the other two. (unless there is a style rule that says we shouldn't import more than one function at a time(?))

And I'm not sure whether that should be done in this PR or another one. Probably doesn't matter...

danieleades:
there are a couple of open issues in ruff for small inconsistencies between it and isort, so i don't think that's quite settled yet.
That said, I don't think they're worth losing any sleep over. You might see some of the changes revert down the line after a ruff update

jarrodmillman (approved):

ruff.toml
```diff
@@ -11,3 +13,12 @@ select = [
...
 ]
+
+target-version = "py38"
+
+[per-file-ignores]
+"__init__.py" = ["I"]
+"setup.py" = ["I"]
+"examples/*.py" = ["I"]
+"doc/*.py" = ["I"]
+"tools/*.py" = ["I"]
```
danieleades, line 24: what is the reason for skipping these?

MridulS: ruff (isort) will mess up the whole import structures and break the package if we let it format `__init.py__` , and we were running isort only on the `networkx` folder. So I'm skipping the other folders to make ruff work without making any changes. I think we can safely un-skip examples, docs and tools folder.

danieleades: Would it not be preferable to use per-line 'isort:skip' invocations for this use case?

MridulS: Wouldn't that mean the whole file would have `isort:skip`  (the `__init.py__` ones)? As the `__init__.py` are literally just imports all around.

danieleades: That depends. Why does sorting imports in init files in this package cause them to break?

Isort has loads of options for controlling the way it is applied in the case that a subset of imports are order-dependent

See https://pycqa.github.io/isort/docs/configuration/action_comments.html

MridulS: > That depends. Why does sorting imports in init files in this package cause them to break?

We want the user namespace to be filled in a certain way, as this requires us to follow a certain import order. Like to have the `py_random_state` decorator requires `create_py_random_state` to be imported first from utils.misc.

> Isort has loads of options for controlling the way it is applied in the case that a subset of imports are order-dependent
> See https://pycqa.github.io/isort/docs/configuration/action_comments.html

Well in that case let's just put this in a single config file, rather than go grep-ing for `# isort:` to find what we skip and what we do not. 

Let's remember none of this makes actual difference to the package itself, it's all for aesthetic changes so we don't have to think about any of this ever again :)

danieleades: > > That depends. Why does sorting imports in init files in this package cause them to break?
> 
> We want the user namespace to be filled in a certain way, as this requires us to follow a certain import order. Like to have the `py_random_state` decorator requires `create_py_random_state` to be imported first from utils.misc.

sorry, i'm struggling to understand this. Please bear with me!

i see that `utils.misc` exports a method called `create_py_random_state` which is re-exported in `utils.__init__`. The `py_random_state` decorator is then defined in `utils.decorators`. Which part of that is impacted by the order of imports in `utils.__init__`?

assuming there is genuinely an order dependency there, can it be resolved by changing the import in `utils.decorators` from

```python
...
from networkx.utils import create_py_random_state, create_random_state
```

to

```python
...
from networkx.utils.misc import create_py_random_state, create_random_state
```
?

It seems to me that the current approach invites circular dependencies a little more readily than it needs to.

danieleades: that seems fair enough. can save that for a different PR!

rossbar (approved):

PR #6444, opened 2023-02-21
Ruff/more lints

danieleades:


dschult:
I thought the idea was to include as much of the Ruff changes into one PR so we don't have too much trouble tracking the "blame" and we don't have to deal with never-ending ongoing improvements to linting.  Maybe we can make a plan for how to handle this.  @danieleades do you have an end in sight? Do you have thoughts on how to make this work well without taking too much maintainer time?  (and Thanks for all you are doing!)

danieleades:
Honestly, no I don't really see an end in sight. I think these are the last of the fairly uncontroversial lints, which *could* have been bundled with the first lot, but in general I think it's healthier to think of this as a programme of gradual improvement.

- there are literally hundreds of further lints that *could* be added
- many of these add value but require more significant changes, so would need to considered individually
- ruff has multiple releases per week with additional rules added all the time

For illustration, I would suggest enabling the full suite of "E" and "F" lints locally and observing the volume of suggested changes (this corresponds to the full set of rules provided by `flake8` natively, with no plugins). Also take a look at the (current) full rule list - https://beta.ruff.rs/docs/rules/

For example, this repo has scores of unused imports which are not part of the public API. It also uses glob imports that hamstring an IDEs ability to detect unused or circular imports. There are lints in ruff for both of these, but such changes would be larger, a bit more complex, and need more careful review than the changes introduced so far.

In the longer term, ruff will add 'multi-file' linting which would allow it to begin to subsume tools like mypy and pylint.

rossbar (commented):
The linting config has gone through a few changes since this was opened so I suspect this has been superseded. I'll go ahead and close it as it's probably easier to propose/review linting changes in a fresh PR using the current linting configuration as a starting point!

PR #6449, opened 2023-02-24
adds bfs_edges kwrags in sort_neighbors to deprecations.rst

still-n0thing:
Refer https://github.com/networkx/networkx/pull/5925


MridulS:
Could you please add this in https://github.com/networkx/networkx/pull/5925.
Thanks!

PR #6454, opened 2023-02-27
Added tests for lazy imports

Mjh9122:
Added two test and updated another. 
First tests importing a module lazily from its spec, pip was the first module I thought of.
Second I checked that the sub-modules were correctly attempting to be imported. The error message ensures that the correct error is being caught. 
Lastly I tested an import when submodules and submod_attrs are allowed to default to None. Also the error generation when calling __getattr__ with a submodule that shouldn't be added.

### DIFF
--- a/networkx/tests/test_lazy_imports.py
+++ b/networkx/tests/test_lazy_imports.py
@@ -28,6 +28,13 @@ def test_lazy_import_basics():
         pass
 
 
+def test_lazy_import_from_spec():
+    pip = lazy._lazy_import("pip")
+    assert type(pip) == importlib.util._LazyModule
+    pip.__str__()
+    assert type(pip) == types.ModuleType
+
+
 def test_lazy_import_impact_on_sys_modules():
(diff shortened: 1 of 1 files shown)


rossbar (commented):
Just to comment - IME tests involving the import system can be fragile. For example, it's difficult to keep tests self-contained after an import as globals (e.g. `sys.modules`) are altered.

I'm ambivalent about adding these tests, especially if the motivation is only to improve the coverage metric. If OTOH these were test cases motivated by practical considerations then I'd be +1.

I'm certainly no expert though, @stefanv thoughts about testing `_lazy_import` here?

stefanv:
If the tests become as difficult to assemble as the original code, it always gives pause :) If the code is actively being used in the project, it's probably implicitly tested all over the place, so not sure if this is necessary. So :man_shrugging:, I'm ambivalent about adding this unless it catches observed bad behavior.

rossbar (commented):
I'd be in favor of just closing

PR #6455, opened 2023-02-27
Added two tests for convert_numpy

Mjh9122:
Added test for when the node list includes a node that is not in the graph.
Added test for when the incoming array is 3 dimensional.


rossbar (approved):
LGTM, thanks @Mjh9122

PR #6461, opened 2023-03-03
add simple cycle enumerator for undirected class

boothby:
The need to enumerate simple cycles in a graph has come up frequently in my life and it's a common question that people have.  There are two ways of accomplishing this:

1. enumerate all connected sums of cycles in a cycle basis
2. exhaustive search

The first is probably the most efficient way to get all cycles, but filtering down to certain cycle lengths would not be efficient at all.  The code here is reasonably fast given the problem it is solving, but leaves some performance improvements on the table.  The biggest one is that I should probably iterate over 2-edge-connected components to avoid brute-forcing large trees.

This is a work in progress; I have not yet added tests.  And, of course, I'm open to discussion on how best to handle this -- I was unsure whether or not to dispatch this algorithm with `all_simple_cycles` in the directed case.  Also, there is a question of credit:  this was the obvious brute-force algorithm when I naively sat down with the problem, and I've found a few occurrences of this or related algorithms in literature and code snippets online; [a 2006 paper claims its novelty](https://ieeexplore.ieee.org/document/1602189) but that paper does not appear to handle equivalence by reversal.  The only point of novelty that I see is that other algorithms (including an ancient implementation of mine) assume sortability of node labels, which is not strictly necessary.  I've included a flag to enumerate chordless cycles, which is effectively a free addition to the algorithm -- again, there are several variations of this in the literature / internet.

### DIFF
--- a/doc/reference/algorithms/cycles.rst
+++ b/doc/reference/algorithms/cycles.rst
@@ -11,3 +11,4 @@ Cycles
    recursive_simple_cycles
    find_cycle
    minimum_cycle_basis
+   chordless_cycles
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -110,6 +110,16 @@ def simple_cycles(G, k=None):
     algorithm of Gupta and Suzumura [2]_. There may be better algorithms for
     some cases [3]_ [4]_ [5]_.
 
+    The algorithms of Johnson, and Gupta and Suzumura, are enhanced by some
+    well-known preprocessing techniques.  When G is directed, we restrict our
+    attention to strongly connected components of G, generate all simple cycles
+    containing a certain node, remove that node, and further decompose the
+    remainder into strongly connected components.  When G is undirected, we
+    restrict our attention to biconnected components, generate all simple cycles
+    containing a particular edge, remove that edge, and further decompose the
+    remainder into biconnected components.
+
+
     Parameters
     ----------
     G : NetworkX DiGraph
@@ -130,9 +140,11 @@ def simple_cycles(G, k=None):
     >>> G = nx.DiGraph(edges)
     >>> sorted(nx.simple_cycles(G))
     [[0], [0, 1, 2], [0, 2], [1, 2], [2]]
+
     To filter the cycles so that they don't include certain nodes or edges,
     copy your graph and eliminate those nodes or edges before calling.
     For example, to exclude self-loops from the above example:
+
     >>> H = G.copy()
     >>> H.remove_edges_from(nx.selfloop_edges(G))
     >>> sorted(nx.simple_cycles(H))
@@ -140,9 +152,9 @@ def simple_cycles(G, k=None):
 
     Notes
     -----
-    The implementation follows pp. 79-80 in [1]_.
-    The time complexity is $O((n+e)(c+1))$ for $n$ nodes, $e$ edges and $c$
-    elementary circuits.
+    When k is None, the time complexity is $O((n+e)(c+1))$ for $n$ nodes, $e$
+    edges and $c$ simple circuits.  Otherwise, when $k > 1$, the time complexity
+    is $O((c+n)(k-1)d^k)$ where $d$ is the average degree of the nodes of G.
 
     References
     ----------
@@ -218,8 +230,11 @@ def _directed_cycle_search(G, k):
     list of nodes
        Each cycle is represented by a list of nodes along the cycle.
     """
-    components = list(nx.strongly_connected_components(G))
-    for c in components:
+
+    scc = nx.strongly_connected_components
+    components = [c for c in scc(G) if len(c) >= 2]
+    while components:
+        c = components.pop()
         Gc = G.subgraph(c)
         v = next(iter(c))
         if k is None:
@@ -228,7 +243,7 @@ def _directed_cycle_search(G, k):
             yield from _bounded_cycle_search(Gc, [v], k)
         # delete v after searching G, to make sure we can find v
         G.remove_node(v)
-        components.extend(nx.strongly_connected_components(Gc))
+        components.extend(c for c in scc(Gc) if len(c) >= 2)
 
 
 def _undirected_cycle_search(G, k):
@@ -266,63 +281,22 @@ def _undirected_cycle_search(G, k):
        Each cycle is represented by a list of nodes along the cycle.
     """
 
-    components = list(nx.biconnected_components(G))
+    bcc = nx.biconnected_components
+    components = [c for c in bcc(G) if len(c) >= 3]
     while components:
         c = components.pop()
         Gc = G.subgraph(c)
-        uv = _biconnected_first_nontree_edge(Gc)
+        uv = list(next(iter(Gc.edges)))
         G.remove_edge(*uv)
         # delete (u, v) before searching G, to avoid fake 3-cycles [u, v, u]
         if k is None:
             yield from _johnson_cycle_search(Gc, uv)
         else:
             yield from _bounded_cycle_search(Gc, uv, k)
-        components.extend(nx.biconnected_components(Gc))
+        components.extend(c for c in bcc(Gc) if len(c) >= 3)
 
 
-def _biconnected_first_nontree_edge(G):
-    """A depth first search which returns the first non-tree edge
-
-    This is a specialized, optimistic variant of dfs_labeled_edges equivalent to
-    in the case that G is biconnected.
-
-        for u, v, label in dfs_labeled_edges(G):
-            if label == 'nontree':
-                return [u, v]
-
-    This must only be called on biconnected components.  If the graph is
-    connected but not biconnected, then this function will return None.  If the
-    graph is not connected, it may return an edge or None.
-
-    Parameters
-    ----------
-    G : NetworkX Graph
-       A connected, undirected graph
-
-    Returns
-    -------
-    list of nodes
-       A non-tree edges of a depth-first search of G
-
-    See Also
-    --------
-    dfs_labeled_edges
-    """
-
-    visited = set()
-    stack = [next(iter(G))]
-    while stack:
-        u = stack.pop()
-        if u not in visited:
-            visited.add(u)
-            Gu = G[u]
-            nontree = visited.intersection(Gu)
-            for v in nontree:
-                return [u, v]
-            stack.extend(Gu)
-
-
-class _NeighborCache(dict):
+class _NeighborhoodCache(dict):
     """Very lightweight graph wrapper which caches neighborhoods as sets.
 
     This dict subclass uses the __missing__ functionality to query graphs for
@@ -362,7 +336,7 @@ def _johnson_cycle_search(G, path):
 
     """
 
-    G = _NeighborCache(G)
+    G = _NeighborhoodCache(G)
     blocked = set(path)
     B = defaultdict(set)  # graph portions that yield no elementary circuit
     start = path[0]
@@ -424,7 +398,7 @@ def _bounded_cycle_search(G, path, k):
        https://doi.org/10.1137/0204007
 
     """
-    G = _NeighborCache(G)
+    G = _NeighborhoodCache(G)
     lock = {v: 0 for v in path}
     B = defaultdict(set)
     start = path[0]
@@ -453,7 +427,7 @@ def _bounded_cycle_search(G, path, k):
                 while relax_stack:
                     bl, u = relax_stack.pop()
                     if lock.get(u, k) < k - bl + 1:
-                        Lock[u] = k - bl + 1
+                        lock[u] = k - bl + 1
                         relax_stack.extend((bl + 1, w) for w in B[u].difference(path))
             else:
                 for w in G[v]:
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -16,6 +16,7 @@ __all__ = [
     "recursive_simple_cycles",
     "find_cycle",
     "minimum_cycle_basis",
+    "chordless_cycles",
 ]
 
 
@@ -96,7 +97,7 @@ def cycle_basis(G, root=None):
     return cycles
 
 
-def simple_cycles(G, length_bound=None, chordless=False):
+def simple_cycles(G, length_bound=None):
     """Find simple cycles (elementary circuits) of a graph.
 
     A `simple cycle`, or `elementary circuit`, is a closed path where
@@ -119,6 +120,11 @@ def simple_cycles(G, length_bound=None, chordless=False):
     containing a particular edge, remove that edge, and further decompose the
     remainder into biconnected components.
 
+    Note that multigraphs are supported by this function -- and in undirected
+    multigraphs, a pair of parallel edges is considered a cycle of length 2.
+    Likewise, self-loops are considered to be cycles of length 1.  We define
+    cycles as sequences of nodes; so the presence of loops and parallel edges
+    does not change the number of simple cycles in a graph.
 
     Parameters
     ----------
@@ -178,9 +184,6 @@ def simple_cycles(G, length_bound=None, chordless=False):
     cycle_basis
     chordless_cycles
     """
-    if chordless:
-        yield from chordless_cycles(G, length_bound=length_bound)
-        return
 
     if length_bound is not None:
         if length_bound == 0:
@@ -469,7 +472,7 @@ def _bounded_cycle_search(G, path, length_bound):
                     B[w].add(v)
 
 
-def chordless_cycles(G, length_bound):
+def chordless_cycles(G, length_bound=None):
     """Find simple chordless cycles of a graph.
 
     A `simple cycle`, is a closed path where no node appears twice.  In a simple
@@ -544,6 +547,9 @@ def chordless_cycles(G, length_bound):
        E. Dias and D. Castonguay and H. Longo and W.A.R. Jradi
        https://arxiv.org/abs/1309.1051
 
+    See Also
+    --------
+    simple_cycles
     """
 
     if length_bound is not None:
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -95,21 +95,29 @@ def cycle_basis(G, root=None):
     return cycles
 
 
-@not_implemented_for("undirected")
-def simple_cycles(G):
-    """Find simple cycles (elementary circuits) of a directed graph.
+def simple_cycles(G, k=None):
+    """Find simple cycles (elementary circuits) of a graph.
 
     A `simple cycle`, or `elementary circuit`, is a closed path where
-    no node appears twice. Two elementary circuits are distinct if they
-    are not cyclic permutations of each other.
+    no node appears twice.  In a directed graph, two simple cycles are distinct
+    if they are not cyclic permutations of each other.  In an undirected graph,
+    two simple cycles are distinct if they are not cyclic permutations of each
+    other nor of the other's reversal.
 
-    This is a nonrecursive, iterator/generator version of Johnson's
-    algorithm [1]_.  There may be better algorithms for some cases [2]_ [3]_.
+    Optionally, the cycles are bounded in length by the parameter k.  In the
+    unbounded case, we use a nonrecursive, iterator/generator version of
+    Johnson's algorithm [1]_.  In the bounded case, we use a version of the
+    algorithm of Gupta and Suzumura [2]_. There may be better algorithms for
+    some cases [3]_ [4]_ [5]_.
 
     Parameters
     ----------
     G : NetworkX DiGraph
        A directed graph
+
+    k : int or None
+       If k is an int, generate all simple cycles of G with length at most k.
+       Otherwise, generate all simple cycles of G.
 
     Yields
     ------
@@ -122,11 +130,9 @@ def simple_cycles(G):
     >>> G = nx.DiGraph(edges)
     >>> sorted(nx.simple_cycles(G))
     [[0], [0, 1, 2], [0, 2], [1, 2], [2]]
-
     To filter the cycles so that they don't include certain nodes or edges,
     copy your graph and eliminate those nodes or edges before calling.
     For example, to exclude self-loops from the above example:
-
     >>> H = G.copy()
     >>> H.remove_edges_from(nx.selfloop_edges(G))
     >>> sorted(nx.simple_cycles(H))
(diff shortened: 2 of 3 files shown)


dschult:
I like the idea of having simple_cycles for undirected graphs. We've got `simple_cycles` for directed graphs. So this is much the same, though I haven't looked hard at that code with this in mind yet. We've had a number of people asking about searching through all cycles, so this would be a reasonable way to address that long term issue.

Of course, tests and docs and stuff are needed. And the `chordless` naming scheme gets switched to `allow_chords` but then used as `chords`, so some fixing is in order.   The paper you point to is 20 years newer than anything in the references for the directed case `simple_cycles`. And this improved version handles cycle reversals and unsortable nodes. So, I'm in favor of fleshing this out.

boothby:
>  The paper you point to is 20 years newer than anything in the references for the directed case simple_cycles. 

Right? I can't imagine that Johnson and contemporaries didn't know of this algorithm.  It is somewhat incredible if something so simple should have eluded their attention.  I haven't dug into these older references.

However, I found a more [recent paper](https://arxiv.org/abs/2105.10094) which does almost everything: fast, bounded enumeration of directed cycles.  Moreover, it's not hard to modify it (or, upon reflection, Johnson's algorithm) to handle undirected graphs: 

1. rather than recurse on prefixes `(s, ...)` we recurse on prefixes `(a, s, ...)`
2. to prevent the edge `(s, a)` from occurring in the search of (1) (as `(a, s, a)` is not a 3-cycle) we need to delete the edge `(a, s)` for the duration of the search of (1)
3. to prevent the reversal of the cycles produced in (1), we still want to delete the edge `(a, s)` for subsequent prefixes `(b, s, ...)` etc.

The simpler algorithm (and Johnson etc) operate by orderly enumerating all cycles containing a fixed node, and then deleting that node.  My modification here works instead by orderly enumerating all cycles containing a fixed edge, and then deleting that edge.

I've implemented the algorithm of Gupta and Suzumura with this modification in its recursive form; I'll patch this PR soonish with a non-recursive one.  What's being lost is my search for chordless cycles -- which is covered [here](https://arxiv.org/abs/1309.1051) and that paper points to an [asymptotically optimal (!) algorithm to enumerate all cycles ](https://arxiv.org/abs/1205.2766).  I'll leaving those alone for the time being.

dschult:
Too bad to lose the search for chordless cycles... but leaving that till later makes sense to me.

It seems that the undirected solution can be built from the fundamental cycle basis. But not much is talked about how to enumerate that. I found a link to an old paper that discussed how to enumerate the elementary cycles from the cycle basis. It's [by Gibbs in 1969](https://dl.acm.org/doi/10.1145/321541.321545) and it seems straight-forward. It'd be cool  to try putting together a version of that to compare with. I'll try to take a stab at that.

dschult:
Here is some code for tests. I had the Gibbs algorithm in mind with these tests, but they might be applicable to other algorithms too.
```python
    simp_cycles = undirected_simple_cycles_from_basis

    testG = nx.cycle_graph(8)
    cyc1 = tuple(range(8)) + (0,)
    assert list(simp_cycles(testG)) == [cyc1]
    testG.add_edge(4, -1)
    nx.add_path(testG, [3, -2, -3, -4])
    assert list(simp_cycles(testG)) == [cyc1]

    testG.update(nx.cycle_graph(range(8, 16)))
    cyc2 = tuple(range(8, 16)) + (8,)
    assert set(map(tuple, simp_cycles(testG))) == {cyc1, cyc2}

    # print("Starting New Test")
    testG.update(nx.cycle_graph(range(4, 12)))
    cyc3 = tuple(range(4, 12)) + (4,)
    expected = {
        tuple(cyc1),
        tuple(cyc2),
        tuple(cyc3),
        tuple(range(5)) + (11, 12, 13, 14, 15, 8, 7, 0),  # cyc1 + cyc2 + cyc3
        tuple(range(4, 9)) + (15, 14, 13, 12, 11, 4),  # cyc2 + cyc3
        tuple(range(5)) + (11, 10, 9, 8, 7, 0),  # cyc2 + cyc3
        # cyc1 + cyc2 is disjoint
    }
    {
        (0, 1, 2, 3, 4, 5, 6, 7, 0),  # cyc1
        (8, 9, 10, 11, 12, 13, 14, 15, 8),  # cyc2
        (4, 5, 6, 7, 8, 9, 10, 11, 4), # cyc3
        (4, 5, 6, 7, 8, 15, 14, 13, 12, 11, 4),  # cyc2 + cyc3
        (0, 1, 2, 3, 4, 11, 10, 9, 8, 7, 0),  # cyc1 + cyc3
        (0, 1, 2, 3, 4, 11, 12, 13, 14, 15, 8, 7, 0),  # cyc1 + cyc2 + cyc3
    }
    answer = set(map(tuple, simp_cycles(testG)))
    # print("Answer",answer)
    # print("Expect",expected)
    assert answer == expected
    assert len(answer) == (2**3 - 1) - 1  # 1 disjoint comb: cyc1 + cyc2

    # Basis size = 5 (2 loops overlapping gives 5 small loops
    #        E        
    #       / \         Note: A-F = 10-15                                        
    #    1-2-3-4-5                        
    #    / |   |  \   cyc1=012DAB -- left                                
    #   0  D   F  6   cyc2=234E   -- top                     
    #   \  |   |  /   cyc3=45678F -- right                                         
    #    B-A-9-8-7    cyc4=89AC   -- bottom                
    #       \ /       cyc5=234F89AD -- middle                                       
    #        C         
    #
    # combinations of 5 basis elements: 2^5 - 1  (one includes no cycles)
    # 
    # disjoint combs: (11 total) not simple cycles
    #   Any pair not including cyc5 => choose(4, 2) = 6
    #   Any triple not including cyc5 => choose(4, 3) = 4
    #   Any quad not including cyc5 => choose(4, 4) = 1
    #
    # we expect 31 - 11 = 20 simple cycles
    #
    testG = nx.cycle_graph(12)
    testG.update(nx.cycle_graph([12, 10, 13, 2, 14, 4, 15, 8]).edges)
    answer = set(simp_cycles(testG))
    assert len(answer) == (2**5 - 1) - 11  # 11 disjoint combinations
```

And the code I'm using. Two functions: 
- one to convert from sets of edge_ids to cycles written as tuples of nodes.
- the other to find the simple cycles as lists of edge_ids and use the 1st function to yields the cycles as tuples of nodes.

```python
import networkx as nx


def path_from_set_of_edges(edges):
    # construct adjacency of the cycle
    nbrs = defaultdict(list)
    for u, v in edges:
        nbrs[u].append(v)
        nbrs[v].append(u)
    # print("nbrs: ", nbrs)

    # Build edges starting at first edge
    start, tip = cycle = list(edges[0])
    while tip != start:
        # cycle guarantees 2 nbrs
        u, v = nbrs[tip]
        tip = u if v == cycle[-2] else v
        cycle.append(tip)
    return tuple(cycle)


def undirected_simple_cycles_from_basis(G):
    edgelist = list(G.edges)
    edge_num = {}
    for i, (u, v) in enumerate(edgelist):
        edge_num[u, v] = edge_num[v, u] = i

    # potentially long
    basis = list(nx.cycle_basis(G))
    #print("cycle_basis:", basis)

    # convert cycles to sets of edge numbers
    Basis = [{edge_num[e] for e in nx.utils.pairwise(C, cyclic=True)} for C in basis]

    # loop over Basis cycles adding to all previously found
    B = Basis.pop()
    yield path_from_set_of_edges([edgelist[i] for i in B])
    all_linear_combs = [B]
    while Basis:
        B = Basis.pop()
        R = []
        new_LC = [B]
        for T in all_linear_combs:
            comb = T ^ B
            new_LC.append(comb)
            # if comb is disjoint cycles T and B (T & B empty) cycle is not simple
            if T & B:
                # if comb contains an existing cycle it is not simple
                if any(comb >= V for V in R):
                    continue
                R = [V for V in R if not (comb <= V)]
                R.append(comb)
        yield path_from_set_of_edges([edgelist[i] for i in B])
        yield from (path_from_set_of_edges([edgelist[i] for i in C]) for C in R)
        all_linear_combs.extend(new_LC)
```
The large memory comes from saving all the linear combinations of basis cycles as sets of edge_ids.  It should be much thinner RAM-wise if we save which basis cycles participate in each cycle, but slower of course. With k basis cycles, we have 2**k -1 combinations to look at, so the enumeration of cycles is exponential in the number of basis cycles no matter what we do. I'm hoping that storing 2**k-1 indexes to the cycles rather than 2**k-1 sets of edge_ids will be thinner. But that is not this code. This code stores the sets of edge_ids for every linear combination of basis cycles.

boothby:
The tests I'm using are somewhat punishing:

```python
#OEIS A002807
print([
    sum(1 for c in LC_CYCLES(nx.complete_graph(n), n))
    for n in range(1, 10)
])

#OEIS A000292 and A050534
for k in (3, 4):
  print([
    sum(len(c) == k for c in LC_CYCLES(nx.complete_graph(n), k))
    for n in range(1, 20)
])
```

Computing A002807 on `range(1, 10)` takes my implementation of Gupta & Suzumura about 0.25 seconds on my laptop.  I loaded up your Gibbs implementation and that takes 105 seconds to compute `range(3, 9)` -- I didn't have the patience for 10.  The cycle basis approach has some aesthetic appeal, but I remain unconvinced that it is worth the trouble except perhaps extremely sparse graphs.



Anyway, here's my rough implementaion of G&S if you want to play with it:

```python

def RELAX_LOCKS(u, k, blen, state):
    Lock, Blist, stack = state.get()

    if Lock[u] < k - blen + 1:
        Lock[u] = k - blen + 1
        for w in Blist[u]:
            if w not in stack:
                RELAX_LOCKS(w, k, blen+1)
  

def cycle_search(G, t, v, k, flen, state):
    Lock, Blist, stack = state.get()
    inf = k + 1

    blen = inf
    Lock[v] = flen
    stack.append(v)
    for w in G[v]:
        if w == t:
            yield stack
            blen = 1
        elif flen + 1 < Lock[w] and flen + 1 < k:
            yield from cycle_search(G, t, w, k, flen+1, state)
            blen = min(blen, 1+state.blen)
    if blen < inf:
        RELAX_LOCKS(v, k, blen, state)
    else:
        for w in G[v]:
           Blist[w].add(v) 
    stack.pop()
    state.blen = blen

def bfs_to_depth(G, v, depth):
    last_level = visited = {v}
    while last_level and depth > 0:
        depth -= 1
        leaves = {w for u in last_level for w in G[u]} - visited
        visited.update(leaves)
        last_level = leaves
    return visited

def LC_CYCLES(G, k):
    G = G.copy()
    inf = k + 1
    class state:
        Lock = defaultdict(lambda: inf)
        Blist = defaultdict(set)
        stack = []
        @classmethod
        def clear(cls):
            cls.Lock.clear()
            cls.Blist.clear()
            cls.stack.clear()
        @classmethod
        def get(cls):
            return cls.Lock, cls.Blist, cls.stack
    Lock, Blist, stack = state.get()
    undirected = not G.is_directed()
    for s in list(G):
        Gs = G.subgraph(bfs_to_depth(G, s, k))
        if undirected:
            Gs = Gs.copy()
            observed = set()
            for a in tuple(G[s]):
                observed.add(a)
                state.clear()
                stack[:] = a,
                Lock[a] = Lock[s] = 0
                Gs.remove_edge(a, s)
                yield from cycle_search(Gs, a, s, k, 0, state)
        else:
            state.clear()
            yield from cycle_search(Gs, s, s, k, 0, state)
        G.remove_node(s)
```

Much to be desired with that code; I'll get it into shape this week.

boothby:
Here's an improved version:
```python

import networkx as nx
from networkx import connectivity
bridge_components = connectivity.edge_kcomponents.bridge_components

def find_cycles(G, k=None):
    if k is None:
        k = len(G)

    if G.is_directed():
        components = nx.strongly_connected_components(G)
        search = _directed_cycle_search
    else:
        components = bridge_components(G)
        search = _undirected_cycle_search

    for c in components:
        yield from search(G.subgraph(c).copy(), k)

def _directed_cycle_search(G, k):
    for v in list(G):
        yield from _cycle_search(G, [v], k)
        G.remove_node(v)

def _undirected_cycle_search(G, k):
    for u, v in list(G.edges):
        G.remove_edge(u, v)
        yield from _cycle_search(G, [u, v], k)

def _cycle_search(G, path, k):
    inf = k+2
    Lock = defaultdict(lambda: inf)
    Blist = defaultdict(set)
    start = path[0]
    for v in path:
        Lock[v] = 0
    pathset = set(path)
    stack = [[inf, len(path)-1, iter(G[path[-1]])]]
    while stack:
        blen, flen, nbrs = top = stack[-1]
        for w in nbrs:
            if w == start:
                yield path
                blen = top[0] = 1
            elif flen + 1 < Lock[w] and flen + 1 < k:
                path.append(w)
                pathset.add(w)
                Lock[w] = flen+1
                stack.append([blen, flen+1, iter(G[w])])
                break
        else:
            v = path[-1]
            if blen < inf:
                relax_stack = [(blen, v)]
                while relax_stack:
                    bl, u = relax_stack.pop()
                    if Lock[u] >= k - bl + 1:
                        continue
                    Lock[u] = k - bl + 1
                    relax_stack.extend((bl, w) for w in Blist[u] if w not in pathset)
            else:
                for w in G[v]:
                    Blist[w].add(v)
            stack.pop()
            pathset.remove(path.pop())
            if stack:
                top = stack[-1]
                top[0] = min(top[0], blen)
```

boothby:
...huh.  It occurs to me that my function `_cycle_search` above is fairly well-described by the name `all_simple_paths`.  I'll have to see how this compares.  I am now somewhat more motivated to implement [Ferriera et al](https://arxiv.org/abs/1205.2766).

The paper bears a nice nugget of wisdom in Lemma 3: don't bother deleting *all* of the edges of the graph, just focus on the nontree edges of a DFS traversal.

```python
def _undirected_cycle_search(G, k):
    for u, v, kind in list(nx.dfs_labeled_edges(G)):
        if kind == 'nontree' and G.has_edge(u, v):
            G.remove_edge(u, v)
            #delete (u, v) before searching G, to avoid fake 3-cycles [u, v, u]
            yield from _cycle_search(G, [u, v], k)
```

The above works very nicely and gives a decent speedup even for complete graphs.

The paper goes on: after you delete an edge, repeat the observation that edges only occur in biconnected components, continue carving up the graph:

```python
def _undirected_cycle_search(G, k):
    components = [list(G)]
    while components:
        c = components.pop()
        Gc = G.subgraph(c)
        for u, v, kind in nx.dfs_labeled_edges(Gc):
            if kind == 'nontree':
                G.remove_edge(u, v)
                #delete (u, v) before searching G, to 
                yield from _cycle_search(Gc, [u, v], k)
                components.extend(biconnected_components(Gc))
                break
```
The performance here is unimpressive -- it tanks dense graphs, so I tested on 3-regular circulant graphs -- there, I didn't see the performance cross over until around 300 nodes.

boothby:
TODO: tests with k not None

boothby:
@dschult ready for review

boothby:
Actually, belay that.  I played with the [Dias et al](https://arxiv.org/abs/1309.1051) chordless cycles algorithm last night and it will be trivial to add it.  While that algorithm doesn't support bounded cycle length, their block/unblock strategy is compatible with bounding 'recursion' depth.  That just means a third dispatch endpoint -- `_johnson_cycle_search`, `_bounded_cycle_search`, and now `_chordless_cycle_search`.

dschult (commented):
This looks great. I especially like the use of dict and subclasses of dict to avoid the slower parts of the NetworkX Graph and View objects (and even utilities like `pairwise` if I understand the `undirected_cycle_set()` function in the tests correctly). These examples could be used as the basis of a tutorial for how to speed up algorithms while retaining python-only functionality. Then we could use that tutorial to prep ourselves for going into other algorithms with an eye toward speed-up.   Thanks!
 
Comments:
- this doesn't consider multiedges. Should it? Either way, we should mention multigraphs in the doc_string somewhere, if only to say that the user will need to figure out how many copies of a cycle should appear when there are multiedges in the cycle.  Maybe we should even describe how to get all cycles with multigraphs -- but that is perhaps another PR.
The previous function ignored multiedges too so this isn't anything new.

Looking forward to the chordless version. :}

boothby:
Thanks for your comments!  I am quite pleased with how this has turned out so far.  However, I hit a stumbling block on chordless cycle enumeration.  In the undirected case, after I tear off an edge, re-compute components and keep going  -- this works very nicely for bounded and unbounded enumeration, but causes an issue with chordless cycles:

```
a - b
| / |
c - d
```

If I explore `(b, c)` first, then I find the chordless cycles `(a, b, c)` and `(b, c, d)`. I then remove the edge `(b, c)`, and... whoops!  `(a, b, d, c)` is now a chordless cycle!  So that's bad.  I tried to incorporate a better strategy into `_undirected_cycle_search` and it does not spark joy.

```def _undirected_cycle_search(G, k, chordless):
    bcc = nx.biconnected_components
    components = [c for c in bcc(G) if len(c) >= 3]
    while components:
        c = components.pop()
        if chordless:
            v = c.pop()
            Gc = G.subgraph(c)
            for u, w in combinations(c.intersection(G[v]), 2):
                if G.has_edge(u, w):
                    yield [u, v, w]
                else:
                    yield from _chordless_cycle_search(Gc, [u, v, w], k)
        else:
            Gc = G.subgraph(c)
            uv = list(next(iter(Gc.edges)))
            G.remove_edge(*uv)
            # delete (u, v) before searching G, to avoid fake 3-cycles [u, v, u]
            if k is None:
                yield from _johnson_cycle_search(Gc, uv)
            else:
                yield from _bounded_cycle_search(Gc, uv, k)
        components.extend(c for c in bcc(Gc) if len(c) >= 3)
```

So, I'm leaning towards not including this condition in the same function.  My preference is to provide a `chordless_cycles` function with the same signature as `simple_cycles`.

With regards to multiple edges, I claim that the code is correct, except in the case of digons (which I should account for).  If the function was returning lists of edges, things would get gross and each node-cycle would occur exponentially many times.  But since I'm returning lists of nodes, and we define simple circuits in terms of their nodes, I claim that my handling of multiedges (less digons) is merely un-documented.

boothby:
Oh, and I found yet another reference, for enumeration of chordless cycles in digraphs.  https://rbisdorff.github.io/documents/chordlessCircuits.pdf

I've been disappointed with many of the references I've found, there are _so many_ typos in their pseudocode; arguments are arbitrarily dropped from function calls, etc.  I've faithfully implemented Bisdorff, but I have not yet chased down all bugs.  It's different enough from all the others that I'd really like to understand it.  I'm also motivated by a modification I needed to make to Dias et al in order for it to correctly handle digraphs (in short: follow forward edges in path-lengthening, use an undirected view of the graph in node-blocking).  It appears that undirected views are painfully slow, even with caching.

My impulse is to calve chordless cycles off to a separate issue, and focus on polishing this PR.  I don't feel any particular rush here, but I'm also thinking about enumeration of bounded-length and chordless paths, which definitely feels out of scope here (despite a large amount of algorithmic overlap).

And, because no discussion is allowed to go on so long without a naming discussion... I really think "induced" is a better term than "chordless".  Feel free to not take that bait.

boothby:
Bisdorff's algorithm, in the end, is nothing more than 

1. pick a vertex `v`
2. exhaustively explore chordless paths rooted at `v`
3. delete `v`; go to 1 unless `G` is empty

dschult:
I think to avoid the slow views you would want to keep one graph for path lengthening and another graph for blocking. Subgraph views are rarely performant unless you only look up each edge once. They save memory, not time. Would keeping two versions of the graph also allow e.g. removing edge `(b, c)` in your example above by removing it from one graph but testing whether the result is chordless by using a version of the graph without the edges removed?  

Maybe we make a copy of the graph to hack at while finding cycles and use the original graph to check for blocking and/or chords? Does that even make sense? Not sure if it si possible.

I like the term `chordless` rather than `induced` because it actually describes what it means (once you learn what "chord" means). The word "induced" is used often for subgraphs, but it confuses many people. I personally think that is because the word doesn't actually mean anything specific on its own. The word "induced" is not a natural word to describe: "all the edges in the graph between these nodes have to occur in the subgraph". Also, you have to think of a cycle as a subgraph of the graph (which it is -- but often isn't thought of that way) and also understand/recall what an induced subgraph is. That said, anyone comfortable with the term "induced subgraph" will be able to work out that a cycle being an induced subgraph means that the only edges between the nodes of the cycle are part of the cycle. So, while my preference is for "chordless", I could be convinced or overruled without too much trouble.

dschult:
I like the approach of returning nodes to represent cycles and leaving multiedges out of the process of finding the cycle-nodes.  Finding all cycles in simple graphs is hard enough. The multiedges "just" duplicate some cycles many times. Hard to see that as a useful feature -- and producing such a list is certainly possible from the cycle's list of nodes.

boothby:
A reconsideration.  Mutigraphs are weird.  Is a pair of parallel edges a chordless cycle?  First impulse: yeah, sure, as long as neither one's got a loop (hack hack hack, we have an implementation).  What about three parallel edges?  Well, I haven't found any literature wherein this is a worthy consideration, but if I push on the definitions, that's a 2-edge-connected graph with zero chordless cycles.  My current implementation does that wrong.  A fix for this is easy, but is there any point?

boothby:
> (I mean the path 1-2-3 has cycle 1-2-3-2-1 if you allow digon cycles.)

That is a cycle, according to some, but it isn't a simple cycle.

Every time I step away from this, I've completed something of additional complexity, and pondered the option of removing a feature / support.  Perhaps if I was serious about that, I would step away *before* getting the implementation done.  Because I am now thinking "wouldn't it be a shame to not support multigraphs, they're just one condition away..."

boothby:
Alright, it's up and running!  Digons and parallel edges turn out to be super detailed and tricky, especially in directed graphs!  So I did the case analysis, added some tests, buttressed the tests with a super-paranoid-cycle-enumeration-algorithm-testing-framework, did a little optimization and simplification and patched an unmentionable number of emergent bugs, and here we are.  I still need some documentation.

Question: since `chordless_cycles` ends up being so very different from `simple_cycles`, should I expose the name to `__all__` and omit the `chordless` option from `simple_cycles`?  I lean towards yes.

dschult:
Yes -- I agree that a new function `chordless_cycles` would be better than having it be an option on `simple_cycles`.

Nice work! I'll do a review of it -- and hold off on the docs till that goes in.

boothby:
Whoops, I had already started on docs... now stepping away.

boothby:
I think I misread your statement of "hold off on docs."  I removed the `chordless` option and this is ready for review.

dschult:
I actually meant that I would hold off on the docs until you were done.  
This works well!   Thanks for getting this ready for review! :)

dschult (approved):
I actually made a Pr to your branch, but now that I look at it, these are very minor suggestions and in the interest in getting this merged for NX3.1 (release candidate hopefully today).  If for some reason you want to undo these minor changes we'll do that separately. Thanks for This!

boothby:
Looks good to me, thanks!

PR #6466, opened 2023-03-05
Ignore weakrefs when testing for memory leak

eriknw:
I started to encounter an error with this when testing `graphblas_algorithms`, b/c it was trying to operate on weakly-reference objects that no longer exist. I have no other context about this, but this change seems "safe" to me.

### DIFF
--- a/networkx/classes/tests/test_graph.py
+++ b/networkx/classes/tests/test_graph.py
@@ -72,6 +72,14 @@ class BaseGraphTester:
         G = self.Graph()
 
         def count_objects_of_type(_type):
+            # Iterating over all objects tracked by gc can include weak references
+            # whose weakly-referenced objects may no longer exist. Calling `isinstance`
+            # on such a weak reference will raise ReferenceError. There are two common
+            # workarounds for this: one is to compare type names instead of `isinstance`
+            # such as `type(obj).__name__ == typename`, and the other is to explicitly
+            # ignore ProxyTypes as we do below.
+            # NOTE: even if this safeguard is deemed unnecessary to pass NetworkX tests,
+            # we should still keep it for maximum safety for other NetworkX backends.
             return sum(
                 1
                 for obj in gc.get_objects()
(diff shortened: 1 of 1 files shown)


eriknw:
Note that we don't use `weakref`'s in `graphblas_algorithms`, so I have no idea why this started happening. Maybe from a dependency?

eriknw:
I did the next best thing: I added an informative code comment.

We need this change for `graphblas_algorithms`, but I'm not inclined to investigate further. I doubt you want to install `graphblas_algorithms` just to test this functionality.

eriknw:
Oh, I'm with you on this and have similar feelings.

I played around a little bit:
```python
Python 3.10.9 | packaged by conda-forge | (main, Feb  2 2023, 20:20:04) [GCC 11.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.11.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import weakref

In [2]: class A:
   ...:     pass
   ...:

In [3]: a = A()

In [4]: wa = weakref.proxy(a)

In [5]: type(wa) == A
Out[5]: False

In [6]: type(a) == A
Out[6]: True

In [7]: isinstance(wa, A)
Out[7]: True

In [8]: isinstance(a, A)
Out[8]: True

In [9]: isinstance(wa, weakref.ProxyTypes)
Out[9]: True

In [10]: isinstance(a, weakref.ProxyTypes)
Out[10]: False

In [11]: del a

In [12]: type(wa) == A
Out[12]: False

In [13]: isinstance(wa, A)
ReferenceError                            Traceback (most recent call last)
Cell In[13], line 1
----> 1 isinstance(wa, A)

ReferenceError: weakly-referenced object no longer exists

In [14]: type(wa) == A
Out[14]: False
```
So, it looks like your alternative implementation works! I don't know which approach to prefer.

I suppose the CPython source code can be a place to learn more:
https://github.com/python/cpython/blob/main/Objects/weakrefobject.c
It would appear that calling `type(x)` on a proxy type does not call `proxy_checkref`.

rossbar (approved):
Thanks for raising, investigating, and documenting this in the code comment @eriknw . I don't have a ton of experience with `weakref` but this LGTM and the proposed solution is the one I prefer out of the listed options.

PR #6467, opened 2023-03-05
Dispatch more BFS-based algorithms

eriknw:
I've already implemented the majority of these in `graphblas_algorithms` and hope to implement the rest soon.


dschult (approved):
This looks good to me. I don't know of any downside to adding the decorator.

rossbar (approved):
Exciting! Thanks @eriknw !

PR #6471, opened 2023-03-07
tweak _dispatch to allow G keyword. Add test.

### LINKED ISSUE #6458: upgrade 2.8.6  -> 3.0 breaks shortest_path()
I just re-installed networkx without specifying version so I was upgraded from 2.8.6 to 3.0
Running unaltered code calling nx.shortest_path(*args, **kwargs) under 2.8.6 runs fine and produces correct result but 3.0 fails with an error, missing arguments.

Looking at the source, the 3.0 version got a decorator which doesn't seem to work properly.

The error is reproduced using stripped-down [lollipop example](https://networkx.org/documentation/stable/auto_examples/basic/plot_properties.html#sphx-glr-auto-examples-basic-plot-properties-py)

### Current Behavior

under 3.0:
```
import networkx as nx
G = nx.lollipop_graph(4, 6)
print(nx.shortest_path(G=G, source=list(G.nodes())[1], target=list(G.nodes())[9]))
```
produces

```
  File "C:\Users\XYZ\AppData\Roaming\JetBrains\PyCharm2022.3\scratches\scratch_1.py", line 7, in <module>
    print(nx.shortest_path(G=G, source=list(G.nodes())[1], target=list(G.nodes())[9]))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\modules\venv\networkx_3\Lib\site-packages\networkx\classes\backends.py", line 134, in wrapper
    graph = args[0]
            ~~~~^^^
IndexError: tuple index out of range
```

### Expected Behavior

under 2.8.6:
```
import networkx as nx
G = nx.lollipop_graph(4, 6)
print(nx.shortest_path(G=G, source=list(G.nodes())[1], target=list(G.nodes())[9]))
```
produces
`[1, 3, 4, 5, 6, 7, 8, 9]`

### Environment
Python version: anaconda 3.10, networkx installed via pip
NetworkX version: 2.8.6 and 3.0

dschult:
The `_dispatch` decorator assumes that the first argument is a positional argument.
This PR allows `G` to be provided as a keyword. Note that `G` is still expected to be
the first positional argument. This only affects how people can call it: 
`nx.pagerank(G=G)` and `nx.pagerank(G)` should be the same.

Fixes #6458

### DIFF
--- a/networkx/classes/backends.py
+++ b/networkx/classes/backends.py
@@ -128,7 +128,10 @@ def _dispatch(func=None, *, name=None):
 
     @functools.wraps(func)
     def wrapper(*args, **kwds):
-        graph = args[0]
+        if args:
+            graph = args[0]
+        else:
+            graph = kwds["G"]  # raise KeyError if not present
         if hasattr(graph, "__networkx_plugin__") and plugins:
             plugin_name = graph.__networkx_plugin__
             if plugin_name in plugins:
--- a/networkx/classes/backends.py
+++ b/networkx/classes/backends.py
@@ -131,7 +131,10 @@ def _dispatch(func=None, *, name=None):
         if args:
             graph = args[0]
         else:
-            graph = kwds["G"]  # raise KeyError if not present
+            try:
+                graph = kwds["G"]
+            except KeyError:
+                raise TypeError(f"{name}() missing positional argument: 'G'")
         if hasattr(graph, "__networkx_plugin__") and plugins:
             plugin_name = graph.__networkx_plugin__
             if plugin_name in plugins:
--- a/networkx/classes/backends.py
+++ b/networkx/classes/backends.py
@@ -134,7 +134,7 @@ def _dispatch(func=None, *, name=None):
             try:
                 graph = kwds["G"]
             except KeyError:
-                raise TypeError(f"{name}() missing positional argument: 'G'")
+                raise TypeError(f"{name}() missing positional argument: 'G'") from None
         if hasattr(graph, "__networkx_plugin__") and plugins:
             plugin_name = graph.__networkx_plugin__
             if plugin_name in plugins:
--- a/networkx/classes/tests/test_backends.py
+++ b/networkx/classes/tests/test_backends.py
@@ -1,0 +1,11 @@
+import pytest
+
+import networkx as nx
+
+
+def test_dispatch_kwds_vs_args():
+    G = nx.path_graph(4)
+    nx.pagerank(G)
+    nx.pagerank(G=G)
+    with pytest.raises(KeyError):
+        nx.pagerank()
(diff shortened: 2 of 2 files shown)


networkx/classes/backends.py
```diff
@@ -128,7 +128,10 @@ def _dispatch(func=None, *, name=None):
 
     @functools.wraps(func)
     def wrapper(*args, **kwds):
-        graph = args[0]
+        if args:
+            graph = args[0]
+        else:
+            graph = kwds["G"]  # raise KeyError if not present
```
eriknw, line 134: I think it would be better to raise `TypeError` with a message like `f"{name}() missing required positional argument: 'G'"` to better match typical errors and messages when incorrectly calling a function.

networkx/classes/backends.py
```diff
@@ -131,7 +131,10 @@ def wrapper(*args, **kwds):
         if args:
             graph = args[0]
         else:
-            graph = kwds["G"]  # raise KeyError if not present
+            try:
+                graph = kwds["G"]
+            except KeyError:
+                raise TypeError(f"{name}() missing positional argument: 'G'")
```
eriknw, line 137: Great! I would also add ` from None` so you don't get chained error messages (i.e., `"During handling of the above exception, another exception occurred:"`):
```suggestion
                raise TypeError(f"{name}() missing positional argument: 'G'") from None
```

MridulS: Hmm, don't we have some places which expects graphs not as `G`, like `graph`?

eriknw: Doesn't look like many, but there are a few in this file such as `to_directed(graph)`.
https://github.com/networkx/networkx/blob/main/networkx/classes/function.py#L512

A couple functions also take multiple graphs, so we'll probably need to update the `_dispatch` decorator to handle these special cases.

Currently, all wrapped functions use `G` for the graph as the first argument. Verified via:
```python
import inspect
import networkx as nx

def _get_nx_func(dispatched_func):
    # There ought to be an easier way to get the original function...
    index = dispatched_func.__code__.co_freevars.index("func")
    cell = dispatched_func.__closure__[index]
    return cell.cell_contents

for name, dispatched_func in nx.backends._registered_algorithms.items():
    func = _get_nx_func(dispatched_func)
    sig = inspect.signature(func)
    first_arg_name = next(iter(sig.parameters))
    assert first_arg_name == "G"
```
So, I think the current PR works for the upcoming NetworkX 3.1 release.

MridulS (approved):

rossbar (approved):
LGTM, thanks @dschult 

I wonder if we can't find a more elegant solution, but this fix makes sense for use-case that is most common to be run into.

Issue #6465, opened 2023-03-05
`negative_edge_cycle` unexpectedly raises for empty graph input


eriknw:
### Current Behavior

```python-traceback
>>> import networkx as nx
>>> G = nx.Graph()
>>> nx.negative_edge_cycle(G)
Traceback (most recent call last):
  File "/home/erik/graphblas_workspace/networkx/networkx/algorithms/shortest_paths/weighted.py", line 2138, in negative_edge_cycle
    bellman_ford_predecessor_and_distance(
  File "/home/erik/graphblas_workspace/networkx/networkx/classes/backends.py", line 143, in wrapper
    return func(*args, **kwds)
  File "/home/erik/graphblas_workspace/networkx/networkx/algorithms/shortest_paths/weighted.py", line 1221, in bellman_ford_predecessor_and_distance
    raise nx.NodeNotFound(f"Node {source} is not found in the graph")
networkx.exception.NodeNotFound: Node -1 is not found in the graph

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/erik/graphblas_workspace/networkx/networkx/classes/graph.py", line 672, in remove_node
    nbrs = list(adj[n])  # list handles self-loops (allows mutation)
KeyError: -1

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/erik/graphblas_workspace/networkx/networkx/classes/backends.py", line 143, in wrapper
    return func(*args, **kwds)
  File "/home/erik/graphblas_workspace/networkx/networkx/algorithms/shortest_paths/weighted.py", line 2144, in negative_edge_cycle
    G.remove_node(newnode)
  File "/home/erik/graphblas_workspace/networkx/networkx/classes/graph.py", line 675, in remove_node
    raise NetworkXError(f"The node {n} is not in the graph.") from err
networkx.exception.NetworkXError: The node -1 is not in the graph.
```

### Expected Behavior
```python
>>> import networkx as nx
>>> G = nx.Graph()
>>> nx.negative_edge_cycle(G)
False
```

### Steps to Reproduce

See above.

### Environment

Python version: 3.10.9
NetworkX version: main branch
```
commit 4a6f2f43508d26d0eb9884a24cba28721d5fb875 (HEAD -> main, origin/main, origin/HEAD)
Author: Ross Barnowski <rossbar@berkeley.edu>
Date:   Thu Mar 2 06:40:41 2023 -0800
```
[...]

PR #6473, opened 2023-03-08
Fix negative edge cycle function raising exception for empty graph and added test function

PurviChaurasia:
In reference to #6465, fixed the `negetive_edge_cycle` function by handling an empty graph. Also added `test_negative_cycle_empty` test function in `test_weighted.py`.

### DIFF
--- a/networkx/algorithms/flow/tests/test_mincost.py
+++ b/networkx/algorithms/flow/tests/test_mincost.py
@@ -438,7 +438,7 @@ class TestMinCostFlow:
         pytest.raises(nx.NetworkXNotImplemented, nx.capacity_scaling, G)
         G = nx.DiGraph()
         pytest.raises(nx.NetworkXError, nx.network_simplex, G)
-        pytest.raises(nx.NetworkXError, nx.capacity_scaling, G)
+        # pytest.raises(nx.NetworkXError, nx.capacity_scaling, G)
         G.add_node(0, demand=float("inf"))
         pytest.raises(nx.NetworkXError, nx.network_simplex, G)
         pytest.raises(nx.NetworkXUnfeasible, nx.capacity_scaling, G)
--- a/networkx/algorithms/shortest_paths/tests/test_weighted.py
+++ b/networkx/algorithms/shortest_paths/tests/test_weighted.py
@@ -342,6 +342,10 @@ class TestWeightedPath(WeightedTestBase):
         G.add_edge(9, 10)
         pytest.raises(ValueError, nx.bidirectional_dijkstra, G, 8, 10)
 
+    def test_negative_edge_cycle_empty(self):
+        G = nx.DiGraph()
+        assert not nx.negative_edge_cycle(G)
+
     def test_negative_edge_cycle_custom_weight_key(self):
         d = nx.DiGraph()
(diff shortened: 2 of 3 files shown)


PurviChaurasia:
Could someone please point out where I'm going wrong with this?

dschult:
You can click on "Show all checks" and then the "Details" link to the right of any check that has a red X indicating that it didn't pass the tests. The "style/format" test usually means that you didn't run the "precommit" suite of style checking programs on your commit.  That can be automated by installing precommit (see the contributer guide at the end of step 1 for how to do that). It looks like in this case, the style checker is complaining about extra parentheses. 

But it also looks like you may have other errors showing up and I'm not sure why...
You can run the test locally using `pytest --doctest-modules weighted.py tests/test_weighted.py` to check for those test errors.  Then commit any changes and push to your branch and it should update the PR and rerun the tests.

PurviChaurasia:
Okay, I'll look into this.

PurviChaurasia:
Hi, so I ran the test suite and here is where it's failing - 
FAILED networkx/algorithms/flow/tests/test_mincost.py::TestMinCostFlow::test_exceptions - Failed: DID NOT RAISE <class 'networkx.exception.NetworkXError'>

PurviChaurasia:
@dschult Any Idea as to why this exception might be raised?

PurviChaurasia:
@MridulS It would be a great, if you could help in figuring out where I'm going wrong and why I'm getting this error.

dschult:
First, can you fix the precommit errors?

Second, other functions use `negative_edge_cycle` and apparently one of them (it looks like `nx.capacity_scaling` relies on the old error behavior -- rather than returning `False`.  So, this change is more delicate than expected.  You will need to find those parts of the code, understand why they used to give a NetworkXError, and make sure that they now give a reasable other type of error.  If that is the case, then you just have to change the error type that **those tests of the other function** look for.  For example if it now raises a `NetworkXUnbounded` error instead of a `NetworkXError` then you will need to change the test that is currently failing to make it look for a `NetworkXUnbounded` error.  Hope this makes sense.

PurviChaurasia:
Hey! Yes it's passing all the test cases now, thanks a lot!

PurviChaurasia:
Woohoo! Finally figured it out. Thanks a lot @dschult and @MridulS :)
Kindly merge the PR!

dschult (approved):
This looks good to me.
Note that the test in `test_mincost` is only commented out because the other tests in that test function are commented out whenever `nx.capacity_scaling` doesn't raise an exception. I'm not sure of the history there, but by commenting it out, we are following whatever reasoning we had when building that test suite. It might be good to go through and delete the commented out tests, but I'm not sure and that's really a separate issue from this PR.

I approve this PR

PurviChaurasia:
Yes, I had a similar question. I noticed that other tests for `nx.capacity_scaling` are also commented out. @dschult Should I raise a separate issue for this? Perhaps I can look into why it is happening and delete/comment out other tests which may be redundant in nature.

rossbar (approved):
I don't understand the test commenting either, but I agree with @dschult that we can leave that for a separate investigation.

LGTM thanks @PurviChaurasia !

PR #6484, opened 2023-03-10
Removing Alpha value Specification

Qudirah:
fixes for #6443 

@dschult and @rossbar please take a look at the PR. Have I done the right thing?


dschult:
The alpha value should be used when it is present, but the "get" function returns `None` by default when the key is not in the dict. So, we need a second argument to the `get` function which provides a default of value 1. See dict.get docs.

PR #6486, opened 2023-03-10
Improve test coverage for mst.py and bug fix in prim_mst_edges()

### LINKED ISSUE #6477: Improve test coverage for MST algorithms
I found that the test coverage for the mst.py file in algorithms/tree is 92.49%

https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/tree/mst.py

I am working on improving this.

Current Behavior

Expected Behavior

navyagarwal:
Closes #6477 

I have added test cases for Kruskal and Prims algorithms. The test coverage for mst.py has now increased to 98%.

image

This is my first time writing tests, so I apologize in advance if there are any errors in the code. I am specifically doubtful of the Prims test when ignore_nan = False and an Error is raised.

Edit - This PR also fixes some bugs based on [this discussion](https://github.com/networkx/networkx/pull/6486#discussion_r1136069166). The bug fix introduces new branches to the code, so the test coverage is now 96%.

### DIFF
--- a/networkx/algorithms/tree/tests/test_mst.py
+++ b/networkx/algorithms/tree/tests/test_mst.py
@@ -307,9 +307,8 @@ class TestPrim(MultigraphMSTTestBase):
         )
 
         # NaN weight edges raise Error when ignore_nan=False
-        edges = nx.minimum_spanning_edges(H, algorithm=self.algo, ignore_nan=False)
         with pytest.raises(ValueError):
-            list(edges)
+            list(nx.minimum_spanning_edges(H, algorithm=self.algo, ignore_nan=False))
 
     def test_multigraph_keys_tree(self):
         G = nx.MultiGraph()
--- a/networkx/algorithms/tree/tests/test_mst.py
+++ b/networkx/algorithms/tree/tests/test_mst.py
@@ -1,6 +1,5 @@
 """Unit tests for the :mod:`networkx.algorithms.tree.mst` module."""
 
-import numpy as np
 import pytest
 
 import networkx as nx
@@ -296,7 +295,7 @@ class TestPrim(MultigraphMSTTestBase):
         """Tests that the edges with NaN weights are ignored or
         raise an Error based on ignore_nan is true or false"""
         H = nx.MultiGraph()
-        H.add_edge(1, 2, key=1, weight=np.NaN)
+        H.add_edge(1, 2, key=1, weight=float("nan"))
         H.add_edge(1, 2, key=2, weight=3)
         H.add_edge(3, 2, key=1, weight=2)
         H.add_edge(3, 1, key=1, weight=4)
--- a/networkx/algorithms/tree/tests/test_mst.py
+++ b/networkx/algorithms/tree/tests/test_mst.py
@@ -1,5 +1,6 @@
 """Unit tests for the :mod:`networkx.algorithms.tree.mst` module."""
 
+import numpy as np
 import pytest
 
 import networkx as nx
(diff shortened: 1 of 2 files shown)


navyagarwal:
I am importing numpy to mst.py to set the value of an edge as NaN which is causing some of the checks to fail.

image

I think I have to mention the imports in some other file as well but I am not able to figure out where. Could someone please help out?

navyagarwal:
@MridulS Could you please help me with this?

BenFaruna:
> I am importing numpy to mst.py to set the weight of an edge as NaN which is causing some of the checks to fail.
> 
> image
> 
> I think I have to mention the imports in some other file as well but I am not able to figure out where. Could someone please help out?

Concerning this, I don't know if this runs on your machine but it seems the environment you're working with does not have numpy installed. So maybe try running python in your virtual environment and importing numpy to see if it works

navyagarwal:
> Concerning this, I don't know if this runs on your machine but it seems the environment you're working with does not have numpy installed. So maybe try running python in your virtual environment and importing numpy to see if it works

It does run on my system and passes all tests, and numpy has been imported as well.

networkx/algorithms/tree/tests/test_mst.py
```diff
@@ -253,6 +254,36 @@ class TestKruskal(MultigraphMSTTestBase):
 
     algorithm = "kruskal"
 
+    def test_key_data_bool(self):
+        """Tests that the keys and data values are included in
+        MST edges based on whether keys and data parameters are
+        true or false"""
+        G = nx.MultiGraph()
```
MridulS, line 261:
```suggestion
        pytest.importorskip("numpy")
        G = nx.MultiGraph()
```

networkx/algorithms/tree/tests/test_mst.py
```diff
@@ -1,5 +1,6 @@
 """Unit tests for the :mod:`networkx.algorithms.tree.mst` module."""
 
+import numpy as np
```
MridulS, line 3:
```suggestion
```

networkx/algorithms/tree/tests/test_mst.py
```diff
@@ -261,6 +292,26 @@ class TestPrim(MultigraphMSTTestBase):
 
     algorithm = "prim"
 
+    def test_ignore_nan(self):
+        """Tests that the edges with NaN weights are ignored or
+        raise an Error based on ignore_nan is true or false"""
+        H = nx.MultiGraph()
```
MridulS, line 298:
```suggestion
        pytest.importorskip("numpy")
        H = nx.MultiGraph()
```

MridulS (commented):
Thanks for the PR!

To import the packages inside the test env you need to import them using pytest.importskip

networkx/algorithms/tree/tests/test_mst.py
navyagarwal: When I do this, tests are failing on my local system because NaN is not recognized. I have tried two ways -

image

image

navyagarwal: @MridulS I realized that I actually didn't have to use Numpy for NaN edge weight, I used `weight=float("nan")` instead and the test cases passed alright on my local system. However, when I push the changes, the checks show that ValueError is never raised. But when I run the example, ValueError is getting raised.

image

I cannot figure out what the problem is here.

dschult: The difference is that the test is being run using Pypy instead of CPython. The test suite has many different configurations and running it on this alternative version of python is one of them. Perhaps they handle `math.isnan` differently. But I haven't looked.

Also, could you cut/paste text into github rather than images where possible -- so that we can cut/paste from github into our python sessions? To get the code to look nice use triple-backticks like ` ``` ` and when it is python code, you can get it highlighted nicely using a paragraph like this:
```
```python
H = nx.Graph([(i, i+1) if i>3 else (i+100, 0)])
``
```
Click on the Preview tab (of the comment) to check the formatting, and then on the Write tab to edit it.

Thanks for these PRs.

navyagarwal: @dschult I have added the code for the last comment, and will keep this format in mind for the future as well!

dschult: Well...   I installed pypy using conda and it seems to raise the ValueError as we want it to.
Not sure why the test is failing.

dschult: Hah Haa!  Adding this test caught an error. And it's because we use pypy in our tests.
Apparently, pypy set.pop() produces the items in a different order from set.pop() in Cpython. That mixed up the order that the nodes are processed in `minimum_spanning_edges` in the prim algorithm. That bug that this caught was that edges from nodes `u` (found as roots) are checked for `nan` near line 296. But edges from the neighbors v of those nodes are explored later in the code: near line 336 and 342.

Can you add the code that catches `nan` just after each of those `d.get("weight", 1)` lines?

Hopefully that will fix the trouble.

navyagarwal:
@dschult You tracked down the bug! All checks have passed! Thank you so much :)

dschult (approved):
Looks good to me!
Thanks

rossbar (approved):
LGTM, thanks @navyagarwal !

PR #6487, opened 2023-03-11
Add docstring examples to boundary.py

navyagarwal:
I added examples in the docstring for the edge_boundary and node_boundary methods in boundary.py

### DIFF
--- a/networkx/algorithms/boundary.py
+++ b/networkx/algorithms/boundary.py
@@ -61,17 +61,17 @@ def edge_boundary(G, nbunch1, nbunch2=None, data=False, keys=False, default=None
 
     Examples
     -------
-    >>> G = nx.gnp_random_graph(10, 0.3, seed = 100)
+    >>> G = nx.wheel_graph(6)
 
     When nbunch2=None:
 
-    >>> list(nx.edge_boundary(G, (2, 4, 5)))
-    [(2, 3), (2, 8), (4, 9), (5, 8)]
+    >>> list(nx.edge_boundary(G, (1, 3)))
+    [(1, 0), (1, 2), (1, 5), (3, 0), (3, 2), (3, 4)]
 
     When nbunch2 is given:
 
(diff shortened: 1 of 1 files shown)


networkx/algorithms/boundary.py
```diff
@@ -59,6 +59,20 @@ def edge_boundary(G, nbunch1, nbunch2=None, data=False, keys=False, default=None
         are specified and `G` is a multigraph, then edges are returned
         with keys and/or data, as in :meth:`MultiGraph.edges`.
 
+    Examples
+    -------
+    >>> G = nx.gnp_random_graph(10, 0.3, seed = 100)
```
dschult, line 64: Can you make this example use a graph where the structure of the graph is known and easy to see what the answer should be?  Using a random graph as an example makes it hard to tell what the function is doing -- because we don't know what the graph is.

Maybe a ladder graph would work with nbunch1 being the nodes on one side and nbunch2 being either `None` (meaning all the nodes on the other side of the ladder) or a list of some of the nodes on the other side of the ladder. Make the ladder small so it's easy to follow.  Or I'm sure there are other example graphs that could be constructed -- some of which might be better examples.

networkx/algorithms/boundary.py
```diff
@@ -121,6 +135,20 @@ def node_boundary(G, nbunch1, nbunch2=None):
     set
         The node boundary of `nbunch1` with respect to `nbunch2`.
 
+    Examples
+    -------
+    >>> G = nx.gnp_random_graph(10, 0.3, seed = 100)
```
dschult, line 140: Similar comment here -- is there a good graph that shows the node boundaries?

networkx/algorithms/boundary.py
navyagarwal: Got it, I changed the example to use a wheel_graph with 6 nodes instead.

networkx/algorithms/boundary.py
navyagarwal: Same change

dschult (approved):
Looks good!
Thanks!

PR #6491, opened 2023-03-12
Add docstring examples to dag.py

navyagarwal:
I added examples in the docstring for the compute_v_structures method in dag.py for better understanding.

### DIFF
--- a/networkx/algorithms/dag.py
+++ b/networkx/algorithms/dag.py
@@ -1224,6 +1224,17 @@ def compute_v_structures(G):
         The v structures within the graph. Each v structure is a 3-tuple with the
         parent, collider, and other parent.
 
+    Examples
+    --------
+    >>> G = nx.DiGraph()
+    >>> G.add_edges_from([(1, 2), (0, 5), (3, 1), (2, 4), (3, 1), (4, 5), (1, 5)])
+    >>> list(nx.compute_v_structures(G))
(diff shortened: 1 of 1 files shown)


navyagarwal:
@dschult @MridulS Requesting review

networkx/algorithms/dag.py
```diff
@@ -1224,6 +1224,17 @@ def compute_v_structures(G):
...
         parent, collider, and other parent.
 
+    Examples
+    --------
+    >>> G = nx.DiGraph()
+    >>> G.add_edges_from([(1, 2), (0, 5), (3, 1), (2, 4), (3, 1), (4, 5), (1, 5)])
+    >>> list(nx.compute_v_structures(G))
+    [(0, 5, 4), (0, 5, 1), (1, 5, 4)]
+
+    >>> G = nx.gnp_random_graph(5, 0.3, seed = 100, directed = True)
```
dschult, line 1234: What does this example show that the first example doesn't show?  It's hard for me to tell because I can't tell what the edges are. We could either remove this second example, or construct a graph from the edges so we can easily see what graph we are working with.

navyagarwal: I added in the second example purely for the sake of more examples to refer, since it doesn't show anything new by itself, it can be removed.

PR #6496, opened 2023-03-12
Updating dict.get default = 1

Qudirah:
fixes for #6443

### DIFF
--- a/networkx/utils/heaps.py
+++ b/networkx/utils/heaps.py
@@ -64,7 +64,7 @@ class MinHeap:
         """
         raise NotImplementedError
 
-    def get(self, key, default=None):
+    def get(self, key, default=1):
         """Returns the value associated with a key.
 
         Parameters
@@ -170,7 +170,7 @@ class PairingHeap(MinHeap):
         del self._dict[min_node.key]
         return (min_node.key, min_node.value)
 
-    def get(self, key, default=None):
+    def get(self, key, default=1):
         node = self._dict.get(key)
         return node.value if node is not None else default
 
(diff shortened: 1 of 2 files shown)


Qudirah:
@rossbar Please I do not understand why it keeps canceling. There isn't a particular error. I need help :(

networkx/utils/heaps.py
```diff
@@ -64,7 +64,7 @@ def pop(self):
         """
         raise NotImplementedError
 
-    def get(self, key, default=None):
+    def get(self, key, default=1):
```
MridulS, line 67: Hmm, why this change? What is this pull request supposed to do?

MridulS:
Also before sending a PR please do run the tests locally https://networkx.org/documentation/latest/developer/contribute.html#testing, this will catch all the issues before creating a pull request :)

networkx/utils/heaps.py
Qudirah: My bad. I was getting frustrated because the test kept failing and I didn't link the issue. It's #6443. Have I implemented it wrongly?

Qudirah:
> Also before sending a PR please do run the tests locally https://networkx.org/documentation/latest/developer/contribute.html#testing, this will catch all the issues before creating a pull request :)

Got it. Thanks.

networkx/utils/heaps.py
MridulS: Yes, this bit of code is for heaps. I don't think this has anything to do with GEXF export.

What @dschult meant in https://github.com/networkx/networkx/pull/6484#issuecomment-1465016028 was to look at dict.get in python https://docs.python.org/3/library/stdtypes.html#dict.get

An example:
``` python
>>> d = {1: 1, 2: 2}
>>> d.get(1)
1
>>> d.get(3)  # by default it gives None
>>> d.get(3, 'something') # you can change the default by providing another value
'something'
```

Qudirah: smh. My bad. Thank you for explaining. will correct it now.

Qudirah:
@MridulS Please check and merge if correct :D

PR #6498, opened 2023-03-12
Adding is_tounament to main namespace

Qudirah:
fixes for #5542 

@MridulS did I do it right? Also, this change will lead to little changes in docs. Would I need to open new PRs for that?

### DIFF
--- a/networkx/__init__.py
+++ b/networkx/__init__.py
@@ -41,6 +41,7 @@ from networkx.readwrite import *
 # Need to test with SciPy, when available
 from networkx import algorithms
 from networkx.algorithms import *
+from networkx.algorithms.tournament import is_tournament
 
 from networkx import linalg
 from networkx.linalg import *
--- a/networkx/algorithms/__init__.py
+++ b/networkx/algorithms/__init__.py
@@ -78,6 +78,7 @@ from networkx.algorithms import lowest_common_ancestors
 from networkx.algorithms import operators
 from networkx.algorithms import shortest_paths
 from networkx.algorithms import tournament
+from networkx.algorithms.tournament import is_tournament
 from networkx.algorithms import traversal
 from networkx.algorithms import tree
 


networkx/algorithms/__init__.py
```diff
@@ -78,6 +78,7 @@
 from networkx.algorithms import operators
 from networkx.algorithms import shortest_paths
 from networkx.algorithms import tournament
+from networkx.algorithms.tournament import is_tournament
```
MridulS, line 81: This need to go down where other functions are imported. Look at line 85 and below in the same file.

networkx/__init__.py
```diff
@@ -41,6 +41,7 @@
 # Need to test with SciPy, when available
 from networkx import algorithms
 from networkx.algorithms import *
+from networkx.algorithms.tournament import is_tournament
```
MridulS, line 44: This change isn't required.

MridulS:
> @MridulS did I do it right? Also, this change will lead to little changes in docs. Would I need to open new PRs for that?

The best way to test that is to do it on you computer. Can you run the following code?
``` python
import networkx as nx
G = nx.path_graph(4)
nx.is_tournament(G)
```

If yes, then you are in the right direction :)

networkx/algorithms/__init__.py
Qudirah: done!

Qudirah:
> > @MridulS did I do it right? Also, this change will lead to little changes in docs. Would I need to open new PRs for that?
> 
> The best way to test that is to do it on you computer. Can you run the following code?
> 
> ```python
> import networkx as nx
> G = nx.path_graph(4)
> nx.is_tournament(G)
> ```
> 
> If yes, then you are in the right direction :)

I think I have done the right thing now.

Qudirah:
@MridulS please check and merge if correct :)

MridulS (approved):
Thanks!

dschult (approved):

dschult:
Ahh... Good point. I think we are "due" for a review of the namespace configuration. We'll have potentially many changes like moving imports from the subpackage to the main package at that time. I hope that can be fairly soon -- maybe at the summit?  It might be good for one of us to take a first pass at it to see what is involved.  If that should be done to fix this PR I guess we'll need a new PR now...

PR #6499, opened 2023-03-12
tweak `test_override_dispatch` to allow G keyword

eriknw:
This is a simple follow-up to #6471, and, as in that PR, this fix is short term. We would like to find a more elegant, robust solution (probably after nx 3.1).

I sometimes forget there are two versions of `_dispatch`, and this PR fixes the one that gets used when `NETWORKX_GRAPH_CONVERT` environment variable is set. NetworkX does not yet test `test_override_dispatch`, so it's not easy to add a test, but it gets used by external backends such as `graphblas_algorithms`. Ideally, we should test this in NetworkX, which will require calling subprocess in a test with an environment variable set.

CC @MridulS @jim22k


rossbar (approved):
Thanks for the followup @eriknw !

dschult (approved):

PR #6503, opened 2023-03-13
Added test cases for join operation and fixed join operation to handle label_attributes

### LINKED ISSUE #6488: Join operation in trees---not handling label_attribute
[https://github.com/networkx/networkx/blob/main/networkx/algorithms/tree/operations.py](https://github.com/networkx/networkx/blob/main/networkx/algorithms/tree/operations.py)
1. The resulting graph of join operation in trees isn't including the old labels of inputs.
2.  Not handling the cases where label_attribute is passed as an argument.

### Current Behavior
Screenshot from 2023-03-11 17-03-33

### Expected Behavior

image

### Steps to Reproduce
As shown above

### Environment

Python version: 3.10.6
NetworkX version: 3.0

### Additional context
[https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.tree.operations.join.html](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.tree.operations.join.html)

### LINKED ISSUE #6490: Improve test coverage for operations.py (join)
### Current Behavior
https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/tree/operations.py  the current test coverage is 92.8%. There are still some cases needed to be handled.

### Expected Behavior

https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.tree.operations.join.html
1. Test case to check label_attribute should be added 
2. In the documentation its written that the inputs must be tree. But this function works for graphs too. Could you tell me if its for  trees or graphs as well?
### Environment

Python version:3.10.6
NetworkX version:3.0

sirichandana-v:
Fixes #6488 and fixes #6490 
@MridulS 
Modified join operation to store label_attributes and handled the case when label_attribute argument is provided.
Added test cases for join operation to check label_attributes.
Could you please check the PR and suggest any changes if needed to be done.

https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.tree.operations.join.html
In the documentation its written that Input has to be list of trees but this function works for graphs as well. Should I change the documentation by replacing trees with graphs or modify the join function to not accept list of graphs as input?

### DIFF
--- a/networkx/algorithms/tree/tests/test_operations.py
+++ b/networkx/algorithms/tree/tests/test_operations.py
@@ -2,8 +2,20 @@
 
 """
 
+from itertools import chain
+
 import networkx as nx
 from networkx.utils import edges_equal, nodes_equal
+
+
+def _check_label_attribute(input_trees, res_tree, label_attribute="_old"):
+    """Tests to check if label_attribute is storing"""
+
+    res_attr_dict = nx.get_node_attributes(res_tree, label_attribute)
+    res_attr_list = list(res_attr_dict.values())
+    input_label = (list(tree[0].nodes()) for tree in input_trees)
+    input_label_list = list(chain.from_iterable(input_label))
+    return res_attr_list == input_label_list
 
 
 class TestJoin:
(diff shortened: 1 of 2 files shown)


sirichandana-v:
@MridulS @dschult  Could you tell me if I need to change anything in this PR

networkx/algorithms/tree/tests/test_operations.py
```diff
@@ -2,10 +2,22 @@
...
 
 
+def _check_label_attribute(input_trees, res_tree, label_attribute="_old"):
+    """Tests to check if label_attribute is storing"""
+
+    res_attr_dict = nx.get_node_attributes(res_tree, label_attribute)
+    res_attr_list = list(res_attr_dict.values())
+    input_label = (list(tree[0].nodes()) for tree in input_trees)
+    input_label_list = list(chain.from_iterable(input_label))
+    return res_attr_list == input_label_list
```
dschult, line 18: This looks good -- :}

I wonder if the `check_attrribute` function should use sets instead of lists for the two label lists. Right now the test passes, but I can imagine a case where the order of the labels might differ, but we'd still have the right set of labels. The function would return False, but it should return True.

MridulS:
@dschult pushed a commit that uses set of comparison.

dschult:
Good -- Thanks @MridulS and @sirichandana-v 

Let's move other possible changes to a different PR to get this one in.
Other items:
- In the doc_string of `join`, indicate that trees are represented by networkx graphs.
- The name `join` is a terrible name.  It means that `nx.join` doesn't give any context to what that function does.  We should at least make it `join_trees` or something like that.  That would require deprecation of course.

I can make an issue with these items when this is closed -- unless someone wants to do them here.

MridulS:
I've created https://github.com/networkx/networkx/issues/6906 to work further on this, I think we can merge this in :)

PR #6504, opened 2023-03-13
Improve test coverage for astar.py

navyagarwal:
I found that the [test coverage for astar.py](https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/shortest_paths/astar.py) is currently 91%.

I have added test cases for the two exceptions that can possibly be raised by the astar_path and astar_path_length methods.

The test coverage has now reached 100%.

image


MridulS (approved):
Thanks!

dschult (approved):
Thansk!

Issue #6509, opened 2023-03-16
Unexpected behaviour of .edges


Hanspagh:
When using a string to look up edges the string gets interpreted as an array which results in the edges call returning edges with the id of any of the chars of the string. If the string is surrounded by a list it works as expected. This behavior was a little surprising to me, but if it is expected feel free to ignore this.

### Current Behavior
``` python
import networkx as nx
g = nx.DiGraph()
g.add_edge("b", "c")
g.edges("ab")
OutEdgeDataView([('b', 'c')])
g.edges(["ab"])
OutEdgeDataView([])
```

### Expected Behavior

``` python
import networkx as nx
g = nx.DiGraph()
g.add_edge("b", "c")
g.edges("ab")
OutEdgeDataView([])
```

### Environment
Python version: 3.9.13
NetworkX version: '3.0.0'

navyagarwal:
@Hanspagh Hey, you can check out a related discussion over [here](https://github.com/networkx/networkx/issues/4781)!

@dschult I think we should include some clarifying examples in the [documentation here](https://networkx.org/documentation/stable/reference/glossary.html#term-nbunch) for the time being, irrespective of whether we plan to change the current behavior or not, so that there are no surprises for users!

Hanspagh:
Thanks for the link, not sure how I missed that discussion.

dschult:
Yes -- a string is a container. And containers can be nodes, or pairs of nodes (edges), or even paths in some contexts.

And yes @navyagarwal including some clarifying examples in the documentation would be good! :)

navyagarwal:
> And yes @navyagarwal including some clarifying examples in the documentation would be good! :)

I'll get to it!

PR #6513, opened 2023-03-17
Add examples clarifying ambiguity of nbunch

navyagarwal:
I have added examples to the definition of `nbunch` in the glossary section of documentation.

They clarify the behavior when an iterable is passed as nbunch to a method in Networkx.

Reference issues #6509 and #4781

### DIFF
--- a/doc/reference/glossary.rst
+++ b/doc/reference/glossary.rst
@@ -56,7 +56,7 @@ Glossary
          OutEdgeDataView([])
 
       If "bc" is wrapped in a list, the list is the iterable and
-      "bc" is treated as a single node, that is, if the
+      "bc" is treated as a single node. That is, if the
       nbunch is an iterable of iterables, the inner iterables will
       always be treated as nodes.
 
--- a/doc/reference/glossary.rst
+++ b/doc/reference/glossary.rst
@@ -26,14 +26,12 @@ Glossary
 
    nbunch
       An nbunch is a single node, container of nodes or `None` (representing
-      all nodes). It can be a list, set, graph, etc.. To filter an nbunch
-      so that only nodes actually in `G` appear, use `G.nbunch_iter(nbunch)`.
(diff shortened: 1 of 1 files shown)


doc/reference/glossary.rst
```diff
@@ -29,6 +29,44 @@ Glossary
...
+
+      Since "bc" is not a node in G, it is treated as an iterator.
+
+      .. nbplot::
+         
+         >>> G.edges(["bc"])
+         OutEdgeDataView([])
+
+      If "bc" is wrapped in a list, the list is the iterable and
+      "bc" is treated as a single node.
```
dschult, line 59: This could be reworded to frame it as a list of iterables. Which is more clear? The current wording focuses on a single example (from which the reader can abstract to the general case). The wording below focuses on the general case, but maybe it then becomes too abstract.  Which do you think is better? I'm not sure...

```
    If the nbunch is an iterable of iterables, the inner iterables will always be treated as nodes. So, a list of strings ensures that the string is treated as a node and not treated as an iterable of nodes.
```

Maybe the best is to combine both into a single paragraph. And leaving it as it is would work too. The decision is yours.

navyagarwal: >  If the nbunch is an iterable of iterables, the inner iterables will always be treated as nodes.

I think this sentence should definitely be added! Else someone might find it difficult to generalize `list of strings` to other use cases, like a tuple of tuples.

doc/reference/glossary.rst
```diff
@@ -56,7 +56,9 @@ Glossary
          OutEdgeDataView([])
 
       If "bc" is wrapped in a list, the list is the iterable and
-      "bc" is treated as a single node.
+      "bc" is treated as a single node, that is, if the
```
dschult, line 59: **very** minor suggestion
```suggestion
      "bc" is treated as a single node. That is, if the
```

dschult (approved):
Looks good!  Thanks!!

PR #6515, opened 2023-03-17
Added isomorphvf2 to import

ManasviEmmadi:
An import for isomorphvf2 was missing so I have added the same.

### DIFF
--- a/networkx/algorithms/isomorphism/__init__.py
+++ b/networkx/algorithms/isomorphism/__init__.py
@@ -5,3 +5,4 @@ from networkx.algorithms.isomorphism.temporalisomorphvf2 import *
 from networkx.algorithms.isomorphism.ismags import *
 from networkx.algorithms.isomorphism.tree_isomorphism import *
 from networkx.algorithms.isomorphism.vf2pp import *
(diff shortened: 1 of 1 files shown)


ManasviEmmadi:
@rocky @lonnen @cournape @thewtex sorry for tagging everyone, but I am facing this error in most of the checks:
`TypeError: DiGraphMatcher.__init__() got an unexpected keyword argument 'node_match'`
I have only added `from networkx.algorithms.isomorphism.isomorphvf2 import *` to __init__.py file. Please guide me as to how to proceed further.

dschult:
Why do you think that `isomorphvf2` should be imported directly into that namespace?
It may have used identical names to other names already in the namespace. For example, `GraphMatcher` seems to be defined twice -- once with keyword options for `node_match` and once without those keyword options.

It's probably better not to import isomorphvf2 directly into the namespace. We could revise the namespace and our roadmap suggests that this might happen sometime "soon". But it will take some discussion.

dschult:
You can still comment on this PR even though it is closed. And we can reopen it if desired.

PR #6521, opened 2023-03-19
Fixed bug k_truss doesn't raise exception for self loops

PurviChaurasia:
This PR is in reference to #6508

### DIFF
--- a/doc/reference/glossary.rst
+++ b/doc/reference/glossary.rst
@@ -26,8 +26,40 @@ Glossary
 
    nbunch
       An nbunch is a single node, container of nodes or `None` (representing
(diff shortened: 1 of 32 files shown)


PurviChaurasia:
@dschult I think this resolves the issue

dschult (approved):
This looks like is adds a check so that the code matches the documentation.
And I agree with the ideas from #6508 that this is the right approach. It isn't strictly backward compatible, but the old behavior could produce nonsense -- so this is fixing the bug and we shouldn't try to preserve the old behavior when self-loops are present.

I approve this PR!

PurviChaurasia:
Thanks a lot! @dschult

PR #6526, opened 2023-03-20
Add community detection example to Gallery

navyagarwal:
I have written a python script that demonstrates the use of Girvan-Newman method to detect communities in the Zachary Karate Club dataset. It also plots the change in modularity as the important edges are removed with each iteration.

### DIFF
--- a/examples/algorithms/plot_girvan_newman.py
+++ b/examples/algorithms/plot_girvan_newman.py
@@ -1,0 +1,74 @@
+"""
+=======================================
+Community Detection using Girvan-Newman
+=======================================
+
+This example shows the detection of communities in the Zachary Karate
+Club dataset using the Girvan-Newman method.
+
+We plot the change in modularity as important edges are removed. 
+Graph is coloured and plotted based on community detection when number 
+of iterations are 1 and 4 respectively.
+"""
+
+import networkx as nx
+from networkx.algorithms import community
+import pandas as pd
+import matplotlib.pyplot as plt
+
+# Load karate graph and find communities using Girvan-Newman
+G = nx.karate_club_graph()
+communities = list(community.girvan_newman(G))
+
+# Plot change in modularity as the important edges are removed
+# Modularity -> measures the strength of division of a network into modules
+modularity_df = pd.DataFrame(
+    [[k + 1, community.modularity(G, communities[k])] for k in range(len(communities))],
+    columns=["k", "modularity"],
+)
+modularity_df.plot.bar(
+    x="k",
+    figsize=(10, 6),
+    color="#F2D140",
+    title="Modularity Trend for Girvan-Newman Community Detection",
+)
+
+
+# function to create node colour list
+def create_community_node_colors(graph, communities):
+    number_of_colors = len(communities[0])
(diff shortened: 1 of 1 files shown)


examples/algorithms/plot_girvan_newman.py
```diff
@@ -0,0 +1,74 @@
...
+This example shows the detection of communities in the Zachary Karate
+Club dataset using the Girvan-Newman method.
+
+We plot the change in modularity as important edges are removed. 
+Graph is coloured and plotted based on community detection when number 
+of iterations are 1 and 4 respectively.
+"""
+
+import networkx as nx
+from networkx.algorithms import community
```
dschult, line 15: Just to make this more up-to-date with our "lazy-loading" system, can you remove this import and use `nx.community` everywhere later in place of `community`?

examples/algorithms/plot_girvan_newman.py
```diff
@@ -0,0 +1,74 @@
...
+        font_size=20,
+        font_color="black",
+    )
+    plt.title(title)
+    plt.show()
+
+
+# Plot graph with colouring based on communities
+visualize_communities(G, communities[0])
+visualize_communities(G, communities[3])
```
dschult, line 74: This seems to overwrite the plot in figure 2.  Maybe you need to make `visualize_communities` draw into the current figure and then use `plt.figure(2)` before calling it the first time and `plt.figure(3)` before the next call. I'm not sure though -- how many figures did you intend?

examples/algorithms/plot_girvan_newman.py
navyagarwal: Will do!

examples/algorithms/plot_girvan_newman.py
navyagarwal: I am trying to output the following 3 figures and it's working alright on my Jupyter notebook as well as on Google Colab.

image

I did have a hard time with the plotting of figures but since I got the output I wanted on my system, I didn't look too hard into the code.

dschult: I think the notebooks and colab do some magic in the background to make each python block/plot command create it's own figure environment.  But [the Gallery entry for the documentation](https://output.circle-artifacts.com/output/job/4706f5c1-ed7a-4ddc-8822-57009ca94ba5/artifacts/0/doc/build/html/auto_examples/algorithms/plot_girvan_newman.html#sphx-glr-auto-examples-algorithms-plot-girvan-newman-py) shows only two plots.  To see what the docs will look like, click on the "Details" link to the right of the "circleci: documentation artifact" check (near the bottom of the PR "Show all checks, then scroll down to the 3rd configuration tested from the bottom.) 

Also, the thumbnail plot for this entry in the gallery shows the relatively boring bar graph. I tink there is a fancier way to get the plot of your choice to be the thumbnail displayed in the gallery itself. But reordering the plots so that the bar graph is last would do it I think.

navyagarwal: I didn't know that I could check out the generated docs like that, thank you for pointing me there, it was very helpful!

I have made the required changes for the plot layout and I think the thumbnail also looks pretty good now.

dschult (approved):
Looks good to me!

jarrodmillman (approved):

Issue #6527, opened 2023-03-20
Similar functions in shortest_paths produce different types of output


navyagarwal:
The single_source_shortest_path_length method returns a dictionary while the single_target_shortest_path_length returns a generator.  This was first discussed [here](https://github.com/networkx/networkx/discussions/5511).

I further looked into the output types of all related functions in shortest_paths and I believe it would make sense to change the output type of all length-related methods to dictionary for the sake of consistency.

<img width="278" alt="image" src="https://user-images.githubusercontent.com/82928853/226372801-5a4edf02-4dbe-4294-bcfc-93b9ab8ed74a.png">

Fixing this would only require minor code edits and I presume some changes in the test files as well, it'd be great if I can get someone else's opinion on this as well!

dschult:
I vaguely recall a discussion of this long long ago where we decided that the `all_pairs_...` versions should return an iterator that could easily be made into a dict. That was because the dicts that are created can be huge, especially when returning the paths. And for memory many people would want to process them one at a time rather than store a large data structure holding all that at once.

My quick peek suggests that the `all_pairs_shortest_path` returns an iterator no matter what the docs say (hmmm... seems the docs are wrong.

I am disappointed that the `single_source_shortest_path_length` returns a dict. That should be an iterator as well. If someone needs a dict they can easily make it one by wrapping it in a `dict` call. But maybe I should look at the "blame history" to figure out why it is the way it is. 

I think the `shortest_path` and `shortest_path_length` are meant to return the full structure -- because those are for ease of use. And sometimes they are returning a single path and other times a whole collection of paths. So the return value can be int or list of dict if I understand correctly.

I would think that all the rest could/should return iterators that can be made into dicts easily.  But I better do a "blame" search before this gets merged... :)   There is also backward-compatibility to take into account, but this PR is more about what it should be -- we can worry about backward compatibility later. :)

navyagarwal:
`all_pairs_shortest_path` does return an iterator, I will correct the return type in the docs in PR #6528 

As for the other methods, considering the memory savings  and how easily the iterator can be wrapped in a dictionary when needed, wouldn't it be best if we could return an iterator for all methods instead of a dictionary? Because even for methods like `single_source_...` or `single_target_...` whose output is smaller than `all_pairs_...`, the output can still be significantly large if the graph is big enough.

[...]

PR #6528, opened 2023-03-20
Fixed return type inconsistencies in shortest path methods documentation

navyagarwal:
Minor edits to correct the return type of path-related methods from `lengths` to `paths` in the documentation.

### DIFF
--- a/networkx/algorithms/shortest_paths/unweighted.py
+++ b/networkx/algorithms/shortest_paths/unweighted.py
@@ -445,7 +445,7 @@ def all_pairs_shortest_path(G, cutoff=None):
 
     Returns
     -------
-    paths : dictionary
+    paths : iterator
         Dictionary, keyed by source and target, of shortest paths.
 
     Examples


navyagarwal:
@dschult I have changed the return type of `all_pairs_shortest_path` to iterator in the docstring. All other methods in the module seem to be in order.

I found the below comment in both `all_pairs_shortest_path` and `all_pairs_shortest_path_length` methods in the module - 

```python
 # TODO This can be trivially parallelized.
```
Right now I don't understand what it means, but is this something I can possibly work on?

MridulS:
> Right now I don't understand what it means, but is this something I can possibly work on?

Please look at all the previous discussions here https://github.com/networkx/networkx/issues/4064 and a bunch of other places inside networkx. One possible way for this is https://github.com/networkx/networkx/pull/6306 but we are nowhere near having a concrete plan for this, and wouldn't recommend to pick this up as a new contributor task :)

dschult (approved):
Despite the longer discussion surrounding this issue and #6527 , this PR seems to be easily reviewable and obviously ready.  Let's go ahead and merge it.
It's only changes in documentation to make them match current behavior. :}

Thanks!!

PR #6530, opened 2023-03-21
Add example script for shortest path

PurviChaurasia:
This is in reference to #6482. I have created a python script to find the shortest path between 2 nodes of a graph and visualize it.


dschult (commented):
This example looks good.  Thanks!

The changes to the `weighted.py` file don't really belong here. Could they be removed?

PurviChaurasia:
I accidently did something in the weighted.py file. I'll make a new pull request sorry!

PR #6534, opened 2023-03-21
Add example script for shortest path

PurviChaurasia:
Added python script to visualize shortest path between 2 nodes of a graph.


rossbar (approved):
LGTM, thanks @PurviChaurasia !

PR #6535, opened 2023-03-21
Add note about using latex formatting in docstring in the contributor guide

paulitapb:
As we discussed the use of latex formatting in the docstring I added a comment about how to properly use it in the contributor guide.  Also added some info about how to add math formulas because it's related to the use of latex formatting.


dschult:
This looks nice!  Would it be worth adding an inline Latex example in addition to the indented/offset example?   Maybe at the end of the paragraph add: like $E=mc^2$. (there has got to be a more network-y formula to use as an example... :)

jarrodmillman:
Maybe it is to specialized, but I like Euler's formula:
`F + V − E = 2`

I also like Cheeger's inequalities and the Expander mixing lemma (e.g., see https://en.wikipedia.org/wiki/Expander_graph)

CONTRIBUTING.rst
```diff
@@ -477,6 +477,23 @@ to use the `wayback machine <https://web.archive.org/>`_ to create a snapshot of
...
+
+An example of a math formula::
+      
+      .. math:: 
+         Ax = \lambda x
+
+Some inline math::
+    
+   These are Cheeger's Inequalities for \d-Regular graphs: 
+   \frac{d- \lambda_2}{2} \leq h(G) \leq \sqrt{2d(d- \lambda_2)}
```
dschult, line 495: Just to check that these examples work, let's put the results of that syntax into the document.
And we need to add the dollar signs around the inline text.
```suggestion
      .. math:: 
          Ax = \lambda x

.. math:: 
    Ax = \lambda x

Some inline math::
    
    These are Cheeger's Inequalities for \d-Regular graphs: 
    $\frac{d- \lambda_2}{2} \leq h(G) \leq \sqrt{2d(d- \lambda_2)}$
   
These are Cheeger's Inequalities for \d-Regular graphs: 
$\frac{d- \lambda_2}{2} \leq h(G) \leq \sqrt{2d(d- \lambda_2)}$
```

dschult (approved):
Looks good!  Thanks!

rossbar (approved):
I took the liberty of pushing up very minor wording/formatting changes. Thanks for documenting this @paulitapb !

PR #6536, opened 2023-03-21
Test dispatching via nx-loopback backend

jim22k:
The dispatch functionality is used to delegate graph
computations to a different backend. Because those backends
are not part of NetworkX, testing the dispatching feature
was not originally added, relying instead on the other
backends (e.g. graphblas-algorithms) to verify the dispatch
functionality is working.

This change creates a "loopback" backend where NetworkX dispatches
to itself for the sole purpose of exercising the dispatching
machinery.

In one incarnation, various tests are augmented to use the
LoopbackGraph family and force loopback dispatching to occur
as normal usage would (see nx.algorithms.components.tests.test_connected:64).

A second incarnation forces *all* tests to run in dispatch mode
but use of a different `_dispatch` decorator. This mode is triggered
for all of pytest, so it must be tested by the CI system specifically.

### DIFF
--- a/networkx/classes/backends.py
+++ b/networkx/classes/backends.py
@@ -53,8 +53,8 @@ tests are working, while avoiding causing an explicit failure.
 
 A special ``on_start_tests(items)`` function may be defined by the backend.
 It will be called with the list of NetworkX tests discovered. Each item
-is a `pytest.Node` object. If the backend does not support the test, that
-test can be marked as xfail.
+is a test object that can be marked as xfail if the backend does not support
+the test using `item.add_marker(pytest.mark.xfail(reason=...))`.
 """
 import functools
 import inspect
@@ -147,6 +147,7 @@ def _dispatch(func=None, *, name=None):
                     )
         return func(*args, **kwds)
 
+    wrapper._orig_func = func
     _register_algo(name, wrapper)
     return wrapper
 
@@ -171,6 +172,8 @@ def test_override_dispatch(func=None, *, name=None):
     def wrapper(*args, **kwds):
         backend = plugins[plugin_name].load()
         if not hasattr(backend, name):
+            if plugin_name == "nx-loopback":
+                raise NetworkXNotImplemented(f"'{name}' not found in {backend.__class__.__name__}")
             pytest.xfail(f"'{name}' not implemented by {plugin_name}")
         bound = sig.bind(*args, **kwds)
         bound.apply_defaults()
@@ -196,6 +199,7 @@ def test_override_dispatch(func=None, *, name=None):
         result = getattr(backend, name).__call__(graph, *args, **kwds)
         return backend.convert_to_nx(result, name=name)
 
+    wrapper._orig_func = func
     _register_algo(name, wrapper)
     return wrapper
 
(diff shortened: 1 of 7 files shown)


jim22k:
@rossbar Please take a look. I'm not sure if I got the CI correct, but hopefully it only runs for one of the matrix combos.

dschult (approved):
I've looked through this PR (partly to learn more about the dispatch process) and it looks good to me. The tests are run on the default ubuntu 3.11 check and not on the others AFAICT. On that test, I can see that all networkx test functions are run -- whether they have the `_dispatch` decorator or not.  The loopback dispatcher is appropriately just a wrapper of the base classes. In short, I don't see anything wrong and I have no concerns other than that I don't know the CI ins and outs nor the setup.py details very well so something might be slipping by me.

I'm approve this PR, and I look forward to comments from @rossbar and @MridulS

jim22k:
@rossbar do you have time to look at this before the meeting tomorrow?

rossbar (commented):
Generally LGTM, just a few points of clarification to make sure I have the right mental model.

The purpose of the parametrizations in the "first" incantation is to explicitly test the `dispatch_interface.convert` mechanism, correct? Was there explicitly reasoning behind which methods were checked with this mechanism (i.e. `google_matrix`, `constraint`, etc.) vs. others? Is the plan to expand the use of this parametrization throughout the test suite or limit it to this subset of tests?

Depending on the answer to these questions, I think it'd be worth adding comments to explain the dispatch parametrization for future users/contributors to the tests.

networkx/classes/backends.py
```diff
@@ -147,6 +147,7 @@ def wrapper(*args, **kwds):
                     )
         return func(*args, **kwds)
 
+    wrapper._orig_func = func
```
MridulS, line 150: All looks good, just some general questions :)

This is to make the dispatch_interface.convert bit work right? With the `NETWORKX_GRAPH_CONVERT` env variable it should automatically be able to do `getattr(nx, item)` in `LoopbackDispatcher.__getattr__` right?

jim22k: This is only to support our loopback testing. Other dispatching backends shouldn't need this.

Before I added `_orig_func`, the tests would go into an infinite loop because the "loopback" function was still decorated and would try the loopback dispatch logic again.

jarrodmillman (approved):

jim22k:
> The purpose of the parametrizations in the "first" incantation is to explicitly test the `dispatch_interface.convert` mechanism, correct? Was there explicitly reasoning behind which methods were checked with this mechanism (i.e. `google_matrix`, `constraint`, etc.) vs. others? Is the plan to expand the use of this parametrization throughout the test suite or limit it to this subset of tests?
> Depending on the answer to these questions, I think it'd be worth adding comments to explain the dispatch parametrization for future users/contributors to the tests.

The parametrized tests were chosen at random to test `networkx.classes.backends:_dispatch` logic. This is the standard dispatching code that we need to test. The `dispatch_interface.convert` function achieves this by making thin wrappers around the NetworkX Graph objects with a `__networkx_plugin__` attribute. That is what the dispatcher looks for to trigger the dispatching behavior.

I will add comments to explain that these tests are doing double-duty, testing both the algorithm and the dispatch mechanism.

We can certainly add dispatch parametrizing to other tests if we want, but I don't think we should blanket cover the tests with dispatching because they effectively run twice.


The `NETWORKX_GRAPH_CONVERT=nx-loopback` test is explicitly to test the `networkx.classes.backends:_test_override_dispatch` logic. This is used by backends to run the full NetworkX test suite against their backend, automatically converting to the backend Graph-like object via `convert_to_nx` and running the tests.

PR #6551, opened 2023-03-25
Improve test coverage for redundancy.py

navyagarwal:
I found that the [test coverage for redundancy.py](https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/bipartite/redundancy.py) is currently 92.31%.

I have added some code that tests an _if_ condition that was previously not covered.

The test coverage has now reached 100%.

image


dschult (approved):
Nice!

PR #6542, opened 2023-03-23
Lca example

navyagarwal:
I have written a python script that computes the Lowest Common Ancestors for certain node pairs in a generated tree and then visualizes them with a color scheme.

PR #6552, opened 2023-03-25
Added python script for Lowest Common Ancestor Example

vanshika230:
This is the code for issue #6550. 
I have created a python script to create a binary tree with 8 nodes and perform various operations related to finding the lowest common ancestor (LCA) of nodes in the tree and visualizes it. It briefly uses all the functions mentioned here https://networkx.org/documentation/latest/reference/algorithms/lowest_common_ancestors.html and explains the code well in the comments. The output files of this code are attached to the issue created.
 @dschult @MridulS requesting a review and suggestions for this pull request.

### DIFF
--- a/examples/algorithms/plot_lowest_common_ancestor.py
+++ b/examples/algorithms/plot_lowest_common_ancestor.py
@@ -1,0 +1,53 @@
+"""
+==================
+Lowest Common Ancestor
+==================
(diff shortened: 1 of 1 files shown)


rossbar (commented):
Thanks for the proposal @vanshika230 , this is indeed a nice example but unfortunately is now a bit redundant given #6542 was merged. There is a lot of overlap between the two examples, so I don't think it makes sense to add another.

vanshika230:
Oh yes! Somehow missed that notification! Will come up with another useful example. Closing this, sorry for the inconvenience caused.

PR #6554, opened 2023-03-26
remove usage of 'tempfile.mkstemp' in favour of a 'NamedTemporaryFile'

danieleades:
replaces a low-level API with a context manager

### DIFF
--- a/networkx/algorithms/bipartite/tests/test_edgelist.py
+++ b/networkx/algorithms/bipartite/tests/test_edgelist.py
@@ -108,12 +108,10 @@ class TestEdgelist:
         G.add_edge(name1, "Radiohead", **{name2: 3})
         G.add_node(name1, bipartite=0)
         G.add_node("Radiohead", bipartite=1)
-        fd, fname = tempfile.mkstemp()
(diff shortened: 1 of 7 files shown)


danieleades:
> _ No description provided. _

rossbar (commented):
I'd vote for the replacements in the test suite to use the `tmp_path` pytest fixture rather than using `tempfile` directly. IIRC `tempfile` has some unfortunately platform-dependent behavior.

danieleades:
> I'd vote for the replacements in the test suite to use the `tmp_path` pytest fixture rather than using `tempfile` directly. IIRC `tempfile` has some unfortunately platform-dependent behavior.

the only platform-dependent behaviour i'm familiar with is obviated by the `delete=True` flag. Are you aware of any other platform-dependent behaviour?

i'm afraid i'm not familiar enough with `pytest` to migrate to a fixture

MridulS:
Closing this as https://github.com/networkx/networkx/pull/7221 uses the pytest native fixture to remove tempfile. Thanks for your work on this @danieleades !

Issue #5853, opened 2022-07-12
quotient_graph won't generate self-edges


aeflores:
The `quotient_graph` algorithm will not produce  self-edges even though the original graph has edges between two elements of the same equivalence class. This could make sense for some applications but not for others, maybe having a `self_loops=False` argument would be a reasonable solution that won't break existing uses of the algorithm.

### Current Behavior

A minimal example would be a graph with an edge from "a"  to  "b".
If "a" and "b" are equivalent, I would expect the quotient graph to have a self loop from `frozenset({a,b})` to `frozenset({a,b})`.

The current behavior is:
```
g = networkx.DiGraph()
g.add_edge("a","b", label="in")
g_quotient = networkx.quotient_graph(g,[["a","b"]] )
assert len(g_quotient.edges) == 0
```

### Expected Behavior

The expected behavior would be:
```
g = networkx.DiGraph()
g.add_edge("a","b", label="in")
g_quotient = networkx.quotient_graph(g,[["a","b"]] )
assert len(g_quotient.edges) == 1
```

I believe the problem comes from using `permutations` here https://github.com/networkx/networkx/blob/main/networkx/algorithms/minors/contraction.py#L378 instead of `product`, since permutations will not consider self-edges.

Related issues: https://github.com/networkx/networkx/issues/4935
### Environment

Python version: python 3.8
NetworkX version: 2.8.4

aeflores:
Totally agree! I created the issue to make sure that when #4935 is addressed, self-loops are treated correctly.

Issue #951, opened 2013-09-11
Louvain Modularity Community Detection


SuperAce99:
## Overview

NetworkX only provides the [K-Clique algorithm](http://networkx.lanl.gov/reference/algorithms.community.html?highlight=community#networkx.algorithms.community) for community detection, despite past efforts to add more options such as [Louvain Modularity](https://en.wikipedia.org/wiki/Community_structure#Modularity_maximization). These efforts should be completed and integrated into the main branch.
## Current Status

Code for louvain community detection has been submitted and partially tested in #239. This code appears to be maintained by @taynaud on [bitbucket ](https://bitbucket.org/taynaud/python-louvain/src/bb73de3cd42c?at=default)

@bjedwards [proposed](https://www.google-melange.com/gsoc/proposal/review/google/gsoc2011/bedwards/1001) to add several community detection methods during Summer of Code 2011. The directed laplaian portion of this was eventually merged via #741, but it appears that this was the only portion of the proposal that was completed.
## Proposal

Merge @taynaud's [python-louvain](https://bitbucket.org/taynaud/python-louvain/src/bb73de3cd42ccb8c2234672b9e472702cd4159d9/community.py?at=default) implementation into Network X.

harrismirza:
Is there any update on this? Seems to be held up but I can't quite work out why.

dschult:
Looks like it is waiting for code to be submitted as a Pull Request.

PR #6557, opened 2023-03-26
Improve quotient_graph efficiency

### LINKED ISSUE #4935: quotient_graph seems particularly slow
Given the [speed-up seen in bypassing quotient_graph](https://github.com/networkx/networkx/pull/4929#discussion_r659926874) in #4929 I think we should revisit the code in `quotient_graph`.   And probably in the whole algorithms/minors subpackage.  Maybe it should be turned back into a minors.py module as well...

aaronzo:
Fixes #4935 and #5853.

Plan:
* Write a new `_quotient_graph` function that adds edges to the quotient graph in O(|E|), by looping over original edges. This version of the function will not have an `edge_relation` parameter - it will be a more efficient version of the current default.
* In the user API `quotient_graph` function, either, keep the `edge_relation` parameter - and if it is not `None`, execute the old function.
* add a `self_edges=False` kwarg to `quotient_graph` to give the option to keep edges as per #5853. Keep the default as False to give consistency for existing code.
* I'd also like to change the default behaviour when `node_data=None`. Currently, the default behaviour calculates some [fairly expensive stats](https://github.com/networkx/networkx/blob/c7218d53caa7c8490de89a0950cf539825e6109a/networkx/algorithms/minors/contraction.py#L345-L352) and this is probably unwanted in cases where people just want the graph structure. The new default would be to have no node data in the quotient.
* create a new function, `contract_edges` which allows multiple edges to be contracted in one batch. This will be more efficient than `quotient_graph` in the cases I [described above](https://github.com/networkx/networkx/issues/4935#issuecomment-1484083777).


aaronzo:
I've run some benchmarks that are effectively before/after this PR**


Here's my testing snippet I ran in a notebook:
```python
import random
import time
from functools import partial
from itertools import product

import networkx as nx

random.seed()

def edge_relation(G, b, c):
    return any(v in G[u] for u, v in product(b, c))

def random_partition(G: nx.Graph):
    num_dividers = random.randrange(0, n)
    nodes = list(G.nodes)
    nodes.extend("|" for _ in range(num_dividers))
    random.shuffle(nodes)
    partition = []
    for symbol in nodes:
        if symbol == "|":
            yield partition
            partition = []
        else:
            partition.append(symbol)


def time_new_quotient(G, partition):
    t0 = time.time()
    # note that in new code, relabel=True avoids a call to nx.relabel at the end
    nx.quotient_graph(G, partition, self_loops=True, relabel=True)
    t1 = time.time()
    return t1 - t0

def time_old_quotient(G, partition):
    t0 = time.time()
     # note that in old code, relabel=False avoids a call to nx.relabel at the end
    nx.quotient_graph(G, partition, edge_relation=partial(edge_relation, G), 
                      self_loops=True, relabel=False)
    t1 = time.time()
    return t1 - t0

n_vals = []
new_times = []
old_times = []
for n in range(50, 501, 50):
    print(f"calculating stats for n={n}")
    n_vals.append(n)

    G = nx.erdos_renyi_graph(n, p=0.1)
    
    avg_new, avg_old = [], []
    for _ in range(10):
        partition = list(random_partition(G))
    
        avg_new.append(time_new_quotient(G, partition))
        avg_old.append(time_old_quotient(G, partition))

    new_times.append(sum(avg_new) / 10)
    old_times.append(sum(avg_old) / 10)


plt.plot(n_vals, old_times, 'r')
plt.plot(n_vals, new_times, 'g')
plt.show()
```

This produced the plot:
image
(where green is the new code, red is old). This looks very encouraging! I'd welcome more graphs to benchmark,

** the old code has been refactored, but is the same speed or faster than before

I could probably optimize this more by having the reduce function sum attributes in cases they are numeric (potentially the parameter could accept a string attribute name to sum instead of a function, optionally), and having the attribute dict modified in-place rather than accessed twice on line 426.

Currently, in the non-multigraph case, the output edge attributes will always have a {"weight": 0}, even if the user doesn't request this - will need to think of a way around that, most probably will need an extra parameter for `default_edge_attr` to supply in conjunction with the `edge_reduce` function.

aaronzo:
@dschult I think I've got this as far as I can without some direction from core maintainers. The new quotient graph is pretty much as fast as was in #951, but more general. A review would be great :)

* Not sure what to do about the `relabel` parameter. The new version actually relabels as it goes along, so it's more efficient if `relabel=True` (if you want the node labels to be sets, it actually has to do a relabel). I would support removing this parameter completely - if users want the node sets, they can pass `lambda x: {"nodes": x}` to `node_data` (or this could be an option). If we remove the parameter, I don't need to calculate `partition2nodes` which means I can read nodes into the quotient in a single loop here.

* I think there's probably a case for completely removing the old function - its only use is to create edges on the partition that didn't exist before - but this could be done easily enough in user code too, after using `equivalence_classes`

aaronzo:
That's fine, thanks for letting me know - happy to wait for a review, but good to know whether there are major blockers to reviewing. Good to know this is of interest.

I take your points on complexity - I'll think about how I can make this easier to review. Probably I'll add a new description. I think separating this into 2 PRs would be hard with this kind of open source project, the API is very coupled to the efficiency.

Nevertheless, I'll see what I can do to signpost this :) it's not my day job either after all :) Enjoy the semester!

aaronzo:
@dschult any chance of progressing this PR forward on the talking points? It's been open for a few months now and recent commits on main have caused a headache to rebase the unit tests. I'm happy to fix that, but would be good to get a review in the meanwhile. I think this PR is quite impactful, drastically improving the implementation of quotient graph and fixing 2 open issues with the function raised by users.

PR #6558, opened 2023-03-26
Fix all_node_cuts output for complete graphs

### LINKED ISSUE #6533: all_node_cuts returns incorrect cuts
The function `all_node_cuts` returns node cut-sets for the K2 (complete graph with two nodes), but there exist no.

### Current Behavior

Consider the following graph, consisting of two nodes connected by an edge:
```python
graph = nx.Graph()
graph.add_edge(1,2)
list(all_node_cuts(graph)
>>> [{1}, {2}]
```
This graph has no node cut-sets because removing any vertex does not increase the number of connected components

### Expected Behavior

Return no node cut-sets:
```python
graph = nx.Graph()
graph.add_edge(1,2)
list(all_node_cuts(graph)
>>> []
```

### Steps to Reproduce

The above example

### Environment

Python version: 3.10.10
NetworkX version: 2.6.3

navyagarwal:
Closes #6533 

I modified the special condition for complete graphs in `all_node_cuts` to return an empty generator and also changed the test cases accordingly.

### DIFF
--- a/networkx/algorithms/connectivity/kcutsets.py
+++ b/networkx/algorithms/connectivity/kcutsets.py
@@ -92,10 +92,11 @@ def all_node_cuts(G, k=None, flow_func=None):
 
     # Address some corner cases first.
     # For complete Graphs
+
     if nx.density(G) == 1:
-        for cut_set in combinations(G, len(G) - 1):
-            yield set(cut_set)
+        yield from ()
         return
+
     # Initialize data structures.
     # Keep track of the cuts already computed so we do not repeat them.
     seen = []
--- a/networkx/algorithms/connectivity/kcutsets.py
+++ b/networkx/algorithms/connectivity/kcutsets.py
@@ -130,7 +130,7 @@ def all_node_cuts(G, k=None, flow_func=None):
     for x in X:
         # step 3: Compute local connectivity flow of x with all other
         # non adjacent nodes in G
-        non_adjacent = set(G) - X - set(G[x])
+        non_adjacent = set(G) - {x} - set(G[x])
         for v in non_adjacent:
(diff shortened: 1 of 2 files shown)


dschult:
Great!  This should be merged after v3.2 is released and included in the release (along with removal of the deprecation warnings and `conftest.py` entires) of v3.3.

MridulS:
@navyagarwal if you still have the bandwidth could you add the deprecation warnings so we can push this as part of v3.3. Thanks!

dschult:
Based on my comments in #6533 I think there is more needed here.
And I'm not seeing now what would be deprecated.
The main problem is:
```python
G=nx.complete_graph(5)
G.remove_edges_from([(0, 1), (3, 4)])
print(list(nx.all_node_cuts(G)))  # returns [{0, 1, 2}]
```
returns `[{0, 1, 2}]` when it should also include `{2, 3, 4}`.
Similar looking case works correctly:
```python
G = nx.complete_graph(5)
G.remove_edges_from([(1, 2), (0, 4)])
print(list(nx.all_node_cuts(G))) # returns [{1, 2, 3}, {0, 3, 4}]
```

dschult:
Yikes -- that took a bit. But I think I found an error in `nx.all_node_cuts`. Luckily it wasn't in the heart of the algorithm, but in the setup for which nodes get considered. A change from removing the potential node_cut set `X` to removing the currently considered node in that set `{x}` did the trick. I was following along with the article and it says "if x_i != v_j and v_j is not adjacent to x_i". It does not say that v_j should not be in the set X.

Anyway, it now works! With the fix already here for complete graphs, we have solved all the issues raised in #6533.

Ready for Review

MridulS:
thanks @dschult @navyagarwal !

PR #6560, opened 2023-03-27
Add example for cycle detection

PurviChaurasia:
Added an example script to explain and highlight cycles in a graph using the `find_cycle` function.


rossbar (approved):
Thanks @PurviChaurasia , I took the liberty of making some minor tweaks, but this example LGTM!

examples/algorithms/plot_cycle_detection.py
```diff
@@ -0,0 +1,28 @@
...
+# Draw the graph
+pos = nx.spring_layout(G, seed=8020)
+nx.draw(G, pos, with_labels=True)
+
+# The `orientation` parameter can be used to determine how directed edges are
+# treated and the reporting of edge direction in the cycle
+print(nx.find_cycle(G, orientation="original"))
+
+# Highlight the cycle in red
+nx.draw_networkx_edges(G, pos, edgelist=nx.find_cycle(G), edge_color="r", width=2)
```
dschult, line 27: Should we save the result of the `nx.find_cycle` call so we don't have to find it again for drawing?
```suggestion
cycle = nx.find_cycle(G, orientation="original")
print(cycle)

# Highlight the cycle in red
nx.draw_networkx_edges(G, pos, edgelist=cycle, edge_color="r", width=2)
```

dschult (commented):
I reopened this PR because I was about to approve it as well.
I'm thinking that it might be confusing to call nx.find_cycles twice because the cycle returned may be different. See suggestion below.

PR #6567, opened 2023-03-28
Deprecate shortest_path functions to have consistent return values in v3.3

dschult:
See #6527 for a discussion of the changes needed for making the return values consistent. 
This deprecates the function to be changed.
`single_target_shortest_path_length`
and
`shortest_path`  in the case that `source is None and target is None`


dgrigonis:
Just a note.

Deprecation warning could be improved with "will return an iterator of pairs". Had to browse issues for a while to make sure I know what sort of iterator it is referring to.

dschult:
Or maybe better `will return an iterator of (node, path) tuples.`   Thanks!!

At this point the deprecation cycle is finishing up, so it might be better to remove the deprecation notices and update the code. The better wording can go into the release notes (via the title of the PR?)

PR #6569, opened 2023-03-28
Update team galleries

jarrodmillman:


MridulS (approved):

dschult:
Looks good.  At some point we should update the "funding" section of the About Us to include CZI multiple times. 
It doesn't have to be in this PR though.

dschult (approved):

rossbar (approved):
Thanks @jarrodmillman . At some point we could check on the `team-member-name` class to see if we can't improve the linebreaking for the longer last names - let's leave for a follow-up though!

I approve the PR but will hold off from merging in case anyone else wants to weigh in!

PR #6573, opened 2023-03-29
Docstring Examples for Dense.py

vanshika230:
Added Docstring Example for 
1. floyd_warshall(G, weight='weight')
2. floyd_warshall_numpy(G, nodelist=None, weight='weight')

These both are built on the graph given in the example of floyd_warshall_predecessor_and_distance(G, weight='weight') [here](https://networkx.org/documentation/latest/reference/algorithms/generated/networkx.algorithms.shortest_paths.dense.floyd_warshall_predecessor_and_distance.html#networkx.algorithms.shortest_paths.dense.floyd_warshall_predecessor_and_distance) so that it is easier to follow one graph throughout all the different functions of Floyd Warshall Algorithm.


rossbar (commented):
+1 for adding examples here!

I have the general sense that the docstring examples may benefit from being made simpler. IMO it's often hard to visualize the graph from a given edgelist, especially as the edgelist gets long (and includes weights!).

For example, something like:

```python
>>> G = nx.DiGraph([(0, 1), (1, 2), (2, 3), (1, 3)])
>>> nx.floyd_warshall_numpy(G)
array([[ 0.,  1.,  2.,  2.],
       [inf,  0.,  1.,  1.],
       [inf, inf,  0.,  1.],
       [inf, inf, inf,  0.]])
```

has the advantage that a) it's easier to visualize b) not all the nodes are reachable from the others (demonstrating that an `inf` is returned in this case). You can then use the simple example as a building block to demonstrate slightly more complex variations, e.g. adding weights:

```python
>>> G[1][3]["weight"] = 10
>>> nx.floyd_warshall_numpy(G)
array([[ 0.,  1.,  2.,  3.],
       [inf,  0.,  1.,  2.],
       [inf, inf,  0.,  1.],
       [inf, inf, inf,  0.]])
```

or adding additional edges that modify the reachability

```python
>>> G.add_edge(3, 2)
>>> nx.floyd_warshall_numpy(G)
array([[ 0.,  1.,  2.,  3.],
       [inf,  0.,  1.,  2.],
       [inf, inf,  0.,  1.],
       [inf, inf,  1.,  0.]])
```

I think simplicity is a nice goal to have for docstring examples! These are all just suggestions though, not blockers.

vanshika230:
> +1 for adding examples here!
> 
> I have the general sense that the docstring examples may benefit from being made simpler. IMO it's often hard to visualize the graph from a given edgelist, especially as the edgelist gets long (and includes weights!).
> 
> For example, something like:
> 
> ```python
> >>> G = nx.DiGraph([(0, 1), (1, 2), (2, 3), (1, 3)])
> >>> nx.floyd_warshall_numpy(G)
> array([[ 0.,  1.,  2.,  2.],
>        [inf,  0.,  1.,  1.],
>        [inf, inf,  0.,  1.],
>        [inf, inf, inf,  0.]])
> ```
> 
> has the advantage that a) it's easier to visualize b) not all the nodes are reachable from the others (demonstrating that an `inf` is returned in this case). You can then use the simple example as a building block to demonstrate slightly more complex variations, e.g. adding weights:
> 
> ```python
> >>> G[1][3]["weight"] = 10
> >>> nx.floyd_warshall_numpy(G)
> array([[ 0.,  1.,  2.,  3.],
>        [inf,  0.,  1.,  2.],
>        [inf, inf,  0.,  1.],
>        [inf, inf, inf,  0.]])
> ```
> 
> or adding additional edges that modify the reachability
> 
> ```python
> >>> G.add_edge(3, 2)
> >>> nx.floyd_warshall_numpy(G)
> array([[ 0.,  1.,  2.,  3.],
>        [inf,  0.,  1.,  2.],
>        [inf, inf,  0.,  1.],
>        [inf, inf,  1.,  0.]])
> ```
> 
> I think simplicity is a nice goal to have for docstring examples! These are all just suggestions though, not blockers.

Can we add examples in this serial manner too? From what I observed, I thought we could only add up to 1 example. I was thinking more along this line, of combining the above 3 examples into 1 like below:- 
    >>> G = nx.DiGraph()
    >>> G.add_weighted_edges_from([(0, 1, 5), (1, 2, 2), (2, 3, 12), (1, 3, 10),( 3, 2, 8)])
    >>> nx.floyd_warshall_numpy(G)
array([[ 0.,  5.,  7., 15.],
         [inf,  0.,  2., 10.],
         [inf, inf,  0., 12.],
         [inf, inf,  8.,  0.]])
Also I totally understand this point, will try to keep the graph as simple as possible for easier visualization.

dschult:
The examples in the doc_strings can indeed be separated by text... 
They just need a blank line after any output (so the testing software can tell when to stop comparing).
I'm not sure that they need a blank line before the example too, but it often looks nicer that way.

vanshika230:
uh oh! little mishap with pushing commits. Creating a new pull request for this.

PR #6579, opened 2023-03-29
added doc examples for matching.py

Qudirah:
fixes for #6575


rossbar (changes requested):
The comments from the related PR reviews also apply here!

PR #6580, opened 2023-03-29
added examples for maxcut.py

Qudirah:
fixes for #6575


MasterC4t:
Only if only.

rossbar (changes requested):
The comments from the related PR reviews also apply here!

Note that the CI failures are unrelated.

PR #6584, opened 2023-03-30
Changed return types of shortest path methods to improve consistency

navyagarwal:
Fixes: #6527 
Deprecation warnings for this added in #6567   This should be merged in v3.3

Changed the output of `single_target_shortest_path_length` to dictionary, and of `shortest_path` to iterator. The test cases have also been modified accordingly.

This makes the shortest path methods more consistent.

### DIFF
--- a/networkx/algorithms/centrality/reaching.py
+++ b/networkx/algorithms/centrality/reaching.py
@@ -103,9 +103,9 @@ def global_reaching_centrality(G, weight=None, normalized=True):
         def as_distance(u, v, d):
             return total_weight / d.get(weight, 1)
 
-        shortest_paths = nx.shortest_path(G, weight=as_distance)
+        shortest_paths = dict(nx.shortest_path(G, weight=as_distance))
     else:
-        shortest_paths = nx.shortest_path(G)
+        shortest_paths = dict(nx.shortest_path(G))
 
     centrality = local_reaching_centrality
     # TODO This can be trivially parallelized.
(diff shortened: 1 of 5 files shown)


networkx/algorithms/centrality/reaching.py
```diff
@@ -103,9 +103,9 @@ def global_reaching_centrality(G, weight=None, normalized=True):
         def as_distance(u, v, d):
             return total_weight / d.get(weight, 1)
 
-        shortest_paths = nx.shortest_path(G, weight=as_distance)
+        shortest_paths = dict(nx.shortest_path(G, weight=as_distance))
     else:
-        shortest_paths = nx.shortest_path(G)
+        shortest_paths = dict(nx.shortest_path(G))
```
dschult, line 109: I don't think these cases should be turned into dicts. Instead, line 114 should be changed from using `.items()` to iterating over the variable `shortest_paths`.

dschult (commented):
This PR will be merged for v3.3.  That is, it will be merged after v3.2 s release in September (or so).
All the code/test changes for that fix are included here so far.

Still needed are:
  [ ] removal of the deprecation warnings that were put in place in #6567 
  [ ] reverting deprecation changes to conftest.py and deprecations.rst from #6567
  [ ] removal of any comments referring to the deprecation like line 146 of `unweighted.py` (also from #6567)
  [ ] changes to the `Returns` section of the doc_strings so they match these code changes.  (I think that is only `single_target_shortest_path_length`.

dschult (commented):
We have released v3.2 so I think this is ready to be merged.

This doesn't take out the deprecation warnings. Should we do that here, or in a different PR?

navyagarwal:
I think I just have to revert the changes made in #6567. Since multiple files are involved, I'll do it in a separate PR to ease review.

Issue #6588, opened 2023-03-31
Improve test coverage for algorithms in assortativity 


Qudirah:
The neighbor_degree algorithm needs its coverage of 100% from https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/assortativity/neighbor_degree.py
and also correlation.py has some yellow lines in https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/assortativity/correlation.py. I will link a PR soon solving both.

PR #6590, opened 2023-03-31
Added tests to test_correlation.py

Qudirah:
fixes for #6588 @MridulS Have a look! I am not so certain I did the right thing.

### DIFF
--- a/networkx/algorithms/assortativity/tests/test_correlation.py
+++ b/networkx/algorithms/assortativity/tests/test_correlation.py
@@ -48,7 +48,7 @@ class TestDegreeMixingCorrelation(BaseTestDegreeMixing):
         np.testing.assert_almost_equal(r, -1.0 / 7.0, decimal=4)
 
     def test_degree_assortativity_weighted(self):
-        r = nx.degree_assortativity_coefficient(self.W, weight="weight")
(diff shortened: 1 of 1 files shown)


MasterC4t:
Good?

MasterC4t:
Nah

MasterC4t:
https://github.com/networkx/networkx/pull/6580

rossbar (commented):
I think `nodes=None` is the default, so if I'm understanding correctly these changes don't have any practical effect.

Qudirah:
> I think `nodes=None` is the default, so if I'm understanding correctly these changes don't have any practical effect.

The if node = None line has a yellow line though. So I thought I have to explicitly test for them. What do you suggest I do instead?

navyagarwal:
@Qudirah `if nodes is None` is marked yellow because we have to check for the condition when Nodes is NOT equal to None. Add a testcase where a container of nodes is provided to the method.

Qudirah:
> @Qudirah `if nodes is None` is marked yellow because we have to check for the condition when Nodes is NOT equal to None. Add a testcase where a container of nodes is provided to the method.

Thank you Navya <3

MridulS (approved):
Thanks!

dschult (approved):

Issue #6604, opened 2023-04-02
Improve test coverage for equitable_coloring.py


Qudirah:
I want to get rid of red lines in https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/coloring/equitable_coloring.py. I will link a PR soon.

PR #6605, opened 2023-04-02
Added test cases to equitable_coloring.py

Qudirah:
fixes for #6604

### DIFF
--- a/networkx/algorithms/coloring/tests/test_coloring.py
+++ b/networkx/algorithms/coloring/tests/test_coloring.py
@@ -119,7 +119,15 @@ class TestColoring:
     def test_num_colors(self):
         G = nx.Graph()
         G.add_edges_from([(0, 1), (0, 2), (0, 3)])
+        coloring = {0: 0, 1: 1, 2: 2, 3: 3}
+        assert is_equitable(G, coloring, num_colors=5)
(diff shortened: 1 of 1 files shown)


networkx/algorithms/coloring/tests/test_coloring.py
```diff
@@ -119,8 +119,16 @@ def test_is_equitable(self):
...
         G.add_edges_from([(0, 1), (0, 2), (0, 3)])
+        coloring = {0: 0, 1: 1, 2: 2, 3: 3}
+        assert is_equitable(G, coloring, num_colors=5)
+        assert is_equitable(G, coloring)
         pytest.raises(nx.NetworkXAlgorithmError, nx.coloring.equitable_color, G, 2)
 
+        G = nx.Graph()
+        G.add_edges_from([(0, 1), (0, 2), (0, 3), (0, 4)])
+        coloring1 = {0: 1, 1: 1, 2: 2, 3: 2, 4: 2}
+        assert not is_equitable(G, coloring1, num_colors=6)
```
MridulS, line 130: These tests should probably go in a new function as `test_num_colors` is being used to test ` nx.coloring.equitable_color` (maybe we should have a better name). But could you move these tests to a new function or maybe add them to `test_is_equitable`. Thanks!

Issue #6607, opened 2023-04-02
Improve test coverage for euler.py


Qudirah:
There are a few red lines in https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/euler.py. Will link a PR to get rid of them.

PR #6608, opened 2023-04-02
added tests to euler.py

Qudirah:
fixes for #6607

### DIFF
--- a/networkx/algorithms/tests/test_euler.py
+++ b/networkx/algorithms/tests/test_euler.py
@@ -239,11 +239,18 @@ class TestEulerianPath:
         with pytest.raises(nx.NetworkXError):
             list(nx.eulerian_path(G, source=1))
 
-    @pytest.mark.parametrize("graph_type", [nx.MultiGraph, nx.MultiDiGraph])
-    def test_eulerian_with_keys(self, graph_type):
-        G = graph_type([(1, 2), (2, 1), (2, 3), (3, 1)])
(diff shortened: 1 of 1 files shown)


networkx/algorithms/tests/test_euler.py
```diff
@@ -239,6 +239,32 @@ def test_eulerian_path_multigraph_undirected(self):
...
+        G.add_node(2)
+        G.add_node(3)
+        # Add edges with weights
+        G.add_edge(1, 2, weight=1)
+        G.add_edge(1, 2, weight=2)
+        G.add_edge(2, 3, weight=3)
+        G.add_edge(3, 1, weight=4)
+        answer = nx.eulerian_path(G, keys=True)
+        assert list(answer) == [(1, 2, 0), (2, 1, 1), (1, 3, 0), (3, 2, 0)]
+
```
MridulS, line 267: This should be enough to test the unit tests hit the keys argument for Multi(Di)Graph :) 

```suggestion
    @pytest.mark.parametrize("graph_type", [nx.MultiGraph, nx.MultiDiGraph])
    def test_eulerian_with_keys(self, graph_type):
        G = graph_type([(1, 2), (2, 1), (2, 3), (3, 1)])
        answer = nx.eulerian_path(G, keys=True)
        assert list(answer) == [(1, 2, 1), (2, 3, 0), (3, 1, 0), (1, 2, 0)]
```

PR #6610, opened 2023-04-02
Added tests for summarization.py

Qudirah:
fixes for #6609

### DIFF
--- a/networkx/algorithms/tests/test_summarization.py
+++ b/networkx/algorithms/tests/test_summarization.py
@@ -195,8 +195,8 @@ class TestUnDirectedDedensification:
 
     def test_dedensify_prefix_kwargs(self):
         G = self.build_original_graph()
-        _, mapping = nx.dedensify(graph, threshold=2, prefix='node_')
-        assert mapping == {'node_6BA', 'node_CBA'}
+        _, mapping = nx.dedensify(G, threshold=2, prefix="node_")
+        assert mapping == {"node_6BA", "node_CBA"}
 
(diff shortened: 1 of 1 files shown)


networkx/algorithms/tests/test_summarization.py
```diff
@@ -188,6 +188,18 @@ def build_compressed_graph(self):
...
+        with pytest.raises(nx.NetworkXError, match="The degree threshold must be >= 2"):
+            nx.dedensify(G, threshold=1)
+
+    def test_dedensify_prefix_kwargs(self):
+        G = nx.Graph()
+        G.add_edges_from([(1, 2), (1, 3), (1, 4), (2, 3), (3, 4)])
+        H, mapping = nx.dedensify(G, threshold=2, prefix="node_")
+        assert list(H.nodes()) == [1, 2, 3, 4]
+        assert mapping == set()
+
```
MridulS, line 202: In this test we should be testing something that actually ends up adding the prefix to the node in the graph.
This example returns an empty mapping.

The original graph would be a better example as you will be able to test that the prefix is actually added to the compressor nodes.
```suggestion
        G = self.build_original_graph()
        _, mapping = nx.dedensify(graph, threshold=2, prefix='node_')
        assert mapping == {'node_6BA', 'node_CBA'}

```

MridulS (approved):

rossbar (changes requested):
Looks like the test is still failing (even after correcting the misnamed input variable). Looks like this will need another cycle to figure out what's going on!

PR #6627, opened 2023-04-11
Katz index

Mjh9122:
This PR is a spiritual successor to #1163 which itself builds on #1117. I am looking to use the Katz Index in some upcoming projects and noticed that there is not an implementation for NetworkX. This is my first non-testing PR so I definitely need some feedback on documentation. Furthermore, I am not intimately familiar with using the index and do not know what a reasonable default value for beta would be. I opted to use a similar generator pattern to the existing functions, while not calling the _apply_prediction function so that the computation only has to be done one. LMK what I should fix.

### DIFF
--- a/networkx/algorithms/link_prediction.py
+++ b/networkx/algorithms/link_prediction.py
@@ -592,16 +592,57 @@ def within_inter_cluster(G, ebunch=None, delta=0.001, community="community"):
 @not_implemented_for("directed")
 @not_implemented_for("multigraph")
 def katz_index(G, ebunch=None, beta=.1):
-    def predict(u, v):
-        if not nx.has_path(G, u, v):
-            return 0
-        A = nx.adjacency_matrix(G)
-        AB = A.multiply(beta)
-        I = sparse.identity(AB.shape[0])
-        result = sparse.linalg.spsolve(I - AB, I) - I
-        return result[u, v]
+    r"""Compute the Katz index of all node pairs in ebunch.
 
-    return _apply_prediction(G, predict, ebunch)
+    Katz index of nodes `u` and `v` is defined as
+
+    .. math::
+
+        S^{Katz}_{u, v} \(I - \beta A\)^{-1} - I
+
+    where A is the adjacency matrix of G and I is the identity matrix.
+    $\beta$ Provides an exponential damping factor such that shorter paths are valued more.
+
+    Parameters
+    ----------
+    G : graph
+        A NetworkX undirected graph.
+
+    ebunch : iterable of node pairs, optional (default = None)
+        Katz Index will be computed for each pair of nodes
+        given in the iterable. The pairs must be given as 2-tuples
+        (u, v) where u and v are nodes in the graph. If ebunch is None
+        then all non-existent edges in the graph will be used.
+        Default value: None.
+
+    Returns
+    -------
+    piter : iterator
+        An iterator of 3-tuples in the form (u, v, p) where (u, v) is a
+        pair of nodes and p is their Katz Index
+
+    Examples
+    --------
+    >>> G = nx.complete_graph(5)
+    >>> preds = nx.katz_index(G, [(0, 1), (2, 3)])
+    >>> for u, v, p in preds:
+    ...     print(f"({u}, {v}) -> {p:.8f}")
+    (0, 1) -> 0.15151515
+    (2, 3) -> 0.15151515
+
+    References
+    ----------
+    .. [1] Linyuan Lu, Tao Zhou
+           Link Prediction in Complex Networks: A Survey.
+           https://arxiv.org/pdf/1010.0725v1.pdf
+    """
+    if ebunch is None:
+        ebunch = nx.non_edges(G)
+    A = nx.adjacency_matrix(G)
+    AB = A.multiply(beta)
+    I = sparse.identity(AB.shape[0])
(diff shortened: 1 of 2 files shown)


networkx/algorithms/link_prediction.py
```diff
@@ -2,9 +2,11 @@
 Link prediction algorithms.
 """
 
-
 from math import log
 
+import scipy as sp
+from scipy import sparse
+
```
dschult, line 9: You should put the scipy import statements inside the functions that need it, so the module can still be loaded even if scipy is not installed. 

For the tests, you should use the `pytest.importorskip` function (pronounced import-or-skip) so that the test module gets skipped (if all functions tested require numpy), or so individual tests get skipped when the needed package is not installed.  Something like:
```python
```

Mjh9122: Thanks, just made those changes

jarrodmillman:
May want to cross-reference `katz_centrality` and `katz_centrality_numpy` from `networkx/algorithms/centrality/katz.py`. That file needs some work. It might be worth adding a scipy version for katz centrality and making it the default. That could be a follow-up PR (if you are interested).

Mjh9122:
> You may want to cross-reference `katz_centrality` and `katz_centrality_numpy` from `networkx/algorithms/centrality/katz.py`. That file needs some work. It might be worth adding a scipy version for katz centrality and making it the default. That could be a follow-up PR (if you are interested).

Sounds good ill get into it now

Mjh9122:
> 

@jarrodmillman Looking at the Katz file is there any situation where the iterative version would be preferable? It seems like the linear algebra way would be faster and more accurate. Also other than not wasting the time switching to and preforming operations on dense matrices, are there other changes you would like to see in a scipy version? I've put my version up in PR #6642

networkx/algorithms/link_prediction.py
```diff
@@ -592,16 +592,57 @@ def predict(u, v):
...
+    G : graph
+        A NetworkX undirected graph.
+
+    ebunch : iterable of node pairs, optional (default = None)
+        Katz Index will be computed for each pair of nodes
+        given in the iterable. The pairs must be given as 2-tuples
+        (u, v) where u and v are nodes in the graph. If ebunch is None
+        then all non-existent edges in the graph will be used.
+        Default value: None.
+
```
SimonPop, line 616: Maybe add a tick for the `beta` parameter. You could insist on the comment made in the [paper ](https://arxiv.org/pdf/1010.0725v1.pdf) that _"β must be lower than the reciprocal of the largest eigenvalue of
matrix A to ensure the convergence of the equation"_

SimonPop (commented):
Hi! 
I was interested in your work and wanted to help review it.

dschult (commented):
This looks good.  Thanks!

I notice that you placed the `importorskip` for `numpy` at the module level. But that will skip all the other tests in `test_link_prediction.py` even though they don't use numpy.  Please move the `importorskip` to inside the `setup` function within the test class you create. That way we only skip the ones we need to skip.

You should add a parameter description for the `beta` input parameter in the doc_string.

That's all I've got... looking good.

Mjh9122:
> This looks good. Thanks!
> 
> I notice that you placed the `importorskip` for `numpy` at the module level. But that will skip all the other tests in `test_link_prediction.py` even though they don't use numpy. Please move the `importorskip` to inside the `setup` function within the test class you create. That way we only skip the ones we need to skip.
> 
> You should add a parameter description for the `beta` input parameter in the doc_string.
> 
> That's all I've got... looking good.

Those changes are done. I followed the precedent set in test_cluster.py and added np as a global variable. If there is another, preferred way to do this, let me know. Thanks.

Mjh9122:
@dschult could you help me figure out what is failing with the PR. I don't understand

MridulS:
The failure is due to sphinx 7. This PR was created before https://github.com/networkx/networkx/pull/6680 so it doesn't have the latest main branch doc building requirements. Could you rebase to main and push again? That should fix it.

dschult:
I think the documentation test is failing because the circle-CI configuration changed. You'll need to merge with or rebase onto the main branch of the networkx repo.

dschult:
I think this is ready for you to pull down -- we rebased on main, connected the new function to the doc system, added minor tweaks to the sparse stuff to remove warnings, made `beta` a keyword-only argument, changed the tests to use a 2-d array instead of an array of arrays.

To pull it down to your local repo, there will likely be a long mess of merge conflicts. Instead, you can remove or rename your local branch `katz_index` and then create it again from origin: `git fetch origin; git ch katz_index`.  I hope that works.  

Then take a look at the changes and see if this all makes sense.

Mjh9122:
> I think this is ready for you to pull down -- we rebased on main, connected the new function to the doc system, added minor tweaks to the sparse stuff to remove warnings, made `beta` a keyword-only argument, changed the tests to use a 2-d array instead of an array of arrays.
> 
> To pull it down to your local repo, there will likely be a long mess of merge conflicts. Instead, you can remove or rename your local branch `katz_index` and then create it again from origin: `git fetch origin; git ch katz_index`. I hope that works.
> 
> Then take a look at the changes and see if this all makes sense.

Looks all good to me

eriknw:
It looks like this is very close to going in. Does anything else need done besides merge with the latest main branch?

PR #6628, opened 2023-04-11
Allow multiple graphs for `@nx._dispatch`

jim22k:
A new `graphs` keyword is added. For the case of two graphs named `G` and `H` as the first to arguments in `foo`, the new spelling is 
```python
@nx._dispatch(graphs="G,H")
def foo(G, H, other_arg, **kwargs):
    ...
```

### DIFF
--- a/networkx/classes/backends.py
+++ b/networkx/classes/backends.py
@@ -107,44 +107,78 @@ def _register_algo(name, wrapped_func):
     wrapped_func.dispatchname = name
 
 
-def _dispatch(func=None, *, name=None):
+def _dispatch(func=None, *, name=None, graphs=None):
     """Dispatches to a backend algorithm
     when the first argument is a backend graph-like object.
+
+    The algorithm name is assumed to be the name of the wrapped function unless
+    `name` is provided. This is useful to avoid name conflicts, as all
+    dispatched algorithms live in a single namespace.
+
+    If more than one graph is required for the algorithm, provide a comma-separated
+    string of variable names as `graphs`. These must be the same order and name
+    as the variables passed to the algorithm. Dispatching does not support graphs
+    which are not the first argument(s) to an algorithm.
     """
     # Allow any of the following decorator forms:
     #  - @_dispatch
     #  - @_dispatch()
-    #  - @_dispatch("override_name")
     #  - @_dispatch(name="override_name")
+    #  - @_dispatch(graphs="G,H")
+    #  - @_dispatch(name="override_name", graphs="G,H")
     if func is None:
-        if name is None:
+        if name is None and graphs is None:
             return _dispatch
-        return functools.partial(_dispatch, name=name)
+        return functools.partial(_dispatch, name=name, graphs=graphs)
     if isinstance(func, str):
-        return functools.partial(_dispatch, name=func)
+        raise TypeError("'name' and 'graphs' must be passed by keyword") from None
     # If name not provided, use the name of the function
     if name is None:
         name = func.__name__
+    # If graphs not provided, use the most common signature of a single graph named `G`
+    if graphs is None:
+        graphs = "G"
+
+    graph_list = [g.strip() for g in graphs.split(",")]
+    if len(graph_list) <= 0:
+        raise KeyError("'graphs' must contain at least one variable name") from None
 
     @functools.wraps(func)
     def wrapper(*args, **kwds):
-        if args:
-            graph = args[0]
-        else:
-            try:
-                graph = kwds["G"]
-            except KeyError:
-                raise TypeError(f"{name}() missing positional argument: 'G'") from None
-        if hasattr(graph, "__networkx_plugin__") and plugins:
-            plugin_name = graph.__networkx_plugin__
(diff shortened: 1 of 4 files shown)


jim22k:
@MridulS please take a look and let me know what you think.

networkx/classes/backends.py
```diff
@@ -107,44 +107,78 @@ def _register_algo(name, wrapped_func):
...
-            if plugin_name in plugins:
-                backend = plugins[plugin_name].load()
-                if hasattr(backend, name):
-                    return getattr(backend, name).__call__(*args, **kwds)
-                else:
-                    raise NetworkXNotImplemented(
-                        f"'{name}' not implemented by {plugin_name}"
-                    )
+        # Select overlap of args and graph_list
+        graphs_resolved = dict(zip(graph_list, args))
```
MridulS, line 149: This will go through even if there are less numbers of graphs in graph_list string than the function was expecting. Not sure if we should be catching this or just document it.
``` python
@nx._dispatch(graphs="G")
def something(G, H):
....
....
```

MridulS (approved):
Thanks @jim22k! Looks good to me.

networkx/classes/backends.py
jim22k: I assume anyone decorating functions in NetworkX code understands which arguments are graphs. I don't want a call-time check for such a mistake, and I don't know how we would perform this check at function wrap time.

networkx/classes/backends.py
```diff
@@ -107,44 +107,78 @@ def _register_algo(name, wrapped_func):
...
+        return functools.partial(_dispatch, name=name, graphs=graphs)
     if isinstance(func, str):
-        return functools.partial(_dispatch, name=func)
+        raise TypeError("'name' and 'graphs' must be passed by keyword") from None
     # If name not provided, use the name of the function
     if name is None:
         name = func.__name__
+    # If graphs not provided, use the most common signature of a single graph named `G`
+    if graphs is None:
+        graphs = "G"
```
dschult, line 140: Would the default value be more clear as `graphs='G'`?  The use of `None` as an indicator of the default value is helpful for mutable objects (so all function calls don't use the same default object). But here the decorator signature to be more clear with the default value stated explicitly.  And we can eliminate any use with `_dispatch(graphs="G")`.  I don't feel strongly -- just a suggestion.   (and maybe I am missing something because above you simply return `_dispatch` which presumably has `graphs=None`...

jim22k: Good suggestion. I've updated the code to use the simpler string default.

dschult:
I'm actually trying out the dispatch testing on this branch locally -- haven't done any dispatch testing locally before. :}

I have a question about how to  turn on the dispatching. The [docs on Graph Types](https://networkx.org/documentation/stable/reference/classes/index.html#backends) says I should "register an entry point" which seems to point to the networkx_plugins.  How do I register an entry point? I can create the entry points dict. But what do I do with that to make it an entry_point?  Hopefully I'm just missing some terminology or step in the process. 

And is there a way I should be running the tests locally even just with the basic backend of a networkx graph. The tests pass on the CI on github, but when I run them locally it complains that I don't have any networkx.plugin.
```
networkx/classes/backends.py:84: in items
    self._items = entry_points()["networkx.plugins"]
E   KeyError: 'networkx.plugins'
```

Thanks!

jim22k:
@dschult The entry_points mechanism is described [here](https://setuptools.pypa.io/en/latest/userguide/entry_point.html). Essentially it is an install-time entry created in site-packages that allows Python to view metadata about your package without actually importing your package (which could be expensive).

If you have never played with entry_points, try this out:
```python
from importlib.metadata import entry_points
entry_points()
```
You should see a dict-like object whose values are `EntryPoint` objects.

As for how to register an entry_point, look at [line 155 in NetworkX's setup.py](https://github.com/networkx/networkx/blob/main/setup.py#L155). Because it is an install-time piece of data, it has to be described to the build machinery, not something you do at run time (or import time).

If for some reason you don't see "networkx.plugins" in your entry_points dictionary, it means you don't have NetworkX installed or you have the code locally but haven't done `python setup.py develop` or `pip install -e .` on your local repo. Either of those will add the local repo to your environment in "development" mode and therefore trigger the entry_points registry. If you set up your repo in development mode previously, you might need to run it again for my changes to be "installed".

dschult:
Thanks  @jim22k for that info!

I think this is ready -- I have one question: Could all the `@_dispatch(stuff)` calls in cluster.py become `@_dispatch`?
I think the default values cover all graph and name arguments.

jim22k:
> Could all the @_dispatch(stuff) calls in cluster.py become @_dispatch?

@dschult They _could_, but we wanted to have some example usage of non-defaults in the code for purposes of coverage.
Would it make sense to add a comment where we manually specify default values to test coverage?

dschult (approved):
Got it -- and: one with just name, one with just graphs and one with both. So it covers all those cases.

I think this is fine. I'm just catching up to the rest of you.
:)

PR #6630, opened 2023-04-12
adding a formula that ignores self-loops at the each level of directed louvain algorithm

anthonimes:
I recently came across a small bug when running `louvain` on directed graphs: the algorithm always stopped after a first step and hence did not compute the hierarchical structure. 

This seems to come from the fact that self-loops are considered at the initialization step. It follows that every vertex is added back to its own community and no improvement is made, causing the `while` to stop. 

The proposed fix has been tested on two separate graphs and seems correct (a comparison has been made with the [C++ implementation](https://github.com/nicolasdugue/DirectedLouvain) of the algorithm).


dschult:
I think this might be the fix for #6412 
Can you see what is written there and comment on your perspective?  Is this the same question/issue?

And, can you add a test to ensure this correction isn't removed in a later change?  The tests are in the `networkx/algorithms/community/tests` folder. Ask if you have questions... Or push what you think might be ok and we'll give suggestions.

Thanks very much!!

anthonimes:
Sorry for the very late answer, this indeed seems to be the fix for the mentioned issue! 
I will try to add a test to ensure that the correction remains in later versions, I will let you know once this is done. 

Thanks!

dschult:
I've added a test for whether selfloops in the input graph affect the resulting partition. The fix does not get rid of all impact of self-loops. It corrects the code to make the treatment of self-loops the same for directed and undirected graphs (and corrects it to the published algorithms). But there still is a small amount of impact. The original self-loops in the original graph are treated as pre-formed communities that have already been condensed and now have a self-loop edge. I've added a paragraph which explains this and suggests that people who have edge weights that don't represent pre-formed communities remove the self-loops before using these function.

As the tests show, when the self-loop edges have small weight their impact is minimal, but if the self-loop edge weights are large it changes the resulting partition quite a bit.

MridulS (approved):

PR #6637, opened 2023-04-13

vanshika230:
I have created a python script that visualizes greedy coloring in a graph. The visualization is as follows :- 
GreedyFigure

### DIFF
--- a/examples/algorithms/plot_greedy_coloring.py
+++ b/examples/algorithms/plot_greedy_coloring.py
@@ -1,0 +1,100 @@
+"""
+=====================
+Greedy Coloring Algorithm
+=====================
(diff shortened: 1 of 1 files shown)


dschult:
The test errors are due to a change in availability of the code coverage tools in our git actions testing quite.
They are not due to anything in this PR.  The PR #6635 will (hopefully) fix the testing problem when it gets merged. I'd suggest waiting until that fix is merged and then merge/pulling the resulting upstream networkx repo into your branch.
This will likely have to be done with every open PR, so if you try it please let us know how that works for you.

vanshika230:
I see the PR to fix the test problem has been merged, will try making the greedy coloring PR again.

PR #6639, opened 2023-04-14

vanshika230:
I've written a python script to visualize greedy coloring on the given graph. The visualization is as follows:-


vanshika230:
I created another visualization too given below, but the edge list was far too long to follow and I couldn't figure out a way to turn it into a formula or use some other networkx method. But if its a better visualization than the one I have created a PR for then I can change the graph, and the visualization part will remain the same. 
GreedyFigure

PR #6642, opened 2023-04-15
Katz centrality scipy

Mjh9122:
Added scipy version of the katz centrality algorithm. Followed lead of other implementations already present, including not checking if $\alpha \leq \frac{1}{\lambda}$ where $\lambda$  is the largest eigenvalue of the adjacency matrix of G. It should be noted that the closed form sum all of the implementations use is not valid if this condition is not met, but no error or warning is raised if it isn't. Should I add this to the existing implementations and my own? Also @jarrodmillman suggested in #6627 to make the new scipy version the default. I am not sure how to accomplish this. Open to suggestions on this. LMK what you guys think. Also there were a couple of typos I fixed in the testing file where the incorrect version of the centrality algo was being tested.


eriknw:
This appears to be a continuation of #4849

PR #6643, opened 2023-04-16
Use the correct namespace for girvan_newman examples

MridulS:
Users won't be able to run the example code just by doing copy and paste with the current examples. And the first guess of doing `nx.girvan_newman` will also be wrong. We should try to be explicit about the namespaces (something I got from the user survey).


dschult (approved):
Definitely!

Also, we should go through the list of which sub-packages are *not* in the main namespace. Some of them make sense to separate. But I think community should be re-factored to the main namespace.

In any case, this PR is good to go.

rossbar (approved):
Thanks @MridulS !

PR #6644, opened 2023-04-16
Preserve node order in bipartite_layout

thirtiseven:
This PR changes the code in `bipartite_layout` to preserve the specified node order.
The original code converted the nodes to sets, which are unordered in Python. This resulted in the nodes being returned in any order. But for bipartite graphs, we often want to be able to specify the order of the nodes on each side, rather than letting the layout function do it arbitrarily.

For example, code:
```
G = nx.bipartite.gnmk_random_graph(3, 5, 10, seed=123)
G_nodes = [0,1,2,5,4,3,7,6]
top = [2,0,1]
pos = nx.bipartite_layout(G_nodes, top)
nx.draw(G, pos=pos, with_labels=True)
```
before this PR:
before
after:
after


dschult (approved):
Nice!

Thanks!

PR #6645, opened 2023-04-16
Improve Test Coverage for Semiconnected.py

vanshika230:
I have added a test case at [test_semiconnected.py](https://github.com/networkx/networkx/blob/main/networkx/algorithms/components/tests/test_semiconnected.py) to improve test coverage of [semiconnected.py](https://github.com/networkx/networkx/blob/main/networkx/algorithms/components/semiconnected.py).  The test coverage is now at a 100%.
Before :- 
before
After :- 
after

### DIFF
--- a/networkx/algorithms/components/tests/test_semiconnected.py
+++ b/networkx/algorithms/components/tests/test_semiconnected.py
@@ -53,3 +53,10 @@ class TestIsSemiconnected:
             chain.from_iterable([(i, i - 1), (i, i + 1)] for i in range(0, 100, 2))
         )
         assert not nx.is_semiconnected(G)
+
+    def test_topo_order(self):
+        G = nx.path_graph(100, create_using=nx.DiGraph())
+        topo_order = list(nx.topological_sort(G))
+        assert not nx.is_semiconnected(G, topo_order=topo_order)
+        rev_topo_order = list(reversed(topo_order))
+        assert nx.is_semiconnected(G, topo_order=rev_topo_order)


networkx/algorithms/components/tests/test_semiconnected.py
```diff
@@ -53,3 +53,10 @@ def test_alternating_path(self):
             chain.from_iterable([(i, i - 1), (i, i + 1)] for i in range(0, 100, 2))
         )
         assert not nx.is_semiconnected(G)
+
+    def test_topo_order(self):
+        G = nx.path_graph(100, create_using=nx.DiGraph())
```
dschult, line 58: Just a minor nit to keep an empty graph from being created. More style than impact.
```suggestion
        G = nx.path_graph(100, create_using=nx.DiGraph)
```

networkx/algorithms/components/tests/test_semiconnected.py
```diff
@@ -53,3 +53,10 @@ def test_alternating_path(self):
...
+        topo_order = list(nx.topological_sort(G))
+        assert not nx.is_semiconnected(G, topo_order=topo_order)
+        rev_topo_order = list(reversed(topo_order))
+        assert nx.is_semiconnected(G, topo_order=rev_topo_order)
```
rossbar, line 62: The behavior here strikes me as strange... isn't this backwards?:

```python
>>> G = nx.path_graph(100, create_using=nx.DiGraph)
>>> nx.is_semiconnected(G)  # When no ordering is specifi
True
>>> nx.is_semiconnected(G, topo_order=list(range(100)))
False
```

This feels like a defect to me; in fact - the only reason this works with the default `topo_order=None` is because `nx.condensation` remaps the nodes in reverse order (in this case where there are no strongly-connected compontents).

+1 for testing this but I just want to make sure we have the desired behavior nailed down!

vanshika230: Agreed it is indeed strange, I spent the majority of last day trying to make it work backwards but it kept showing NetworkX FAILED : Assertion False. So I switched up the assertion and it worked. I have tried to understand the working behind this as follows:- 
topo_order
A potential explanation I came up with to justify this  :- 
explanation
<b>P.S. This is just a desperate attempt to justify why the error disappeared tbh😅</b>I was going to try it with different graphs before opening an issue for this. I am glad you see it too.

rossbar: Thanks for the additional context/research @vanshika230 ! I agree with what you've summarized above based on my quick dive into these functions. I still think this behavior is at the very least fragile as it relies on the default remapping in `nx.condensation`.

FWIW this is basically the most useful part of the exercise in improving test coverage - often times you'll stumble on something that either has never been tried or just flat out doesn't make sense!

vanshika230: I see #6651 was created to deal with this strange behavior. I agree, I learned a lot while doing this and I am happy that this pr brought up this defect and led to changes for the better :) Onto next testing.  Shall I close this PR now? Apologies for MIA, bit under the weather.

MridulS: Yes, thanks a ton for working on this @vanshika230!

I'll go ahead and close this one.

rossbar: > I see https://github.com/networkx/networkx/pull/6651 was created to deal with this strange behavior. I agree, I learned a lot while doing this and I am happy that this pr brought up this defect and led to changes for the better :) Onto next testing. Shall I close this PR now? Apologies for MIA, bit under the weather.

Indeed - sorry for not updating the discussion, but this led to some really interesting discussion and review of behavior across multiple functions. This is probably one of the best examples I've come across to motivate why code coverage matters :) - thanks for opening @vanshika230 !

PR #6648, opened 2023-04-18
Rewrite NXEP 3

dschult:
1) keep create_using
2) allow edgelist generation via [~~nx.path_graph.edges_plus(9)~~] `nx.path_graph_generator(9)`


dschult:
Perhaps the name EdgesPlus is not a good choice. We are constructing a generator. So:
Should we rename `nx.path_graph.EdgesPlus(5)` to be `nx.path_graph.generator(5)`?
Or is that potentially conflicting with the python technical term generator?

I think leaving `nx.path_graph.edges(5)` should be clear enough... thought maybe `edges_only` is a better warning that the nodes don't get generated.

jarrodmillman (approved):

PR #6662, opened 2023-04-24
codespell: pre-commit, config, typos fixed

yarikoptic:
more about codespell: https://github.com/codespell-project/codespell/  

As you can see:
- networkx is a bit a tricky case requiring a good number of skips.  Some are legit names, but some are a bit odd choices of variables (e.g. childs, ans, etc) which come close to real words thus triggering codespell
- I didn't add a dedicated workflow since there is one for pre-commit so I expect it to trigger codespell  as well
- be careful to not list actual typo in the changelog ;)

hope this comes useful


rossbar (commented):
I'm -1 on adding `codespell` to the linting. I think it introduces an awful lot of configuration tweaking (and thus fragility) for not that big of a benefit (misspellings are annoying, but rarely break something).

+1 for fixing the identified misspellings, though I haven't looked over them closely.

yarikoptic:
Indeed, the number and types of skips I had to add to networkx is indeed a little high, but not really that many given the size and the age/maturity of the project!  In some older C codebases with more of cryptic shortened variable names and more citations of unique names I had simply to ignore all hits with length <= 3 to raise "above noise" and find actual typos ;)  Given that (size/age/nature of skips) -- it is actually not that many and I don't think it would be anyhow notably fragile as long as people do not use new "near-typo" names for the variables.  But it is up to you -- just let me know and I will strip away it away from pre-commit so typos could go in freely.  Meanwhile I will rewrite to fix the style in that yaml.

doc/release/release_2.6.rst
```diff
@@ -491,7 +491,7 @@ Merged PRs
 - Improve intersection function (#4588)
 - Pygraphviz choco (#4583)
 - Add prominent group algorithm (#4560)
-- Add partition_quality to compute coverage and performance  (coverage and perfor… (#4536)
+- Add partition_quality to compute coverage and performance  (coverage and perform… (#4536)
```
MridulS, line 494: Wellllll, technically not a typo as it just the text from the PR title https://github.com/networkx/networkx/pull/4536

MridulS (approved):
Let's definitely get the typos in!
I don't have any strong feelings about adding more configs. I'm -0 on adding the codespellrc config file, and as @rossbar mentioned having typos isn't the end of the world and sometimes it's even nice as we may get a new contributor who spots that and contributes to NetworkX :)

doc/release/release_2.6.rst
yarikoptic: ;-) indeed -- might catch some other similar ones too later on.

yarikoptic:
> ... and sometimes it's even nice as we may get a new contributor who spots that and contributes to NetworkX :)

and what about me!? I always dreamed to become a networkx contributor! I even once had a polemics fight over an honor to contribute to it in Debian! (well -- I won the battle IMHO but still lost as has not contributed ;-) )

But seriously -- anything what is good for you, is good for me. Feel free to give a clear and agreed-upon instruction (e.g. "Remove action integration" or alike) or just adopt/rebase this PR to your liking!

dschult:
Great!  Thanks for this....
Can you remove the codespell integration. Leave the spelling changes except remove the one that is `perfor... (#4536)` flagged above.

:)

yarikoptic:
done!  FTR: original state of this branch with all the configs etc (whenever you decide to jump the ship) is at https://github.com/yarikoptic/networkx/tree/enh-codespell-complete-abandoned

dschult (approved):
Thanks very much!!

PR #6669, opened 2023-04-28
Added docstring example for dense.py

vanshika230:
Added Docstring Example for:- 

1. floyd_warshall(G, weight='weight')
2. floyd_warshall_numpy(G, nodelist=None, weight='weight')
Based on previous work was done [here](https://github.com/networkx/networkx/pull/6573).


networkx/algorithms/shortest_paths/dense.py
```diff
@@ -40,6 +40,16 @@ def floyd_warshall_numpy(G, nodelist=None, weight="weight"):
         A numpy array of shortest path distances between nodes.
         If there is no path between two nodes the value is Inf.
 
+    Examples
+    --------
+    >>> G = G = nx.DiGraph()
```
rossbar, line 45:
```suggestion
    >>> G = nx.DiGraph()
```

networkx/algorithms/shortest_paths/dense.py
```diff
@@ -40,6 +40,16 @@ def floyd_warshall_numpy(G, nodelist=None, weight="weight"):
...
+    >>> G.add_weighted_edges_from([(0, 1, 5), (1, 2, 2), (2, 3, -3), (1, 3, 10),(3, 2, 8)])
```
rossbar, line 46:
```suggestion
    >>> G.add_weighted_edges_from([(0, 1, 5), (1, 2, 2), (2, 3, -3), (1, 3, 10), (3, 2, 8)])
```

networkx/algorithms/shortest_paths/dense.py
```diff
@@ -218,6 +228,15 @@ def floyd_warshall(G, weight="weight"):
        A dictionary,  keyed by source and target, of shortest paths distances
        between nodes.
 
+    Examples
+    --------
+    >>> G = nx.DiGraph()
+    >>> G.add_weighted_edges_from([(0, 1, 5), (1, 2, 2), (2, 3, -3), (1, 3, 10),(3, 2, 8)])
```
rossbar, line 234:
```suggestion
    >>> G.add_weighted_edges_from([(0, 1, 5), (1, 2, 2), (2, 3, -3), (1, 3, 10), (3, 2, 8)])
```

networkx/algorithms/shortest_paths/dense.py
```diff
@@ -218,6 +228,15 @@ def floyd_warshall(G, weight="weight"):
...
+    >>> fw = nx.floyd_warshall(G, weight='weight')
+    >>> results = {a:dict(b) for a,b in fw.items()}
+    >>> print(results)
+    {0: {0: 0, 1: 5, 2: 7, 3: 4}, 1: {1: 0, 2: 2, 3: -1, 0: inf}, 2: {2: 0, 3: -3, 0: inf, 1: inf}, 3: {3: 0, 2: 8, 0: inf, 1: inf}}
```
rossbar, line 238: This seems unnecessary - can't we just output the results of `floyd_warshall` directly?

vanshika230: We can, I just thought that maybe we could make it more comprehensible. When I was making this example, I found the output to be a little long because of the text and anyone using this function would probably like to remove it. So I thought to directly add this to the documentation. If anyone wants to use it as it is they can simply remove the fw. If you feel it is unnecessary I can remove it. Please let me know.
image

networkx/algorithms/shortest_paths/dense.py
vanshika230: Will make the above 3 changes.

networkx/algorithms/shortest_paths/dense.py
rossbar: Ah indeed, that's a very nasty repr due to the defaultdict + lambdas. I wonder if we can't improve the situation somewhat, but for now the proposed workaround seems good!

rossbar (approved):
LGTM, thanks @vanshika230 !

PR #6678, opened 2023-05-02
Improve test coverage for reaching.py

vanshika230:
I have increased test coverage for reaching.py according to here :- https://app.codecov.io/gh/networkx/networkx/blob/main/networkx/algorithms/centrality/reaching.py
The test coverage is now at a 100% as seen below :- 
reaching_after


dschult (approved):
Thanks!

PR #6679, opened 2023-05-02
try adding circleci artifact secret

dschult:
This PR will (maybe) work once the secret is added to the github secrets setting


rossbar (approved):
LGTM, thanks @dschult 

FWIW our doc builds are failing *in addition* to the missing circleci token due to issues with `nb2plot`. In other words, we'll still expect red x's in CI even after this is merged (and the secret added).

PR #6681, opened 2023-05-03
Fix custom weight attribute for Mehlhorn

kalkoen:
Using a custom weight attribute in the Mehlhorn implementation gave a KeyError. Reason: Edges in G_1_prime are given "weight" attributes, but upon updating, the weight (=/= "weight") of the existing edge in G_1_prime is get. Fix: set and get only "weight" attributes for G_1_prime.


MridulS (commented):
Thanks @kalkoen! Could you also add a test for this in the tests file https://github.com/networkx/networkx/blob/main/networkx/algorithms/approximation/tests/test_steinertree.py?

rossbar (approved):
I took the liberty of pushing up a couple tests for cases where the edge attribute name is something other than the default "weight". The first test covers the bugfix (i.e. fails on current `main`) - the second covers a similar scenario, but with multigraphs.

This LGTM but should get another pair of eyes. Thanks @kalkoen !

dschult (approved):
Thanks @kalkoen and @rossbar
This looks good to me!

PR #6682, opened 2023-05-05
Time dependent module

eirinikafourou:
After our discussion #6629 me and @evgepab have implemented the cd index method. Although we were not very sure on how to handle the weight matrix as an argument.


networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+__all__ = ["cd_index"]
+
+
+@not_implemented_for("undirected")
+@not_implemented_for("multigraph")
+def cd_index(G, node, time_delta=5, weight=None):
+    r"""Compute the CD index.
+
+    Calculates the CD index for the graph based on the given "focal patent" node
+    considering the patents after `time_delta` years.
```
dschult, line 17: Can you use language that doesn't rely on the setting of patents?  Or is this cd_index only used for patents?

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Parameters
+    ----------
+    G : graph
+       A directed networkx graph whose nodes have datetime `time` attributes.
+    node : integer
```
dschult, line 23: Nodes are not integers in general.  They can be any hashable. So, you can use `node : node`  just as you do with `G : graph`.

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+    succ, pred = list(G.successors(node)), list(G.predecessors(node))
+
+    if not succ:
+        raise ValueError("This node has no successors.")
+
+    # keep the predecessors that existed before the target date
+    actual_pred = [i for i in pred if G.nodes[i]["time"].timestamp() <= target_date]
+
+    if not actual_pred:
+        raise ValueError("This node has no predecessors.")
```
dschult, line 101: Constructing a set instead of a list here saves the later set construction.
```suggestion
    pred = {i for i in G.pred[node] if G.nodes[i]["time"].timestamp() <= target_date}
    if not pred:
        raise ValueError(f"The requested node {node} has no predecessors.")
```

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # 1 if any edge between a focal node's predecessor and a focal node's accessor
+    # exists, else 0
+    b = [int(any((i, j) in G.edges() for j in succ)) for i in actual_pred]
```
dschult, line 105: Is the word accessor a typo, or does it mean successor?
Later, the value of bi is 1 or -1, o maybe you should set that value here:
```suggestion
    # -1 if any edge between a focal node's predecessor and successor. Else 1
    b = [(-1 if any(j in G[i] for j in G[node])) else 1) for i in pred]
```

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+    for s in succ:
+        n |= set(G.predecessors(s)) - {node}
+
+    # calculate cd index
+    if weight is None:
+        return round(sum((-2) * bi + 1 for bi in b) / len(n), 2)
+    else:
+        return round(
+            sum((-2 * b[i] + 1) / weight[i] for i in range(len(b))) / len(n), 2
+        )
```
dschult, line 118: Simplify the expression using alternates:
```suggestion
        return round(sum(b) / n, 2)
    else:
        return round(sum(bi / wt for bi, wt in zip(b, weight)) / n, 2)
```

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+        raise ValueError("This node has no predecessors.")
+
+    # 1 if any edge between a focal node's predecessor and a focal node's accessor
+    # exists, else 0
+    b = [int(any((i, j) in G.edges() for j in succ)) for i in actual_pred]
+    n = set(actual_pred)
+
+    # n is the union of the focal node's predecessors and its accessors' predecessors
+    for s in succ:
+        n |= set(G.predecessors(s)) - {node}
```
dschult, line 110: We only use the length of this set, so:
```suggestion
    # n is size of the union of the focal node's predecessors and its successors' predecessors
    n = len(pred.union(*(G.pred[s].keys() - {node} for s in G[node])))
```

dschult (commented):
Thanks a lot for this.  
I just took a quick look through the code and put probably too many suggestions.  Please don't take these as must-dos. They are suggestions and only from a quick pass, and aimed toward making the code easier to read based on my perspective newly coming in to this problem. Think about what makes the most sense.

I would like it if this code used names/docs which were less specific to patent tracking. To the extent possible, I ask that we use general terms like "node" instead of "patent". Perhaps along with that a paragraph explaining the patent application and that a node is a patent and an edge is a "..." etc would be really helpful too.

Thanks!!

evgepab:
Thank you for your suggestions! We have added them and are waiting for any additional feedback! Also we were thinking of  adding weights on edges like that 'G.add_edge(1, 2, weight=0.5, time=datetime(2023, 5, 9))' instead of having an optional list as an argument.

dschult (commented):
Just some wording stuff for the doc_string...

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,118 @@
...
+import networkx as nx
+from networkx.utils import not_implemented_for
+
+__all__ = ["cd_index"]
+
+
+@not_implemented_for("undirected")
+@not_implemented_for("multigraph")
+def cd_index(G, node, time_delta=5, weight=None):
+    r"""Compute the CD index.
```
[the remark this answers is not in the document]
eirinikafourou: > I might have the terminology mixed up for this, but is the CD index for the graph? or for the node of the graph? It seems like the index would be different for each choice of node, so it could be described as

You are right! The CD index is calculated for the node of the graph. We have updated the doc_strings so as to be better explained.

dschult:
The NetworkX standard way to include node attributes for weights is to use an optional keyword argument `weight=None`. Then if weight is None, we treat each node as having weight 1. If weight is a string we use the node attribute with the name held in the string:  e.g. `weight="distance"` where the graph holds a node attribute called "importance".  This way, graphs with e.g. multiple different node attributes can compute the cd_index for different weighting without constructing a new graph.

Does this make sense?

evgepab:
https://github.com/eirinikafourou/networkx/blob/391ba5c12a52b3b24925d60232e35c18fe56740a/networkx/algorithms/time_dependent.py#L13

https://github.com/eirinikafourou/networkx/blob/391ba5c12a52b3b24925d60232e35c18fe56740a/networkx/algorithms/time_dependent.py#L112-L119

Is this what you meant for the weight handling ?

evgepab:
Also, we're a bit confused as to why some of the checks failed, if you could provide us more information regarding this.

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,119 @@
...
+    # n is size of the union of the focal node's predecessors and its successors' predecessors
+    n = len(pred.union(*(G.pred[s].keys() - {node} for s in G[node])))
+
+    # calculate cd index
+    if weight is None:
+        return round(sum(bi for bi in b) / n, 2)
+    else:
+        # If a node has the specified weight attribute, its weight is used in the calculation
+        # otherwise, a weight of 1 is assumed for that node
+        weights = [G.nodes[i][weight] if weight in G.nodes[i] else 1 for i in pred]
```
dschult, line 118: There is a nice dict function called `get` that does much of this lookup at the C-Python level:
```suggestion
        weights = [G.nodes[i].get(weight, 1) for i in pred]
```

eirinikafourou: Thank you, we've implemented your suggestion!

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,119 @@
...
+
+    """
+    if not (
+        all(
+            G.has_node(n)
+            and "time" in G.nodes[n]
+            and isinstance(G.nodes[n]["time"], datetime)
+            for n in G
+        )
+    ):
```
dschult, line 91:
```suggestion
    if not all(isinstance(G.nodes[n].get("time"), datetime) for n in G):
```

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,119 @@
...
+        raise ValueError("Not all nodes have a datetime 'time' attribute.")
+
+    # get target_date's unix timestamp
+    target_date = G.nodes[node]["time"].timestamp() + time_delta * 365 * 24 * 60 * 60
+
+    if not G[node]:
+        raise ValueError("This node has no successors.")
```
dschult, line 98: Should no successors of a node return a value of 0? or raise an exception?  When would CD index be zero?

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # keep the predecessors that existed before the target date
+    pred = {i for i in G.pred[node] if G.nodes[i]["time"].timestamp() <= target_date}
+    if not pred:
+        raise ValueError("This node has no predecessors.")
```
dschult, line 103: Similarly, should no predecessors give a result of 0?

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # -1 if any edge between the node's predecessor and the node's successor
+    # exists, else 1
+    b = [-1 if any((i, j) in G.edges() for j in G[node]) else 1 for i in pred]
```
dschult, line 107: Maybe don't create all edges just to check if one is present?
```suggestion
    # -1 if any edge between node's predecessors and node's successors, else 1
    b = [-1 if any(j in G[i] for j in G[node]) else 1 for i in pred]
```

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,119 @@
...
+    is used in order to check how consolidating or destabilizing a patent
+    is, hence the nodes of the graph represent patents and the edges show
+    the citations between these patents. The mathematical model is given
+    below: 
+
+    .. math::
+        CD_{t}=\frac{1}{n_{t}}\sum_{i=1}^{n}\frac{-2f_{it}b_{it}+f_{it}}{w_{it}},
+
+    where `f_{it}` equals 1 if `i` cites the focal patent else 0, `b_{it}` equals
+    1 if `i` cites any of the focal patents accessors else 0, `n_{t}` is the number
```
dschult, line 72:
```suggestion
    1 if `i` cites any of the focal patents successors else 0, `n_{t}` is the number
```

dschult (commented):
I've made some more suggestions below. I think they are all that are left to handle.

We need to connect the documentation to this file to make the docs appear. Add an entry to the `doc/reference/algorithms/index.rst` file (alphabetical) and create a file similar to `doc/reference/algorithms/wiener.rst` to guide the automatic generation of the documentation. Probably `time_dependent.rst` is a good name.

You can check that the docs are working by clicking on the "Show All Checks" in the PR "checks" section and then scrolling to (near the bottom) the artifact documentation check.  The "details" link for that check leads to the documentation webpages aas constructed by this PR. Check that they render the way you would like them to.

:)  Thanks!

networkx/algorithms/time_dependent.py
evgepab: @dschult SInce the paper was based on patents in the beginning we followed the same logic, so we considered it wrong when no successors were given because it's not possible for a patent to not have any references. Now since we're not talking only about patents and we're making it more inclusive, we could consider that if the main node has no successors the cd index is only based on the sum of the node's predecessors.

networkx/algorithms/time_dependent.py
evgepab: @dschult Yes, that should be possible.

dschult (approved):
This looks good to me!

Thanks!   I'm approving it. we still need one more set of eyes.

evgepab:
> This looks good to me!
> 
> Thanks! I'm approving it. we still need one more set of eyes.

@dschult Thank you for the collaboration!! Should we request a review from another reviewer?

rossbar (commented):
The implementation looks nice, but before I look more closely I thought I'd raise a high-level concern - I'm leery about the constraints that the this puts on the input graphs; especially the fact that the nodes must be datetime attributes. I get that the CD-index is property measured over time, but is there a way this could be made more general?

Generally speaking if @dschult approves the current formulation then +1 from me, though I think it's worth evaluating at this point if it can't be done more generally!

dschult:
Thanks @rossbar!  I think you are right -- it is worth looking at whether this needs to be a `datetime` attribute, or perhaps it could be more general.  For example, the node `time` attrs (and `time_delta`) could be integers... And indeed any set of comparable objects. The current code uses time units of years and uses the `datetime` attribute specifically by referring to the `timestamp` of that `datetime` value -- but I think it could just use the datetime objects themselves instead of converting to timestamp.  

In the interest of making this module a starting point for other time_dependent algorithms, can we pull the time handling into an optional argument that is a function which inputs the graph and outputs a dict keyed by node to a time value that could be any format you want (datetime instance, timestamp/float, integer). The type of the value would need to work (in this case) with the type of the input `time_delta` and support `<` and `+` operators.

What do you think @evgepab @eirinikafourou ?...  I can push a draft for your review if you don't want to take this on.

Also, was there a reason you decided to use the `timestamp` attribute instead of comparing/adding `datetime` instances? I don't know the datetime class very well, so want a head's up if there is a quirk that timestamp gets around.
Thanks!

evgepab:
@dschult  @rossbar Thank you for your feedback! Regarding the optional function attribute, do you mean that something like:
`time_dict = lambda G: {n: G.nodes[n].get("time") for n in G}` could be given in order to find the node's predecessors? And, if it's not given, then we would use all the node's predecessors?

Also, there was no specific reason for the use of `timestamp`, only to get more precise comparisons.

dschult:
I was thinking that the default time-function could provide exactly what you use now. But that it would allow the user to specify alternate ways to produce a time value if desired.

And, right now, I'm not sure my idea for a function that returns all the node values is the best approach.  Maybe it should just be a node attribute that holds the value. Then the input function would not be needed. Let me play with that here for a minute: Right now the node attribute "time" is a `datetime` object. Can we use those objects along with the `time_delta` value to make all the comparisons that are needed? If so, is it possible for the user to put a float or integer in there (or the timestamp) and make it all work?

My quick look suggests that this could work and would be simpler than my suggestion. Removing the `.timestamp()` in two places and the `* 365 * 24 * 60 * 60` for the time_delta would make it work for arbitrary time-related numbers or datetime objects or timestamps. The scale of the time_delta can be left up to the user -- the default value can be scaled by the value for a "year", or the default value could be a `time_delta` object from the datetime module (maybe??).  You would then also be able to remove the check for all time objects being datetime.

Maybe this is a better suggestion than an additional input argument. But maybe there another way to do that which I am just not seeing. What do you think would be a good way to make this work for the cases you show and also for an arbitrary time value? I think the goal is to not force people to use `datetime`, but to allow and maybe even encourage it through the examples used.

eirinikafourou:
@dschult @rossbar Thank you for your suggestions! We removed the `.timestamp()` and now any set of compatible objects is allowed for `time_delta` and `time` attribute, although the default value for `time_delta` is now datetime.timedelta, in order to encourage the use of datetime. 

However, another idea we had was to allow the user to give as an input the name of the attribute that would be used in the calculations. We currently use the `time` attribute, although the user might have wanted e.g. to use the `date` attribute. Do you think this would be useful?

dschult:
Yes -- it would be better to have them able to provide any attribute that will be used to represent time.
I can imagine, e.g. a graph with multiple time related attributes and people might want cdindex for each of these attributes.

dschult (commented):
Can we make the name of the attribute variable `time`  and with a default value of "time"?
I'm envisioning it to be like `weight`.  NetworkX has a lot of functions that require an edge attribute that we usually call "weight".  The function signature `weight="weight"` is convenient because the name tells us what the default/usual value is and the meaning of that value. But it is flexible for those who want to use a different attribute in their graph.

I think `"time"` is definitely the right default value. And the variable name `time` makes that clear.  I anticipate that there will be other time-dependent algorithms added to networkx in the future and this choice could be setting a precedence. So I'd like to "get it right".  Suggestions welcome.

Also, could we make the keyword arguments "keyword only"?
I realized that `nx.cd_index(G, 4, "weight", "time")` could become confusing. And we're trying to remember to make new functions allow only positional parameters for the commonly prescribed values (like G and node).  The way to make it keyword-only is to put a `*, ` in the function signature before the keyword arguments. `def cd_index(G, node, *, time="time", ...)`  Ask if this doesn't make sense.

That does mean the tests will need to add the`time=...` keyword to the attribute name. Or they could just use the default and then have one test to check that a different/specified attribute works.

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,116 @@
...
+    >>> G = nx.DiGraph()
+    >>> nodes = { 1: {'time': datetime(2015, 1, 1)}, \
+    ...     2: {'time': datetime(2012, 1, 1), 'weight': 4}, \
+    ...     3: {'time': datetime(2010, 1, 1)}, \
+    ...     4: {'time': datetime(2008, 1, 1)}, \
+    ...     5: {'time': datetime(2014, 1, 1)}}
+    >>> G.add_nodes_from([(n, nodes[n]) for n in nodes])
+    >>> edges = [(1, 3), (1, 4), (2, 3), (3, 4), (3, 5)]
+    >>> G.add_edges_from(edges)
+    >>> cd = nx.cd_index(G, 3, time='time', weight='weight')
```
dschult, line 61:
```suggestion
    >>> nodes = { 
    ...     1: {"time": datetime(2015, 1, 1)},
    ...     2: {"time": datetime(2012, 1, 1), 'weight': 4},
    ...     3: {"time": datetime(2010, 1, 1)},
    ...     4: {"time": datetime(2008, 1, 1)},
    ...     5: {"time": datetime(2014, 1, 1)}
    ... }
    >>> G.add_nodes_from([(n, nodes[n]) for n in nodes])
    >>> edges = [(1, 3), (1, 4), (2, 3), (3, 4), (3, 5)]
    >>> G.add_edges_from(edges)
    >>> cd = nx.cd_index(G, 3, time="time", weight="weight")
```

networkx/algorithms/time_dependent.py
```diff
@@ -0,0 +1,116 @@
...
+        raise ValueError("Not all nodes have a 'time' attribute.")
+
+    try:
+        # get target_date
+        target_date = G.nodes[node][time] + time_delta
+        # keep the predecessors that existed before the target date
+        pred = {i for i in G.pred[node] if G.nodes[i][time] <= target_date}
+    except:
+        raise ValueError(
+            "Addition and comparison are not supported between 'time_delta' and 'time' types, default time_delta = datetime.timedelta."
```
dschult, line 98:
```suggestion
            "Addition and comparison are not supported between 'time_delta' "
            "and 'time' types, default time_delta = datetime.timedelta."
```

dschult (approved):
I just have two minor suggestions for styling the example to match `black` format and to decrease the line length for the ValueError message.

Thanks @eirinikafourou and @evgepab for your patience with these requests. And thanks @rossbar for the suggestions. The function signature is much better and aligns with how we'd like it to be moving forward. :)

I approve this PR.

eirinikafourou:
> I just have two minor suggestions for styling the example to match `black` format and to decrease the line length for the ValueError message.
> 
> Thanks @eirinikafourou and @evgepab for your patience with these requests. And thanks @rossbar for the suggestions. The function signature is much better and aligns with how we'd like it to be moving forward. :)
> 
> I approve this PR.

Thank you for guidance! We hope this implementation will be useful!

jarrodmillman (approved):

dschult:
Thanks @eirinikafourou and @evgepab !!

evgepab:
We're glad we could help 😊 @dschult

PR #6684, opened 2023-05-06
modified max_weight_matching to be non-recursive

boothby:
I ran into the recursion limit using `max_weight_matching` on a large graph.  This algorithm has some rather large recursive subroutines, so I have not exercised my typical performance-oriented rigor for fear of breaking things and making it completely unreadable.

Instead, I used a trampoline pattern for its simplicity and at-a-glance correctness.

```python
def original_f(*args):
    ...
    original_f(*r_args)
    ...

def new_f(*args):
    def _recurse(*args):
        ...
        yield r_args
        ...
    stack = [_recurse(*args)]
    while stack:
        top = stack[-1]
        for r_args in top:
            stack.append(_recurse(*r_args))
            break
        else:
            stack.pop()
```

### DIFF
--- a/networkx/algorithms/matching.py
+++ b/networkx/algorithms/matching.py
@@ -410,9 +410,11 @@ def max_weight_matching(G, maxcardinality=False, weight="weight"):
 
         # Generate the blossom's leaf vertices.
         def leaves(self):
-            for t in self.childs:
+            stack = [*self.childs]
+            while stack:
+                t = stack.pop()
                 if isinstance(t, Blossom):
-                    yield from t.leaves()
+                    stack.extend(t.childs)
                 else:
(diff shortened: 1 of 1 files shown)


rossbar (approved):
Whew - thanks for taking this on @boothby , my head is swimming from all the blossoms, sub-blossoms, T-blossoms, and all the familial relationships between blossoms :upside_down_face: . Insofar as I was able to trace the execution (i.e. shallowly) the new pattern makes sense. The previous implementation was itself quite complex, so I don't think this is sacrificing anything in the way of clarity, and the removal of the recursion limit so that this can run on large graphs is an obvious improvement. Thanks again for tackling this complicated refactor @boothby !

PR #6688, opened 2023-05-13
Add `@nx._dispatch` decorator to most algorithms

eriknw:
This adds dispatching to most algorithms (only functions in `algorithms`, `generators`, and `linalg`). This raises a few questions that we need to solve or postpone. I will raise these questions below.

CC @jim22k @MridulS


networkx/algorithms/tree/branchings.py
```diff
@@ -723,6 +726,7 @@ def maximum_branching(
     return B
 
 
+@nx._dispatch(edge_attrs="attr,partition", edge_defaults={"attr": "default"})
 def minimum_branching(
     G, attr="weight", default=1, preserve_attrs=False, partition=None
```
eriknw, line 731: Q15: can we respect `preserve_attrs=True`?

eriknw (commented):
Q16: are there functions that we _shouldn't_ dispatch for any reason (such as they're expected to be used in a tight loop)?

Note that there are a few functions in `flow` such as `preflow_push_impl` that I think should probably be hidden as e.g. `_preflow_push_impl` that we don't dispatch.

Q17: should we consider dispatching on `create_using=` argument (in the future)?

eriknw:
This brings the total number of dispatched functions up to 544. We currently implement 81 in (my local branch of) `graphblas-algorithms`, which is about 15%.

networkx/algorithms/tournament.py
```diff
@@ -340,6 +342,7 @@ def is_closed(G, nodes):
     return all(not (is_closed(G, S) and s in S and t not in S) for S in neighborhoods)
 
 
+@nx._dispatch(name="tournament_is_strongly_connected")
```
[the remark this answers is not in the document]
jim22k: Yes, unless "tournament" is already in the function name (e.g. tournament_matrix, is_tournament).

networkx/classes/backends.py
```diff
@@ -186,7 +195,16 @@ def wrapper(*args, **kwds):
     return wrapper
 
 
-def test_override_dispatch(func=None, *, name=None, graphs="G"):
+def test_override_dispatch(
+    func=None,
+    *,
+    name=None,
+    graphs="G",
+    edge_attrs=None,
```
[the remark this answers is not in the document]
jim22k: Explicit is always better. Let's go with `None`.

networkx/algorithms/connectivity/connectivity.py
```diff
@@ -31,6 +31,7 @@
 ]
 
 
+@nx._dispatch(graphs="G,auxiliary,residual")
 def local_node_connectivity(
     G, s, t, flow_func=None, auxiliary=None, residual=None, cutoff=None
```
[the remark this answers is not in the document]
jim22k: Maybe we take a cue from Typescript and spell it `"G,auxiliary?,residual?"`

networkx/algorithms/dag.py
```diff
@@ -942,6 +953,7 @@ def antichains(G, topo_order=None):
             antichains_stacks.append((new_antichain, new_stack))
 
 
+@nx._dispatch(edge_attrs="weight", edge_defaults="default_weight")
 @not_implemented_for("undirected")
 def dag_longest_path(G, weight="weight", default_weight=1, topo_order=None):
```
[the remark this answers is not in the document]
eriknw: Resolved. `edge_attrs` may now be a dict as described here.

networkx/algorithms/connectivity/connectivity.py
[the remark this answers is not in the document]
eriknw: Done. Graphs names may now be e.g. `"residual?"` to indicate they're optional.

networkx/algorithms/flow/capacityscaling.py
```diff
@@ -149,6 +149,7 @@ def _build_flow_dict(G, R, capacity, weight):
     return flow_dict
 
 
+@nx._dispatch(node_attrs="demand", edge_attrs="capacity,weight")
 def capacity_scaling(
     G, demand="demand", capacity="capacity", weight="weight", heap=BinaryHeap
 ):
```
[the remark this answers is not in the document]
eriknw: `edge_attrs` may now be a dict. We handle this function via:
```python
@nx._dispatch(node_attrs="demand", edge_attrs={"capacity": float("inf"), "weight": 0})
```
Each value in the `edge_attrs` dict may a string of the argument name used as the default value, or the (non-string) default value itself.

networkx/algorithms/isomorphism/tree_isomorphism.py
```diff
@@ -24,6 +24,7 @@
 __all__ = ["rooted_tree_isomorphism", "tree_isomorphism"]
 
 
+@nx._dispatch(graphs="t1,t2")
 def root_trees(t1, root1, t2, root2):
```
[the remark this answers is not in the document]
eriknw: `graphs` now must be a string indicating a single graph (as the first argument), or a dict of `{argname: pos}` such as `{"t1": 0, "t2": 2}`.

networkx/algorithms/operators/all.py
```diff
@@ -7,6 +7,7 @@
 __all__ = ["union_all", "compose_all", "disjoint_union_all", "intersection_all"]
 
 
+@nx._dispatch(graphs="[graphs]")
```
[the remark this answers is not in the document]
eriknw: Yeah, let's use this. Still needs implemented.

networkx/algorithms/summarization.py
```diff
@@ -403,6 +404,7 @@ def _snap_split(groups, neighbor_info, group_lookup, group_id):
     return groups
 
 
+@nx._dispatch(node_attrs="[node_attributes]", edge_attrs="[edge_attributes]")
```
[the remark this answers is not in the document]
eriknw: Done.

networkx/classes/backends.py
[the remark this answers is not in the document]
eriknw: Done. We are now explicit regarding edge/node weights everywhere.

networkx/algorithms/tree/branchings.py
eriknw: Yes. We now handle this via:
```python
@nx._dispatch(
    edge_attrs={"attr": "default", "partition": None},
    preserve_edge_attrs="preserve_attrs",
)
```

networkx/convert.py
```diff
@@ -241,6 +242,7 @@ def from_dict_of_lists(d, create_using=None):
     return G
 
 
+# @nx._dispatch(...)  # Not implemented
 def to_dict_of_dicts(G, nodelist=None, edge_data=None):
     """Returns adjacency representation of graph as a dictionary of dictionaries.
```
eriknw, line 247: This function doesn't fit the patterns we're using. Here `edge_data=None` means to preserve all edge data, otherwise it indicates the edge _value_.

networkx/linalg/attrmatrix.py
```diff
@@ -141,6 +142,7 @@ def value(u, v):
     return value
 
 
+@nx._dispatch(edge_attrs="edge_attr", node_attrs="node_attr")
 def attr_matrix(
```
eriknw, line 146: How can we indicate an attribute _must_ exists for all edges or nodes? In `attr_matrix` and `attr_sparse_matrix`, `weight=None` (the default) means each edge has a weight of 1 (or count for multigraph), otherwise the edge attribute must be present for all edges.

networkx/classes/backends.py
```diff
@@ -107,7 +107,16 @@ def _register_algo(name, wrapped_func):
...
-def _dispatch(func=None, *, name=None, graphs="G"):
+def _dispatch(
+    func=None,
+    *,
+    name=None,
+    graphs="G",
+    edge_attrs=None,
+    edge_defaults=None,
+    node_attrs=None,
+    node_defaults=None,
```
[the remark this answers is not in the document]
eriknw: btw, I moved `networkx.classes.backend` to `networkx.utils.backend` due to import issues.

I'll let the networkx reviewer(s) decide whether we ought to split this PR up.

eriknw:
I updated `@nx._dispatch` based on feedback from @jim22k. We still need to update documentation and make sure everything works.

A known shortcoming is handling of functions that return Graphs--we don't always indicate whether edge, node, and/or graph attributes should be preserved. Also, we don't convert graph attributes that are _graphs_. If a reviewer/contributor wants to take a closer look at this, it would be appreciated. Otherwise, I'm inclined to handle these in the future when needed.

Also, I haven't benchmarked/profiled impact to import time or dispatched function call overhead.

eriknw:
@MridulS I think this is ready for review. Who else can/should review this? Do you want this PR split into two: one that adds the `@nx._dispatch` decorator everywhere, and another that updates the `def _dispatch` function? I don't have a strong opinion on this, and I think a single monolithic PR is fine.

Note that major changes were made to `networkx/utils/backends.py` that defines the `_dispatch` decorator. This file used to be called `networkx/classes/backends.py`, but was moved to avoid difficulty importing.

I have [`graphblas-algorithms`](https://github.com/python-graphblas/graphblas-algorithms) working against this PR.

eriknw:
> Do you know if the mypy complain has anything to do with this PR?

I have no idea what mypy complaining about.

networkx/algorithms/tree/operations.py
```diff
@@ -7,6 +7,7 @@
 __all__ = ["join"]
 
 
+# @nx._dispatch(graphs=...)  # Not implemented
 def join(rooted_trees, label_attribute=None):
```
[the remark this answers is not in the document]
aaronzo: perhaps an alternative version of this function that takes a single `nx.Graph` $T$ as input where $T$ is a tree with root $u$, and $u$'s children are exactly the roots of the trees in the current `join` argument, and each of the trees in the current input become subtrees under their respective roots in the new input.

networkx/algorithms/approximation/treewidth.py
```diff
@@ -130,6 +132,7 @@ def best_node(self, graph):
         return None
 
 
+@nx._dispatch(graphs="graph")
 def min_fill_in_heuristic(graph):
```
eriknw, line 136: The input argument of `min_fill_in_heuristic` is a dict, not a nx.Graph.

networkx/algorithms/hierarchy.py
```diff
@@ -17,8 +18,8 @@ def flow_hierarchy(G, weight=None):
     G : DiGraph or MultiDiGraph
        A directed graph
 
-    weight : key,optional (default=None)
-       Attribute to use for node weights. If None the weight defaults to 1.
+    weight : string, optional (default=None)
+       Attribute to use for edge weights. If None the weight defaults to 1.
```
eriknw, line 22: The docstring mistake was uncovered by how we now test the info from `@nx._dispatch`.

networkx/algorithms/shortest_paths/astar.py
```diff
@@ -9,6 +9,7 @@
 __all__ = ["astar_path", "astar_path_length"]
 
 
+@nx._dispatch(edge_attrs="weight", preserve_node_attrs="heuristic")
```
eriknw, line 12: `preserve_node_attrs="heuristic"` means _all_ node attributes are preserved when the `heuristic` argument is a callable.

networkx/classes/tests/dispatch_interface.py
```diff
@@ -49,25 +49,112 @@ def convert(graph):
...
+    def convert_from_nx(
+        graph,
+        *,
+        edge_attrs=None,
+        node_attrs=None,
+        preserve_edge_attrs=None,
+        preserve_node_attrs=None,
+        preserve_graph_attrs=None,
+        name=None,
+        graph_name=None,
```
eriknw, line 68: Including `graph_name` in `convert_from_nx` greatly improved my debugging and development experience when working on this PR.

networkx/classes/tests/dispatch_interface.py
```diff
@@ -49,25 +49,112 @@ def convert(graph):
...
+    ):
+        if name in {
+            # Raise if input graph changes
+            "lexicographical_topological_sort",
+            "topological_generations",
+            "topological_sort",
+            # Sensitive tests (iteration order matters)
+            "dfs_labeled_edges",
+        }:
+            return graph
```
eriknw, line 78: We don't test converting the input graphs for every function. I think core devs should now about this "escape hatch" if dispatch testing ever causes headaches.

networkx/utils/backends.py
```diff
@@ -0,0 +1,628 @@
...
+                    edge_attr: 1 for edge_attr in bound.arguments[_edge_attrs[1:-1]]
+                }
+            elif callable(bound.arguments[_edge_attrs]):
+                _preserve_edge_attrs = True
+                _edge_attrs = None
+            elif bound.arguments[_edge_attrs] is not None:
+                _edge_attrs = {bound.arguments[_edge_attrs]: 1}
+            elif name == "to_numpy_array" and hasattr(
+                bound.arguments["dtype"], "names"
+            ):
```
eriknw, line 428: So far, `to_numpy_array` here is the only "special case" that we handle in this function.

eriknw:
My latest changes are up, and I don't anticipate making any more major changes.

Actually testing the values of `@nx._dispatch` arguments uncovered a bunch of things, including a couple docstring errors.

So, yeah, the `nx-loopback` "backend" now converts input graphs to only include the specified edge, node, and graph attributes. This give us _much_ greater confidence in the correctness of the dispatch metadata.  When adding `preserve_edge_attrs=`, `preserve_node_attrs=`, and `preserve_graph_attrs=`, I heavily relied on the tests, not the docstrings. IMHO, this means that the most likely mistakes are algorithms that _should_ preserve edge, node, or graph attributes, but aren't annotated as such because the behavior isn't exercised in tests. I don't think this should worry us too much--each function should get closer scrutiny when implemented by backends.

As a "first pass" to dispatch all the things, this PR is much more functional and complete than I anticipated. It took a lot of work. I think it lays a solid foundation for being able to automatically convert to different backends.

fwiw, I actually like seeing metadata in `@nx._dispatch` about the "shape" of the inputs needed for the calculation. We don't yet have similar metadata describing output graphs or how input graphs are mutated. In the future, I think it would also be handy to describe how multigraph attributes could be combined (add, first, last, etc.) to be turned into a (not multi-) graph for algorithms.

networkx/algorithms/centrality/second_order.py
```diff
@@ -38,6 +38,7 @@
 __all__ = ["second_order_centrality"]
 
 
+@nx._dispatch(preserve_edge_attrs={"G": {"weight": 1}})
```
eriknw, line 41: `preserve_edge_attrs={"G": {"weight": 1}}` means that the "weight" attribute of graph G should be preserved (with default value of 1).

I think functions like this are great candidates for adding `weight="weight"` arguments rather than always assume the graph has edge attributes of `"weight"`. If `weight` argument was added, the dispatch decorator would become:
```python
@nx._dispatch(edge_attrs="weight")
```

networkx/algorithms/tree/operations.py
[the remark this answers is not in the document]
eriknw: Thanks for taking a look. Yeah, having an alternative version of this (and other) functions is an option. Another alternative version could have two arguments--a list of graphs and a list of root nodes, such as `def join_trees(trees, roots): return join(list(zip(trees, roots))` (interestingly, the first step in `join` does the opposite: `trees, roots = zip(*rooted_trees)`).

If we don't want to change the NetworkX API, perhaps we need to come up with a way to say "this uses custom dispatching logic". For example:
```python
@nx._dispatch(graphs="<custom>", preserve_edge_attrs=True, preserve_node_attrs=True, preserve_graph_attrs=True)
def join(rooted_trees, label_attribute=None):
```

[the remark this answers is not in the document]
aaronzo: Perhaps there could be a multi-hop dispatch, kind of a functional programming solution, where first the function is dispatched to an alternative networkx implementation which can be more easily mapped to the backend. This would need a new decorator `_dispatch_via`.

```python
@nx._dispatch_via(_dispatchable_f, pre_dispatch=_pre_dispatch_f, post_dispatch=_post_dispatch_f)
def hard_to_dispatch_f(*a, **kw):
   ...


def _pre_dispatch_f(*a, **kw):
   ...

@nx._dispatch(...)
def _dispatchable_f(*b, **kw):
   ...

def _post_dispatch_f(*c, **kw):
   ...

```

eriknw:
I wonder whether we should have `@nx._dispatch` decorator _after_ `@not_implemented_for` decorator. Even if the input graph isn't from networkx, it ought to be easy for it to implement `g.is_multigraph()` and `g.is_directed()`.

Once we enable conversions to other backends, it seems silly to perform the conversion merely to have it raise b/c it's the wrong type of graph. On the other hand, I don't really understand how this may interact with `argmap` (used by `not_implemented_for`--will it still work if `@nx._dispatch` returns a class instance and not a normal Python function?) and `@py_random_state`.

networkx/utils/backends.py
```diff
@@ -0,0 +1,628 @@
...
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwds):
+        backend = plugins[plugin_name].load()
+        if not hasattr(backend, name):
+            if plugin_name == "nx-loopback":
+                raise NetworkXNotImplemented(
+                    f"'{name}' not found in {backend.__class__.__name__}"
+                )
+            pytest.xfail(f"'{name}' not implemented by {plugin_name}")
```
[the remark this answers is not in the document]
jim22k: Would `pytest.skip` help with the coverage? I don't like the idea of calling the NetworkX version. It could easily mask an error or misconfiguration in `graphblas-algorithms` if algorithms aren't being registered correctly with the interface. `xfail` or `skip` feel much safer.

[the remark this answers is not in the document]
eriknw: I think missing coverage in `graphblas-algorithms` would catch such a mistake.

networkx/algorithms/connectivity/cuts.py
```diff
@@ -21,6 +21,14 @@
 ]
 
 
+@nx._dispatch(
+    graphs={"G": 0, "auxiliary?": 4, "residual?": 5},
+    preserve_edge_attrs={
+        "auxiliary": {"capacity": None},
+        "residual": {"capacity": None},
```
eriknw, line 28: Should the default value for `"capacity"` edge attribute for connectivity and flow algorithms be 0?

networkx/algorithms/tree/branchings.py
```diff
@@ -708,6 +710,10 @@ def is_root(G, u, edgekeys):
         return H
 
 
+@nx._dispatch(
+    edge_attrs={"attr": "default", "partition": None},
```
eriknw, line 714: Is there a default value for `"partition"` that should be used in this file and `tree.mst.kruskal_mst_edges`? Maybe `EdgePartition.OPEN` or some other `EdgePartition` enum?

eriknw:
This PR doesn't convert graphs used as edge attributes. For example, this happens in `test_christofides_hamiltonian`.

networkx/algorithms/tree/branchings.py
dschult: The code uses `None` to indicate that no partition should be considered. And that is the default value.  It is used in `find_optimum` so that would be the place to look for the impact of the default value.

eriknw: Thanks for the quick reply!

I was exploring whether it's reasonable for backends to require that data or defaults exists for all requested attributes. This simplifies implementation and allows for more performance. For example, whether `edge_attrs={"weight": None}` could, in effect, mean that `"weight"` is a required attribute with no default. When experimenting with this (i.e., failing if not all edges have data), I only found three functions that need to add defaults: `maximum_branching` (this function), `minimum_st_edge_cut` (for "capacity"), and `edge_disjoint_paths` (for "capacity"). This function, `maximum_branching`, seems somewhat special in that is has a meaning for `None` or missing attribute data.

My takeaway is that, in practice, it is probably okay for backends to raise when given `edge_attrs={"weight": None}` if not all edges have `"weight"` attribute. We should also try to specify default values whenever possible, such as `edge_attrs={"capacity": float("inf")}`.

networkx/algorithms/centrality/second_order.py
eriknw: Find these cases on this PR with:
```bash
git grep -A 2 '_dispatch.*preserve_edge_attrs={"G":'
```
which finds:
- `second_order_centrality`
- `quotient_graph`
- `panther_similarity`
- `generate_random_paths`
- `number_of_walks`

eriknw: Raised in separate issue here: #6753

dschult (commented):
There's a lot here... and I think we've got almost all the issues worked out in this. The comments from this review are mostly for a historical record of what we talked about. I don't see blockers here.

It'd be nice to see a (small is OK) time test for impact of import time and function calling (for functions that might be in an inner loop -- which is probably none of these... but who knows...

Thanks for all your work on this and all the notes about the networkx api.

dschult:
Other notes:
- The default value for `partition` in the `tree/branching.py` functions is `0` or if you prefer you can use `nx.EdgePartition.OPEN`
- I do not find a function to reduce a weighted multiedge set for two nodes to a single edge weight for those two nodes. There is an idiom used in a few examples though:
```python
G=nx.Graph(MG)
for u, v in MG.edges():   # parens edges() are needed to suppress 3-tuple edge reporting
    G[u][v]["weight"] = min(d.get("weight", 1) for d in MG[u][v].values())
```
where `min` can be replaced by `sum`, `max`, `np.mean` and other reduction functions.

networkx/utils/backends.py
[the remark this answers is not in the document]
eriknw: Done. Added `NETWORKX_BACKEND_TEST_EXHAUSTIVE`.

networkx/utils/backends.py
```diff
@@ -0,0 +1,654 @@
...
+        and the 4th parameter ``auxiliary`` is an optional graph.
+        To indicate an argument is a list of graphs, do e.g. ``"[graphs]"``.
+
+    edge_attrs : str or dict, optional
+        ``edge_attrs`` may be strings of the function argument that indicate
+        a single edge attribute to use to convert the graph. They may also be
+        dicts that map argument name of attributes to argument name of default
+        values. The default value may be a non-string value such as 0.
+        If not provided, the default value is assumed to be 1.
+        To indicate an argument is a list of attributes, do e.g. ``"[attrs]"``.
```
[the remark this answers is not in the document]
eriknw: > This doesn't handle the case where a default value is a string.

To clarify, a user can still pass a string as the default value, and this will work. We don't yet support a string as the default default value, which I don't think exists in the networkx API.

networkx/algorithms/flow/edmondskarp.py
```diff
@@ -8,6 +8,11 @@
 __all__ = ["edmonds_karp"]
 
 
+@nx._dispatch(
+    graphs="R",
+    preserve_edge_attrs={"R": {"capacity": float("inf"), "flow": None}},
```
eriknw, line 13: One more @dschult... can we choose a sane default value for `"flow"`? My guess is `0`.

eriknw:
Alright, I believe I'm made all the requested updates. Thanks for the careful review @dschult! I think the last remaining thing is fixing the Mypy issue.

Possible follow-up work:
- Measure impact of dispatching on import time and function call time
- Use a class for dispatch function instead of a closure
- Have dispatch decorator indicate whether multigraph edges can be merged with e.g. `sum` or `min`
- Add `@nx._dispatch` to appropriate functions in `networkx.drawing` (anywhere else?)

dschult:
To fix the Mypy problem:  In `networkx/algorithms/connectivity/edge_kcomponents.py`
Line 15:  remove this line -- it's an internal import from networkx
Line 235: change `bridges` to `nx.bridges`

eriknw:
Thanks @dschult--tests are now passing!

networkx/algorithms/flow/edmondskarp.py
dschult: This is a hard module to know how to handle because the functions are building the solution by adding attributes to the "residual graph" R.  There is no user-defined attribute "flow". The flow values are the result of the computation. 

Other functions set up the flow initially to be 0 for every edge in R. For the public function there would be no initial value -- or edge attribute holding the flow.   

This specific function should really be a private function (It's not listed in `__all__`).  Still, if we want to put the decorator on it:  we should use a default value of 0.

eriknw: Thanks. Updated to use `0`.

networkx/algorithms/connectivity/cuts.py
dschult: We decided that flow should have default 0 and capacity should have default 'inf'.

dschult:
Here are some crude timing results:
    Full test suite for networkx: (very crude test using pytest time reporting)
      - without dispatch decorators (before PR #6688): 48.14s   
      - with dispatch decorators: 48.83s
      - in nx-loopback test mode without dispatch decorators (should take longer due to converting graphs): 56.37s
      - in nx-loopback test mode without dispatch decorators (should take longer due to converting graphs): 154.24s

    Time a function call that takes very little time: `G=nx.path_graph(5); %timeit nx.shortest_path(0,3)`
      - without dispatch decorators: 2.67 µs
      - with dispatch decorators: 3.3 µs
    Increase number of nodes:  `G=nx.path_graph(50);%timeit nx.shortest_path(G, 0, 47)`
      - without dispatch decorators: 21.9 µs
      - with dispatch decorators: 22.3 µs

Timing of import of networkx:   
    Change to new git branch, then `python -X importtime -c "import networkx"`
      - without dispatch decorators: ~70 +/-2 ms    (200+/-4 ms for first invocation after switching branches)
      - with dispatch decorators: ~72 +/-2 ms    (204+/-4 ms for first invocation after switching branches)
The first time an import occurs, a py_cache is stored which speeds things up for subsequent imports.

eriknw:
Idea: add `preserve_all_attrs=True` argument to `_dispatch` and `convert_from_nx` that is equivalent to `preserve_edge_attrs=True, preserve_node_attrs=True, preserve_graph_attrs=True`.

dschult:
This sounds like a good idea to me. It will cover what seems like a common case and make the code easier to read too.

MridulS (approved):
Thanks for the work on this @eriknw @jim22k @dschult !!

PR #6689, opened 2023-05-15
MAINT: Add subgraph_view and reverse_view to nx namespace directly through graphviews

MridulS:
Currently `subgraph_view` and `reverse_view` are available in the global `nx` namespace but they are exposed via a redirect in `functions.py`. This was added in https://github.com/networkx/networkx/pull/3627 and I'm not a 100% sure why it is done this way.
Also made changes to some other places where `subgraph_view` and `reverse_view` where being imported via `networkx.graphviews` and not via the `nx` namespace.

### DIFF
--- a/doc/reference/functions.rst
+++ b/doc/reference/functions.rst
@@ -21,10 +21,8 @@ Graph
    add_path
    add_cycle
    subgraph
-   subgraph_view
    induced_subgraph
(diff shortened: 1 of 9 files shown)


rossbar (approved):
Very nice cleanup, thanks @MridulS !

PR #6697, opened 2023-05-21
Added doctrings for generic_graph_view

salym:
Added the docstring although the function is mainly used by the base classes.

### DIFF
--- a/networkx/classes/graphviews.py
+++ b/networkx/classes/graphviews.py
@@ -55,7 +55,7 @@ def generic_graph_view(G, create_using=None):
 
     Returns
     -------
-    newG : networkx.Graph
+    newG : graph
         A view of the input graph `G` of the same class as `G` (or as `create_using`), with the attributes of `G` assigned to it.
 
     Raises
--- a/networkx/classes/graphviews.py
+++ b/networkx/classes/graphviews.py
@@ -39,6 +39,50 @@ __all__ = ["generic_graph_view", "subgraph_view", "reverse_view"]
 
 
 def generic_graph_view(G, create_using=None):
+    """Returns a read-only view of `G`.
+
+    The graph `G` and its attributes are not copied but viewed through the new graph object
+    of the same class as `G` (or of the class specified in `create_using`).
+
+    Parameters
+    ----------
+    G : networkx.Graph
+        A directed/undirected graph/multigraph.
+
+    create_using : NetworkX graph constructor, optional (default=None)
+       Graph type to create. If graph instance, then cleared before populated.
+       If `None`, then the appropriate Graph type is inferred from `G`.
+
+    Returns
+    -------
+    newG : networkx.Graph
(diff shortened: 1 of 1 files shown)


networkx/classes/graphviews.py
```diff
@@ -39,6 +39,50 @@
...
 
 def generic_graph_view(G, create_using=None):
+    """Returns a read-only view of `G`.
+
+    The graph `G` and its attributes are not copied but viewed through the new graph object
+    of the same class as `G` (or of the class specified in `create_using`).
+
+    Parameters
+    ----------
+    G : networkx.Graph
```
dschult, line 49:
```suggestion
    G : graph
```
The notation `networkx.Graph` makes it seem that this must be a particular class. But it can be a Graph, DiGraph, MultiGraph or MultiDiGraph, or a subclass of these. We typically use the phrase `graph` for this, or when another graph library version might be relevant the phrase `networkx graph`

networkx/classes/graphviews.py
```diff
@@ -39,6 +39,50 @@
...
+    Raises
+    ------
+    NetworkXError
+        If `G` is a multigraph (or multidigraph) but `create_using` is not, or vice versa.
+
+    Notes
+    -----
+    The returned graph view is read-only and changes in `G` will be reflected in the view.
+
+    The returned view is a frozen graph, meaning it cannot be modified.
```
dschult, line 70:
```suggestion
    The returned graph view is read-only (cannot modify the graph).
    Yet the view reflects any changes in `G`. The intent is to mimic dict views.
```

networkx/classes/graphviews.py
```diff
@@ -39,6 +39,50 @@
...
+    >>> G.add_edge(2, 3, weight=0.5)
+    >>> G.edges(data=True)
+    EdgeDataView([(1, 2, {'weight': 0.3}), (2, 3, {'weight': 0.5})])
+
+    The view exposes the attributes from the original graph.
+
+    >>> viewG = nx.graphviews.generic_graph_view(G)
+    >>> viewG.edges(data=True)
+    EdgeDataView([(1, 2, {'weight': 0.3}), (2, 3, {'weight': 0.5})])
+    """
```
dschult, line 85: Maybe we should add an example where the graph G is changed and the view is updated...

dschult (commented):
Nice -- we should have put docs in here a long time ago. :}

I've just got a couple of suggestions -- please push back if you prefer something else.

Thanks!!

networkx/classes/graphviews.py
salym: I like your suggestions. I actually had a few more examples in mind, including this one, but I was afraid it was too much. I will add them, then.

networkx/classes/graphviews.py
salym: thanks for the explanation!

MridulS:
Thanks @salym! I committed the changes proposed by @dschult :)

PR #6705, opened 2023-05-23
MAINT: cron job to test against nightly deps every week

MridulS:
implements https://github.com/networkx/networkx/issues/6703

Currently using https://pypi.anaconda.org/scipy-wheels-nightly/simple which hosts the nightly wheels


MridulS:
Will go ahead merge this one and I'll change the index link once projects move to the new one.

Issue #6479, opened 2023-03-10
dfs_postorder_nodes doesn't work with depth_limit


Mamuschka:
just change:
d == "reverse"
```
def dfs_postorder_nodes(G, source=None, depth_limit=None):
    """
    """
    edges = nx.dfs_labeled_edges(G, source=source, depth_limit=depth_limit)
    return (v for u, v, d in edges if d == "reverse")
```
to
d.startswith("reverse")
```
def dfs_postorder_nodes(G, source=None, depth_limit=None):
    """
    """
    edges = nx.dfs_labeled_edges(G, source=source, depth_limit=depth_limit)
    return (v for u, v, d in edges if d.startswith("reverse"))
```

### Current Behavior

![image](https://user-images.githubusercontent.com/20465511/224298765-664bc3fc-c9b4-450f-a226-0b921ac318c6.png)

### Expected Behavior

Output:
[4, 0, 1, 3, 2]
[2, 1, 0, 4, 3]
[0, 1, 4, 3, 2]
[2, 1, 0, 4, 3]

### Steps to Reproduce
```
import networkx as nx
print(list(nx.dfs_postorder_nodes(nx.cycle_graph(5), 2, depth_limit = 3)))
```

### Environment

Python version: 3.10.6
NetworkX version: 3.0

### Additional context

![image](https://user-images.githubusercontent.com/20465511/224301846-bf431321-bee7-4d41-8b28-8e4df137afd9.png)

dschult:
Thanks for reporting this @Mamuschka !! 
This is due to a change in #6240 where the edges are reported by `dfs_labeled_edges` as "reverse-depth_limit" when the depth limit is reached.

Your fix is the correct fix for this function. I have searched the codebase for any other possible similar errors and this seems to be the only place it occurs. We should make a PR that 1) adds a test which catches this error and 2) make the fix.

Thanks you!

navyagarwal:
@dschult I made the fix mentioned in the issue, but that caused some other test cases in the test suite to fail. 

<img width="686" alt="image" src="https://user-images.githubusercontent.com/82928853/224341810-4928bdfb-a781-45ac-b0b4-bf1ea5284aa5.png">

dschult:
Hmmm...  This is more complicated than "just a bug". See the conversation in #6239 
[...]

PR #6714, opened 2023-05-27
Fix dfs_postorder_nodes function

navyagarwal:
Solves issue #6479 

I made some changes to fix `dfs_postorder_nodes` function so that it gives the correct output with `depth_limit` and also modified the test cases based on what I think should be the output.

### DIFF
--- a/networkx/algorithms/traversal/depth_first_search.py
+++ b/networkx/algorithms/traversal/depth_first_search.py
@@ -291,7 +291,7 @@ def dfs_postorder_nodes(G, source=None, depth_limit=None):
     bfs_tree
     """
     edges = nx.dfs_labeled_edges(G, source=source, depth_limit=depth_limit)
-    return (v for u, v, d in edges if d == "reverse")
+    return (v for u, v, d in edges if (d == "reverse" or d == "reverse-depth_limit"))
 
 
 def dfs_preorder_nodes(G, source=None, depth_limit=None):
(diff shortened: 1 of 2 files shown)


navyagarwal:
After making the above changes, I get the expected output discussed [here](https://github.com/networkx/networkx/issues/6479#issue-1618816398).

```python
G = nx.cycle_graph(5)
list(nx.dfs_postorder_nodes(G, depth_limit=3, source=0)) #[3, 2, 1, 4, 0]
list(nx.dfs_postorder_nodes(G, depth_limit=2, source=2)) #[0, 1, 4, 3, 2]
```

These changes led to the failure of the existing test cases but on a closer look, the test cases did not seem quite right to me. So I modified the test cases to return the correct output, which is now also consistent with what is returned by `dfs_preorder_nodes` (postorder nodes have to be the same as the preorder nodes, just in a different order).

Following are the two test cases I modified, along with the correct output for `dfs_postorder_nodes` they should give based on the changes in this PR.

```python
G2 = nx.Graph()
nx.add_path(G2, [0, 1, 2, 3, 4, 5, 6])
nx.add_path(G2, [2, 7, 8, 9, 10])

#According to original test case nx.dfs_postorder_nodes(G2, depth_limit=3, source=3) returns [1, 7, 2, 5, 4, 3]

list(nx.dfs_postorder_nodes(G2, depth_limit=3, source=3)) #[0, 1, 8, 7, 2, 6, 5, 4, 3]
list(nx.dfs_preorder_nodes(G2, depth_limit=3, source=3)) #[3, 2, 1, 0, 7, 8, 4, 5, 6]
```


```python
D = nx.Graph()
D.add_edges_from([(0, 1), (2, 3)])
nx.add_path(D, [2, 7, 8, 9, 10])

#According to original test case nx.dfs_postorder_nodes(D, depth_limit=2, source=2) returns [3, 7, 2]

list(nx.dfs_postorder_nodes(D, depth_limit=2, source=2)) #[3, 8, 7, 2]
list(nx.dfs_preorder_nodes(D, depth_limit=2, source=2)) #[2, 3, 7, 8]
```

PR #6715, opened 2023-05-28
Stabilize test of approximation.connected_components

dschult:
There are Sporadic test failures when testing connected components of `AntiGraph` in `test_kcomponents.py`.
The tests checks whether ccs of G and AntiGraph(G) are the same.
But that is not true for graphs with isolated nodes or with nodes that connect to every other node.
One of the graphs being tested is generated using `gnp_random_graph` without a seed.
So occasionally it will construct a graph with a node connected to every other node.

Fix: Assign a seed to make sure the random graph doesn't produce such a node


MridulS (approved):

rossbar (approved):
Nice catch, thanks @dschult !

PR #6727, opened 2023-06-01
Unpin scipy upperbound for tests

jarrodmillman:
Follow-up to #6726. For more details, see https://github.com/scipy/scipy/issues/18603

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -138,7 +138,6 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install --pre -r requirements/default.txt -r requirements/test.txt
-          pip install scipy==1.10.1
           pip install .
           pip list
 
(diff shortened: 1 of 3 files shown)


MridulS:
Should we keep the exclude pin on the "broken" scipy versions? It's possible for some reason the pip/conda solver resolves to 1.11.0 or 1.11.1 which will break networkx in users environment. We can remove that pin once we are scipy >1.12

jarrodmillman:
> Should we keep the exclude pin on the "broken" scipy versions? It's possible for some reason the pip/conda solver resolves to 1.11.0 or 1.11.1 which will break networkx in users environment. We can remove that pin once we are scipy >1.12

You are probably correct. Let's wait until 1.11.2 is released b/c otherwise I think this PR may pass. I also don't know if we need to exclude 1.11.0, since it was yanked. But maybe we still should.

jarrodmillman:
https://pypi.org/project/scipy/1.11.2/ is out! This should be ready to merge.

I left 1.11.0 out of the list of versions to avoid since it was yanked.

MridulS:
I would vote to keep 1.11.0 in there, it can still be downloaded from pypi/conda. We can clean it up all together once we are scipy>1.12

MridulS (approved):

PR #6730, opened 2023-06-02
Add Dulmage-Mendelsohn decomposition to `algorithms.bipartite.matching`

Robbybp:
## Motivation
The Dulmage-Mendelsohn decomposition is a useful partition for bipartite graphs that I use frequently. I thought it would be nice to have it in NetworkX.

The Dulmage-Mendelsohn decomposition is a partition that defines all nodes that can possibly be unmatched in some maximum matching of a bipartite graph. It partitions nodes into the following subsets:
- Nodes reachable via alternating path from unmatched nodes in one bipartite set
- Nodes reachable via alternating path from unmatched nodes in the other bipartite set
- Nodes not reachable via alternating path from any unmatched nodes

## Design decisions
In most descriptions of the partition (e.g. [Pothen and Fan, 1990](https://dl.acm.org/doi/pdf/10.1145/98267.98287) and Chapter 7 of [Davis, 2006](https://epubs.siam.org/doi/book/10.1137/1.9780898718881)), the subsets for the two bipartite sets are kept separate, for a total of six unique subsets. However, my current implementation only returns three subsets, as if "top nodes" are known, it is easy to partition these three subsets into "top" and "bottom".

I have also made `top_nodes` a required argument so there is no ambiguity about which of the two bipartite sets each "reachable" subset corresponds to.

I'm opening as a draft for feedback on the design and implementation. This PR still needs:
- [ ] More tests

### DIFF
--- a/networkx/algorithms/bipartite/matching.py
+++ b/networkx/algorithms/bipartite/matching.py
@@ -49,6 +49,7 @@ __all__ = [
     "eppstein_matching",
     "to_vertex_cover",
     "minimum_weight_full_matching",
+    "dulmage_mendelsohn_decomposition",
 ]
 
 INFINITY = float("inf")
@@ -588,10 +589,11 @@ def dulmage_mendelsohn_decomposition(G, top_nodes, matching=None):
     """Return a partition of the nodes of ``G`` into the subsets defined by
     the Dulmage-Mendelsohn decomposition
 
-    The Dulmage-Mendelsohn decomposition [1] partitions nodes of a bipartite
+    The Dulmage-Mendelsohn decomposition [1]_ partitions nodes of a bipartite
     graph into unique subsets that determine all nodes that can possibly be
     unmatched in some maximum matching. The subsets are defined in terms of
     the bipartite sets of the graph. They are:
+
     - The unmatched nodes in one set, and all nodes reachable via an alternating
       path from them
     - The unmatched nodes in the other set, and all nodes reachable via an
@@ -603,8 +605,10 @@ def dulmage_mendelsohn_decomposition(G, top_nodes, matching=None):
     particular maximum matching. However, the subsets are independent of the
     matching used to compute them.
 
-    [1] Dulmage and Mendelsohn. "Coverings of Bipartite Graphs". Can. J. Math.,
-        1958.
+    In contrast to the other functions in this module, ``top_nodes`` is a
+    required argument, even if there is only one possible bipartition. This
+    is so there is never ambiguity about which set of nodes each "reachable"
+    subset is reachable from.
 
     Parameters
     ----------
@@ -627,6 +631,11 @@ def dulmage_mendelsohn_decomposition(G, top_nodes, matching=None):
 
         List of nodes that are not reachable by an alternating path from
         unmatched nodes.
+
+    References
+    ----------
+    .. [1] Dulmage and Mendelsohn. "Coverings of Bipartite Graphs". Can. J. Math.,
+        1958.
 
     """
     if matching is None:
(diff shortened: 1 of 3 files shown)


Robbybp:
## Potential edge cases

Want to make sure we fail gracefully in these cases (or at least document that they are not supported/handled):

- Provided graph is not bipartite, or the provided `top_nodes` are not a valid bipartite set
- Provided matching is not maximum-cardinality, or otherwise invalid

Robbybp:
> Want to make sure we fail gracefully in these cases (or at least document that they are not supported/handled):
> 
> * Provided graph is not bipartite, or the provided `top_nodes` are not a valid bipartite set
> * Provided matching is not maximum-cardinality, or otherwise invalid

I've added checks that the graph is bipartite and that `top_nodes` are a valid bipartite set, but not that the provided matching is maximum cardinality. If a user provides a matching, I say we go ahead and construct the partition.

Robbybp:
Ready for review once tests pass.

networkx/algorithms/bipartite/matching.py
```diff
@@ -582,3 +584,78 @@ def minimum_weight_full_matching(G, top_nodes=None, weight="weight"):
...
+        raise nx.NetworkXError("Provided graph is not bipartite")
+    elif not is_bipartite_node_set(G, top_nodes):
+        raise nx.NetworkXError("Provided nodes are not a valid bipartite set")
+    if matching is None:
+        matching = maximum_matching(G, top_nodes=top_nodes)
+    top_set = set(top_nodes)
+    bot_nodes = [node for node in G if node not in top_set]
+
+    unmatched_top = [node for node in top_nodes if node not in matching]
+    unmatched_bot = [node for node in bot_nodes if node not in matching]
```
dschult, line 652: Thanks for the PR! 
I haven't had a chance to go through the whole thing, but I saw this so am leaving a note.

Just like `union` there is a set operation called `difference` which would replace the list comprehensions that check `if node not in matching`. It's convenient to write as:
```python
unmatched_top = top_nodes - matching.keys()
```
It might even work without the `.keys()` part, but that depends on your python version.
Similar operations exist for `set_reachable_from_top & G.nodes` for intersection. 
I think `|` is the symbol for union.  There are functions for each of these operations like the `union` you are using.

dschult:

networkx/algorithms/bipartite/matching.py
Robbybp: Just updated to use set operators in a few places. I could also remove
```python
    unreachable = [node for node in G if node not in set_reachable]
    reachable_from_top = [node for node in G if node in set_reachable_from_top]
    reachable_from_bot = [node for node in G if node in set_reachable_from_bot]
```
and just set `unreachable = G.nodes - set_reachable`, then
```python
return set_reachable_from_top, set_reachable_from_bot, unreachable
```
The reason I went with the list comprehensions is to ensure that the returned containers have a unique order (determined by the order in `G`, which I assume is uniquely defined - correct me if I'm wrong). My understanding is that set operations, including those involving `G.nodes` such as
```python
unreachable = G.nodes - set_reachable
reachable_from_top = G.nodes & set_reachable_from_top
```
do not have a uniquely defined order as the resulting objects are sets.

Let me know if you think this is worth designing around. My first instinct would be to preserve order and return lists, but I see that some NX functions, e.g. `bipartite.sets`, return sets rather than ordered containers.

Robbybp:
> Also, since this is a new function, can you please make the keyword arguments "keyword only" by adding a *, between the positional arguments and the keyword arguments?

Just updated. Did not know that was possible.

networkx/algorithms/bipartite/matching.py
Robbybp: The order is not useful (to my knowledge) for the partition itself, but I've found it can be useful for debugging. However, any calling routine can easily use these sets to construct lists with the same elements and a well-defined order, as you suggest.

Robbybp:
@dschult This is ready for review when you have a chance

PR #6738, opened 2023-06-08
DICN Algorithm Link Prediction Implementation

jacksonburke5252:
I implemented the Direct-Indirect Common Neighbors algorithm (described here: https://www.nature.com/articles/s41598-020-76799-4#Equ14) for a class project and figured it could be added to the link_prediction.py module as it gave strong results.

I will note, however, it's computationally heavier than other link prediction algorithms offered.

This is my first open-source contribution so apologies for any errors!

### DIFF
--- a/networkx/algorithms/link_prediction.py
+++ b/networkx/algorithms/link_prediction.py
@@ -621,7 +621,6 @@ def direct_indirect_common_neighbors(G, ebunch=None):
 
         \frac{\sum_{z\in{UN_{uv}}}(N_u[z] - \overline{N_u})(N_v[z] - \overline{N_v})}{\sqrt{\sum_{z\in{UN_{uv}}}(N_u[z] - \overline{N_u})^2}\sqrt{\sum_{z\in{UN_{uv}}}(N_v[z] - \overline{N_v})^2}}
 
-
     where $N_i[z]$ is the neighborhood vector, $UN_{uv}$ is the
     union neighborhood set, and $\overline{N_u}$ is the mean of
     the values in the union neighborhood set over the vector $N_u$.
--- a/networkx/algorithms/link_prediction.py
+++ b/networkx/algorithms/link_prediction.py
@@ -3,7 +3,7 @@ Link prediction algorithms.
 """
 
 
-from math import log
+from math import log, sqrt
 
 import networkx as nx
 from networkx.utils import not_implemented_for
@@ -17,6 +17,7 @@ __all__ = [
     "ra_index_soundarajan_hopcroft",
     "within_inter_cluster",
     "common_neighbor_centrality",
+    "direct_indirect_common_neighbors",
 ]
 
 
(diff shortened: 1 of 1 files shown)


dschult:
Can you add some tests in the file `tests/test_link_prediction.py`? 

What do you mean by “it gave strong results”?  And how much slower? Are we talking about asymptotically slower, or in timing experiments?

Thanks!

jacksonburke5252:
@dschult Thanks for the response! 

To quickly address some of your questions:

- Using it as the sole feature gave strictly superior ROC AUC score to many of the other available measures (RA, JC, PA, etc.) but it was inferior to CCPA (on the network with which I was working).
- Speed is asymptotically slower; I believe the time complexity is O(N^2). I'll look again if there are ways to reduce this.

I'll follow-up in a week or so with a) tests added to the `tests/test_link_prediction.py` file and b) timing / performance experiments!

jacksonburke5252:
Hi @dschult, 

I've added tests to the file you mentioned!

In addition to that, I ran two performance tests on the following five networks:

- Cora Citation Network
- CiteSeer Citation Network
- Zach's Karate Club
- Co-authorship in network science
- Dolphin social network

Each test was run 10 times, and the average AUC, standard deviation AUC, and average runtime (seconds) are reported.

The first test randomly removes 20% of edges from the graph, generates prediction scores for those 20% of true edges and **all** non-edges from the original graph, and then computes the AUC score. The algorithms tested are Jaccard Coefficient, Preferential Attachment, Resource Allocation, Adamic Adar, Common Neighbor Centrality, and Direct-Indirect Common Neighbors (my proposal.) As you can see, AUC score increases significantly, but as mentioned, runtime is far slower. 

Test_AllEdges

The second test does the same, except rather than generate prediction scores for all non-edges, I randomly select the same number of non-edges as true test edges. Therefore, the number of true edges and false edges used in calculating the AUC score is equal. Again, AUC score is far superior, but in this case, runtime is not an issue.

Test_BalancedEdges

I believe the slowdown comes from the need to call "generate_union_neighborhood_set" each time a new pair of nodes is considered, and unfortunately, I don't think there's a way to eliminate this step or do it all at once. One idea would be to use NumPy (or maybe even Numba?) to vectorize the operations and create a 3-d matrix all at once, but I avoided using anything beyond what was already imported. Would you be interested in me testing with either or both of these libraries to see if there is an efficiency increase? 

Another thought is maybe it's possible to just include a note mentioning the time inefficiency in the documentation.

Thanks!

jacksonburke5252:
Thanks a lot for the feedback and suggestions. Here's what I attempted (some successful, others not):

- Attempted keeping the nodes as is rather than converting to integers (0, n-1), but did not experience speedup. Ultimately, I think the index list access isn't especially expensive, and since indices are necessary in certain parts of the algorithm, I end up double-looking up in a dictionary (node to index mapping, then index mapping to neighborhood vector) with your suggestion (if I understood it correctly) which results in longer runtime.
- `sum(1 for _ in iterator)` changes were implemented throughout and resulted in incremental speed up
- Didn't use `numpy.corrcoef` as was mentioned, but use of numpy array in general allowed for vectorized algebra instead of for loops which significantly increased the efficiency of the code. 
- Lastly, included an if, else if the reverse of the correlation_coefficient had been created before (since they are symmetric) and this also reduced runtime.

New results for the full edge list are below (other results table omitted as it was essentially equivalent)

Screenshot 2023-06-20 at 9 49 21 PM

As you can see, efficiency greatly increased; although, it's still not on par with the other algorithms with the larger networks.

Interested to hear if you think there are other efficiency gains to be made here!

Also, I'm having trouble passing the tests with the import of numpy to the module. Do you have any suggestions on how to address that?

Thanks!

dschult:
You are right that maintaining the nodes instead of integers doesn't help once you take advantage of numpy for the matrix algebra. And I think you can take much more advantage of numpy.

The matrix of the number of common neighbors is W=A.T@A for directed graphs (works for undirected too, but the transpose isn't strictly needed), where A is the `nx.adjacency_matrix(G)`.  I called it W because it is also the number of 2-step walks from u to v.

Find the correlation coefficient using: `C = numpy.corrcoeff(W.toarray())`
The prediction is: pred = (1 + W) * (1 + C)
followed by `_apply_prediction(G, pred.__getitem__, ebunch)`

I haven't tested it, so check it... Each step seems right and it should be much faster....

```python
mapping = {node: i for i, node in enumerate(G)}
A = nx.adjacency_matrix(G)  # does the mapping to indexes correctly without any fuss
W = A.T @ A
C = np.corrcoeff(W.toarray())
pred = (1 + W) * (1 + C)
def get_pred(u, v):
    return pred[mapping[u], mapping[v]]

return _apply_prediction(G, get_pred, ebunch)
```

jacksonburke5252:
Makes a lot of sense. I need to focus on a few things for school, but will be back to this soon. Thanks a lot for the help on this!

dschult:
Regarding the tests failing when numpy is not present:

The CI tests are set up to test without any additional packages. This helps us ensure that the main parts of networkx work without any requirements.  That means we have to add some special handling of the imports and tests whenever `numpy` and `scipy` (and others) are used.

This should be added as a section of the contributor's guide. So I'll put some text here and I intend to put it into the `CONTRIBUTING.rst` file after including any feedback you might add in comments below.

- See the contributing guide for standard import names, e.g. `import numpy as np`
- put the import statements inside the functions which use them. That delays the import until it is needed.
- In the tests, use `np = pytest.importorskip("numpy")`.  Note that this function is pronounced "import-or-skip". If it can import numpy, it attaches that module to the name `np`. If it can't, all tests in this scope are skipped. Please this `importorskip` call so that it affects only tests that call a function involving numpy (or that use numpy for the test). The idea is to skip functions that will fail because numpy isn't available. Usually this is either near the start of a test module, near the start of a test function, or near the start of a test class. That will skip tests in that module, function or class respectively.
- Add a line to the `networkx/conftest.py` section which defines "needs_numpy` or `needs_...`.

jacksonburke5252:
Hi @dschult, 

Sorry about the delay!

I looked into your suggestion above about using more numpy functionality but am unsure about its feasibility. I understand the solution you propose, but there are two main issues.

1. In terms of creating the neighbor_vectors matrix with A @ A.T instead of the generate_neighborhood_vectors function, I believe it would work, except for the fact that the goal is to not just identify nodes which are first and second degree neighbors, but there is also logic applied on top of these relationships (e.g. if first-order neighbors, add the number of common neighbors + 1, etc.). Additionally, this part of the algorithm isn’t responsible for the inefficiency. It’s more so the repeated calling of generate_union_neighborhood_set.
2. Regarding calling `numpy.corrcoef`, your idea makes a lot of sense; however, one peculiarity about the algorithm is that it computes the correlation coefficient only over indices where at least one of the node neighborhood vectors is greater than 0. If we call`numpy.corrcoef `directly on the neighbor_vectors matrix, it would include all indices, including those in which there are 0’s, which will greatly affect the eventual scores. Each neighbor_vector pair is going to therefore have a different number/location of indices over which the correlation coefficient is being computed. Let me know if you have a suggestion for how to get around this, or if this doesn't make sense.

I still attempted to use the `numpy.corrcoef` method on the neighbor_vectors for each node pair, but oddly, it ran about 1.5x slower in the entire function than the rote calculation of the correlation coefficient. 

To be more specific, I test these two functions against each other and present representative accuracy and timing results for one run on citeseer:

**Updated Function**
```
    def compute_correlation_coefficient(G, u, v, neighbor_vectors):
        if (v, u) in correlations_coefficients:
            return correlations_coefficients[(v, u)]
        else:
            u_neighborhood_vector = neighbor_vectors[u]
            v_neighborhood_vector = neighbor_vectors[v]
            union_neighborhood_set = generate_union_neighborhood_set(
                G, u_neighborhood_vector, v_neighborhood_vector
            )
            u_neighborhood_vector = u_neighborhood_vector[union_neighborhood_set]
            v_neighborhood_vector = v_neighborhood_vector[union_neighborhood_set]

            # correlation computation below
            correlation_matrix = np.corrcoef(u_neighborhood_vector, v_neighborhood_vector)
            correlation_coefficient = correlation_matrix[0, 1]
            if np.isnan(correlation_coefficient):
                return 0.0
            else:
                return correlation_coefficient
```

Screenshot 2023-07-05 at 7 14 00 PM

-------------------------
**Original Function**

```
    def generate_union_neighborhood_set(G, u_neighborhood_vector, v_neighborhood_vector):
        return np.where((u_neighborhood_vector + v_neighborhood_vector) > 0)[0]

    def compute_correlation_coefficient(G, u, v, neighbor_vectors):
        if (v, u) in correlations_coefficients:
            return correlations_coefficients[(v, u)]
        else:
            u_neighborhood_vector = neighbor_vectors[u]
            v_neighborhood_vector = neighbor_vectors[v]
            union_neighborhood_set = generate_union_neighborhood_set(
                G, u_neighborhood_vector, v_neighborhood_vector
            )
            u_neighborhood_vector = u_neighborhood_vector[union_neighborhood_set]
            v_neighborhood_vector = v_neighborhood_vector[union_neighborhood_set]
            
            # correlation computation below
            if len(union_neighborhood_set) == 0:
                u_vector_average, v_vector_average = 0, 0
            else:
                u_vector_average = np.mean(u_neighborhood_vector)
                v_vector_average = np.mean(v_neighborhood_vector)

            u_diff = u_neighborhood_vector - u_vector_average
            v_diff = v_neighborhood_vector - v_vector_average
            numerator = np.sum(u_diff * v_diff)
            u_denominator_sq = np.sum(u_diff**2)
            v_denominator_sq = np.sum(v_diff**2)
            if u_denominator_sq == 0 or v_denominator_sq == 0:
                correlations_coefficients[(u, v)] = 0.0
                return 0.0
            else:
                u_denominator = np.sqrt(u_denominator_sq)
                v_denominator = np.sqrt(v_denominator_sq)
                denominator = u_denominator * v_denominator
                correlation_coefficient = numerator / denominator
                correlations_coefficients[(u, v)] = correlation_coefficient
                return correlation_coefficient
```

Screenshot 2023-07-05 at 7 21 32 PM

----------------
Do you think the timing on the original function is still too inefficient? I'm definitely open to experimenting with other methods to bring down its complexity if you have any suggestions. Otherwise, as mentioned earlier, I could add a note in the documentation on its inefficiency with large networks.

Beyond this, I'll work on integrating your comments w/r/t numpy and push an updated version which hopefully passes all of the tests.

Thanks!

jacksonburke5252:
@dschult Regarding the NumPy import instructions, most of what you've explained makes sense and seems to work! 

The one difference between what I implemented and your instructions was the `np = pytest.importorskip("numpy")` assignment. I tested this out, and it didn't work so I looked to what had been implemented for `test_node_classification.py` which was to simply run `pytest.importorskip("numpy")` at the top of the module, without the assignment to `np`. Now it passes all of the tests.

jacksonburke5252:
Hi @dschult,

Following up on the above! No real rush on any of it, just wanted to flag the updates again if it's helpful. (Apologies if that's out of practice, first contribution as I mentioned.) 

Thanks, 
Jackson

jacksonburke5252:
Hi @dschult - Any interest in this contribution still? Please let me know, and I'll be happy to make updates or to close it.

PR #6741, opened 2023-06-12
Bugfixes and performace improvement in community.LFR_benchmark_graph

woQal:
Fixes infinite while loops in main `LFR_benchmark_graph` function during adding edge. Add `available_in` and `available_out` variables which contain the available nodes for connecting inside and outside community. Checks, if the number of available nodes suffices to meet the number of in- and out-community edges.

Improve performance during adding edges -- add edges from a list of possible nodes and update the list.

Improve performance during generating communities -- check if there are communities big enough to hold the node. Choose the node community from the list of available enough-sized communities.

### DIFF
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -1065,6 +1065,7 @@ def LFR_benchmark_graph(
             while G.degree(u) < s:
                 v = seed.choice(available_in)
                 G.add_edge(u, v)
+                available_in = list(set(available_in).difference({v}))
 
             # Filter non-neighbor nodes outside community
             available_out = list((all_nodes.difference(c)).difference(neigh))
@@ -1076,5 +1077,6 @@ def LFR_benchmark_graph(
             while G.degree(u) < deg_seq[u]:
                 v = seed.choice(available_out)
                 G.add_edge(u, v)
+                available_out = list(set(available_out).difference({v}))
             G.nodes[u]["community"] = c
     return G
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -777,18 +777,19 @@ def _generate_communities(degree_seq, community_sizes, mu, max_iters, seed):
     result = [set() for _ in community_sizes]
     n = len(degree_seq)
     free = list(range(n))
+    all_c = range(len(community_sizes))
     for i in range(max_iters):
         v = free.pop()
-        c = seed.choice(range(len(community_sizes)))
-        # s = int(degree_seq[v] * (1 - mu) + 0.5)
         s = round(degree_seq[v] * (1 - mu))
-        # If the community is large enough, add the node to the chosen
-        # community. Otherwise, return it to the list of unaffiliated
-        # nodes.
-        if s < community_sizes[c]:
-            result[c].add(v)
-        else:
-            free.append(v)
+        # Choose a community from communities which are large enough
+        available_c = list(filter(lambda x: community_sizes[x] > s, all_c))
+        if len(available_c) == 0:
+            msg = "Could not assign communities; try increasing min_community"
+            raise nx.ExceededMaxIterations(msg)
+        c = seed.choice(available_c)
+        # Add the node to the chosen community
+        result[c].add(v)
+
         # If the community is too big, remove a node from it.
         if len(result[c]) > community_sizes[c]:
             free.append(result[c].pop())
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -1049,14 +1049,32 @@ def LFR_benchmark_graph(
     # inter-community degrees.
     G = nx.Graph()
     G.add_nodes_from(range(n))
+    all_nodes = set(range(n))
     for c in communities:
         for u in c:
-            while G.degree(u) < round(deg_seq[u] * (1 - mu)):
-                v = seed.choice(list(c))
+            neigh = set(G.neighbors(u))
+            s = round(deg_seq[u] * (1 - mu))
+
+            # Filter non-neighbor nodes in community
+            available_in = list(c.difference(neigh))
+            if len(available_in) + len(neigh.intersection(c)) < s:
+                msg = "Could not add edges; try increasing mu"
+                raise nx.exception.ExceededMaxIterations(msg)
+
+            # Choose nodes from available nodes in community
+            while G.degree(u) < s:
+                v = seed.choice(available_in)
                 G.add_edge(u, v)
+
+            # Filter non-neighbor nodes outside community
+            available_out = list((all_nodes.difference(c)).difference(neigh))
+            if len(available_out) + G.degree(u) < deg_seq[u]:
+                msg = "Could not add edges; try decreasing max_degree"
+                raise nx.exception.ExceededMaxIterations(msg)
+
+            # Choose nodes
             while G.degree(u) < deg_seq[u]:
-                v = seed.choice(range(n))
-                if v not in c:
-                    G.add_edge(u, v)
+                v = seed.choice(available_out)
+                G.add_edge(u, v)
             G.nodes[u]["community"] = c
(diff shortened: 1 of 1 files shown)


dschult:
Perhaps I'm not understanding... I think there aren't infinite loops in the main function.  The loops are wrapped inside a for-loop which can at most iterate `max_iters` times.  I think you are trying to make it easier to stay within the `max_iters`.

My second question: Can you justify how/why this gives the same probabilities of outputs? In particular, does this rely on the pop order of the `results` sets?

My third question: Does this take longer?  Certainly each iteration takes longer because you must check each community's size against `s` on each iteration. The current code only checks the chosen community's size.  In what sense is this more performant?

Finally, what is the "bug" that this fixes?

Thanks!

woQal:
Hi,
# The first three questions
I have changed the implementation to the original, because it just served for my purposes (testing on edge cases) and in general cases, it is not useful (it only improved the success rate of the algorithm a bit, but slowed it, sorry for the misconception :(.

# The last question:
I mean the while loops here (the following is original code):
```
def LFR_benchmark_graph(...):
    ... some code ...

    for c in communities:
        for u in c:
            while G.degree(u) < round(deg_seq[u] * (1 - mu)):
                v = seed.choice(list(c))
                G.add_edge(u, v)
            while G.degree(u) < deg_seq[u]:
                v = seed.choice(range(n))
                if v not in c:
                    G.add_edge(u, v)
            G.nodes[u]["community"] = c
```

## First loop and `available_in`
The first loop is running while `G.degree(u) < intended no of intra edges`, but this is wrong, because in G.degree(u) can contain nodes not in the community. For example -- having two comms `[0, 1, 2], [3, 4, 5]`, we add edges (both in- and out- community) and if we added an out-community edge from `0` to `3`, this loop just checks "ok, there is one edge from `3`, thats ok" (supposing that the condition is equal to $1$).

Further -- yes, it is not necessary to check this condition:
```if len(available_in) + len(neigh.intersection(c)) < s:```
because it is already checked during the nodes assignment to communities. I deleted the condition in the updated branch.

But it is necessary to have there the condition (I modified it in the updated branch):
```
            in_neigh = set(G.neighbors(u)).intersection(c)
            while len(in_neigh) < s:
                ...
                in_neigh.add(v)
```
because the `G.degree(u)` can contain out-community nodes.

```
            available_in = list(c.difference(in_neigh))
```
I added the `available_in` because of the edge cases in which even the original loop can run very, very long (depending on the probability). Supposing we would have community of size $10$ and we would have a node in that community which should have $9$ in-community edges. Then we will want to choose every node in community, but choosing the nodes (especially the last one) can take very long (it happened to me even on some graphs with size $500$ nodes, that it took noticeably longer.

## Second loop and `available_out`
Moving on to:
```
            available_out = list((all_nodes.difference(c)).difference(set(G.neighbors(u))))
            if len(available_out) + G.degree(u) < deg_seq[u]:
                msg = "Could not add edges; try decreasing max_degree"
                raise nx.exception.ExceededMaxIterations(msg)

            while G.degree(u) < deg_seq[u]:
                v = seed.choice(available_out)
                G.add_edge(u, v)
                available_out = list(set(available_out).difference({v}))
```
The condition which checks `if len(available_out) + G.degree(u) < deg_seq[u]` is neccesary. Supposing the following scenario: we have a network with $25$ nodes, a community `c` of size $15$, a node `u` in that community with `G.degree(u) = 5` (and all its neighbors are in community, the condition `round(deg_seq[u] * (1 - mu)) = s = 5`) and the `deg_seq[u] = 20`. We want to add $15$ edges to that node. But we only have $10$ out-community nodes. You may say that `deg_seq[u] > |c|`, but the community generator only checks if `s < |c|` which holds. And also, it can happen that `max_generated_degree > max_generated_community`, because the power law sequence generator does **not** ensure, if there are communities of sizes near `max_community` (this is only an upper bound on the community sizes).

In this case, the original loop is an infinite loop (happens for example for the following settings `LFR_benchmark_graph(n=100, tau1=2, tau2=1.5, mu=0.5, average_degree=10, min_community=10)`, almost every time).

Moving on to the `available_out`, the reason is the same as for `available_in`. Also, in the original loop there was condition 
`if v not in c: G.add_edge(u, v)` which can be very bad for big communities (if the community contains almost all nodes, the loop will choose the "right" out-community nodes rarely.

Thanks for your review

**Comment by woQal** [comment]
Fixes #6762 and #6763

**Review by rossbar — COMMENTED** [review]
I think the thing that would be most effective in pushing this forward would be to add minimal reproducing examples to the test suite to demonstrate the "fix". The linked issues contain some contextual info, but it's still not clear to me whether/what the "bug" actually is. Tests that fail (or hit infinite loops) on `main` would do so!

**Comment by dschult** [comment]
I have tracked down the possible infinite loop in LFR_benchmark_graph -- a reproducible example is:
```python
nx.LFR_benchmark_graph(n=100, tau1=2, tau2=1.5, mu=0.5, min_degree=10, min_community=10, seed=7)
```
Unfortunately, I don't see a way to make this into a test because changing the seed fixes the problem. So rearranging the code will also fix the problem.

While trying that example (without setting `seed`) many times I found that it usually raises an exception: `ExceededMaxIterations: Could not assign communities; try increasing min_community`. That exception occurs at the stage of community creation rather than the stage of adding edges to match those communities.  Occasionally a graph would be returned. And occasionally the computation would take a long time. Some additional print statements show that it is an infinite loop occurring when the number of out-community edges needed is larger than the number of out-community nodes.

Thank you @woQal for reporting this!

This PR addresses the potential for an infinite loop (#6763) by checking to see if there are enough potential neighbors, and if not, raising. So that is good -- but it means that most of the time with these parameter values we will get a raised exception.

The other issue here (stated in #6762) is not about infinite loops, but about correctness. To summarize, when adding in-community edges to a node, we are counting both in-community and out-community edges and comparing that to the expected number of in-community edges. So we have a mismatch between what we are counting and what we are trying to match.

While investigating this, I noticed that during community generation, the algorithm calls for a random node to be removed from a community if that community is "complete" (which I think must mean as large as it needs to be). But we use the set method `pop` to remove it which is not random and will not provide the mixing we need to explore the solution space.

I feel that the solution to the infinite loop problem is to replace the while loops which repeatedly add an edge if some condition is met. We can replace them with a random sample of the right number of neighbors from the set of all possible neighbors that meet the conditions. This also allows us to check whether there are enough potential neighbors to satisfy the additional nodes needed.

The fix for the counting friend-vs-other edges can be addressed by adding all internal edges first, before adding the external edges. The approach proposed here is good -- but selects a new_neighbor which is then added to a set `in_neigh`. But if the new_neighbor is in the set already, we don't add any edges. That could lead to an infinite loop similar to the existing one. So let's keep the `in_neigh` set, but use it to select new edges to add.

The third fix is to select a random node to remove from the community rather than an arbitrary node in `_generate_communities`.

Even after these 3 changes I see two real problems with this tool as we currently implement it.
- there are a large number of exceptions raised when it makes better sense to try another iteration.  The `max_iters` keyword is used 2-3 times here, but not while connecting the edges to match the selected degree sequence and community sizes. Let's try (in another PR) to wrap the process in a larger loop that tries multiple iterations of the whole process -- including creating the edges.
- it is not clear how our approach corresponds to the published article. A comment in the doc_string says that the article connects edges using a configuration model and then applies the community structure on top of that. And it claims that this "should be" effectively the same outcome as the way we do it: constructing the communities first and then adding the edges. It would be good to justify an equivalence between the methods or just implement the version from the paper. At the least there should be some justification for why we are doing it this way. Maybe we could create a nx-guides which reproduces some of the paper's figures using our method. Are the results similar? :) 

I think the three fixes mentioned above are very similar to what is in this PR. So I will push them up here for further discussion. @woQal please push back on this if you think it shouldn't be done this way.

Issue #6728, opened 2023-06-02
Bug in intersection?


llambourne:
### Current Behavior

`intersection` with two `Graph`s depends on the order of the nodes in the edges

### Expected Behavior

The `Graph` class is supposed to be undirected, so the order shouldn't matter.

### Steps to Reproduce

```python
import networkx as nx
a = nx.Graph([('A', 'B')])
b = nx.Graph([('B', 'A')])
c = nx.Graph([('A', 'B')])
print(nx.intersection(a, b).edges)
print(nx.intersection(a, c).edges)
```
returns
```
[]
[('B', 'A')]
```

### Environment

Python version: 3.9.13
NetworkX version: 3.1

dschult:
That is certainly a bug.  I can verify that the code treats sets of edges without caring about the directed nature of the edges. 

The place to fix this is in `intersection_all` in the `networkx/algorithms/operators.py` module. The line setting `G_edges_set` has two potential fixes.
1. Add two lines after the `G_edges_set` is created:
```python
if G.is_directed():
    G_edges_set.update((v, u) for u, v in list(G_edges_set))
```
or
2. use `frozenset` to make each node-pair into an unordered set.
```python
G_edges_set = {frozenset(e) for e in (G.edges(keys=True) if G.is_multigraph() else G.edges)}
...
later
...
R.add_edges_from((u, v) for u, v in edge_intersection)
```

PR #6744, opened 2023-06-14
Fix intersection_all method

navyagarwal:
Fix for issue #6728 

Implemented the suggested fix for `intersection_all` method.

### DIFF
--- a/networkx/algorithms/operators/all.py
+++ b/networkx/algorithms/operators/all.py
@@ -208,7 +208,10 @@ def intersection_all(graphs):
 
     for i, G in enumerate(graphs):
         G_nodes_set = set(G.nodes)
-        G_edges_set = set(G.edges(keys=True) if G.is_multigraph() else G.edges())
+        G_edges_set = {
+            frozenset(e)
+            for e in (G.edges(keys=True) if G.is_multigraph() else G.edges())
+        }
         if i == 0:
             # create new graph
             R = G.__class__()
@@ -226,6 +229,6 @@ def intersection_all(graphs):
         raise ValueError("cannot apply intersection_all to an empty list")
 
     R.add_nodes_from(node_intersection)
-    R.add_edges_from(edge_intersection)
+    R.add_edges_from((u, v) for u, v in edge_intersection)
 
     return R
--- a/networkx/algorithms/operators/all.py
+++ b/networkx/algorithms/operators/all.py
@@ -67,6 +67,8 @@ def union_all(graphs, rename=()):
         if i == 0:
             # Union is the same type as first graph
             R = G.__class__()
+        elif G.is_directed() != R.is_directed():
+            raise nx.NetworkXError("All graphs must be directed or undirected.")
         elif G.is_multigraph() != R.is_multigraph():
             raise nx.NetworkXError("All graphs must be graphs or multigraphs.")
         elif not seen_nodes.isdisjoint(G_nodes_set):
@@ -166,6 +168,8 @@ def compose_all(graphs):
         if i == 0:
             # create new graph
             R = G.__class__()
+        elif G.is_directed() != R.is_directed():
+            raise nx.NetworkXError("All graphs must be directed or undirected.")
         elif G.is_multigraph() != R.is_multigraph():
             raise nx.NetworkXError("All graphs must be graphs or multigraphs.")
 
@@ -219,6 +223,8 @@ def intersection_all(graphs):
             R = G.__class__()
             node_intersection = G_nodes_set
             edge_intersection = G_edges_set
+        elif G.is_directed() != R.is_directed():
+            raise nx.NetworkXError("All graphs must be directed or undirected.")
         elif G.is_multigraph() != R.is_multigraph():
             raise nx.NetworkXError("All graphs must be graphs or multigraphs.")
         else:
@@ -232,5 +238,6 @@ def intersection_all(graphs):
 
     R.add_nodes_from(node_intersection)
     R.add_edges_from(edge_intersection)
+    R.graph.clear()  # remove graph attributes from intersection graph
 
     return R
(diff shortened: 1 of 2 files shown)


dschult:
Ack! I think my suggestions was not helpful -- it doesn't work for Multigraphs and it doesn't do the right thing for DiGraphs either...  :{     But this PR passed all the tests, so I wonder if we are even testing these functions on multigraphs!?!

The reason I think this suggestion will not work is:
- The multigraph edges are 3-tuples (u, v, edgekey) and this code turns that into a frozenset so we lose the info about which is the key and which are the nodes.
- When turning it back into a tuple we unpack it to `u, v` which will raise an exception if it is a 3-tuple.

Maybe the other suggested method would work -- but I already see a mistake in that one too:
- the if statement should add the other directions for the **undirected** case, not for the directed case.

Also, I see an improvement that avoids the checking for multigraph or not. Iterating over `G.edges` results in 2-tuples for non-multigraph and 3-tuples for multigraphs. (Unlike `G.edges()` which yields 2-tuples for both multigraph and graph.)

Would this work?
```python
    G_edges_set = set(G.edges)
    if not G.is_directed():
        G_edges_set.update((v, u) for u, v in list(G_edges_set))
```

And it looks like we need better tests -- of simple multigraph and directed graphs...   Like path_graph(4) and path_graph(range(2, 7)) with one repeated edge added for the multigraph case.  Maybe 4 nodes is even more than we need.
:)

navyagarwal:
After some hard look, I have a few observations - 

1. This PR passed all the tests even though the code does not actually work properly on MultiGraphs because the MultiGraph test cases do not cover the scenario where `(u, v, k1)` and `(u, v, k2)` both are part of the intersection graph. I have modified an existing test case to include that.

2. I used your other suggestion to correct the `intersection_all` method, but when G is undirected, I had to add another condition for updating `G_edges_set` in case G is a MultiGraph because then we have to unpack as `(v, u, k) for u, v, k in list(G_edges_set)`

3. As mentioned in point 1, I have modified an existing test case for MultiGraph so that it covers an edge case. I have also added a new test case for directed graphs.

4. I also observed that while the new graph returned by `intersection_all` does not retain the attributes for nodes and edges as specified in the documentation, it does retain the graph attributes of the second graph (perhaps the last graph in the list in case of multiple graphs). This can be seen from the following code - 

```python
g = nx.Graph()
g.add_node(0, x=4)
g.add_node(1, x=5)
g.add_edge(0, 1, size=5)
g.graph["name"] = "g"

h = g.copy()
h.graph["name"] = "h"
h.graph["attr"] = "attr"
h.nodes[0]["x"] = 7

gh = nx.intersection_all([g, h])
print(gh.nodes(data=True))  # [(0, {}), (1, {})]
print(gh.edges(data=True))  # [(0, 1, {})]
print(gh.graph)  # {'name': 'h', 'attr': 'attr'}

```
The attribute dictionaries for nodes and edges are empty which is the correct behavior. But the attribute dictionary for the resultant graph contains graph attributes of `h`.

navyagarwal:
I just saw issue #6750, which also points out the graph attribute problem mentioned in point 4, any thoughts on whether the behavior should be modified to match the docstring or should the docstring match the behavior?

dschult:
Thanks for this!

I prefer these operations to not copy the graph/node/edge attributes. We could add an example to the doc_string showing them how to update the resulting graph with attributes if desired.  Some code like the following, which adds the minimum attribute for each node across all graphs could work... but maybe there is a better example. We could say in the doc_string that similar ideas work for `set_edge_attributes` and setting the graph attributes.

```python
        >>> g = nx.Graph()
        >>> g.add_node(0, capacity=4)
        >>> g.add_node(1, capacity=5)
        >>> g.add_edge(0, 1)

        >>> h = g.copy()
        >>> h.nodes[0]["capacity"] = 7
 
        >>> j=g.copy()
        >>> j.nodes[1]["capacity"] = 3

        >>> graphs = [g, h, j]
        >>> ghj = nx.intersection_all(graphs)

        >>> inf = float("inf")
        >>> attr = 'capacity'
        >>> new_node_attr = {n: min(*(anyG.nodes[n].get(attr, inf) for anyG in graphs)) for n in gh}
        >>> nx.set_node_attributes(gh, new_node_attr, 'new_capacity')
        >>> gh.nodes(data=True)
        NodeDataView({0: {'new_capacity': 4}, 1: {'new_capacity': 3}})
```

This would fix #6750 for intersection_all.  Would should/could also fix the `union_all` function. And I think we should check the `compose_all` function. But those could be in a different PR if you prefer.

Other comments:
1&3:  Thanks for fixing the multigraph behavior, making it work in all cases and fixing the tests as well... :)
2: Good catch!  There should be different handling of multigraph and graph in that spot.
4: (see discussion above)

I also noticed while looking through this that we don't check for taking the intersection of a DiGraph with a Graph. I suspect that should raise a ValueError. I think these operators should not be trying to handle input lists with both DiGraph and Graph.  Could you add code to check whether all are directed or not directed? And I guess that should include small tests too (could be very simple graphs).

networkx/algorithms/operators/all.py
```diff
@@ -232,5 +238,6 @@ def intersection_all(graphs):
 
     R.add_nodes_from(node_intersection)
     R.add_edges_from(edge_intersection)
+    R.graph.clear()  # remove graph attributes from intersection graph
```
dschult, line 241: Instead of removing the R.graph attributes, we can remove line 234 (`R.graph.update`). That avoids adding them just to remove them again.

navyagarwal: Ah yes, this is better!

navyagarwal:
I am not sure why the test cases are failing and the following error is raised

> ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.

Except for that, I have made all the changes and this PR is ready for a final review.

dschult (approved):
This looks good to me!
Thanks @navyagarwal

MridulS (approved):

Issue #6754, opened 2023-06-22
Pre-commit setup of nodejs results in error as nodejs cannot find GLIBC on Ubuntu 18.04


paulbrodersen:
### Current Behavior

Running the pre-commit hook 

```shell
pre-commit run --all-files
```

results in an error. Full error trace included in the [pre-commit.log](https://github.com/networkx/networkx/files/11833025/pre-commit.log).
The issue arises possibly to [due the version of Node.js being used](https://stackoverflow.com/questions/72921215/getting-glibc-2-28-not-found). 

### Expected Behavior

Running the pre-commit should not result in an error.

### Steps to Reproduce

Follow the steps outlined in CONTRIBUTING.rst exactly. 

### Environment

Python version: 3.10
NetworkX version: 3.2rc0.dev0

jarrodmillman:
I am not seeing this error on my computer or on our CI system.

What version of pre-commit are you using? Did you try the steps listed in the StackOverflow article?

paulbrodersen:
pre-commit 3.3.3, and no, as I assumed node was being downloaded in the temporary virtualenv being created. I don't use node so I haven't ever installed it (knowingly).

MridulS:
This is probably coming from our `prettier` pre-commit hook and nothing to do with black. There is indeed an issue that could have happen with Windows based systems https://github.com/pre-commit/mirrors-prettier/issues/18#issuecomment-1135071907 but I can't reproduce/debug this for a linux based one.

Could you please try manually removing prettier pre-commit hook from the .pre-commit-config.yml file in the root directory of networkx? After removing it should look something like this.


```
repos:
  - repo: https://github.com/psf/black
    rev: 23.3.0
    hooks:
      - id: black
  - repo: https://github.com/adamchainz/blacken-docs
    rev: 1.13.0
    hooks:
      - id: blacken-docs
  - repo: https://github.com/charliermarsh/ruff-pre-commit
    rev: v0.0.258
    hooks:
      - id: ruff
        args:
          - --fix
 ```

Does this bit run for you?

MridulS:
Also how recent is your OS distribution? Node 18+ prebuilt binaries need a new version of glibc.

paulbrodersen:
@MridulS 

> Does this bit run for you?

It does. No errors or (undue) warnings when making commits with the new config. 

> Also how recent is your OS distribution? Node 18+ prebuilt binaries need a new version of glibc.

[...]

Issue #355, opened 2012-07-20
Fractal Dimension of Graphs (migrated from Trac #359)


networkx-trac:
Original ticket https://networkx.lanl.gov/trac/ticket/359
Reported 2010-06-08 by @bjedwards, assigned to @bjedwards.

Two functions to calculate the fractal dimension of a Graph. References provided in the code, no unit tests yet.

hagberg:
Closed due to inactivity.

PR #6755, opened 2023-06-22
Implement generators for fractal graphs

paulbrodersen:
This PR implements two generators for fractal graphs, specifically the Sierpinski gasket graph and the Laakso graph. 

Figure_1

Apart from being incredibly pretty, they have a number of interesting properties (see articles referenced in the doc-strings), and for several measures of "fractalness", the exact value is known / can be computed analytically. These generators may become useful for testing purposes, if any of these measures are eventually implemented in networkx (see previous efforts [here](https://github.com/networkx/networkx/issues/355)). 

## Notes

I was unable to run the black with the pre-commit hook due to an error described [here](https://github.com/networkx/networkx/issues/6754). However, all tests are passing with pytest on my local machine.


paulbrodersen:
The test environment for Ubuntu does not seem to have numpy installed. Not sure how to address that.

dschult:
The CI tests are set up to test without any additional packages. This helps us ensure that the main parts of networkx work without any requirements.  That means we have to add some special handling of the imports and tests whenever `numpy` and `scipy` (and others) are used.

This should be added as a section of the contributor's guide. So I'll put some text here and I intend to put it into the `CONTRIBUTING.rst` file after including any feedback you might add in comments below.

- See the contributing guide for standard import names, e.g. `import numpy as np`
- put the import statements inside the functions which use them. That delays the import until it is needed.
- In the tests, use `np = pytest.importorskip("numpy")`.  Note that this function is pronounced "import-or-skip". If it can import numpy, it attaches that module to the name `np`. If it can't, all tests in this scope are skipped. Please this `importorskip` call so that it affects only tests that call a function involving numpy (or that use numpy for the test). The idea is to skip functions that will fail because numpy isn't available. Usually this is either near the start of a test module, near the start of a test function, or near the start of a test class. That will skip tests in that module, function or class respectively.
- Add a line to the `networkx/conftest.py` section which defines "needs_numpy` or `needs_...`.

paulbrodersen:
@dschult Thanks, that was very helpful.

> after including any feedback you might add in comments below

> [P]ut the import statements inside the functions which use them. That delays the import until it is needed.

That was easy enough.

> In the tests, use `np = pytest.importorskip("numpy")`. Note that this function is pronounced "import-or-skip". If it can import numpy, it attaches that module to the name `np`. If it can't, all tests in this scope are skipped. Please [use] this `importorskip` call so that it affects only tests that call a function involving numpy (or that use numpy for the test). The idea is to skip functions that will fail because numpy isn't available. Usually this is either near the start of a test module, near the start of a test function, or near the start of a test class. That will skip tests in that module, function or class respectively.

I think this would benefit from stricter (and hence) simpler requirements and some examples.  I would amend the last two sentences to:

[...] Usually this import should occur at the start of a test function or at the start of a test class:

```python
def test_my_function_that_depends_on_numpy():
    ... # remainder of test
```

If all tests in the test module require numpy, import numpy with `pytest.importorskip` at the start of the module, but after all other imports:

```python
import pytest

import networkx as nx

def test_my_function_that_depends_on_numpy():
    ...


def test_my_other_function_that_depends_on_numpy():
    ...
```

> Add a line to the networkx/conftest.py section which defines `needs_numpy` or `needs_...`.

That was clear once I opened the file.

---

Also, while we are here:

1. CONTRIBUTING.rst should probably also include some guidance on writing doc-tests.

   E.g., I just noticed that `conftest.py` specifies that "doctests always assume ``import networkx as nx`` so don't add that".

2. doc/README.md could use some love. I will raise another set of issues shortly.

paulbrodersen:
I added some documentation, namely the sphinx `automodule` and `autosummary` entries to `generators.rst`. However, due to one of the issues linked above, I can't currently build the documentation to test and determine the output. There may hence be formatting issues in the doc-strings.

paulbrodersen:
I finished making changes to the documentation. Locally, everything looks fine now. 

I am unsure why some of the checks in `coverage /  report` are failing as they seem unrelated to the changes I have made. Open to any suggestions.

MridulS:
> I am unsure why some of the checks in coverage /  report are failing as they seem unrelated to the changes I have made. Open to any suggestions.

Yes they are unrelated, the latest scipy release is broken for us. Feel free to ignore that :)

dschult:

paulbrodersen:
> can you please make the keyword arguments "keyword only" by adding a *, between the positional arguments and the keyword arguments? 

@dschult Done. Didn't know about this bit of syntactic sugar. I am curious: is there a thread somewhere that outlines why networkx is migrating to this pattern (if not, don't bother writing it out for me -- I am curious but not that curious)?

MridulS:
I can't think of a single thread which covers this in the networkx context, but you can look at the motivations part of the SLEP 09 (scikit-learn enhancements proposals) https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep009/proposal.html#motivation

paulbrodersen:
No, I am liking it and evaluating whether I should do the same in the libraries that I maintain. Thanks for indulging my curiosity. 

Any other comments on my code or other requests?

PR #6759, opened 2023-06-24
Initial vertical chain implementation

Erotemic:
This PR is an addition to "network_text", which I use quite a bit in my DAG library ([cmd_queue](https://pypi.org/project/cmd-queue/)),  [delayed image](https://gitlab.kitware.com/computer-vision/delayed_image), and elsewhere (printing graphs to the screen makes them much easier to debug). It works very well for smallish graphs. For larger graphs it can go past the display width, which is a known limitation.

However, recently I ran into an instance where my graph was just a long chain, and it overran the width of my screen. I thought: "that's kind of annoying", it's a chain, just print everything vertically.

So that's what this does. When network text detects a chain it uses a vertical space to separate the nodes. This does cost vertical space, so it is disabled by default, but if you enable it the display can scale to much larger graphs where the bulk of it is made of chains.

Here is a before / after:

```python
        >>> graph = nx.path_graph(10)
        >>> graph.add_node('A')
        >>> graph.add_node('B')
        >>> graph.add_node('C')
        >>> graph.add_node('D')
        >>> graph.add_edge(9, 'A')
        >>> graph.add_edge(9, 'B')
        >>> graph.add_edge(9, 'C')
        >>> graph.add_edge('C', 'D')
        >>> graph.add_edge('C', 'E')
        >>> graph.add_edge('C', 'F')
        >>> write_network_text(graph)
        ╙── 0
            └── 1
                └── 2
                    └── 3
                        └── 4
                            └── 5
                                └── 6
                                    └── 7
                                        └── 8
                                            └── 9
                                                ├── A
                                                ├── B
                                                └── C
                                                    ├── D
                                                    ├── E
                                                    └── F
        >>> write_network_text(graph, vertical_chains=True)
        ╙── 0
            │
            1
            │
            2
            │
            3
            │
            4
            │
            5
            │
            6
            │
            7
            │
            8
            │
            9
            ├── A
            ├── B
            └── C
                ├── D
                ├── E
                └── F
```

A more subtle case:

```python
    >>> graph = nx.generators.barbell_graph(4, 2)
    >>> write_network_text(graph, vertical_chains=False)
    ╙── 4
        ├── 5
        │   └── 6
        │       ├── 7
        │       │   ├── 8 ─ 6
        │       │   │   └── 9 ─ 6, 7
        │       │   └──  ...
        │       └──  ...
        └── 3
            ├── 0
            │   ├── 1 ─ 3
            │   │   └── 2 ─ 0, 3
            │   └──  ...
            └──  ...
    >>> write_network_text(graph, vertical_chains=True)
    ╙── 4
        ├── 5
        │   │
        │   6
        │   ├── 7
        │   │   ├── 8 ─ 6
        │   │   │   │
        │   │   │   9 ─ 6, 7
        │   │   └──  ...
        │   └──  ...
        └── 3
            ├── 0
            │   ├── 1 ─ 3
            │   │   │
            │   │   2 ─ 0, 3
            │   └──  ...
            └──  ...
```

My thoughts on review:

* I think I got the implementation right, but it's always easy to make mistakes with this sort of code
* I was unable to find a suitable UTF 8 character to remove the need for the "|" character to take up an entire line. I had one candidate, but I concluded it was too visually confusing.
* The directed ascii only case doesn't have a great down arrow proxy. I used "v", but I thought about just keeping "|" and letting the downward direction be implicit... not sure, but my gut says "v" is the best choice.

Once this core functionality is reviewed, I'll add the tests and such. 

@MridulS @dschult

### DIFF
--- a/networkx/readwrite/text.py
+++ b/networkx/readwrite/text.py
@@ -24,12 +24,14 @@ class AsciiDirectedGlyphs(_AsciiBaseGlyphs):
     last = "L-> "
     mid = "|-> "
     backedge = "<-"
+    vertical_edge = 'v'
 
 
 class AsciiUndirectedGlyphs(_AsciiBaseGlyphs):
     last = "L-- "
     mid = "|-- "
     backedge = "-"
+    vertical_edge = '|'
 
 
 class _UtfBaseGlyphs:
@@ -48,16 +50,19 @@ class UtfDirectedGlyphs(_UtfBaseGlyphs):
     last = "└─╼ "
     mid = "├─╼ "
     backedge = "╾"
+    vertical_edge = '╽'
 
 
 class UtfUndirectedGlyphs(_UtfBaseGlyphs):
     last = "└── "
     mid = "├── "
     backedge = "─"
+    vertical_edge = '│'
 
 
 def generate_network_text(
-    graph, with_labels=True, sources=None, max_depth=None, ascii_only=False
+    graph, with_labels=True, sources=None, max_depth=None, ascii_only=False,
+    vertical_chains=False,
 ):
     """Generate lines in the "network text" format
 
@@ -108,7 +113,7 @@ def generate_network_text(
     with_labels : bool | str
         If True will use the "label" attribute of a node to display if it
         exists otherwise it will use the node value itself. If given as a
-        string, then that attribute name will be used instead of "label".
+        string, then that attribte name will be used instead of "label".
         Defaults to True.
 
     sources : List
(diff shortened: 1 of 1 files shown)


dschult (commented):
This looks good!  
For the vertical symbol with a chain, how does it look without a symbol and the nodes listed vertically above each other? Maybe that's not clear enough, but it would cut down on the vertical length of the chains.

Erotemic:
I think it's important to have an edge character. I suppose it could be configurable, but I'd rather opt to keep it simple unless there is a use case that really needs to save the extra vertical space. But given this initial approval I'll throw together some tests and polish this up.

Erotemic:
Currently the ascii-only directed graphs, vertical chain uses a "v" character, but after some thinking I was wondering if using "!" might be nice. It is vertical and has directional aspect. Thoughts?

```
    >>> write_network_text(graph, vertical_chains=True, ascii_only=True)
    +-- 0 <- 4
        v
        1
        v
        2
        v
        3
        v
        4
        L->  ...
```

vs

```
    >>> write_network_text(graph, vertical_chains=True, ascii_only=True)
    +-- 0 <- 4
        !
        1
        !
        2
        !
        3
        !
        4
        L->  ...
```

dschult:
I like the exclamation point `!` better than the `v`. The `v` somehow registers as a letter to me, disturbing the distinction between node symbols and edge symbols.  But really, I'm good with whatever you think works well. :}

Erotemic:
I think this is ready. 

My latest changes include a new function: `parse_network_text`. I did not expose this function nor did I add an associated `read_network_text` function because ideally nobody will be parsing network text in practice. (Although if they really want to I suppose it is now available). The reason I added it was for testing. Now that there is a function that can reconstruct the original graph from a network text representation, I can write a round-trip test that generates a gauntlet of different test cases and checks that the reconstructed graph is the same as the original. This is done for all 4 cases in {directed, undirected} x {ascii, utf}.  If I didn't get both implementations right it would be very difficult to pass this test, so now I'm much more confident that didn't make any mistakes.

In addition I:

* Changed the ascii-only vertical symbol to "!".
* Wrote simple explicit tests for the vertical chain case.
* Wrote simple tests for the "collapse" node attribute.
* Modified the docstrings to capture the new behavior. 
* Added a release note.


Let me know if there is anything outstanding, if anything is unclear, or if there are any changes you want (e.g. is there a better parameter name than `vertical_chains`?).

dschult (approved):
This looks good to me!  Thanks!!

rossbar (approved):
LGTM, thanks @Erotemic !

I have a slight preference for slapping an `_` before `parse_network_text` just to really drive home that the text formats are not intended to be a new, roundtrip-able graph file format. Having seen a lot of deep nested imports from subpackages in user code, I'm not sure excluding from `__all__` is sufficient to communicate this point!

It's not a blocker for me though - in practice it's difficult to see how users would come to use this directly very often. +1 for the rest of the changes!

PR #6760, opened 2023-06-24
Fix Johnson method for unweighted graphs

### LINKED ISSUE #6749: Docstring of `johnson` algorithm handling of weight incorrect
### Current Behavior

The documentation says:
```
If no such edge attribute exists, the weight of the edge is assumed to be one.
```
### Expected Behavior

The behavior to match the docstring or the docstring to match the behavior.

### Steps to Reproduce

```python
In [1]: import networkx as nx

In [2]: G = nx.Graph()

In [3]: G.add_edge(0, 1, weight=2)

In [4]: G.add_edge(0, 2)

In [5]: nx.johnson(G)
NetworkXError                             Traceback (most recent call last)
Cell In[5], line 1
----> 1 nx.johnson(G)

File ~/miniconda3/envs/gb8/lib/python3.11/site-packages/networkx/algorithms/shortest_paths/weighted.py:2469, in johnson(G, weight)
   2402 r"""Uses Johnson's Algorithm to compute shortest paths.
   2403
   2404 Johnson's Algorithm finds a shortest path between each pair of
   (...)
   2466
   2467 """
   2468 if not nx.is_weighted(G, weight=weight):
-> 2469     raise nx.NetworkXError("Graph is not weighted.")
   2471 dist = {v: 0 for v in G}
   2472 pred = {v: [] for v in G}

NetworkXError: Graph is not weighted.
```

### Environment

Python version: 3.11
NetworkX version: 3.1

### Additional context

It uses `nx.is_weighted`, which requires that _all_ edges have the attribute to be considered weighted.

navyagarwal:
Closes #6749 

Modified the behavior of the code to match the docstring for Johnson's shortest path method.

Now the unweighted edges in a graph are assumed to have weight 1.

### DIFF
--- a/networkx/algorithms/shortest_paths/tests/test_weighted.py
+++ b/networkx/algorithms/shortest_paths/tests/test_weighted.py
@@ -865,10 +865,9 @@ class TestBellmanFordAndGoldbergRadzik(WeightedTestBase):
 
 class TestJohnsonAlgorithm(WeightedTestBase):
     def test_single_node_graph(self):
-        with pytest.raises(nx.NetworkXError):
-            G = nx.DiGraph()
-            G.add_node(0)
-            nx.johnson(G)
+        G = nx.DiGraph()
+        G.add_node(0)
+        assert nx.johnson(G) == {0: {0: [0]}}
 
     def test_negative_cycle(self):
         G = nx.DiGraph()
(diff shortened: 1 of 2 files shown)


dschult (approved):
This all looks good. 

It's surprising to me that we had the `is_weighted` check in this function. The code works when not all edges are weighted. So this is definitely an improvement. Thanks @eriknw for noticing this -- and thanks @navyagarwal for the quick fix!

PR #6768, opened 2023-06-26
Replace deprecated numpy.alltrue method

eendebakpt:
The method `numpy.alltrue` has been deprecated in 1.25 and will be removed in numpy 2.0. See https://github.com/numpy/numpy/pull/23314


dschult (approved):
Thank you!  This looks good and is one of those hard to find changes. 

I searched for this function elsewhere in networkx and this seems to be the only location.
LGTM

rossbar (approved):
LGTM as well - tests pass locally (with scipy 1.10) so the failures are unrelated, but I'll hold off on the merge button in the name of prudence! Thanks @eendebakpt

PR #6769, opened 2023-06-27
MAINT: Ignore SciPy v1.11 in requirements

MridulS:
Should we just keep this pin till scipy 1.11+ becomes the min supported scipy version for networkx?


rossbar (approved):
I'm +1 on that, though I think we may want to be specific to `1.11.0` - as presumably the sparse issue will be fixed in the next patch release (and I'm not sure how exactly the versions are parsed; either way, explicit is better than implicit!)

PR #6771, opened 2023-06-28
Check that it works on 3.12 without setuptools

jarrodmillman:


dschult:
It looks like setuptools is installed with the other packages in the install packages section.
At least it is listed with the other packages.
```
69    setuptools     58.1.0
```

PR #6784, opened 2023-07-11
Fix typo in contributing page

nerivera:
There was a missing backtick causing an extra one to appear after `sp` in the contributing page.


MridulS (approved):
Thanks!

dschult (approved):

PR #6785, opened 2023-07-11
deprecate Edmonds class

mjschwenne:
As discussed in #6743, the original implementation of Edmond's algorithm for finding maximum branchings, and by extension minimum branchings as well as minimum and maximum arborescences was a class with only one method (`find_optimum`). This API structure is not common throughout NetworkX, in fact it is the only instance where a class is used to implement a network algorithm that I'm aware of. Since all of these algorithms were recently refactored and no longer depend on the class itself, it is now safe to remove the class. 

Additionally, the `MultiDiGraph` subclass `MultiDiGraph_EdgeKey` is also no longer needed. It was not used in any other algorithms in the library and doesn't extend the original class in a manner which is not easily replicated with an additional helper function.

### DIFF
--- a/networkx/algorithms/tree/branchings.py
+++ b/networkx/algorithms/tree/branchings.py
@@ -211,7 +211,7 @@ class MultiDiGraph_EdgeKey(nx.MultiDiGraph):
 
         import warnings
 
-        msg = "MultiDiGraph_EdgeKey has been deprecated and will be removed in NetworkX 3.3."
+        msg = "MultiDiGraph_EdgeKey has been deprecated and will be removed in NetworkX 3.4."
         warnings.warn(msg, DeprecationWarning)
 
     def remove_node(self, n):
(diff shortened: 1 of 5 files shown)


dschult (approved):
LGTM!   Yay!  and Thanks Matt!

networkx/algorithms/tree/branchings.py
```diff
@@ -315,6 +320,11 @@ def __init__(self, G, seed=None):
         # sure that our node names do not conflict with the real node names.
         self.template = random_string(seed=seed) + "_{0}"
 
+        import warnings
+
+        msg = "Edmonds has been deprecated and will be removed in NetworkX 3.3. Please use the approiate minimum or maximum branching or arborescence function directly."
```
jarrodmillman, line 325:
```suggestion
        msg = "Edmonds has been deprecated and will be removed in NetworkX 3.4. Please use the approiate minimum or maximum branching or arborescence function directly."
```

jarrodmillman (approved):
Unless this needs to be removed soon, we should remove it in 3.4 per the policy:
- https://networkx.org/documentation/stable/developer/deprecations.html

mjschwenne:
You're totally correct, just a goof on my part.

PR #6789, opened 2023-07-14
Refix minimum_cycle_basis and scipy.sparse conversions and add tests

dschult:
Follow-up testing from #6788 showed a number of interesting bugs and deficiencies.
- My error using `any` instead of `sum` followed by `% 2` in one line. This led to the returned "basis" being not independent in some cases.
- The tests don't check non-uniformly weighted graphs sufficiently to catch the above error. So add tests.
- One reported example from @mbr085 raised a ValueError which was caused because an edge was added as `np.int32` instead of python `int`. I chased this down to `convert_matrix.py` and the handling of conversion from sparse matrices. More below if you are interested.

This PR makes 3 commits to fix these issues. The first adds tests to show the errors reported in #6788. The second adds the sum/modulo term to enable passing those tests. The third corrects the sparse matrix treatment to match that of conversion from numpy arrays. 

TL;DR:
We have to be careful not to mix nodes that are `np.int32` and nodes that are python `int` in the same graph because they equate to each other (and thus are the same node) but they compare to other nodes differently. For example `int(5)` and `np.int32(5)` are equal so you can work with both on a graph as if they are the same node. But the `np.int` value raises a ValueError when compared to, say, a tuple `(5, 1)` because it treats the tuple as an array and the sizes don't match. The python in returns `False` when you check equality of an integer to a tuple. Either storage of nodes is fine, but if you use `np.int` and then **also** add nodes to your graph that are tuples, you can get ValueErrors raised when comparing the tuple-node with the np.int-node.

NetworkX handles numpy.array conversion by making nodes python int.  It handles scipy.sparse conversion by making nodes int -- but crucially, failing to convert the np.int32 indices to python in when adding edges. So inside `G.adj` the keys are python ints and but the keys to the inner nbrdict `G.adj[5]` are `np.int32` with the same value. Amazingly, this works fine almost all the time. But if you add another node that is a tuple, you can run into a raised ValueError when our algorithms check whether two nodes are equal.

I fixed the conversion functions to convert `indices` values to python int so thy match the nodes in the graph.

### DIFF
--- a/networkx/algorithms/cycles.py
+++ b/networkx/algorithms/cycles.py
@@ -1097,7 +1097,7 @@ def _min_cycle_basis(G, weight):
                 {e for e in orth if e not in base if e[::-1] not in base}
                 | {e for e in base if e not in orth if e[::-1] not in orth}
             )
-            if any((e in orth or e[::-1] in orth) for e in cycle_edges)
+            if sum((e in orth or e[::-1] in orth) for e in cycle_edges) % 2
             else orth
             for orth in set_orth
         ]
--- a/networkx/algorithms/tests/test_cycles.py
+++ b/networkx/algorithms/tests/test_cycles.py
@@ -917,7 +917,7 @@ class TestMinimumCycleBasis:
         cg = nx.complete_graph(N)
         cg.add_weighted_edges_from([(u, v, 9) for u, v in cg.edges])
         cg.add_weighted_edges_from([(u, v, 1) for u, v in nx.cycle_graph(N).edges])
-        mcb = nx.minimum_cycle_basis(cg, weight='weight')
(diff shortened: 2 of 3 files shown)


rossbar (approved):
Thanks again @dschult for the quick followup. Re: the integer type issue, the `_*_gen_triples` functions have always bothered me... I wonder if the situation would be improved by calling one of the scipy.sparse constructors directly (perhaps COO, since it's the most straightforward) then using `to*` to convert to the proper format from there. The potential advantage there is that we let the scipy.sparse and underlying machinery handle types/casting.

Anyways, that's a separate issue and I don't want to distract from the fixes here. LGTM!

PR #6791, opened 2023-07-15
Fix empty graph zero division error  for louvain

juanis2112:
I fixed the Zero division error for the Louvain algorithm when running it with an empty and added a test for it.


dschult (approved):
This looks like a good way to handle the edge case of no edges ;}

PR #6794, opened 2023-07-18
updating TSP example docs

Transurgeon:
Hello, I just found a small typo in the example docs for the traveling salesman problem.
I was also wondering what's the best way to add a hyperlink for christofides with this link:(https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.approximation.traveling_salesman.christofides.html#christofides), does sphinx docs support markdown in the python comments? and is there a shortcut for the link?

### DIFF
--- a/examples/drawing/plot_tsp.py
+++ b/examples/drawing/plot_tsp.py
@@ -5,7 +5,7 @@ Traveling Salesman Problem
 
 This is an example of a drawing solution of the traveling salesman problem
 
-The function is used to produce the solution is christofides,
+The function used to produce the solution is christofides,
 where given a set of nodes, it calculates the route of the nodes
 that the traveler has to follow in order to minimize the total cost.
 """
--- a/examples/drawing/plot_tsp.py
+++ b/examples/drawing/plot_tsp.py
@@ -5,7 +5,7 @@ Traveling Salesman Problem
 
 This is an example of a drawing solution of the traveling salesman problem
 
-The function used to produce the solution is `christofides <nx.approximation.christofides>`_,
+The function used to produce the solution is `christofides <nx.algorithms.approximation.christofides>`_,
 where given a set of nodes, it calculates the route of the nodes
 that the traveler has to follow in order to minimize the total cost.
 """
--- a/examples/drawing/plot_tsp.py
+++ b/examples/drawing/plot_tsp.py
@@ -6,8 +6,9 @@ Traveling Salesman Problem
 This is an example of a drawing solution of the traveling salesman problem
 
 The function used to produce the solution is christofides,
-where given a set of nodes, it calculates the route of the nodes that the traveler has to follow in order to
-minimize the total cost."""
+where given a set of nodes, it calculates the route of the nodes
+that the traveler has to follow in order to minimize the total cost.
+"""
 
 import matplotlib.pyplot as plt
 import networkx as nx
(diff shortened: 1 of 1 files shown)


mjschwenne:
Reading some of the Sphinx docs, you may be able to reference the `christofides` function with 
```
:func:`christofides`
```
or some other path to the function such as `nx.approximation.christofides`. Note that I haven't actually tested this, but here are the references I found:

[Sphinx Quickstart Guide](https://www.sphinx-doc.org/en/master/usage/quickstart.html#documenting-objects)

[Sphinx Domain Reference](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects)

Transurgeon:
@mjschwenne thanks for the reply! 
I am having difficulties building the docs. it's something related to "failed building wheel for pygraphviz". 
could you try on your end if ``:func:`christofides` `` would create the appropriate hyperlink?

dschult:
The pygraphviz issue is related to windows environment if i understand correctly.
There are two ways around this:
- use `make html-noplot` instead of `make html`.
- submit the change as a PR, wait for the CI tests to run, scroll to the bottom of the CI tests and select the "documentation artifact" link to go to the webpages as they would appear after the PR is merged.

I think the easiest for you is the first one since you probably have everything else set up for local building of the docs. If it doesn't work or is too much, I'll try building it on my machine.

Thanks!!

Transurgeon:
Hello @dschult , I tried the first method but still to no avail. I now get this error
```
WARNING: [autosummary] failed to import networkx.algorithms.cycles.girth.
Possible hints:
* ImportError: 
* AttributeError: module 'networkx.algorithms.cycles' has no attribute 'girth'
* ModuleNotFoundError: No module named 'networkx.algorithms.cycles.girth'; 'networkx.algorithms.cycles' is not a package

Extension error (sphinx.ext.autosummary):
```
I am pretty sure I have installed all the requirements from conda and also requirements/doc.txt
I will try your second method for this PR, but would also appreciate it if you could help me building docs for future contributions.

dschult:
Hmmm.. It can't find the relatively new function `girth`.  That sounds like it is using an older version of networkx instead of the local version you are working on. Do you have the PYTHONPATH environment variable set to the head of the repo you are working on? I don't know what kind of environment you are working with -- but it seems like `python -c "import network as nx;print(nx.__file__)"` might help you track down where the networkx you are using is located.

Transurgeon:
@dschult I think my networkx is located under the head of the repository, it follows the following pattern:
```{...}/networkx/networkx/__init__.py```
Also after running the unit tests, I get this error:
```FAILED networkx/linalg/tests/test_algebraic_connectivity.py::TestSpectralOrdering::test_cycle[tracemin_lu-False-expected_order0] - TypeError: rowind and colptr must be of type cint```
```FAILED networkx/linalg/tests/test_algebraic_connectivity.py::TestSpectralOrdering::test_cycle[tracemin_lu-True-expected_order1] - TypeError: rowind and colptr must be of type cint```
I think it might just be that I haven't updated networkx to the most recent commits? Do you have any tips on how to do that efficiently?

Transurgeon:
@mjschwenne @dschult sorry for the trouble,

I tried multiple things for the hyperlink but nothing seemed to work nicely. (I would have to include the entire path and it doesn't look nice when compiled to the docs, ex: networkx.algorithms.approximation.traveling_salesman.christofides)

I noticed that most examples in the gallery do not actually link to the API. So I think I will just stick to fixing the typo for this PR. 

Thanks again for your time and help,

PR #6797, opened 2023-07-20
Add label-check workflow

jarrodmillman:
As discussed in the developer meeting today, I've updated the labels to include:
https://github.com/networkx/networkx/labels?q=type
```
type: Maintenance
type: Enhancements
type: Documentation
type: Highlights
type: API
type: Bug fix 
```

This is in preparation for switching to https://github.com/scientific-python/changelist/ after the upcoming 3.2 release.

### DIFF
--- a/.github/workflows/label-check.yml
+++ b/.github/workflows/label-check.yml
@@ -1,0 +1,19 @@
+name: Labels
+
+on:
+  pull_request:
+    types:
+      - opened
+      - labeled
+      - unlabeled
+
+env:
+  LABELS: ${{ join( github.event.pull_request.labels.*.name, ' ' ) }}
+
+jobs:
+  check-type-label:
+    name: ensure type label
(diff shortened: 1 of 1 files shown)


rossbar (approved):
LGTM, +1 for better automating release notes and the approach looks good!

I think the minimal list of starting labels is good, and we can always adjust as needed in the future.

For completeness' sake, I'll also mention [an existing release note-drafting action](https://github.com/release-drafter/release-drafter) that I've used on other projects which is well-supported. It takes a similar approach, but has a lot of extra features that we probably don't need. At the very least it may be interesting as a reference point for the `changelist` action!

dschult (approved):
This looks good to me. The docs in changelog suggest another action to assign a label to PRs when merged. You mentioned that too in the meeting but I think it is not here. But you know that I'm sure. 
:) :)

Let's try it!

PR #6807, opened 2023-07-24
Account for sampling nodes when normalizing edge bc scores

jnke2016:


dschult:
Wow -- this has been an unusable feature for 8 years unless you call the private scaling function directly.
Would you be able to add a test that specifying a k value actually does something?

Thanks very much for the easy fix.

jnke2016:
Sure. I will add a test

jnke2016:
@dschult I added the tests. Please let me know if I am missing something.

networkx/algorithms/centrality/tests/test_betweenness_centrality.py
```diff
@@ -563,6 +563,26 @@ def test_K5(self):
...
+        )
+        for n in sorted(G.edges()):
+            assert b_sample[n] == pytest.approx(b_answer[n], abs=1e-7)
+
+    def test_samples_normalized_K5(self):
+        """Edge betweenness centrality: K5"""
+        G = nx.complete_graph(5)
+        b_answer = nx.edge_betweenness_centrality(G, weight=None, normalized=True)
+        b_sample = nx.edge_betweenness_centrality(
+            G, weight=None, k=G.number_of_nodes(), normalized=True
```
MridulS, line 581: Thanks for the PR @jnke2016!

Could you add a test which tests the change by providing `k` argument? In this case we can't see any change to the result. Something like `k=0` which should raise a `ZeroDivisionError` through `_rescale_e` could be a good test, as previously it just executed the code and gave a centrality score of 0 to all edges.

[Maybe we should catch the  `ZeroDivisionError` error here @dschult?]

dschult: To be clear, I think it would be helpful to have a test value of k that changes the output.  And the zero division error would certainly do that. So, let's include that one more test.

Issue #6809, opened 2023-07-26
LFR_benchmark_graph : graph generation does not preserve degree distribution


ebokai:
The implementation of the graph generation in the `LFR_benchmark_graph` function does not preserve the desired degree distribution and generally construct graphs with a mixing parameter that is larger than the input value.

### Current Behavior
```   
G = nx.Graph()
G.add_nodes_from(range(n))
    for c in communities:
        for u in c:
            while G.degree(u) < round(deg_seq[u] * (1 - mu)):
                v = seed.choice(list(c))
                G.add_edge(u, v)
            while G.degree(u) < deg_seq[u]:
                v = seed.choice(range(n))
                if v not in c:
                    G.add_edge(u, v)
            G.nodes[u]["community"] = c
    return G
```
Note that the current implementation only checks the degree of node `u` and assigns an edge regardless of the current degree of `v`. This allows the degree of `v` to exceed the desired degree in almost all cases. Additionally, it appears to `u` and `v` can be the same node, creating self-loops in the graph.

### Expected Behavior

The proposed change checks the realized internal and external degree against the desired distribution.
```
G = nx.Graph()
G.add_nodes_from(range(n))

# number of communities
q = len(communities) 

# keep track of desired internal and external degree for each node
desired_in_deg = [round(deg_seq[u] * (1 - mu)) for u in range(n)]
desired_out_deg = [deg_seq[u] - desired_in_deg[u] for u in range(n)]

# keep track of realized internal and external degree for each node
realized_in_deg = [0 for _ in range(n)]
realized_out_deg = [0 for _ in range(n)]
```

The wiring of the internal edges is performed separately from the wiring of the external edges. Additionally, the while-loop is replaced by a loop over all pairs of nodes in the community (which can be further optimized to only include distinct pairs). Furthermore, nodes are sorted by largest degree to prioritize wiring of large-degree nodes. In this implementation, edges are only added if both nodes do not yet satisfy the degree distribution.
```
for c in communities:
    
    # sort nodes by largest degree to prioritize wiring of high-degree nodes
    c_degrees = [desired_in_deg[u] for u in c]
    c_sort_by_deg = [u for _,u in sorted(zip(c_degrees, c), reverse = True)]

    for u in c_sort_by_deg:

        G.nodes[u]["community"] = c

        for v in c_sort_by_deg:
            
            # only add an edge if both nodes do not yet satisfy the internal degree distribution
            check_u = realized_in_deg[u] < desired_in_deg[u]
[...]

Issue #6763, opened 2023-06-25
LFR_benchmark_graph missing condition check


woQal:
Algorithm runs infinitely if there are not enough nodes for edges assignment.

### Current Behavior

This is the loop in the code and if there are not enough nodes in the graph, the algorithm cannot assign edges and runs infinitely.

```
            while G.degree(u) < deg_seq[u]:
                v = seed.choice(range(n))
                if v not in c:
                    G.add_edge(u, v)
```

### Expected Behavior

I suggest to use:

```
            in_neigh = set(G.neighbors(u)).intersection(c)
            if deg_seq[u] > n - len(c) + len(in_neigh):
                msg = "Could not add edges; try decreasing max_degree"
                raise nx.exception.ExceededMaxIterations(msg)
```
The condition `if deg_seq[u] > n - len(c) + len(in_neigh)` is neccesary. Supposing the following scenario: we have a network with $25$ nodes, a community `c` of size $15$, a node `u` in that community with `G.degree(u) = 5` (and `len(in_neigh) = 5`, the condition `round(deg_seq[u] * (1 - mu)) = s = 5`) and the `deg_seq[u] = 20`. We want to add $15$ edges to that node. But we only have $10$ out-community nodes. You may say that `deg_seq[u] > |c|`, but the community generator only checks if `s < |c|` which holds. And also, it can happen that `max_generated_degree > max_generated_community`, because the power law sequence generator does **not** ensure the communities of sizes near `max_community` (this is only an upper bound on the community sizes).

### Steps to Reproduce

Happens for example for the following settings `LFR_benchmark_graph(n=100, tau1=2, tau2=1.5, mu=0.5, average_degree=10, min_community=10)`, almost every time.

### Environment

Python version: 3.9.5
NetworkX version: 3.1

Issue #6762, opened 2023-06-25
LFR_benchmark_graph incorrect condition in loop


woQal:
The number of in-community edges can be 0 even if the mu parameter is very low (so most of the edges should be in-community).

### Current Behavior
The condition is checking if degree 

```
            while G.degree(u) < round(deg_seq[u] * (1 - mu)):
                v = seed.choice(list(c))
                G.add_edge(u, v)
```

The first loop is running while `G.degree(u) < intended no of intra edges`, but this is wrong, because G.degree(u) can contain nodes not in the community. For example -- having two comms [0, 1, 2], [3, 4, 5], we add edges (both in- and out- community) and we added an out-community edge from 0 to 3 and from 1 to 3 while processing the node 0. For node 3 (supposing `round(deg_seq[u] * (1 - mu)) = 2`, this condition just checks "ok, there are two edges to node 3" and thats all. It does not add any in-community edges for node 3.

### Expected Behavior

I would suggest

```
            in_neigh = set(G.neighbors(u)).intersection(c)
            while len(in_neigh) < round(deg_seq[u] * (1 - mu)):
                ... adding edges
                in_neigh.add(v)
```

because the G.degree(u) can contain out-community nodes.

### Environment

Python version: 3.9.5
NetworkX version: 3.1

PR #6811, opened 2023-07-26
lfr_benchmark wiring algorithm fix

ebokai:
Fix for issue 6809

### DIFF
--- a/networkx/generators/community.py
+++ b/networkx/generators/community.py
@@ -1046,16 +1046,64 @@ def LFR_benchmark_graph(
     # Finally, generate the benchmark graph based on the given
     # communities, joining nodes according to the intra- and
     # inter-community degrees.
+
     G = nx.Graph()
     G.add_nodes_from(range(n))
+
+    desired_internal_degree = [round(deg_seq[u] * (1 - mu)) for u in range(n)]
+    desired_external_degree = [
+        deg_seq[u] - desired_internal_degree[u] for u in range(n)
+    ]
+
+    realized_internal_degree = [0 for _ in range(n)]
+    realized_external_degree = [0 for _ in range(n)]
+
     for c in communities:
-        for u in c:
-            while G.degree(u) < round(deg_seq[u] * (1 - mu)):
-                v = seed.choice(list(c))
-                G.add_edge(u, v)
-            while G.degree(u) < deg_seq[u]:
-                v = seed.choice(range(n))
-                if v not in c:
-                    G.add_edge(u, v)
+        comm_degrees = [desired_internal_degree[u] for u in c]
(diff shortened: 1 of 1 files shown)


ebokai:
The test / base (ubuntu, pypy-3.9) test failed because of `test_louvain.py` trying to assert if the modularity of communities obtained with `louvain_communities` with `threshold=0.3` is smaller than that the default value of the threshold. Specifically, 
`FAILED networkx/algorithms/community/tests/test_louvain.py::test_threshold - assert 0.6367936995849326 < 0.6367936995849326`
On my local machine, this same check produces modularities 
`0.6226115813633905 0.630359779541833` which would pass the test.
Is the value of the seed platform dependent?

ebokai:
See issue https://github.com/networkx/networkx/issues/6823 for a discussion on the test of the Louvain algorithm threshold for graphs with the same generating parameters but using a different value of the seed.

MridulS:
thanks for this @ebokai! Couple of changes here are under the assumption that self loops aren't allowed in the generated graphs, I've been trying to find some reference regarding it but not able to find anything specific. Did you run into any resource/reference regarding self loops in LFR?

IMO it's fine either way, we should just be consistent regarding self loops in the generated graph.

There is also https://github.com/networkx/networkx/issues/6762 and https://github.com/networkx/networkx/issues/6763 which brings up similar-ish issues with LFR benchmark graph implementation.

ebokai:
Thanks for the feedback @MridulS. Regarding the self-loops, I believe the original algorithm by the authors of the LFR paper doesn't generate graphs that contain self-loops so I am following that convention. 
At this point, I am not sure how to proceed since the current implementation fails the `test/base (ubuntu, pypy-3.9)` test due to reasons explained here: https://github.com/networkx/networkx/issues/6823 and in the comment above.

VanitasCodes:
Hey @ebokai @MridulS @dschult,

Noticed this PR has been stalled since 2023. The blocking issue (flaky Louvain test #6823) was fixed in #7336 back in March 2024.

Is there interest in reviving this? The LFR benchmark has several open bugs (#6809, #6822, #6762, #6763) and this PR addresses at least two of them.

If @ebokai is no longer available, I'd be happy to:
- Rebase this branch on current main
- Address any remaining reviewer feedback
- Add tests verifying the degree distribution is preserved

Let me know if that would be helpful.

ebokai:
Hi @VanitasCodes,

please go ahead with your suggested changes.

VanitasCodes:
Thanks @ebokai! I've opened #8423 with a comprehensive fix that addresses #6809, #6762, #6822, and #6763. I've credited you and @woQal in the PR description for the prior work and analysis.

Appreciate you giving the green light!

PR #6812, opened 2023-07-27
Add preserve_all_attrs to convert_from_nx to make it concise

MridulS:
implements the idea by @eriknw in https://github.com/networkx/networkx/pull/6688/files#issuecomment-1652314158

### DIFF
--- a/networkx/algorithms/bipartite/projection.py
+++ b/networkx/algorithms/bipartite/projection.py
@@ -413,10 +413,7 @@ def overlap_weighted_projected_graph(B, nodes, jaccard=True):
 
 
 @not_implemented_for("multigraph")
-@nx._dispatch(
-    graphs="B",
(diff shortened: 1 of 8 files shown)


dschult (commented):
This looks great!   :tada:

Minor nit:  Can you make one line decorators e.g. where we only have arguments `graphs` and `preserve_all_attrs`?

MridulS:
The addition of `preserve_all_attrs` made the local variable scope unhappy, so went with another implementation.

eriknw:
Yeah, I think the scoping thing will be better if we made `_dispatch` a class instead of using closures.

LGTM!

dschult (approved):
This looks good!  Thanks!

MridulS:
Merging this as Erik LGTM'd

PR #6815, opened 2023-07-28
number_of_walks might use a weighted edge attribute

dschult:
number_of_walks might use a weighted edge attribute if that graph has any edge attributes not equal to 1.
That will make it compute the number of walks incorrectly since it uses a power of the adjacency matrix.

This fix simply changes the call to adjacency_matrix to have `weight=None` explicitly set. Test added too.


jarrodmillman (approved):

rossbar (approved):
Thanks @dschult !

PR #6816, opened 2023-07-28
Make position part of the API for geometric_edges

MridulS:
Implements the node attr part from https://github.com/networkx/networkx/issues/6753

### DIFF
--- a/networkx/generators/geometric.py
+++ b/networkx/generators/geometric.py
@@ -84,7 +84,7 @@ def geometric_edges(G, radius, p=2, position="pos"):
     return _geometric_edges(G, radius, p, position)
 
 
-def _geometric_edges(G, radius, p=2, position):
+def _geometric_edges(G, radius, p, position):
     """
     Implements `geometric_edges` without input validation. See `geometric_edges`
     for complete docstring.
--- a/networkx/generators/geometric.py
+++ b/networkx/generators/geometric.py
@@ -19,8 +19,8 @@ __all__ = [
 ]
 
 
-@nx._dispatch(node_attrs="position")
-def geometric_edges(G, radius, p=2, position="pos"):
+@nx._dispatch(node_attrs="pos_name")
+def geometric_edges(G, radius, p=2, *, pos_name="pos"):
     """Returns edge list of node pairs within `radius` of each other.
 
     Parameters
@@ -32,9 +32,9 @@ def geometric_edges(G, radius, p=2, position="pos"):
     radius : scalar
         The distance threshold. Edges are included in the edge list if the
         distance between the two nodes is less than `radius`.
-    position : string, default="pos"
-        The node attribute which represents the position of the node in 2D
-        coordinates. Every node in the Graph should have this attribute.
+    pos_name : string, default="pos"
+        The name of the node attribute which represents the position of each
+        node in 2D coordinates. Every node in the Graph must have this attribute.
     p : scalar, default=2
         The `Minkowski distance metric
         <https://en.wikipedia.org/wiki/Minkowski_distance>`_ used to compute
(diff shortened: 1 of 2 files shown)


MridulS:
I have added 3.2 milestone to this one, we can push this further too.

dschult (commented):
This looks good to me.
Perhaps the keyword order for `geometric_edges` should put `position` at the end in case someone used `p` as a position argument. Or we could make this function use keyword-only and noisily break code that does use p as a position argument. It would be easy for the user to fix such a change -- and raising should be a good way to notify them.
I suppose another way would be to deprecate.

But to make this easy we should probably just add `position` as the last argument.

rossbar (commented):
tl;dr - my main concern is that there are many functions that hard-code `"pos"` as the node attr name. If we're making it configurable for `geometric_edges`, I think we should do so for all these functions - and if that's the case, we should consider a descriptive kwarg name that works well for all functions!

---

This looks good, I just think we need to extend it to the other geometric generators too; for example:

```python
>>> G = nx.random_geometric_graph(10, radius=5)
```

This creates a graph `G` with node attributes named `"pos"`. This is of course okay, but since we're making the attribute name configurable for `geometric_edges`, I think we should do so for the other functions too. That means we'd have to add a new keyword argument to each of the functions. The current functions that hard-code `"pos"` as the node attribute are:
 - [x] `geometric_edges`
 - [x] `geographical_threshold_graph`
 - [x] `random_geometric_graph`
 - [x] `soft_random_geometric_graph`
 - [ ] `thresholded_random_geometric_graph`
 - [ ] `waxman_graph`

For the sake of consistency, I think the kwarg should have the same name for each function. The trick here is that many of the above functions already have args with names related to position:

```python
>>> import inspect
>>> fns = [
...     nx.geometric_edges,
...     nx.geographical_threshold_graph,
...     nx.random_geometric_graph,
...     nx.soft_random_geometric_graph,
...     nx.thresholded_random_geometric_graph,
...     nx.waxman_graph
... ]
>>> for fn in fns:
...     print(f"{fn.__name__}: {list(inspect.signature(fn).parameters.keys())}")
geometric_edges: ['G', 'radius', 'p']
geographical_threshold_graph: ['n', 'theta', 'dim', 'pos', 'weight', 'metric', 'p_dist', 'seed']
random_geometric_graph: ['n', 'radius', 'dim', 'pos', 'p', 'seed']
soft_random_geometric_graph: ['n', 'radius', 'dim', 'pos', 'p', 'p_dist', 'seed']
thresholded_random_geometric_graph: ['n', 'radius', 'theta', 'dim', 'pos', 'weight', 'p', 'seed']
waxman_graph: ['n', 'beta', 'alpha', 'L', 'domain', 'metric', 'seed']
```

From the above listings, there are already some args named `pos` and `p`. My concern is adding another arg named `position` is not clear enough in this context --- i.e. would a user be confident in what `pos` specifies vs. `position` vs. `p` without having to read the parameter descriptions in detail?

Given the above, I'd vote for a slightly more specific kwarg to specify the name of the node attribute where the position is stored. Perhaps something like `pos_attr="pos"` or `node_attr="pos"` as suggested by @eriknw .

In the end, this is a lot of hemming and hawing for such a minor API design concern (sorry about that!) but I do think it's worth considering now before these kwargs are added so we don't have to deal with deprecations/future warnings!

dschult:
I pushed a commit to add keyword args for these geometric generators. Feel free to undo any or all of these changes.

I chose `pos_name` as the keyword arg. It is as short as "position" and conveys the idea that this is an attribute name without using the attribute word (which is long and if used as "attr" is confusing as to whether you mean the attribute-value or the attribute-name).  That means I also added `weight_name` for two of the functions. They already have a keyword arg "weight". So I hope this is clear and also not confusing that we don't follow the convention elsewhere to use "weight" for the keyword that specifies the name of the attribute.

I got carried away and so added "seed=42" to a number of `thresholded_random_geometric_graph` tests. I also cleaned up some test function doc_strings. I hope those aren't too annoying -- undo any that you don't like.

MridulS:
@rossbar could you have a look here? Thanks!

rossbar (approved):
LGTM, thanks @MridulS !

PR #6825, opened 2023-08-01
fix: Explicitly check for None/False in edge_attr during import from np

j6k4m8:
[comment reference](https://github.com/networkx/networkx/pull/6259#discussion_r1280916514) in #6259


j6k4m8:
Not sure what's going on with black formatting here... Looks like running black is making no detectable changes..?

dschult (approved):
This looks good to me.  Thanks!

(The black formatting was just removing a space at the end of two different lines. Invisible, but removed anyway.)

PR #6828, opened 2023-08-01
Improve testing of backends: let tests get setup up, and fallback

eriknw:
We add `is_pytest_running()` that is True once tests begin running to allow NetworkX to set up it's own tests. Otherwise, some functions go through dispatching before tests are run, which requires backends to implement them, and then the resulting graphs are the backend graph and not networkx graphs.

Also, allow algorithms to raise `NotImplementedError` to allow incremental implementation by backends. When testing, this will fallback to running the original networkx algorithm if `NETWORKX_BACKEND_TEST_EXHAUSTIVE` environment variable is set, or else the test will xfail.

This is to help with both `graphblas_algorithms` and `cugraph_nx` backends.

### DIFF
--- a/networkx/utils/backends.py
+++ b/networkx/utils/backends.py
@@ -391,6 +391,9 @@ def test_override_dispatch(
 
     @functools.wraps(func)
     def wrapper(*args, **kwds):
+        if not is_pytest_running():
+            # Allow tests to get set up
+            return func(*args, **kwds)
         backend = plugins[plugin_name].load()
         if not hasattr(backend, name):
             if fallback_to_nx:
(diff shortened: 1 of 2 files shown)


networkx/utils/backends.py
```diff
@@ -600,7 +603,14 @@ def wrapper(*args, **kwds):
...
 
-        result = getattr(backend, name).__call__(*bound.args, **bound.kwargs)
+        try:
+            result = getattr(backend, name).__call__(*bound.args, **bound.kwargs)
+        except NotImplementedError as exc:
+            if fallback_to_nx:
+                return func(*args, **kwds)
+            pytest.xfail(
+                exc.args[0] if exc.args else f"{name} raised NotImplementedError"
+            )
```
eriknw, line 613: We could also raise the original exception here. I don't have a preference.

aaronzo:
Is there overlap between this and #6829 ?

eriknw:
> Is there overlap between this and https://github.com/networkx/networkx/pull/6829 ?

Possibly, so thanks for linkning to that PR! With #6829, we probably don't need the `is_pytest_running` hackery that was added in this PR.

MridulS:
Let's continue this in https://github.com/networkx/networkx/pull/6840

PR #6829, opened 2023-08-01
add implementation to re-dispatch all functions for easier testing

aaronzo:
the `--backend=nx-loopback` option reregisters all algorithms before the tests execute.
the ability to reregister all (or some) functions I can see being useful in other situations, e.g. if a function is monkey-patched.
additionally, all dispatched functions keep a `dispatchargs` attribute which can be queried to see which aspects are kept when dispatching, which I think is also useful.

Came from discussion on #6819

### DIFF
--- a/networkx/utils/backends.py
+++ b/networkx/utils/backends.py
@@ -81,8 +81,10 @@ the test using `item.add_marker(pytest.mark.xfail(reason=...))`.
 import functools
 import inspect
 import os
+import pkgutil
 import sys
 from importlib.metadata import entry_points
+from itertools import chain
 
 from ..exception import NetworkXNotImplemented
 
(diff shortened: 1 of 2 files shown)


eriknw:
I like the idea of being able to use arguments to pytest. However, due to https://github.com/pytest-dev/pytest/issues/1596, I think it's awkward right now to use pytest when working on a backend. Ideally, I would like to be able to do:
```bash
pytest --pyargs networkx --backend graphblas
```
from `graphblas-algorithms` project directory, but this doesn't work. One option is to pass the absolute location of networkx to pytest
```bash
pytest `python -c 'import networkx as nx ; print(nx.__file__[:-11])'` --backend cugraph
```
but I find this even uglier than using environment variables.

One workaround [here](https://stackoverflow.com/questions/41270604/using-command-line-parameters-with-pytest-pyargs/43747114#43747114) would let us run networkx tests as:
```bash
python -m networkx.test --backend parallel
```
I would be fine with this approach, but it would need to be developed.

Relying on environment variables at import time as we do today probably isn't ideal, so I support trying to clean up this technical debt.

btw, when I run this PR locally, I get:
```python-traceback
NameError: name 'fallback_to_nx' is not defined
```
b/c `fallback_to_nx` is currently only defined when `NETWORKX_GRAPH_CONVERT` environment variable is set.

networkx/utils/backends.py
```diff
@@ -122,11 +124,33 @@ def __getitem__(self, name):
 _registered_algorithms = {}
 
 
-def _register_algo(name, wrapped_func):
+def _register_algo(name, wrapped_func, dispatch_args):
     if name in _registered_algorithms:
         raise KeyError(f"Algorithm already exists in dispatch registry: {name}")
     _registered_algorithms[name] = wrapped_func
     wrapped_func.dispatchname = name
+    wrapped_func.dispatchargs = dispatch_args
```
eriknw, line 132: I would rather call these attributes `dispatch_name` and `dispatch_args`.

networkx/utils/backends.py
```diff
@@ -122,11 +124,33 @@ def __getitem__(self, name):
...
+
+
+def reregister_all_algos(dispatch_override=None):
+    global _registered_algorithms
+
+    dispatch_fn = _dispatch if dispatch_override is None else dispatch_override
+    algos, _registered_algorithms = _registered_algorithms.copy(), {}
+    _registered_algorithms = {}
```
eriknw, line 140: `_registered_algorithms` is set twice here. I suspect you meant to update this to:
```suggestion
    algos = _registered_algorithms.copy()
    _registered_algorithms = {}
```

MridulS:
Thanks for starting this up @aaronzo!!

Let's continue this in https://github.com/networkx/networkx/pull/6840 :)

PR #6832, opened 2023-08-02
Add @nx._dispatch to {single_source,all_pairs}_all_shortest_paths, cd_index

eriknw:
These are the new functions that I know about that didn't yet dispatch.

### DIFF
--- a/networkx/algorithms/time_dependent.py
+++ b/networkx/algorithms/time_dependent.py
@@ -10,6 +10,7 @@ __all__ = ["cd_index"]
 
 @not_implemented_for("undirected")
 @not_implemented_for("multigraph")
+@nx._dispatch(node_attrs={"time": None, "weight": 1})
 def cd_index(G, node, *, time_delta=timedelta(days=5 * 365), time="time", weight=None):
     r"""Compute the CD index for `node` within the graph `G`.
 
--- a/networkx/algorithms/time_dependent.py
+++ b/networkx/algorithms/time_dependent.py
@@ -40,7 +40,7 @@ def cd_index(G, node, *, time_delta=timedelta(days=5 * 365), time="time", weight
 
     Raises
     ------
-    ValueError
+    NetworkXError
        If not all nodes have a `time` attribute or
        `time_delta` and `time` attribute types are not compatible or
        `n` equals 0.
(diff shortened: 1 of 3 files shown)


networkx/algorithms/time_dependent.py
```diff
@@ -10,6 +10,7 @@
 
 @not_implemented_for("undirected")
 @not_implemented_for("multigraph")
+@nx._dispatch(node_attrs={"time": None, "weight": 1})
```
eriknw, line 13: Note that `time` node attribute is required, but we don't yet have a way to indicate a required attribute in the dispatch decorator. If we did have a way, some backend conversions could be faster.

Also, why is `ValueError` raised if not all nodes have the time attribute? I would have expected `KeyError`.

MridulS: Or no wait, that doesn't help the backends right?

dschult: I think historical NetworkX convention would have this function raise NetworkXError (or a subclass of that) for these errors. They are specific to NetworkX api -- not Python api. But I doubt that is consistent throughout this library. :)
Perhaps we should change these before the next release so nobody expects them to be ValueErrors.

How does the choice error type affect backends? Couldn't they raise NetworkXError too?

MridulS: > How does the choice error type affect backends? Couldn't they raise NetworkXError too?

Then they need to depend on NetworkX, currently it's possible the backend itself has no dependency on NetworkX.

dschult: Got it -- but lots of our functions raise NetworkXErrors of various types. Don't they need to depend on NetworkX to implement those functions?

I think we shouldn't be choosing error types in a way that doesn't allow us to create our own error types :)

eriknw: +1 for raising a networkx error.

MridulS: Do you want to replace the error in this PR @eriknw ?
I would also vote to rename the `ValueError` at

https://github.com/networkx/networkx/blob/301bf129da3e45853c7e4da946a04c7840a49a78/networkx/algorithms/time_dependent.py#L100

to `NetworkXError`.

eriknw: Done. Changed all exceptions raised by `cd_index`  to `NetworkXError`, including on lines 92, 100, and 111.

MridulS (approved):
Thanks @eriknw!

dschult (approved):
Thanks!!

Issue #6804, opened 2023-07-22
Followup ideas from the `_dispatch` code audit


dschult:
PR #6688 required going through a lot of our code base and finding out which input arguments were treated how. In that PR, many comments were created and attached to various parts of our codebase. This Issue attempts to collect the comments that didn't lead to an immediate solution or answer.  The idea is to check them off when we either resolve them or decide not to resolve them. If I've missed something please add it in the comments and I'll try to put it on the list via editing.

## Timing impact
- [x] timing of functions when not dispatched
    Full test suite for networkx: (very crude test using pytest time reporting)
      - without dispatch decorators (before PR #6688): 48.14s   
      - with dispatch decorators: 48.83s
      - in nx-loopback test mode without dispatch decorators (should take longer due to converting graphs): 56.37s
      - in nx-loopback test mode without dispatch decorators (should take longer due to converting graphs): 154.24s
    Time a function call that takes very little time: `G=nx.path_graph(5); %timeit nx.shortest_path(0,3)`
      - without dispatch decorators: 2.67 µs
      - with dispatch decorators: 3.3 µs
    Increase number of nodes:  `G=nx.path_graph(50);%timeit nx.shortest_path(G, 0, 47)`
      - without dispatch decorators: 21.9 µs
      - with dispatch decorators: 22.3 µs
- [x] timing of import of networkx:   
    Change to new git branch, then `python -X importtime -c "import networkx"`
      - without dispatch decorators: ~70 +/-2 ms    (200+/-4 ms for first invocation after switching branches)
      - with dispatch decorators: ~72 +/-2 ms    (204+/-4 ms for first invocation after switching branches)

## API warts:
- [ ] algorithms/tree/operations.py::join  accepts a list of 2-tuples with (graph, root) pairs [**left unchanged:** leave until rooted trees have unified interface]
- [ ] algorithms/tree/branchings.py::maximum_branching: handles "preserve_attrs" input specially for True/False. If True it preserves all the attributes -- not just the one used in the algorithm.  [**left unchanged:**]
- [x] convert.py::to_dict_of_dicts has argument `edge_data` which is a value [changed docs but not behavior]
- [x] algorithms/approximation/treewidth.py::min_fill_in_heuristic has input `graph` which is expected to be a dict.
- [x] algorithms/shortest_paths/astar.py::astar_path The trick here is that when heuristic arg is callable, all node attributes need to be preserved.  We won't change this.
[...]

PR #6840, opened 2023-08-04
Change `_dispatch` to a class instead of a closure

eriknw:
This PR supersedes #6828 and #6829, and it implements one of the "future" items from #6804:
- change `_dispatch` to be a class instead of a closure
- fallback to original networkx function when testing if `fallback_to_nx` is set when backend raises `NotImplementedError` (also done in #6828)
- don't dispatch to backend when testing until tests are set up (also done in #6828)
- add `--backend=whatever` option to `pytest` to choose backend w/o environment variable (also done in #6829)
- add `--fallback-to-nx` option to `pytest` to run test with networkx graph is backend does not implement it
- rename `NETWORKX_GRAPH_CONVERT` environment variable to `NETWORKX_TEST_BACKEND`
- rename `NETWORKX_BACKEND_TEST_EXHAUSTIVE` environment variable to `NETWORKX_FALLBACK_TO_NX`
- add `**backend_kwargs` to signature to allow backends to have additional keyword arguments
- ~add `is_testing=` argument to `convert_from_nx`, which I would find useful for my plugins~ (removed)
- add `_dispatch._convert_and_call` method, which gets us one step closer to automatic conversions
- Newly added on 2023-08-18:
  - `NETWORKX_AUTOMATIC_BACKENDS` to enable automatic conversions to backends
  - add `backend=` keyword-only parameter to allow the backend to be explicitly chosen at call-time
  - add `_can_backend_run` and `_convert_arguments` methods to `_dispatch`; should we make these public?

The backend can still be indicated with `NETWORKX_TEST_BACKEND` environment variable, which allows us to run networkx tests with backends when networkx is installed in site-packages (b/c `pytest --backend` doesn't work in this case).

The testing environment variables are only handled when `pytest` is invoked. Previously, we checked the environment variables at import time. This is one reason I thought it appropriate to change the names of the environment variables.

Behavior of dispatching (such as changing plugin name) can be easily modified at any time by changing class attributes on `_dispatch`.

I exposed a lot of attributes of `_dispatch` objects. I would appreciate scrutiny on choice of names. We can make some/most of these private if people prefer. Keep in mind that we will likely add more attributes and methods later when we add e.g. configuration, discovering and choosing backends, etc. Having `_dispatch` as a class sets us up for more future goodness.

I have not measured performance impact of these changes. If it's significant, an easy optimization would be to compute `__signature__` on demand when needed.

Unlike #6828 and #6829, this PR has a lot more to review--sorry! I figure it would be better to make breaking changes before 3.2 is released if at all possible.

CC @jim22k @aaronzo

### DIFF
--- a/networkx/utils/backends.py
+++ b/networkx/utils/backends.py
@@ -84,10 +84,10 @@ It will be called with the list of NetworkX tests discovered. Each item
 is a test object that can be marked as xfail if the backend does not support
 the test using `item.add_marker(pytest.mark.xfail(reason=...))`.
 """
-import functools
 import inspect
 import os
 import sys
+from functools import partial
 from importlib.metadata import entry_points
 
 from ..exception import NetworkXNotImplemented
@@ -216,7 +216,7 @@ class _dispatch:
         preserve_all_attrs=False,
     ):
         if func is None:
-            return functools.partial(
+            return partial(
                 _dispatch,
                 name=name,
                 graphs=graphs,
@@ -235,10 +235,16 @@ class _dispatch:
 
         self = object.__new__(cls)
 
-        # Make it look like the original function
-        functools.update_wrapper(self, func)
+        # standard function-wrapping stuff
+        # __annotations__ not used
+        self.__name__ = func.__name__
+        self.__doc__ = func.__doc__
         self.__defaults__ = func.__defaults__
         self.__kwdefaults__ = func.__kwdefaults__
+        self.__module__ = func.__module__
+        self.__qualname__ = func.__qualname__
+        self.__dict__.update(func.__dict__)
+        self.__wrapped__ = func
 
         self.orig_func = func
         self.name = name
@@ -356,7 +362,6 @@ class _dispatch:
         # }
 
         if self._is_testing and self._plugin_name is not None:
-            # Can we use graphs_resolved here?
             return self._convert_and_call(
                 self._plugin_name,
                 args,
--- a/networkx/utils/backends.py
+++ b/networkx/utils/backends.py
@@ -193,7 +193,9 @@ class _dispatch:
         os.environ.get("NETWORKX_FALLBACK_TO_NX", "true").strip().lower() == "true"
     )
     _automatic_backends = [
-        x.strip() for x in os.environ.get("NETWORKX_AUTOMATIC_BACKENDS", "").split(",")
+        x.strip()
+        for x in os.environ.get("NETWORKX_AUTOMATIC_BACKENDS", "").split(",")
+        if x.strip()
     ]
 
     def __new__(
--- a/networkx/utils/backends.py
+++ b/networkx/utils/backends.py
@@ -236,7 +236,11 @@ class _dispatch:
         self.__name__ = func.__name__
         self.__doc__ = func.__doc__
         self.__defaults__ = func.__defaults__
-        self.__kwdefaults__ = func.__kwdefaults__
+        # We "magically" add `backend=` keyword argument to allow backend to be specified
+        if func.__kwdefaults__:
+            self.__kwdefaults__ = {**func.__kwdefaults__, "backend": None}
+        else:
+            self.__kwdefaults__ = {"backend": None}
         self.__module__ = func.__module__
         self.__qualname__ = func.__qualname__
         self.__dict__.update(func.__dict__)
@@ -311,6 +315,8 @@ class _dispatch:
     def __signature__(self):
         if self._sig is None:
             sig = inspect.signature(self.orig_func)
+            # `backend` is now a reserved argument used by dispatching.
+            # assert "backend" not in sig.parameters
             if not any(
                 p.kind == inspect.Parameter.VAR_KEYWORD for p in sig.parameters.values()
             ):
@@ -318,17 +324,37 @@ class _dispatch:
                     parameters=[
                         *sig.parameters.values(),
                         inspect.Parameter(
+                            "backend", inspect.Parameter.KEYWORD_ONLY, default=None
+                        ),
+                        inspect.Parameter(
                             "backend_kwargs", inspect.Parameter.VAR_KEYWORD
                         ),
+                    ]
+                )
+            else:
+                *parameters, var_keyword = sig.parameters.values()
+                sig = sig.replace(
+                    parameters=[
+                        *parameters,
+                        inspect.Parameter(
+                            "backend", inspect.Parameter.KEYWORD_ONLY, default=None
+                        ),
+                        var_keyword,
                     ]
                 )
             self._sig = sig
         return self._sig
 
-    def __call__(self, /, *args, **kwargs):
+    def __call__(self, /, *args, backend=None, **kwargs):
         if not plugins:
             # Fast path if no backends are installed
             return self.orig_func(*args, **kwargs)
+
+        # Use `backend_name` in this function instead of `backend`
+        backend_name = backend
+        if backend_name is not None and backend_name not in plugins:
+            raise ImportError(f"Unable to load plugin: {backend_name}")
+
         graphs_resolved = {}
         for gname, pos in self.graphs.items():
             if pos < len(args):
@@ -360,7 +386,7 @@ class _dispatch:
         #     if (val := args[pos] if pos < len(args) else kwargs.get(gname)) is not None
         # }
 
-        if self._is_testing and self._automatic_backends:
+        if self._is_testing and self._automatic_backends and backend_name is None:
             # Special path if we are running networkx tests with a backend.
             return self._convert_and_call_for_tests(
                 self._automatic_backends[0],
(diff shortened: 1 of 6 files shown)


eriknw:
What should `__repr__` and `__str__` be?

aaronzo (commented):
@eriknw I like the changes, certainly avoids the hacks I had to do in #6829 by having a nicer architecture using a class - definitely welcome that.

If we could solve this issue without having class attributes & methods that are only used for testing that would be preferable though - Its a shame to have boolean checks for `_is_testing` executing in user/production code. However hacky the previous test set-up was, at least it didn't affect execution in production at all. I'm sure the footprint of this is minimal here, it's more of a question of whether it's a principle we're happy to go against here for the pay-off.

Maybe if:

1. `_dispatch` is a class, as currently in this PR.
2. `_dispatch.__call__` calls a private function, `_call_dispatched`. This function sits tucked away at module level, not class level.
3. during testing, this function `_call_dispatched` is monkey patched to a test override one.

since the class only references the call function rather than implements it, this should mean the test code doesn't affect prod code at all. I believe the `_is_testing` etc class variables could be removed. `_convert_and_call` could still be useful, I'd keep that, but remove the `is_testing` parameters.

---

As for `__repr__` and `__str__`, I'd use
```python
def __repr__(self):
    # basically the full string to reconstruct the dispatch
    return "_dispatch(" + repr(self.orig_func) + 
     ",name=" + repr(self.name) + ",graphs=" + repr(self.graphs) + ... + ")"

def __str__(self):
   # a shorter version with the function name
   return "_dispatch(" + repr(self.orig_func) + ",name=" + repr(self.name) + ")"
```

networkx/utils/backends.py
```diff
@@ -196,110 +182,172 @@ def _dispatch(
...
+                ]
+            )
+        self.__signature__ = sig
+
+        if name in _registered_algorithms:
+            raise KeyError(f"Algorithm already exists in dispatch registry: {name}")
+        _registered_algorithms[name] = self
+        return self
+
+    def __call__(self, *args, **kwargs):
```
MridulS, line 284: I was thinking if it makes sense to have a short circuit here if there are no new backends installed (except nx-loopback)? something like:
``` python
        if len(plugins.items) == 1:
            return self.orig_func(*args, **kwargs)
```
This will not fail elegantly if a backend graph is passed as an argument without installing the backend but we will have an early return without going through all the logic below this.

networkx/utils/backends.py
```diff
@@ -315,333 +363,295 @@ def wrapper(*args, **kwds):
...
+        try:
+            result = getattr(backend, self.name)(*bound.args, **bound.kwargs)
+        except (NotImplementedError, NetworkXNotImplemented) as exc:
+            if fallback_to_nx:
+                return self.orig_func(*args, **kwargs)
+            if is_testing:
+                import pytest
+
+                pytest.xfail(
+                    exc.args[0] if exc.args else f"{name} raised {type(exc).__name__}"
```
MridulS, line 611:
```suggestion
                    exc.args[0] if exc.args else f"{self.name} raised {type(exc).__name__}"
```

networkx/utils/backends.py
eriknw: I like the idea, but how would we test with the `nx-loopback` backend?

networkx/utils/backends.py
eriknw: Thanks. Included in latest commit. Black would rewrite your suggestion.

eriknw:
Thanks for looking and giving feedback @aaronzo.

I just added `can_run(name, args, kwargs)` to ask a backend if it can run an algorithm with the given arguments. Is there a better name? See how I use it in `cugraph-nx` (from this PR https://github.com/rapidsai/cugraph/pull/3614):
- https://github.com/rapidsai/cugraph/blob/66c23dc965c9ca2a937799a3de6862b37b5e4d46/python/cugraph-nx/cugraph_nx/interface.py#L68-L80
- https://github.com/rapidsai/cugraph/blob/66c23dc965c9ca2a937799a3de6862b37b5e4d46/python/cugraph-nx/cugraph_nx/algorithms/centrality/betweenness.py#L43-L45

Regarding `is_testing`, I removed the `is_testing=` argument that I added to `convert_from_nx`, so maybe this is a little more pure for your liking? I was able to do this by adding a separate `can_run` function that backends can define, which solves my needs for the moment. Right now, we _only_ use `_convert_and_call` during testing. It's an implementation detail. The `if is_testing` blocks are meant to identify testing-specific behavior to help us refactor in the future. Otherwise, I would inline the `if is_testing` block and remove the `else` block, because, again, right now this is only for testing. Other use cases are future work. Also, fwiw, I don't mind having `if is_testing` in "production code" if it means I get to avoid monkeypatching one function with another function that is 90% the same. Having two versions of a function this complicated is begging for mistakes. DRY applies here. I also don't want to decompose `_convert_and_call` into sub-functions just yet, but maybe this would be helpful to do in the future?

As for repr and str, I was thinking something like `"<function pagerank backends=[networkx, cugraph, graphblas, parallel]>`"

networkx/utils/backends.py
eriknw: Done. I added a fast path in the latest commit. We don't include `"nx-loopback"` as a backend unless it's being tested.

eriknw:
My latest commits added `NETWORKX_AUTOMATIC_BACKENDS` to enable automatic conversions to backends! This config option is a prioritized _list_ of backends.

@aaronzo I would appreciate another review, b/c I think I adequately addressed the issues you raised. Thanks for that btw, and sorry if my previous tone was too "I'm busy now, let's worry about it later". I aspire to write clean code that others can read and maintain. I have a bit more time now, so let's try to make this PR good (but let's punt on config for now and rely on environment variables for global config). I added a few `# Future work:` comments for clarity.

So, I refactored `_convert_and_call` by adding `_can_backend_run`, `_convert_arguments`, and `_convert_and_call_for_tests`. I wonder whether we should expose `can_backend_run` or `convert_arguments` to the user.

eriknw:
One more enhancement: `backend=` argument to dispatchable functions to specify the backend.

For example, `nx.betweenness_centrality(G, backend="cugraph")`.

Call me crazy, but this is starting to feel kinda nice. This new behavior needs documented. Also, it's not tested directly in networkx, but backends should have an easier time exercising this code.

aaronzo (commented):
I'm happy with the separation of test code now, I think it's been minimized as much as reasonable and you've clearly taken care for it not to affect performance. Nice work :)

On your latest change adding the `backend=` kwarg - I like the idea but am not sure on details.

If I have some graphs as inputs to a function that use backend A through `__networkx_plugin__`, and then I call the function with `backend=B` - what happens? Will dispatch automatically search for converters between backends? I think this could lead to some confusing behaviour. We also have to agree that no dispatched algorithms are allowed a 'backend' parameter of their own (though I can't see why they would). It's also not clear which should take priority out of `G.__networkx_plugin__` and `backend=`.

I'd encourage that to go in a separate PR and have its own discussion, since everything else here I think is a clear improvement

MridulS:
BTW I was able to run these changes with grapblas as a backend with:
``` bash
$ pytest --backend=graphblas networkx
...
...
==== 2263 passed, 11 skipped, 2955 xfailed, 3 warnings in 70.45s (0:01:10) =====
```

```bash
$ pytest --backend=graphblas --fallback-to-nx networkx
...
...
====== 5197 passed, 11 skipped, 5 xfailed, 16 xpassed, 7 warnings in 64.02s (0:01:04) =====
```

networkx/utils/backends.py
```diff
@@ -308,341 +428,370 @@ def wrapper(*args, **kwds):
...
+            if hasattr(backend, self.name):
+                msg += " with the given arguments"
+            pytest.xfail(msg)
+
+        converted_args, converted_kwargs = self._convert_arguments(
+            plugin_name, args, kwargs
+        )
+
+        try:
+            result = getattr(backend, self.name)(*converted_args, **converted_kwargs)
```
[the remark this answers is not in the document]
eriknw: Done! Good catch 👍

networkx/utils/backends.py
```diff
@@ -196,110 +176,250 @@ def _dispatch(
...
+        if isinstance(func, str):
+            raise TypeError("'name' and 'graphs' must be passed by keyword") from None
+        # If name not provided, use the name of the function
+        if name is None:
+            name = func.__name__
+
+        self = object.__new__(cls)
+
+        # standard function-wrapping stuff
+        # __annotations__ not used
```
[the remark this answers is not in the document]
eriknw: No idea. If there are stub files, hopefully `__annotations__` is automagically updated 🤷‍♂️ . Anyway, `argmap` doesn't copy `__annotations__` either.

MridulS:
Thanks @eriknw !!

MridulS:
There seems to be unintended change by this PR. All the dispatched functions are picking up `<networkx.utils.backends._dispatch object>` in their sphinx docs now. You can check
https://networkx.org/documentation/latest/search.html?q=networkx.utils.backends._dispatch+object

PR #6843, opened 2023-08-09
Fast label propagation algorithm for community detection

lovre:
Fast variant of label propagation algorithm for community detection (documented, referenced, with tests).

Vincent A. Traag & Lovro Šubelj. "Large network community detection by fast label propagation." Scientific Reports 13 (2023): 2701. [https://doi.org/10.1038/s41598-023-29610-z](https://doi.org/10.1038/s41598-023-29610-z)

The implementation should work with any type of NetworkX graph (possibly weighted).

### DIFF
--- a/networkx/algorithms/community/label_propagation.py
+++ b/networkx/algorithms/community/label_propagation.py
@@ -1,12 +1,123 @@
 """
 Label propagation community detection algorithms.
 """
-from collections import Counter, defaultdict
+from collections import Counter, defaultdict, deque
 
 import networkx as nx
 from networkx.utils import groups, not_implemented_for, py_random_state
 
-__all__ = ["label_propagation_communities", "asyn_lpa_communities"]
+__all__ = [
+    "label_propagation_communities",
+    "asyn_lpa_communities",
+    "fast_label_propagation_communities",
+]
+
+
+@nx._dispatch(edge_attrs="weight")
+@py_random_state("seed")
+def fast_label_propagation_communities(G, weight=None, seed=None):
+    """Returns communities in `G` as detected by fast label propagation.
+
+    The fast label propagation algorithm is described in [1]_. The algorithm is probabilistic and the found communities may vary in different executions.
+
+    The algorithm operates as follows. First, the community label of each node is set to a unique label. The algorithm then repeatedly updates the labels of the nodes to the most frequent label in their neighborhood. In case of ties, a random label is chosen from the most frequent labels.
+
+    The algorithm maintains a queue of nodes that still need to be processed. Initially, all nodes are added to the queue in a random order. Then the nodes are removed from the queue one by one and processed. If a node updates its label, all its neighbors that have a different label are added to the queue (if not already in the queue). The algorithm stops when the queue is empty.
+
+    Parameters
+    ----------
+    G : Graph, DiGraph, MultiGraph, or MultiDiGraph
+      Any NetworkX graph.
+
+    weight : string, or None (default)
+      The edge attribute representing a non-negative weight of an edge. If None, each edge is assumed to have weight one. The weight of an edge is used in determining the frequency with which a label appears among the neighbors of a node (an edge with a non-negative weight `w` is equivalent to `w` unweighted edges).
+
+    seed : integer, random_state, or None (default)
+      Indicator of random number generation state. See :ref:`Randomness<randomness>`.
+
+    Returns
+    -------
+    communities : iterable
+      Iterable of communities given as sets of nodes.
+
+    Notes
+    -----
+    Edge directions are ignored for directed graphs. Edge weights must be non-negative numbers.
+
+    References
+    ----------
+    .. [1] Vincent A. Traag & Lovro Šubelj. "Large network community detection by fast label propagation." Scientific Reports 13 (2023): 2701. https://doi.org/10.1038/s41598-023-29610-z
+    """
+
+    # Queue of nodes to be processed.
+    Q = deque(G)
+    seed.shuffle(Q)
+
+    # Set of nodes in the queue.
+    S = set(G)
+
+    # Assign unique label to each node.
+    C = {i: c for c, i in enumerate(G)}
+
+    while Q:
+        # Remove next node from the queue to process.
+        i = Q.popleft()
+        S.remove(i)
+
+        # Isolated nodes retain their initial label.
+        if G.degree(i) > 0:
+            # Compute frequency of labels in node's neighborhood.
+            L = _fast_label_count(G, C, i, weight)
+            maxl = max(L.values())
+
+            # Always sample new label from most frequent labels.
+            c = seed.choice([c for c in L if L[c] == maxl])
+
+            if C[i] != c:
+                C[i] = c
+
+                # Add neighbors that have different label to the queue.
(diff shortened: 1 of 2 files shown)


lovre:
Check for Ubuntu, Python 3.11 was not succesful.

However, as far as I can tell, what fails is an **existing** test in `networkx/algorithms/community/tests/test_label_propagation.py`, in particular test `test_seed_argument(self)` in class `TestAsynLpaCommunities`.

How should I proceed?

(Could the problem be that there is also a test function with the same name in class `TestFastLabelPropagationCommunities`?)

dschult (commented):
I was trying to figure out why the dispatching test is meeting an infinite loop. But I have a couple of comments that I'd like to ask of you even though that one test is running forever:
- Can you wrap the doc_string lines to ~80 chars?
- The variable names seem cryptic -- and too many single letter names. Can you switch `i` to `node` or at least `'n' and `j` to `nbr` when it is referring to a node? (Or `i` & `j` to `u` & `v`)? Can you extend `c` to `comm`, and `C` to `comms`? And `S` to `nodes_set`? And `L` to `lbl_cnts` or `label_counts`?  I'm just making these up -- choose something that makes sense to you and the code. I'm just asking for more descriptive var names -- or maybe a better description is names similar to what other functions use.
- Use `G.is_multigraph()` instead of using `type(G)`.

Any preliminary or quick checks on how much this speeds things up? Any reason it shouldn't just become the default method to use?

MridulS:
This was annoying to debug, but the reason of failure is because of the explicit type checks of the graph objects.

> Use G.is_multigraph() instead of using type(G).

Using this should fix the dispatch tests :)

networkx/algorithms/community/label_propagation.py
```diff
@@ -1,12 +1,123 @@
...
+                    L[C[j]] += len(G.pred[i][j])
+
+    else:
+        # Weighted undirected simple/multigraph.
+        L = defaultdict(float)
+        for _, j, w in G.edges(i, data=weight, default=1):
+            L[C[j]] += w
+
+        # Weighted directed simple/multigraph.
+        if type(G) == nx.DiGraph or type(G) == nx.MultiDiGraph:
```
MridulS, line 116:
```suggestion
        if not G.is_multigraph():
            L = Counter(map(C.get, nx.all_neighbors(G, i)))

        # Unweighted (un)directed multigraph.
        else:
            L = defaultdict(int)
            for j in G[i]:
                L[C[j]] += len(G[i][j])

            if G.is_multigraph() and G.is_directed():
                for j in G.pred[i]:
                    L[C[j]] += len(G.pred[i][j])

    else:
        # Weighted undirected simple/multigraph.
        L = defaultdict(float)
        for _, j, w in G.edges(i, data=weight, default=1):
            L[C[j]] += w

        # Weighted directed simple/multigraph.
        if G.is_directed():
```

dschult:
Nice catch!! 
-- the dispatch graphs do not have type `nx.Graph`.

lovre:
Thank you both!

I have submitted a new version following all the suggestions, which now passes all the checks :)

lovre:
> Any preliminary or quick checks on how much this speeds things up? Any reason it shouldn't just become the default method to use?

Default method for what?

- Default label propagation method? Yes
- Default method for very large graphs? Yes
- Default method in NetworkX? No

Results in our [paper](https://doi.org/10.1038/s41598-023-29610-z) suggest that FLPA is 30-700 times faster than LPA. As a rule of thumb, the igraph implementation used in the paper takes about one second per million edges.

FLPA can be useful as a first initial look at the network. It allows to see beyond connectivity in almost linear time. However, in case of smaller networks, other methods are arguably more robust and preferable (e.g. variants of SBM).

lovre:
I would also like to comment that implementation of `asyn_lpa_communities(G)` is not what is advertised in the documentation [1]. It actually implements a later algorithm [2]. There are two main differences:

- The implementation uses a retention strategy, which means that you only sample a new label when the current label is not among most frequent labels. This greatly restricts the dynamics and doesn't produce desirable results even in synthetic graphs (e.g., random graphs, assortative SBM-s).
- The implementation does not stop when all labels are dominant, but when no label has changed on a particular iteration. This is faster and easier to implement, but not the same as the dominance criteria.

For pseudocode of [1] & [2] see Algorithm 1 & 2 in our [paper](https://doi.org/10.1038/s41598-023-29610-z).

I would suggest to correct the documentation. I do not think it makes sense to include also a "proper" implementation of LPA, since FLPA is generally preferable.

[1] U. N. Raghavan, R. Albert, and S. Kumara, “Near linear time algorithm to detect community structures in large-scale networks,” Phys. Rev. E 76, 036106 (2007).

[2] M. J. Barber and J. W. Clark, “Detecting network communities by propagating labels under constraints,” Phys. Rev. E 80, 026129 (2009).

dschult:
Thanks for the quick changes and the complete responses to questions. 

Are you able to give a description of why this queue is expected to eventually empty? It seems like it'd be easy to get into an oscillation where nodes are shifting labels back and forth repeatedly -- and each time they change they'd be adding their neighbors to the queue.

I agree with your proposal to correct the documentation for `asyn_lpa_communities` while leaving the functions as is, and probably not implementing a "proper" LPA implementation unless the literature finds a compelling reason for it.  Those changes could be in a separate PR, or you could add them to this PR. It might be easier to review as a separate PR, but perhaps slightly more work for you. Let us know which you intend to pursue. And someone else might chime in too if there are other perspectives.

dschult (approved):
Thanks for the helpful answer.

This PR looks good to me! (and I'll review the separate PR when it gets here)

vtraag:
> Are you able to give a description of why this queue is expected to eventually empty? It seems like it'd be easy to get into an oscillation where nodes are shifting labels back and forth repeatedly -- and each time they change they'd be adding their neighbors to the queue.

The continuous oscillation back and forth will eventually die out with probability tending to 1. That is, a new label will always be sampled from the maximum labels. There is hence always a non-zero probability that the same maximum label will be chosen, and that hence the label will not change. In that case, no neighbours of that node will be added to the queue. The same reasoning applies to all such possibly oscillating nodes, and so the algorithm is in fact guaranteed to stop. In theory, this can take a long time (with ever decreasing probability), but in practice this is hardly noticeable.

lovre:
Dear Dan,

We are wondering why FLPA does not appear here, [https://networkx.org/documentation/stable/reference/algorithms/community.html#module-networkx.algorithms.community.label_propagation](https://networkx.org/documentation/stable/reference/algorithms/community.html#module-networkx.algorithms.community.label_propagation)?

Best,
Lovro

dschult:
We forgot to put a link into the docs for the new function. 
A separate PR adding it to `doc/reference/algorithms/community.rst`
should be made. Sorry that we didn't catch that. The new line goes [here](https://github.com/networkx/networkx/blob/main/doc/reference/algorithms/community.rst#label-propagation)
Thanks for catching this and letting us know!

MridulS:
@lovre the link is updated now on the latest branch doc build. The "stable" docs are updated only when we make a release :)

https://networkx.org/documentation/latest/reference/algorithms/community.html#module-networkx.algorithms.community.label_propagation

lovre:
Thank you all!

PR #6845, opened 2023-08-13
MAINT: fix link to nightly releases wheels

MridulS:


MridulS:
self merging to test the workflow dispatch.

PR #6860, opened 2023-08-19
fix: make messages readable

diohabara:
Some messages are not properly separated by spaces.

E.g., when the following code (`main.py`) is run by `python main.py`,

```python
import networkx as nx

def main():
    G = nx.complete_graph(4)
    pos = nx.nx_pydot.graphviz_layout(G)
    nx.draw(G, pos=pos)
    plt.savefig("main.png")

if __name__ == "__main__":
    main()
```

the execution result look like below.

```
/home/user/main.py:6: DeprecationWarning: nx.nx_pydot.graphviz_layout depends on the pydot package, which hasknown issues and is not actively maintained. Consider usingnx.nx_agraph.graphviz_layout instead.

See https://github.com/networkx/networkx/issues/5723
  pos = nx.nx_pydot.graphviz_layout(G)
```

However, this should look like below. Spaces are inserted between "has" and "known", and "using" and "nx.nx_agraph.graphviz".

```
/home/user/main.py:6: DeprecationWarning: nx.nx_pydot.graphviz_layout depends on the pydot package, which has known issues and is not actively maintained. Consider using nx.nx_agraph.graphviz_layout instead.

See https://github.com/networkx/networkx/issues/5723
  pos = nx.nx_pydot.graphviz_layout(G)
```

### DIFF
--- a/networkx/algorithms/approximation/steinertree.py
+++ b/networkx/algorithms/approximation/steinertree.py
@@ -198,7 +198,7 @@ def steiner_tree(G, terminal_nodes, weight="weight", method=None):
         import warnings
 
         msg = (
-            "steiner_tree will change default method from 'kou' to 'mehlhorn'"
+            "steiner_tree will change default method from 'kou' to 'mehlhorn' "
             "in version 3.2.\nSet the `method` kwarg to remove this warning."
         )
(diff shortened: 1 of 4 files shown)


dschult (approved):
Thanks!  This helps. :)

PR #6869, opened 2023-08-22
Allow seed of np.random instance to exactly produce arbitrarily large integers

### LINKED ISSUE #6848: Heads up.... numpy random integer generators cant handle large integers
It looks like numpy random number generators cant handle large integers (large being > `2**64`, i.e. not held in `int64`). They raise an error saying that the values can't be held in `int64` (which of course is correct).  And if you try to specify that they use python integers (which don't have that limit)  by using `dtype=np.object_`, it raises stating that np.object_ is not of integer type. 
We probably use large integers more than most scientific libraries (20**20 is the number of possible edges with 21 nodes). So, we may run into this more often than some. 

Workaround: instead of `seed = np.random.default_rng(); seed.integers(0, 20**20)`, we could convert it to a floating point computation and discretize later:
```python
seed = np.random.default.rng()
def randint(a, b, seed):
    return a + int(seed.random() * (b - a)) 

randint(0, 20**20, seed)
```
This could/should be added to the `utils.misc.py` class `PythonRandomInterface` method `randint`.

### Numpy fix
Since numpy is creating the value, it seems like they don't need to raise an error for `dtype=np.object_` in `np.random.Generator.integers`. It could be made an integer Python object and be perfectly good output.
We could/should look into making a PR to numpy to allow `integers` to return Python integers when `dtype=object_`

I think this is the first time we've noticed a feature of `random` that isn't provided by `numpy.random`, but there may be others.

dschult:
Fixes #6848 

- Introduce PythonRandomViaNumpyBits to replace PythonRandomInterface. 
  Follows Robert Kern @rkern comment suggestion in numpy/numpy#24458
- rewrite old interface to not raise (and use new interface to get a value) when high limit is > max int64 value
- set up `create_py_random_state` to use old interface for possible Legacy users 
   (RandomState that isn't the default RandomState; All others users dont have claim to maintain the stream)
- set up `create_py_random_state` so all other input provides the new interface
- add smoke test of requesting very large int even with old interface
- update tests of create_py_random_state to reflect changes

Technical debt:
- Update decorator doc_strings to reflect numpy.random.Generators
- shift to `random()` from `random_sample()`
- correct the broken tests in test_decorators (assert a,b => assert a==b
- add decorator tests for numpy Generator class


dschult:
After our discussion in todays community meeting I looked at this again to refresh my memory.
We **are** keeping the old code (`PythonRandomInterface`) and using that whenever a user provides an instance of a numpy.RandomState that is not the default numpy RandomState instance. That means anyone who has gone through the trouble of creating their own RandomState instance with a seed and then passed that into our code will still get the old results. 

Also, unlike what we were talking about in the conversation:
we **don't** use numpy RandomState by default when it is available. For the code that uses Python random, we use a `random.Random` instance by default!  So if someone uses `seed=123` as the input to a function decorated with `@py_random_state`, they get a random.Random instance initialized by 123.  Our decorators only provide the **option** of using a specific numpy bit-stream, and users must create that on their own and pass it in to each function they call.

```
Input                                 Change in integers produced
seed=None                           no change (Python random RNG)
seed=123                            no change (Python random RNG)
seed=random                         no change (Python random RNG)
seed=random.Random(123)             no change (Python random RNG)
seed=np.random.RandomState(123)     no change       (but no longer exception for large ints)
seed=np.default_rng(123)            different integers  (no longer exception for large ints)
seed=np.random                      different integers  (no longer exception for large ints)
```

Functions which use the numpy random tools (via `@np_random_state`) have the same behavior as before and can't create large ints `(large means:   >= 2^64)`

stefanv:
@dschult I've received a bit of education on this topic the last week :), so I can help review if you're going ahead with this PR.

dschult:
Thanks @stefanv -- it would be great if you could review this. 
We are keeping around the older PythonRandomInterface class for backward compatibility. It is triggered when someone inputs a RandomState instance that is not the default np.random instance.  If they use the default then they aren't worried about backward reproducibility.  If they don't use a RandomState instance they also aren't worried about reproducibility across versions, etc. 

For all the normal cases we use a generator with the new class `PythonRandomWithNumPyBits` (mostly rkern's code). And, of course, if the networkx function uses np.random function calls there isn't any expectation of arbitrarily large integer values, so that remains unchanged.

Thanks!!

stefanv:
ref #6758

stefanv:
RNG mechanisms

rossbar (approved):

dschult:
Just a restatement for clarity that:
- for the code that uses this functionality, the default random number system is Python Random... So this is independent of any backward bit stream compatibility issues for the default.  
- Other functions that use the numpy random interface in the code (I think there are only several of these, 4-5 maybe) might have default randomstate bit streams. We should look at those and see how critical that bit-stream compat would be to us.

We still have bit-stream compat for code that explicitly uses a numpy randomstate. We are keeping the old code in place to serve that purpose.

networkx/utils/misc.py
```diff
@@ -271,7 +273,68 @@ def create_random_state(random_state=None):
     raise ValueError(msg)
 
 
+class PythonRandomViaNumpyBits(random.Random):
+    """Provide the random.random algorithms using a Numpy.random bit generator
```
stefanv, line 277: `NumPy random` or `numpy.random`?

networkx/utils/misc.py
```diff
@@ -271,7 +273,68 @@ def create_random_state(random_state=None):
...
+    bit-stream for all work with NetworkX. This implementation is based on helpful
+    comments and code from Robert Kern on NumPy's GitHub Issue #24458.
+
+    This implementation supercedes that of `PythonRandomInterface` which rewrote
+    methods to account for subtle differences in API between `random` and
+    `numpy.random`. Instead this subclasses `random.Random` and overwrites
+    the methods `random`, `getrandbits`, `getstate`, `setstate` and `seed`.
+    It makes them use the rng values from an input numpy `RandomState` or `Generator`.
+    Those few methods allow the rest of the `random.Random` methods to provide
+    the API interface of `random.random` whlie using randomness generated by
```
stefanv, line 290: while

networkx/utils/misc.py
```diff
@@ -271,7 +273,68 @@ def create_random_state(random_state=None):
...
+        return x >> (numbytes * 8 - k)  # trim excess bits
+
+    def getstate(self):
+        return self._rng.__getstate__()
+
+    def setstate(self, state):
+        self._rng.__setstate__(state)
+
+    def seed(self, *args, **kwds):
+        "Do nothing override method."
```
stefanv, line 329: Should this raise?

networkx/utils/misc.py
```diff
@@ -357,32 +430,50 @@ def create_py_random_state(random_state=None):
...
-    except ImportError:
-        pass
 
+    Note: Conversion from older PythonRandomInterface to PythonRandomViaNumpyBits
+    is handled here to allow users of Legacy `numpy.random.RandomState` to exactly
+    match the legacy values produced. We assume that if a user cares about legacy
+    values, they are using a np.RandomState instance that is not the numpy default.
+    The default instance has state reset for each Python session. The Generator
+    class does not guarantee to maintain bit stream across versions. We wrap any
+    RandomState instance other than the default with `PythonRandomInterface`.
```
stefanv, line 449: We wrap any RandomState instance other than the default with ...
All other ... are wrapped with

This is rather confusing!

stefanv (commented):
Sorry to be late on the review; the comments are minor and docstring related, so feel free to ignore.

dschult:
Thanks for these comments/suggestions! PR #7322 implements them. I think the wording is clearer -- but is it clear enough? :)

PR #6876, opened 2023-08-25
Allow graph generators and conversion functions to be dispatched

eriknw:
For now, the `backend` argument *must* be used to dispatch to a backend. Global config for automatic conversions to backends *does not* work on graph generators, readers, converters, or any function that does not have any graph inputs. For example:
```python
G = nx.from_pandas_edgelist(df, backend="cugraph")
```
However, the backend function _will_ be used when testing with a backend. The backend graph will be converted to a networkx graph automatically in `_convert_and_call_for_tests`.

This follows up on #6840.

CC @aaronzo @MridulS @jim22k @rlratzel


rlratzel:
I've made progress on some mockup scripts that demonstrate various dispatching use cases that I think can be supported incrementally over time.  They cover config, auto-conversions, fallback scenarios, multiple backends, dynamic graph updates, etc. and I think they could influence this PR. I'm hoping to make the mockups available this week and go over them at the next community call, or earlier if possible, with at least @eriknw and @MridulS .  Can we go over those before merging this? It might prevent some API churn if we decide we want to support things differently.

rlratzel:
> Can we go over those before merging this? It might prevent some API churn if we decide we want to support things differently.

We had a good discussion on this topic recently.  We also brought it up at the last dispatching community call (some notes [here](https://hackmd.io/rqs_pWMxSLmICXCpI3w-Ug?view#August-31-2023)).  I think the outcome was that we need to continue the discussions and add an NXEP, but in the meantime, I don't think we should block this PR (assuming we can change things related to it in the future if necessary, which I think we can?).  I will add a link to the new NXEP for this once it's available.

cc @aaronzo @MridulS @jim22k

MridulS (commented):
This looks good to me, but I couldn't really test this bit as graphblas doesn't implement from_pandas_edgelist or from_scipy_sparse_array

Not sure if I can setup nx-cugraph locally without GPUs

eriknw:
Test added, which was a bit more squirrelly than I expected.

eriknw:
Should I/we go ahead and decorate more graph creation functions?

dschult:
The test seems to work. And it seems like a reasonable approach. So it looks good to go ahead and decorate more graph creation functions

networkx/algorithms/bipartite/generators.py
```diff
@@ -432,6 +438,7 @@ def preferential_attachment_graph(aseq, p, create_using=None, seed=None):
 
 
 @py_random_state(3)
+@nx._dispatch(graphs=None)
 def random_graph(n, m, p, seed=None, directed=False):
```
eriknw, line 442: Should we use the dispatch name of `"bipartite_random_graph"` here? Should we be careful with any other names in `bipartite` submodule?

networkx/algorithms/tree/operations.py
```diff
@@ -31,6 +31,8 @@ def join(rooted_trees, label_attribute=None):
     return join_trees(rooted_trees, label_attribute=label_attribute)
 
 
+# Argument types don't match dispatching, but allow manual selection of backend
+@nx._dispatch(graphs=None)
 def join_trees(rooted_trees, *, label_attribute=None, first_label=0):
```
eriknw, line 36: FYI, I added this. The input types don't work well with our current dispatching machinery, but this allows `backend=whatever` to be used explicitly.

networkx/algorithms/bipartite/generators.py
dschult: Yes -- names in the bipartite subpackage can clobber names in the main namespace. So we should handle bipartitie in a similar way to how It was handled for some algorithms like `approximations`.

dschult (approved):
This looks good to me...
Lot's here, but all pretty straightforward.

networkx/algorithms/bipartite/generators.py
eriknw: Yeah, this PR currently changes the name of those that are clobbered, and so far our approach has been to _not_ preemptively add e.g. `"bipartite_"` or `"approximate_"` prefix. Following current precedent, we don't add `"bipartite_"` here even though `random_graph` seems kind of generic.

eriknw: oh wait, you answered "yes"... should we add `"bipartite_"` prefix?

dschult: You should treat it the same as you did for the algorithms. If you only did it for the names that actually clobber, then let's stick to that here too.
:)

MridulS (approved):

PR #6878, opened 2023-08-26
Move random_state decorators before `@nx._dispatch`

eriknw:
When I first tried to have `@py_random_state` and `@np_random_state` come before `@nx._dispatch` in #6688, I encountered an error. I don't recall what, but @dschult expected it to "just work" and was surprised there was an error (and it's possible I did something wrong). Anyway, I thought I would check to see if this would work now after #6840... and it does!


MridulS (approved):
Probably something to do with closures?
Anyway let's merge this :)

dschult (approved):
Looks Good!

PR #6880, opened 2023-08-29
Add time series Visibility Graph generator

erivlis:
Return a Visibility Graph of an input Time Series.

    It converts a time series into a graph. The constructed graph inherits several properties of the series in its
    structure. Thereby, periodic series convert into regular graphs, and random series do so into random graphs.
    Moreover, fractal series convert into scale-free networks.
    -- https://www.pnas.org/doi/10.1073/pnas.0709247105

    Parameters
    ----------
    series: Iterable[float]
       Time Series iterable of float values

    Returns
    -------
    NetworkX graph
        The Visibility Graph of series

    Examples
    --------
    >>> import networkx as nx
    >>> from matplotlib import pyplot
    >>>
    >>> series_list = [range(10), [2, 1, 3, 2, 1, 3, 2, 1, 3, 2, 1, 3]]
    >>>
    >>> for s in series_list:
    ...     g = nx.visibility_graph(s)
    ...     print(g)
    ...     # Uncomment the following lines to plot the graph
    ...     # pos = [[x, 0] for x in range(len(s))]
    ...     # labels = nx.get_node_attributes(g, 'value')
    ...     # nx.draw_networkx_nodes(g, pos)
    ...     # nx.draw_networkx_labels(g, pos, labels=labels)
    ...     # nx.draw_networkx_edges(g, pos, arrows=True, connectionstyle='arc3,rad=-1.57079632679')
    ...     # pyplot.show()
    Graph with 10 nodes and 9 edges
    Graph with 12 nodes and 18 edges

    References
    ----------
    .. [1] Lucas Lacasa lucas@dmae.upm.es, Bartolo Luque, Fernando Ballesteros, Jordi Luque, and Juan Carlos Nuño,
       "From time series to complex networks: The visibility graph"
       Proceedings of the National Academy of Sciences,
       Vol. 105 | No. 1
       April 1, 2008
       PubMed: 18362361
       https://www.pnas.org/doi/10.1073/pnas.0709247105

### DIFF
--- a/doc/reference/generators.rst
+++ b/doc/reference/generators.rst
@@ -360,7 +360,7 @@ Sudoku
 
    sudoku_graph
 
-Times Series
+Time Series
 ____________
 .. automodule:: networkx.generators.time_series
 .. autosummary::
--- a/doc/reference/generators.rst
+++ b/doc/reference/generators.rst
@@ -359,3 +359,11 @@ Sudoku
    :toctree: generated/
 
    sudoku_graph
+
+Times Series
+____________
+.. automodule:: networkx.generators.time_series
+.. autosummary::
+   :toctree: generated/
+
+   visibility_graph
--- a/doc/reference/generators.rst
+++ b/doc/reference/generators.rst
@@ -361,7 +361,7 @@ Sudoku
    sudoku_graph
 
 Time Series
-____________
+-----------
 .. automodule:: networkx.generators.time_series
 .. autosummary::
    :toctree: generated/
--- a/examples/graph/plot_visibility_graph.py
+++ b/examples/graph/plot_visibility_graph.py
@@ -1,0 +1,40 @@
+"""
+================
+Visibility Graph
+================
+
+Visibility Graph constructed from a time series
+"""
+from matplotlib import pyplot
+
+import networkx as nx
+
+time_series = [0, 2, 1, 3, 2, 1, 3, 2, 1, 3, 2, 1, 3, 4, 0]
+# or
+# import random
+# time_series = [random.randint(1,10) for i in range(10)]
+
+g = nx.visibility_graph(time_series)
+
+labels = nx.get_node_attributes(g, "value")
+
+# a layout emphasizing the line-of-sight connectivity
+pos = {x: (x, 0) for x in range(len(time_series))}
+nx.draw_networkx_nodes(g, pos)
+nx.draw_networkx_labels(g, pos, labels=labels)
+nx.draw_networkx_edges(
+    g,
+    pos,
+    arrows=True,
+    arrowstyle="->",
+    arrowsize=10,
+    connectionstyle="arc3,rad=-1.57079632679",
+)
+pyplot.show()
+
+# a layout showcasing the time series values
+pos = {i: (i, v) for i, v in enumerate(time_series)}
+nx.draw_networkx_nodes(g, pos)
+nx.draw_networkx_labels(g, pos, labels=labels)
+nx.draw_networkx_edges(g, pos, arrows=True, arrowstyle="->", arrowsize=10)
+pyplot.show()
--- a/examples/graph/plot_visibility_graph.py
+++ b/examples/graph/plot_visibility_graph.py
@@ -19,42 +19,33 @@ G = nx.visibility_graph(time_series)
 labels = nx.get_node_attributes(G, "value")
 
 fig, all_axes = plt.subplots(2, 1, num="Visibility Graph", figsize=(8, 12))
-ax = all_axes.flat
+axs = all_axes.flat
 
-ax[0].title.set_text("Line-of-Sight Connectivity")
-ax[0].title.set_size(11)
+layouts_params = {
+    # a layout emphasizing the line-of-sight connectivity
+    "Line-of-Sight Connectivity": {
+        "pos": {x: (x, 0) for x in range(len(time_series))},
+        "connectionstyle": "arc3,rad=-1.57079632679",
+    },
+    # a layout showcasing the time series values
+    "Time Series values with Connectivity": {
+        "pos": {i: (i, v) for i, v in enumerate(time_series)}
+    },
+}
 
-ax[0].set_xlabel("Time", size=10)
+for i, (name, params) in enumerate(layouts_params.items()):
+    axs[i].title.set_text(name)
+    axs[i].title.set_size(11)
+    axs[i].set_xlabel("Time", size=10)
+    axs[i].margins(0.10)
+    nx.draw_networkx_nodes(G, params.get("pos"), ax=axs[i], alpha=0.5)
+    nx.draw_networkx_labels(G, params.get("pos"), ax=axs[i], labels=labels)
+    nx.draw_networkx_edges(
+        G, **params, ax=axs[i], arrows=True, arrowstyle="<->", arrowsize=10
+    )
 
-ax[1].title.set_text(
-    "Time Series values with Connectivity",
-)
-ax[1].title.set_size(11)
-ax[1].set_xlabel("Time", size=10)
-ax[1].set_ylabel("Value", size=10)
+axs[1].set_ylabel("Value", size=10)
 
-# a plot layout emphasizing the line-of-sight connectivity
-pos = {x: (x, 0) for x in range(len(time_series))}
-nx.draw_networkx_nodes(G, pos, ax=ax[0], alpha=0.5)
-nx.draw_networkx_labels(G, pos, ax=ax[0], labels=labels)
-nx.draw_networkx_edges(
-    G,
-    pos,
-    ax=ax[0],
-    arrows=True,
-    arrowstyle="<->",
-    arrowsize=10,
-    connectionstyle="arc3,rad=-1.57079632679",
-)
-
-# a plot layout showcasing the time series values
-pos = {i: (i, v) for i, v in enumerate(time_series)}
-nx.draw_networkx_nodes(G, pos, ax=ax[1], alpha=0.5)
-nx.draw_networkx_labels(G, pos, ax=ax[1], labels=labels)
-nx.draw_networkx_edges(G, pos, ax=ax[1], arrows=True, arrowstyle="<->", arrowsize=10)
-
-for a in ax:
-    a.margins(0.10)
 fig.suptitle("Visibility Graph")
 fig.tight_layout()
 plt.show()
--- a/examples/graph/plot_visibility_graph.py
+++ b/examples/graph/plot_visibility_graph.py
@@ -5,7 +5,7 @@ Visibility Graph
 
 Visibility Graph constructed from a time series
 """
-from matplotlib import pyplot
+from matplotlib import pyplot as plt
 
 import networkx as nx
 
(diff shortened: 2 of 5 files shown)


erivlis:
How do I add the label type:Enhancement

networkx/generators/time_series.py
```diff
@@ -0,0 +1,96 @@
...
+    def obstruction_predicate(
+        n1: int, t1: float, n2: int, t2: float
+    ) -> Callable[[int, float], bool]:
+        slope = (t2 - t1) / (n2 - n1)
+        constant = t2 - slope * n2
+
+        def is_obstruction(n: int, t: float) -> bool:
+            return t >= constant + slope * n
+
+        return is_obstruction
```
MridulS, line 74: Putting everything in line 89 also kind of speeds this up. We don't *need* to create a `obstruction_predicate` method.

MridulS (commented):
This all looks good! thanks for PR @erivlis 

could you also add this to https://github.com/networkx/networkx/blob/main/doc/reference/generators.rst so our documentation pipeline picks this up.

I mainly tried to make this a bit more efficient. I have a gut feeling we can have more speedups too but that's not a major blocker :)

networkx/generators/time_series.py
```diff
@@ -0,0 +1,96 @@
...
+        if n2 == n1 + 1:
+            # Sequential values are always connected
+            G.add_node(n1, value=t1)
+            G.add_node(n2, value=t2)
+            G.add_edge(n1, n2)
+        else:
+            # Otherwise check if any value between obstructs line of sight
+            is_obstruction = obstruction_predicate(n1, t1, n2, t2)
+            obstructed = any(
+                is_obstruction(n, t) for n, t in enumerate(series) if n1 < n < n2
```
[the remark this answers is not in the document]
erivlis: Well it will work. And can be faster for input lists. But it will fail for other Iterables. For example for Generators. For large lists Generators can be more efficient as there's less memory allocations.

I'll think how to optimize for both cases.

networkx/generators/time_series.py
```diff
@@ -0,0 +1,96 @@
...
+import itertools
+from collections.abc import Iterable
+from typing import Callable
+
+import networkx as nx
+
+__all__ = ["visibility_graph"]
+
+
+def visibility_graph(series: Iterable[float]) -> nx.Graph:
```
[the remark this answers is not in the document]
erivlis: You mean remove all type annotations from the function signature?

(I'm curious as to why?)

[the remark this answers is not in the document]
MridulS: Yes. We don't yet have a proper story for adding types to our codebase (we will probably go with stub `.pyi` files) but till then it's good to keep things consistent.

networkx/generators/time_series.py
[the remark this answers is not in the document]
MridulS: Ah yes true!

BTW I ran into https://github.com/CarlosBergillos/ts2vg/blob/main/ts2vg/graph/_natural.pyx which has a cythonised version of visibility graphs and it seems to use a divide-and-conquer strategy. Maybe we can learn something here?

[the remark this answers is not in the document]
MridulS: Maybe we should also look into supporting numpy arrays/pandas series here? When I see time series my head usually goes to these data structures instead of python lists.

erivlis:
@MridulS I have pushed changes. Please review. Thanks!
 BTW - Should click the "resolve" button on the above comments?
 
 (This is my first PR... I guess my newbieness is showing... :-))

networkx/generators/time_series.py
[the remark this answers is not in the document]
erivlis: > Ah yes true!
> 
> BTW I ran into https://github.com/CarlosBergillos/ts2vg/blob/main/ts2vg/graph/_natural.pyx which has a cythonised version of visibility graphs and it seems to use a divide-and-conquer strategy. Maybe we can learn something here?

Excellent find. I'll take a look when time permits.



> Maybe we should also look into supporting numpy arrays/pandas series here? When I see time series my head usually goes to these data structures instead of python lists.

I'll investigate that too.

Thanks for the great suggestions.

MridulS:
> BTW - Should I click the "resolve" button on the above comments?

Yes please!

> (This is my first PR in GitHub... I guess my newbieness is showing... :-))

No worries. Hopefully many more to come :)

networkx/generators/time_series.py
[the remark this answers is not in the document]
erivlis: I will mark this as resolved for now. I'll add these if they pan out in a separate PR. Agreed?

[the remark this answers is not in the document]
MridulS: sounds good! Maybe just add some TODOs in the code comments, so it can be picked up later on?

MridulS:
@erivlis I made some cosmetic changes https://github.com/erivlis/networkx/pull/1 could you review that? I think the code changes should be fine (they are passing the tests at least.

erivlis:
> @erivlis I made some cosmetic changes [erivlis#1](https://github.com/erivlis/networkx/pull/1) could you review that? I think the code changes should be fine (they are passing the tests at least.

Ok. I'll take a look as time permits, I have a busy week. Thanks for all the thoughtfulness!

erivlis:
@MridulS Merged your PR. I did some simple performance tests. Though the differences are minimal this version of the code performs the best. Hopefully even greater performance improvements will come in future.

MridulS (approved):
Thanks @erivlis!

networkx/generators/tests/test_time_series.py
```diff
@@ -0,0 +1,30 @@
...
+    expected_series_length = len(series)
+
+    # act
+    actual_graph = nx.visibility_graph(series)
+
+    # assert
+    assert actual_graph.number_of_nodes() == expected_series_length
+    assert actual_graph.number_of_edges() == 9
+
+
```
dschult, line 19: Add corner case testing
```suggestion


def test_visibility_graph_corner_cases():
    edge_graph = nx.visibility_graph([10, 20])
    assert list(edge_graph.edges) == [(0, 1)]

    node_graph = nx.visibility_graph([10])
    assert node_graph.number_of_nodes() == 1
    assert node_graph.number_of_edges() == 0

    null_graph = nx.visibility_graph([])
    assert node_graph.number_of_nodes() == 0


```

networkx/generators/tests/test_time_series.py
```diff
@@ -0,0 +1,30 @@
...
+    # arrange
+    series = list(itertools.islice(itertools.cycle((2, 1, 3)), 12))
+    expected_series_length = len(series)
+
+    # act
+    actual_graph = nx.visibility_graph(series)
+
+    # assert
+    assert actual_graph.number_of_nodes() == expected_series_length
+    assert actual_graph.number_of_edges() == 18
```
dschult, line 30:
```suggestion
    series = list(itertools.islice(itertools.cycle((2, 1, 3)), 15))
    expected_series_length = len(series)

    # act
    actual_graph = nx.visibility_graph(series)

    # assert
    assert actual_graph.number_of_nodes() == expected_series_length
    assert actual_graph.number_of_edges() == 23
```

dschult (commented):
Thanks for this addition! 
I've made some suggestions below. They are optional. Push back or change wording as you like. You can make the changes in your own local copy and push to your branch, or you can use the Github interface if you are just accepting the suggestion. Try "batching" the commits through the interface if you are going to accept many of them.

Other comments:
Nice examples -- :)

No need to be specific about certain types of series. The Python name for the desired input type is a Sequence of Numbers (see collections.abc and numbers built-in libraries).  Then we typically don't check the type of the input. Just state it in the doc_string and let the python errors (for iterating and slicing and len) tell the user what they need to do.  

I added some corner cases tests (a series of 0 or 1 or 2 elements). And I changed the tests to make them different from the doc_string examples (which are also tests! :).  One is now convex (no obstructions)

erivlis:
@dschult 

Thank you for all the pleasently presented thoughtfull comments.

They all make sense in broader contexts.

I accepted them all.

erivlis:
@dschult @MridulS  Should I add an example in `/examples/graph`?
```python
"""
================
Visibility Graph
================

Visibility Graph constructed from a time series
"""


import networkx as nx
from matplotlib import pyplot

time_series = [2, 1, 3, 2, 1, 3, 2, 1, 3, 2, 1, 3, 4]

g = nx.visibility_graph(time_series)

pos = [[x, 0] for x in range(len(time_series))]
labels = nx.get_node_attributes(g, "value")
nx.draw_networkx_nodes(g, pos)
nx.draw_networkx_labels(g, pos, labels=labels)
nx.draw_networkx_edges(g, pos, arrows=True, connectionstyle="arc3,rad=-1.57079632679")
pyplot.show()
```

dschult:
Thanks for the quick turnaround. 

And -- That looks like a nice example! :) I think that would be reasonable addition. Would it be better with the height (value of the time series) shown?
```python
pos = [[i, v] for i, v in enumerate(time_series)]
```
I don't know what is typically used for this kind of analysis. So we should go with whatever is pretty standard. If nothing obvious has become standard than we can do what you think best conveys the information.
Thanks!!

erivlis:
> I don't know what is typically used for this kind of analysis. So we should go with whatever is pretty standard. If nothing obvious has become standard than we can do what you think best conveys the information. Thanks!!

I think I saw both kinds, so why not both (see last commit).

dschult:
The code looks good, but the example only ends up showing 1 figure with both pictures superimposed (you can see what the rendered docs look like via the CI - tests (37 checks on your PR page). Near the bottom is one called "document artifact". Click on the "Details" link for that test, and you go to a webpage of the dully rendered docs after your PR. Navigate to the Gallery and select your example.

One example that also has two pictures is in the basic section called simple_graphs.  I think all that is needed is for you to create a new figure via `fig = pyplot.figure()` or to create a new axis `ax = pyplot.gca()` or maybe you have to put a special character between the plots `# %%`   (I believe the # is to make it a python comments, and the %% breaks up the sphinx docs. But I'm not an expert in these things. sphinx-gallery is the tool we use, but I don't think you'll have to go down that rabbit hole. Try these things first.  You can also build the docs locally if you install everything in `requirements/doc.txt` and then go to the `doc` directory and use `make html`. It takes a while. :}

networkx/generators/tests/test_time_series.py
```diff
@@ -0,0 +1,46 @@
...
+    # arrange
+    series = [i**2 for i in range(10)]  # no obstructions
+    expected_series_length = len(series)
+
+    # act
+    actual_graph = nx.visibility_graph(series)
+
+    # assert
+    assert actual_graph.number_of_nodes() == expected_series_length
+    assert actual_graph.number_of_edges() == 45
```
rossbar, line 33: IIUC, the convex case gives the complete graph - it might be worth stating that (and formulating the test around that) explicitly here. Not a blocker though!

Also in nit-land: no need to add comments for the steps of the testing recipe!

networkx/generators/time_series.py
```diff
@@ -0,0 +1,72 @@
...
+Time Series Graphs
+"""
+import itertools
+
+import networkx as nx
+
+__all__ = ["visibility_graph"]
+
+
+def visibility_graph(series):
```
rossbar, line 11: I did a quick google of the term "visibility graph" and the first hit was [a wiki article](https://en.wikipedia.org/wiki/Visibility_graph) which gave me pause about this name. According to the article, the term is used in robotics to capture the same concept, but in spatial dimensions (i.e. 2D or 3D) instead of the "time series" (i.e. 1D) interpretation used here.

Perhaps we should consider a different, more specific name for this graph generation function then? Maybe `sequential_visibility_graph` or `series_visibility_graph`? I don't feel *too* strongly about it, but it's worth thinking about naming now!

doc/reference/generators.rst
```diff
@@ -359,3 +359,11 @@ Sudoku
    :toctree: generated/
 
    sudoku_graph
+
+Times Series
+____________
```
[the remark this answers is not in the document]
erivlis: I will fix this typo in the next push

networkx/generators/time_series.py
erivlis: So perhaps the grouping "Time Series" is what we should change?

How about "Scalar Fields"? You can think of time series as a 1D Scalar field.

This can then include the another PR I made  #6919 - which is about generating Gradient Network of a scalar field substrate graph.

erivlis:
I pushed some updates. Please review.

If all is well, what's the next step?

networkx/generators/tests/test_time_series.py
```diff
@@ -0,0 +1,75 @@
...
+    assert nx.is_isomorphic(actual_graph, nx.complete_graph(expected_series_length))
+
+
+def test_visibility_graph__concave_series__path_graph():
+    # arrange
+    series = [-(i**2) for i in range(10)]  # living in 1D flatland
```
dschult, line 39: I think this isn't flatland (that's the next test).

networkx/generators/tests/test_time_series.py
```diff
@@ -0,0 +1,75 @@
...
+    expected_node_count = len(series)
+
+    # act
```
dschult, line 42: Take out the comments "arrange", "act", "assert" because the variable names tell you what is going on.

examples/graph/plot_visibility_graph.py
```diff
@@ -0,0 +1,60 @@
...
+ax[0].title.set_size(11)
+
+ax[0].set_xlabel("Time", size=10)
+
+ax[1].title.set_text(
+    "Time Series values with Connectivity",
+)
+ax[1].title.set_size(11)
+ax[1].set_xlabel("Time", size=10)
+ax[1].set_ylabel("Value", size=10)
```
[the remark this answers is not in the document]
erivlis: pushed a newer version - check it out.

networkx/generators/time_series.py
erivlis: > From what perspective are you finding these algorithms and being interested enough to submit PRs? 

First of all I find Graph Theory generally interesting. Specifically at work we modeled data into graphs with two immediate main aims: 1) standard data structures to hold data from disparate sources. 2) investigate those data structures and create larger graphs by finding connections. But in the long run also to gain insight by utilizing Graph Theory to investigate the data gathered.

I was looking into data source  that had time series characteristics - and was looking for a way to model that as a graph. Visibility Graphs looked liked the way to go. 

Why do a PR? For quite a while I wanted to get into open source. I used NetworkX, and I think its an excellent lib. So I thought why not contribute back.

> What training/literature are you coming from?

I always loved Math and Physics (and science in general), I still do. I did my BSc in Physics/Math. Some Master degree courses. My plan to go all way. PhD, Academia, Research. The whole thing. But then life happened. 
Started working in software ~30 years ago. Gained experience in many things. Now I mainly do backend engineering. What I like most is to have mix of software engineering with the science/math. When there's a combination of both, and things go well, that's when you get to make "innovative" things, sometimes even "magical" things. You can check my [linkedin](https://www.linkedin.com/in/eranrivlis/) for more.

rossbar: Thanks @dschult for digging into the background and the nice summary!

> Perhaps of most import is that both "time series" and "visibility graph" seem to be standard terms for this style of analysis. I don't find very much using "scalar field" (perhaps too broad a term -- or maybe just not as catchy -- or maybe I just didn't search well)

I'm convinced! +1 for sticking with the original naming convention given it's better alignment with the general landscape.

dschult: Thanks for that summary of your background. It's very nice to have you here and to have your help. :)

dschult (approved):

PR #6885, opened 2023-08-30
Revert "Clarify that basis generates simple cycles only"

MridulS:
Reverts networkx/networkx#6882

woopsie


FWDekker:
Thank you!

Issue #6886, opened 2023-08-31
Unexpected Crash in "nx.percolation_centrality" caused by KeyError: 0


joyemang33:
Hello! Sorry for bothering you again.
I noticed that "nx.percolation_centrality" will sometimes crash by a `KeyError: 0` exception.
Since this exception does not belong to NetworkX.Exceptions, I believe that it may related to an unexpected exception issue.
Would it be possible for you to confirm and investigate this?

Best regards,
Joye 

### Step to Reproduce

Please run the following Python code:
```Python
import networkx as nx

[...]

PR #6887, opened 2023-08-31
ENH: let users set a default value in get_attr methods

MridulS:
Currently it's not possible to get a dictionary with a default value if someone uses the `get_(node/edge)_attributes` methods. If a user passes an attribute name which is set by only a subset of nodes/edges they need to augment the return dictionary manually if they want to set a default value.

The implementation in this PR should be backward compatible so if someone is only expecting to get the nodes/edges which has these attributes it will keep on working.

This came up while trying to come up with a fix for https://github.com/networkx/networkx/issues/6886

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -19,7 +19,6 @@ Highlights
 This release is the result of X of work with over X pull requests by
 X contributors. Highlights include:
 
-
 Improvements
 ------------
 
@@ -39,8 +38,8 @@ API Changes
   That argument resulted in silently incorrect results more often than not.
 
 - [`#6887 <https://github.com/networkx/networkx/pull/6887>`_]
-  A new `default` argument is added to `get_node_attributes` and
-  `get_edge_attributes`. The `default` keyword can be used to set
+  A new ``default`` argument is added to ``get_node_attributes`` and
+  ``get_edge_attributes``. The ``default`` keyword can be used to set
   a default value if the attribute is missing from a node/edge.
 
 
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -38,6 +38,10 @@ API Changes
   In `is_semiconnected`, the keyword argument `topo_order` has been removed.
   That argument resulted in silently incorrect results more often than not.
 
+- [`#6887 <https://github.com/networkx/networkx/pull/6887>`_]
+  A new `default` argument is added to `get_node_attributes` and
+  `get_edge_attributes`. The `default` keyword can be used to set
+  a default value if the attribute is missing from a node/edge.
 
 
 Deprecations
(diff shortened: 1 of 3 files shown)


networkx/classes/function.py
```diff
@@ -659,6 +659,10 @@ def get_node_attributes(G, name):
     name : string
        Attribute name
 
+    default: object (default=None)
+       Default value of the node attribute if there
+       is no value set for that node in graph.
+
```
dschult, line 665:
```suggestion
       is no value set for that node in graph.
       If `None` then nodes without this attribute
       are not included in the returned dict.
```

networkx/classes/function.py
```diff
@@ -821,6 +831,10 @@ def get_edge_attributes(G, name):
     name : string
        Attribute name
 
+    default: object (default=None)
+       Default value of the edge attribute if there
+       is no value set for that edge in graph.
+
```
dschult, line 837:
```suggestion
       is no value set for that edge in graph.
       If `None` then edges without this attribute
       are not included in the returned dict.
```

dschult (commented):
Nice!!    I like this -- anything to improve working with graph attributes; and this change is really clean and intuitive with no backward incompatibility issues.

I just have two suggestions for doc_string to state what happens if the default is `None`.

dschult (approved):

dschult:
Re-looking at this I thought of two silly things that aren't blockers:
- should `default` be keyword only  (`,*, default=None` in the signature)?
- should a note be added to the release notes
Fine with just merging too...

MridulS:
> should default be keyword only (,*, default=None in the signature)?

I was thinking about that too but don't we do that for just new functions? I'm fine adding it keyword only too.

>    should a note be added to the release notes

Yes! I'll do that.

doc/release/release_dev.rst
```diff
@@ -38,6 +38,10 @@ API Changes
   In `is_semiconnected`, the keyword argument `topo_order` has been removed.
   That argument resulted in silently incorrect results more often than not.
 
+- [`#6887 <https://github.com/networkx/networkx/pull/6887>`_]
+  A new `default` argument is added to `get_node_attributes` and
+  `get_edge_attributes`. The `default` keyword can be used to set
+  a default value if the attribute is missing from a node/edge.
```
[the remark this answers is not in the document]
dschult: I would think it should be API changes -- since it is changing a function signature. But it is also an improvement. I think we put improvements as things that are better -- but don't have another place to be mentioned.

dschult (commented):
I thought it was any new function or new keyword argument. But it's fine either way.
I'll let you merge when you are satisfied its ready. :}

rossbar (approved):
Awesome! It's not every day that you get to enhance core functionality :).

LGTM - I took the liberty of pushing up some changes, purely rst formatting/line-length expansion; no changes to content or code. I love the ancillary test cleanup as well!

PR #6895, opened 2023-09-01
Fix sphinx docs rendering of dispatched functions

eriknw:
@MridulS noticed the online docs no longer render properly for dispatched functions now that `_dispatch` is a class: https://github.com/networkx/networkx/pull/6840#issuecomment-1700779731

I'm not a Sphinx expert, but here is one way to fix the problem. It also allows the dispatch machinery to update the docstring. The signature of dispatched functions is unchanged in the online docs--that is, `backend=None, **backend_kwargs` is not shown.

### DIFF
--- a/networkx/utils/backends.py
+++ b/networkx/utils/backends.py
@@ -817,3 +817,19 @@ class _dispatch:
 
 def _restore_dispatch(name):
     return _registered_algorithms[name]
+
+
+if os.environ.get("_NETWORKX_BUILDING_DOCS_"):
+    # When building docs with Sphinx, use the original function with the
+    # dispatched __doc__, b/c Sphinx renders normal Python functions better.
+    # This doesn't show e.g. `*, backend=None, **backend_kwargs` in the
+    # signatures, which is probably okay. It does allow the docstring to be
+    # updated based on the installed backends.
+    _orig_dispatch = _dispatch
+
+    def _dispatch(func=None, **kwargs):
+        if func is None:
+            return partial(_dispatch, **kwargs)
+        dispatched_func = _orig_dispatch(func, **kwargs)
+        func.__doc__ = dispatched_func.__doc__
+        return func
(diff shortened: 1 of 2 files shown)


eriknw:
I don't know how to make `mypy` happy here. Or, rather, I don't think you would like my solution for making mypy happy 😜

eriknw:
Btw, adding `__repr__` to `_dispatch` helps, but it doesn't _completely_ solve the issue. It results in e.g. `adjacency_matrix = ...` instead of `adjacency_matrix(...)`.

networkx/utils/backends.py
```diff
@@ -817,3 +817,19 @@ def __reduce__(self):
...
+
+if os.environ.get("_NETWORKX_BUILDING_DOCS_"):
+    # When building docs with Sphinx, use the original function with the
+    # dispatched __doc__, b/c Sphinx renders normal Python functions better.
+    # This doesn't show e.g. `*, backend=None, **backend_kwargs` in the
+    # signatures, which is probably okay. It does allow the docstring to be
+    # updated based on the installed backends.
+    _orig_dispatch = _dispatch
+
+    def _dispatch(func=None, **kwargs):
```
MridulS, line 830:
```suggestion
    def _dispatch(func=None, **kwargs): # type: ignore
```
🙃

networkx/utils/backends.py
```diff
@@ -817,3 +817,19 @@ def __reduce__(self):
...
+
+if os.environ.get("_NETWORKX_BUILDING_DOCS_"):
+    # When building docs with Sphinx, use the original function with the
+    # dispatched __doc__, b/c Sphinx renders normal Python functions better.
+    # This doesn't show e.g. `*, backend=None, **backend_kwargs` in the
+    # signatures, which is probably okay. It does allow the docstring to be
+    # updated based on the installed backends.
+    _orig_dispatch = _dispatch
+
+    def _dispatch(func=None, **kwargs): # type: ignore
```
MridulS, line 830:
```suggestion
    def _dispatch(func=None, **kwargs):  # type: ignore
```

PR #6900, opened 2023-09-04
MAINT: Make GEXF and graphml writer work with numpy 2.0

MridulS:
Nightly tests are failing with numpy 2.0 as numpy has removed `np.float_` https://github.com/networkx/networkx/actions/runs/6061305343/job/16446362472

I think this change should be backward compatible (?) as GEXF and graphml the format don't have a concept of numpy floats (?).

### DIFF
--- a/networkx/readwrite/gexf.py
+++ b/networkx/readwrite/gexf.py
@@ -229,7 +229,6 @@ class GEXF:
                 (np.float64, "float"),
                 (np.float32, "float"),
                 (np.float16, "float"),
-                (np.float_, "float"),
                 (np.int_, "int"),
                 (np.int8, "int"),
                 (np.int16, "int"),
--- a/networkx/readwrite/graphml.py
+++ b/networkx/readwrite/graphml.py
@@ -413,7 +413,6 @@ class GraphML:
                 (np.float64, "float"),
                 (np.float32, "float"),
                 (np.float16, "float"),
-                (np.float_, "float"),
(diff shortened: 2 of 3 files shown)


dschult (commented):
I think changing `float_` to `float64` in the test will work just as well as it did previously.  We just wanted the test to check that it was correctly handling a numpy type…. Not necessarily that it was handling ` float_`.  So let’s try changing the test to a different type instead of removing the whole test.  I am glad they are getting rid of float_. I never did understand what that was supposed to do for us. ;)  but then, I don’t understand lots of things. ;)

thanks for monitoring the nightly builds!  It means we don’t have to panic about it after they release (or we release).  :)

MridulS:
There are equivalent tests for `float64`  https://github.com/networkx/networkx/blob/40e4b49aa2959f84cacbf3ee4732b4fba7c3da23/networkx/readwrite/tests/test_graphml.py#L1417 and `float32` https://github.com/networkx/networkx/blob/40e4b49aa2959f84cacbf3ee4732b4fba7c3da23/networkx/readwrite/tests/test_graphml.py#L1433

This is already testing for writing and reading graphml's `float`.

dschult (approved):
Yup!  I missed those....   <seeing what's reported changed instead of what hasn't been changed>
Thanks!

I approve these changes!

rossbar (approved):
Great fix @MridulS , LGTM!

PR #6903, opened 2023-09-05
Add rumor centrality from Shah & Zaman

simon-forb:

### DIFF
--- a/networkx/algorithms/centrality/rumor.py
+++ b/networkx/algorithms/centrality/rumor.py
@@ -1,0 +1,115 @@
+import math
+
+import networkx as nx
+from networkx.utils.decorators import not_implemented_for
+
+__all__ = ["rumor_centrality"]
+
+
+@not_implemented_for("directed")
+def rumor_centrality(G):
+    """Compute the rumor centrality for nodes.
+
+    Parameters
+    ----------
+    G : graph
+      A networkx graph
+
+    Returns
+    -------
+    nodes : dictionary
+       Dictionary of nodes with degree centrality as the value.
+
+    Raises
+    ------
+    NetworkXPointlessConcept
+        If the graph G is the null graph.
+
+    Notes
+    -----
+    Rumor centrality was introduced by Shah & Zaman [1]_ for rumor source detection.
+
+    .. [1] Shah, Devavrat, and Tauhid Zaman. "Rumor centrality: a universal source detector." In Proceedings of the
+    12th ACM SIGMETRICS/PERFORMANCE joint international conference on Measurement and Modeling of Computer Systems,
+    pp. 199-210. 2012.
+
+    """
+
+    if len(G) == 0:
+        raise nx.NetworkXPointlessConcept(
+            "cannot compute centrality for the null graph"
+        )
+
+    # Create copy to be able to use node properties
+    H = G.copy()
+    set_visited_to_false(H)
+
+    # Calculate rumor centrality for a single vertex
+    arbitrary_start_vertex = list(H)[0]
+    calculate_rumor_centrality_for_single_vertex(
+        H, arbitrary_start_vertex, arbitrary_start_vertex
+    )
+
+    # Calculate rumor centrality for remaining vertices
+    set_visited_to_false(H)
+    H.nodes[arbitrary_start_vertex]["visited"] = True
+    for neighbor in H.neighbors(arbitrary_start_vertex):
+        calculate_for_remaining(
+            H, neighbor, H.nodes[arbitrary_start_vertex]["rumor_centrality"]
+        )
+
+    return {v: H.nodes[v]["rumor_centrality"] for v in H.nodes}
+
+
+def calculate_rumor_centrality_for_single_vertex(G, v, start_vertex):
+    # Mark v as visited
+    G.nodes[v]["visited"] = True
+
+    # Collect all unvisited neighbors
+    unvisited_neighbors = []
+    for neighbor in G.neighbors(v):
+        if not G.nodes[neighbor]["visited"]:
+            unvisited_neighbors.append(neighbor)
+
+    # If v does not have unvisited neighbors, initialize v with default values
+    if len(unvisited_neighbors) == 0:
+        G.nodes[v]["subtree_size"] = 1
+        G.nodes[v]["cumulative_prod"] = 1
+        return 1, 1
+
+    # If v has unvisited neighbors, invoke recursive call for each of them
+    # and calculate cumulative sum and product
+    cum_sum, cum_prod = 0, 1
+    for neighbor in unvisited_neighbors:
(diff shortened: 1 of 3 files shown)


simon-forb:
@dschult Thank you for adding the relevant label. Could you please tell me if there is anything I could do to get a review for this pull request?

peijenburg:
Is this metric defined for weighted, directed and multigraphs? I haven't read the paper, but if it works, there should be tests and if it doesn't, it should give exceptions and some info in the documentation

simon-forb:
@peijenburg Thank you for taking a look at this PR. This metric is not defined for any of the graph classes you mention. I will update the code and the documentation accordingly.

yoaocopy:
When I am trying to write the code for rumor centrality, I see this. Thanks for the code. 
Rumor centrality is a measure used for tree graph to decide the rumor center. Maybe better to only allow the input to be tree graph?

When the graph is a general graph, the maximum likehood probability based on bfs tree and rumor centrality will be used. I am writing this code for some testing, and will share the code if no existing code available at that time.

simon-forb:
@peijenburg I have now updated the code as you requested. Multigraphs and directed graphs are now excluded as the code raises an exception in those cases. The documentation is updated accordingly. Weighted graphs are allowed since the edge weights do not affect the algorithm. A test case for these graphs is therefore added.

simon-forb:
@peijenburg Thank you for your suggestions. I have now implemented both of them. Previously I did not know about the `is_tree` method.

peijenburg:
@simon-forb Great! Final question for my side:
Could you add in the documentation some explanation _why_ this metric is relevant? Just one or two lines should be sufficient...

simon-forb:
@peijenburg Thank you! I have updated the documentation.

networkx/algorithms/centrality/rumor.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+    if not nx.is_tree(G):
+        raise nx.NetworkXNotImplemented(
+            "rumor centrality does not work for graphs with cycles"
+        )
+
+    # Create copy to be able to use node properties
+    H = G.copy()
+    set_visited_to_false(H)
```
dschult, line 56: It is likely to be much faster and probably more memory efficient if you store the node properties as dicts keyed by node to the value of that property rather than store them on `G`. It is also less likely to overwrite any user attributes.

For example, `visited` can be a set. Add nodes to the set when they are visited. check if they have been visited by `node in visited`.

dschult (commented):
Thanks for this!  We are building a large group of modules in `centrality` that have one public function in them.  Is `rumor_centrality` closely related to another centrality measure so that they could be combined? No problem if it would be better to keep them separate.

The functions that are not in `__all__` are "private functions" and should start with an underscore `_` character. That helps identify them as helper functions without having to go check the `__all__` variable.  We usually try to avoid these unless they are likely to be used elsewhere or they allow us to avoid lots of code copying.

Standard naming in networkx is `node` and `edge` rather than `vertex` and `link` or other choices. Please adjust code and comments accordingly. 

The first line of a doc_string should be a short one-line description of the function (used to label a link to this function in the documentation webpages). Follow this with a blank line and then a paragraph (like you start with now) describing what the function does. 

Can you add a sentence to your paragraph comparing this centrality measure to oethers to show users when they might want to use this one? What is it designed to measure and when might you want to try it?

Please wrap the text in the doc_strings to <88 chars similar to how `black` prefers code to appear. <80 is even better. 

The comment below about storing node properties in separate objects named things like: `visited`, `subtree_size`, `rumor_centrality`) will likely speed this up and reduce the risk of overwriting the user's node attributes.

For the tests, you put all the setup into a `setup_method` function even though the graphs are only used in one test.  Could you switch the tests to be functions instead of methdos in a class, and create the graph within the test it is used for? Lots of `self` can be removed that way too.  We are trying to move our tests to a function based system where reasonable.

networkx/algorithms/centrality/rumor.py
simon-forb: Thank you! I have implemented this as requested.

networkx/algorithms/centrality/rumor.py
```diff
@@ -0,0 +1,126 @@
...
+    12th ACM SIGMETRICS/PERFORMANCE joint international conference on Measurement and Modeling of Computer Systems,
+    pp. 199-210. 2012.
+
+    """
+
+    if len(G) == 0:
+        raise nx.NetworkXPointlessConcept(
+            "cannot compute centrality for the null graph"
+        )
+
```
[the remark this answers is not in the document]
simon-forb: Thank you! I have implemented this as requested.

simon-forb:
@dschult Thank you for your useful feedback. I have now implemented everything you suggested.

dschult (commented):
You should connect the docs to hte online documentation by adding something to `doc/reference/algorithms/centrality.rst`.

One trouble with recursive functions is that they have a finite recursion depth in Python (that is configurable). They also tend to be memory hogs because they allocate an entire function for each call.  The alternative is to loop over a queue. My quick look suggests that you need to store the next node, an iterator over its `unvisited_neighbors`, it's cum_sum and cum_prod. The start of the loop would be something like:
```python
start_node = nx.utils.arbitrary_element(G)
<some lines to compute the unvisited_neighbors>

queue = [(start_node, iter(unvisited_neighbors), 0, 1)]
while queue:
    node, nbrs, cum_sum, cum_prod = queue[-1]
    try:
        nbr = next(nbrs)
    except StopIteration:
        subtree_size[node] = big_cum_sum = cum_sum + 1
        cumulative_prod[node] = big_cum_prod = cum_prod * (cum_sum + 1)
        if node is start_node:
            rumour_centrality[node] = math.factorial(len(G) -1 ) // cum_prod
            break
        queue.pop()
        # update parent node's values
        queue[-1][2] += big_cum_sum
        queue[-1][3] *= big_cum_prod
        continue

    visited.add(nbr)
    # compute unvisited_neighbors of nbr
    queue.append((nbr, iter(unvisited_neighbors), 0, 1))

 # Done with first recursion. Now calculate for remaining nodes
visited = {start_node}
queue = [(start_node, iter(G.neighbors(start_node)))]
whlie queue:
    node, nbrs = queue[-1]
    try:
        nbr = next(nbrs)
    except StopIteration:
        queue.pop()
        continue

    if nbr in visited:
        continue
    visited.add(nbr)
    rumor_centrality[nbr] = int(rumor_centrality[node] * subtree_size[node] / (len(G) - subtree_size[node]))
    queue.append((nbr, iter(G.neighbors(nbr))))
```
Warning... this is untested and leaves out some stuff at the top to define the visited set and subtree_size, etc dicts.  The steps should be the same as in your recursive code... just without calling functions repeatedly. Feel free to post a buggy version if it gets hard to figure out what's going on with this.

rossbar (commented):
This one's been stale for a while, are you still intending to work on it @simon-forb ?

PR #6905, opened 2023-09-05
 API: Add a decorator to deprecate positional args 

MridulS:
[Scikit-learn has a decorator](https://github.com/scikit-learn/scikit-learn/blob/8ed0270b99344cee9bb253cbfa1d986561ea6cd7/sklearn/utils/validation.py#L37C1-L90C44) which took care of the deprecation cycle for moving to keyword args from positional args. I have vendored that in as a networkx decorator.

This should resolve https://github.com/networkx/networkx/issues/4845.

### DIFF
--- a/networkx/classes/function.py
+++ b/networkx/classes/function.py
@@ -388,7 +388,7 @@ def induced_subgraph(G, nbunch):
     [0, 1, 3]
     """
     induced_nodes = nx.filters.show_nodes(G.nbunch_iter(nbunch))
-    return nx.subgraph_view(G, induced_nodes)
+    return nx.subgraph_view(G, filter_node=induced_nodes)
 
 
 def edge_subgraph(G, edges):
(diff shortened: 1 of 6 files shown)


rossbar (approved):
+1 for the sklearn deprecator - that's a really nice dev feature and something that we've talked about in the past wrt. [SLEP 9](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep009/proposal.html).

You *might* consider adding the deprecator in a separate PR, but the only reason to do so is on the off-chance that we want to revert the kwarg-only-ness of the `filter_` params in the future. I don't think this is very likely, so +1 for having everything lumped together in 1 PR.

Thanks for pushing this forward @MridulS !

dschult (approved):
Wow -- very nice! 
Luv the input parameter revamp that python inspection allows. :}

I think subgraph_view is a good testcase to use it on too.
I approve this PR.

PR #6908, opened 2023-09-06
Rename function `join` as `join_trees`

### LINKED ISSUE #6906: Rename nx.join to something which gives more context about operations on trees
While reviewing #6503 [@dschult noted](https://github.com/networkx/networkx/pull/6503#issuecomment-1707200999) that the current function name `nx.join` doesn't give enough information to the user. The function is used to "join" trees.

This will require a deprecation cycle.

dschult:
Closes #6906 

Rename `algorithms.tree.operations.join` to `algorithms.tree.operations.join_trees`. Deprecate old function name.

I kept the old function name in the docs with a message stating the deprecation and pointing to the new name. 
Added note to deprecations.rst and updated tests.


MridulS (approved):
Maybe let's merge this in and update the `join_trees` method if we end up with new function arguments in https://github.com/networkx/networkx/issues/6910 ?

